PR #89, opened 2019-06-07
Fix rpy2 version requirement

lcallot:
*Issue #, if available:*

*Description of changes:*
`rpy2 >=3.*` introduces breaking changes that are not compatible with the R wrapper. Tightening the version requirement.


szha:
Job PR-89/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-89/1/index.html

lostella:
Do you know what are the breaking changes? Because another option would be to update our code and restrict to `rpy2 >= 3.*`

lcallot:
The breaking changes I've noticed happened in the handling of errors and the definition of interface functions. I don't think updating would be hugely difficult, I don't know what immediate benefits moving to `rpy2>=3.*` would bring besides not requiring an older version.

Issue #94, opened 2019-06-12
Dynamical Feature with Different Length will Throw Exception


houchangtao:
System: Ubuntu 16.04
Python: 3.6.4
mxnet: 1.4.1

Code to reproduce:

```py
import pandas as pd
url = "https://raw.githubusercontent.com/numenta/NAB/master/data/realTweets/Twitter_volume_AMZN.csv"
df = pd.read_csv(url, header=0, index_col=0)
from gluonts.dataset.common import ListDataset
training_data = [
    {"start": pd.Timestamp(df.index[0], freq='5min'), "target": df.value[:"2015-04-05 00:00:00"], 
     "feat_dynamic_real": pd.to_datetime(df[:"2015-04-05 00:00:00"].index).dayofweek.values},
    {"start": pd.Timestamp(df.index[0], freq='5min'), "target": df.value[:"2015-04-10 00:00:00"], 
     "feat_dynamic_real": pd.to_datetime(df[:"2015-04-10 00:00:00"].index).dayofweek.values}
]
from gluonts.model.deepar import DeepAREstimator
from gluonts.trainer import Trainer
from gluonts.distribution import NegativeBinomialOutput, StudentTOutput
estimator = DeepAREstimator(freq="5min", prediction_length=12, distr_output=NegativeBinomialOutput(),
                            trainer=Trainer(epochs=10))
predictor = estimator.train(training_data=training_data)
```

Exceptions:

```
KeyError                                  Traceback (most recent call last)
<ipython-input-9-7512b68a283e> in <module>()
      1 estimator = DeepAREstimator(freq="5min", prediction_length=12, distr_output=NegativeBinomialOutput(),
      2                             trainer=Trainer(epochs=10))
----> 3 predictor = estimator.train(training_data=training_data)

/usr/local/lib/python3.6/site-packages/gluonts-0.1.1-py3.6.egg/gluonts/model/estimator.py in train(self, training_data)
    187     def train(self, training_data: Dataset) -> Predictor:
    188 
--> 189         training_transformation, trained_net = self.train_model(training_data)
    190 
    191         # ensure that the prediction network is created within the same MXNet

/usr/local/lib/python3.6/site-packages/gluonts-0.1.1-py3.6.egg/gluonts/model/estimator.py in train_model(self, training_data)
    180             net=trained_net,
    181             input_names=get_hybrid_forward_input_names(trained_net),
--> 182             train_iter=training_data_loader,
    183         )
    184 

[...]

PR #125, opened 2019-06-19
Add support for feat_dynamic_real in DeepAREstimator

lostella:
*Issue #, if available:* #94 

*Description of changes:* This PR enables the `DeepAREstimator` to consume the `"feat_dynamic_real"` field. This is done involves two changes, essentially:
1. Added flags enabling optional feature fields. By default optional feature fields are not used (hence removed if present). **Note:** this makes for a breaking change, in that the `DeepAREstimator` so far was automatically using the `"feat_static_cat"` feature if provided.
2. A `RemoveFields` transformation is added, that gets rid of the optional feature fields which are not enabled.

Bonus:
* refactored/expanded tests for `DeepAREstimator` with a few combinations of features
* fixed some docstrings
* renamed the `compute_date` utility function into `shift_timestamp` which I believe is a more appropriate name


szha:
Job PR-125/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-125/1/index.html

lostella:
> I'm wondering what would be the best way to test the estimators with different input combinations (i.e. `feat_dynamic_real` given or not). Maybe we can write a fancy test class for that at some point.

I'm on it ;-)

szha:
Job PR-125/2 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-125/2/index.html

szha:
Job PR-125/3 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-125/3/index.html

src/gluonts/transform.py
```diff
@@ -515,7 +520,11 @@ def __init__(
         )
 
     def transform(self, data: DataEntry) -> DataEntry:
-        r = [data[fname] for fname in self.input_fields]
+        r = [
+            data[fname]
+            for fname in self.input_fields
+            if data[fname] is not None
+        ]
         output = np.vstack(r)
```
geoalgo: Perhaps we could add `assert len(r) > 0, "no valid field found"`

src/gluonts/model/deepar/_estimator.py
```diff
@@ -186,9 +186,16 @@ def create_transformation(self) -> Transformation:
                     pred_length=self.prediction_length,
                     log_scale=True,
                 ),
+                SetFieldIfNotPresent(
+                    field=FieldName.FEAT_DYNAMIC_REAL, value=None
+                ),
```
aalexandrov: I am not sure I am convinced that this is the best way do implement this.

What will happen if some lines have `feat_dynamic_real` and some not? If I read this correctly, the output feature `time_feat` will have a variable dimension. 

Another issue is performance. In the current implementation, we will add a field `feat_static_real` with a default value `None` only to ignore it later in the `VstackFeatures` implementation.

I think a better way to do this is to have an `__init__` parameter in the estimator that indicates whether `feat_dynamic_real` is expected and should be used. Based on this flag we can do multiple things, e.g.

- infer a schema of a particular estimator instance, 
- throw an error if `feat_dynamic_real` is missing or has malformed input
- statically decide whether the `input_fields` argument of the `VstackFeatures` transformation below should contain `FieldName.FEAT_DYNAMIC_REAL` at the time when we construct the transformation pipeline.

test/model/test_deepar_smoke.py
```diff
@@ -0,0 +1,179 @@
...
+from gluonts.dataset.common import ListDataset
+from gluonts.model.deepar import DeepAREstimator
+from gluonts.transform import FieldName
+from gluonts.trainer import Trainer
+
+NUM_TS = 5
+START = "2018-01-01"
+FREQ = "D"
+MIN_LENGTH = 20
+MAX_LENGTH = 50
```
geoalgo: Sorry for this cherrypicking but if this test is not very fast, can you reduce min_length/max_length to smaller values?

test/model/test_deepar_smoke.py
```diff
@@ -0,0 +1,179 @@
...
+        ListDataset(data_iter=data_iter_test, freq=freq),
+    )
+
+
+@pytest.mark.parametrize(
+    "estimator, datasets",
+    [
+        # No features
+        (
+            DeepAREstimator(
```
geoalgo: ``DeepAREstimator(freq="D", prediction_length=PRED_LENGTH, trainer=Trainer(epochs=3, 
num_batches_per_epoch=2))`` is repeated 6 times. Could you do something like:

``
common_args = {'freq': 'D', 'prediction_length: PRED_LENGTH, trainer=...}
DeepAREstimator(cardinality=[3, 10, 42], **common_args)
``

src/gluonts/model/deepar/_estimator.py
lostella: Yeah, that's the other way of doing it. I was trying to avoid those additional hyperparameters, but it's true that they make everything very explicit.

geoalgo: @aalexandrov, this is already the behavior for FEAT_STATIC_CAT if I am not mistaken (see line 200). Perhaps we could work on letting the user specify a schema in a separate PR?

mbohlkeschneider: > What will happen if some lines have feat_dynamic_real and some not? If I read this correctly, the output feature time_feat will have a variable dimension.

Maybe we can add this case to the tests?

szha:
Job PR-125/4 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-125/4/index.html

src/gluonts/model/deepar/_estimator.py
vafl: Wouldn't it be nicer to have the estimator pick the features up automatically if they are present? I can imagine its easy to forget to switch this on.

lostella: @vafl we discussed the two options today. Roughly speaking, either:
1. The estimator, by default, uses *all* available features in the data, unless they are explicitly disabled.
2. The estimator, by default, does *not* use any of the features, regardless of the fact that they are present or not in the dataset, and to use features one must explicitly enable it.

(The option to enable/disable features is potentially useful for experimentation purposes, since this way one does not need to e.g. reformat the data in order to exclude some features.)

Option 1 was implemented in the [previous commit](https://github.com/awslabs/gluon-ts/pull/125/files/2a55317b6c1d52ce0c7b846acbd4d2362896e9ad), in case you want to check it out.

Potential drawbacks that I see in option 1:
* All seems automatic, but some things like `cardinality` aren't: the user may get the false impression that he's doing the right thing but if he doesn't set `cardinality` he's probably doing just as good as if the categorical features were ignored. Unless we go for a fully automatic mode (one pass through the data, validating the schema, understanding which fields are/aren't there, checking min/max of categorical features and so on), in which case everything is beautiful.
* If the first entry in the dataset has a some features that the second doesn't have, these will probably be transformed somewhat differently, and a weird and unreadable numpy or mxnet error will be thrown.

On the other hand, with option 2:
* You have to enable all optional features (i.e., fields which are not `"start"` or `"target"`) and setting the right `cardinality` is part of that.
* It's clear from the start what fields the transformation chain will expect to find: asserting that each data entry has those should be quite easy and yield a very clear error message in case the user is missing something.

I'm happy to discuss this further, since it's a delicate matter and a design choice that we should carry over to all other built-in estimators.

src/gluonts/model/deepar/_estimator.py
```diff
@@ -127,9 +128,12 @@ def __init__(
             num_eval_samples > 0
         ), "The value of `num_eval_samples` should be > 0"
         assert dropout_rate >= 0, "The value of `dropout_rate` should be >= 0"
-        assert all(
-            [c > 0 for c in cardinality]
-        ), "Elements of `cardinality` should be > 0"
+        assert (
+            cardinality is not None or not use_feat_static_cat
+        ), "You must set `cardinality` if `use_feat_static_cat=True`"
```
geoalgo: Could we just use categorical features when cardinality is set and drop the field otherwise?
Currently `use_feat_static_cat` is redundant.

src/gluonts/model/deepar/_estimator.py
```diff
@@ -164,6 +170,25 @@ def __init__(
     def create_transformation(self) -> Transformation:
         return Chain(
             [
+                RemoveFields(
```
geoalgo: For the block `
                   (
                        []
                        if self.use_feat_dynamic_real
                        else [FieldName.FEAT_DYNAMIC_REAL]
                    )
`

This alternative might be more readable:

```
remove_field_names = [FieldName.FEAT_DYNAMIC_CAT, FieldName.FEAT_STATIC_REAL]
if self.use_feat_dynamic_real:
    remove_field_names.append(FieldName.FEAT_DYNAMIC_REAL)

return Chain(
    [
        RemoveFields(field_names=remove_field_names),
        ...
```

(if you disagree, just discard the comment)

src/gluonts/model/deepar/_estimator.py
lostella: we could, but this would be a bit asymmetric

szha:
Job PR-125/5 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-125/5/index.html

PR #120, opened 2019-06-18
include cdf method in distribution

vafl:
*Issue #, if available:*

*Description of changes:*

This includes a `cdf` method for (some of) the distribution.
Also fixes a couple of bugs in distributions.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #129, opened 2019-06-19
Fix the seed of a flaky `test_sampling` test

aalexandrov:
*Issue #, if available:*

*Description of changes:*

Fixes the seed of the `test_sampling` which otherwise might be flaky.

### DIFF
--- a/test/distribution/test_distribution_sampling.py
+++ b/test/distribution/test_distribution_sampling.py
@@ -67,6 +67,7 @@ test_cases = [
 DISTRIBUTIONS_WITH_CDF = [Gaussian, Uniform, Laplace, Binned]
 
 
+@pytest.mark.seed(1960264129)  # fix a static seed for a flaky test
 @pytest.mark.parametrize("distr_class, params", test_cases)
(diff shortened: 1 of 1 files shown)


szha:
Job PR-129/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-129/1/index.html

test/distribution/test_distribution_sampling.py
```diff
@@ -67,6 +67,7 @@
 DISTRIBUTIONS_WITH_CDF = [Gaussian, Uniform, Laplace, Binned]
 
 
+@pytest.mark.seed(1960264129)  # fix a static seed for a flaky test
```
jaheba: Where does this magic number come from?

jaheba:
I think this particular issue was fixed in #120.

The issue here was that `log(0)` was run in some cases which evaluates to `-inf` in `mx.nd`.

aalexandrov:
You are right, the issue only occurs in versions up to 0.1.4

PR #130, opened 2019-06-19
Bump down `pytest-cov` and `pytest-xdist` versions

aalexandrov:
*Issue #, if available:*

Reverts the versions of `pytest-cov` and `pytest-xdist` to older values (otherwise pytest fails if `pytest < 3.6`).

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,6 +1,6 @@
 flake8==3.4.*
 pytest==3.*
-pytest-cov==2.6.*
+pytest-cov==2.4.*
 pytest-runner<2.12
-pytest-xdist==1.27.*
+pytest-xdist==1.20.*
 pytest-timeout==1.2.*


szha:
Job PR-130/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-130/1/index.html

jaheba:
What I don't understand, we didn't have issues with our CI system with these versions, or do we?

And, what is the actual problem these versions caused?

requirements-test.txt
```diff
@@ -1,6 +1,6 @@
 flake8==3.4.*
 pytest==3.*
```
lostella: Can't we fix `pytest` to something >= 3.6 instead?

aalexandrov:
Will try another workaround internally.

PR #138, opened 2019-06-21
ignoring version.py

lostella:
*Description of changes:* A `version.py` file is created in `src/gluonts` which should not be edited nor committed: this PR adds it to the `.gitignore. Otherwise I think it's easy to stage this by mistake (especially when committing changes to many files).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -159,3 +159,6 @@ conda
 
 # vscode
 .vscode
+
+# version.py
+src/gluonts/version.py


jaheba:
This is specifically for the recent changes Alex did?

lostella:
> This is specifically for the recent changes Alex did?

I believe so, it wasn't there before

szha:
Job PR-138/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-138/1/index.html

aalexandrov:
This is no longer needed, with my fix from yesterday I believe that the file is no longer created.

lostella:
I'll revert this then, and remove the file locally. Sorry for the confusion.

PR #139, opened 2019-06-21
include safety check that catches infinite loops in transformations

vafl:
*Issue #, if available:*

*Description of changes:*

include safety check that catches infinite loops in transformations

### DIFF
--- a/src/gluonts/transform.py
+++ b/src/gluonts/transform.py
@@ -363,7 +363,7 @@ class FlatMapTransformation(Transformation):
                     f'This means the transformation looped over '
                     f'GLUONTS_MAX_IDLE_TRANSFORMS={GLUONTS_MAX_IDLE_TRANSFORMS} '
                     f'inputs without returning any output.\n'
-                    f'The transformation this occurred in is:\n{self}'
+                    f'This occurred in the following transformation:\n{self}'
                 )
 
     @abc.abstractmethod
--- a/src/gluonts/transform.py
+++ b/src/gluonts/transform.py
@@ -27,6 +27,7 @@ from gluonts.core.exception import GluonTSDateBoundsError, assert_data_error
 from gluonts.dataset.common import DataEntry, Dataset
 from gluonts.dataset.stat import ScaleHistogram
 from gluonts.time_feature import TimeFeature
+from gluonts.runtime_params import GLUONTS_MAX_IDLE_TRANSFORMS
 
 
 def serialize_data_entry(data_entry: DataEntry) -> Dict:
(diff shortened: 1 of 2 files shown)


szha:
Job PR-139/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-139/1/index.html

src/gluonts/transform.py
```diff
@@ -298,6 +299,7 @@ def __call__(
         self, data_it: Iterator[DataEntry], is_train: bool
     ) -> Iterator:
         for data_entry in data_it:
+            did_yield = True
```
jaheba: Can this be removed?

src/gluonts/transform.py
```diff
@@ -345,14 +347,25 @@ class FlatMapTransformation(Transformation):
     def __call__(
         self, data_it: Iterator[DataEntry], is_train: bool
     ) -> Iterator:
+        num_idle_transforms = 0
         for data_entry in data_it:
+            num_idle_transforms += 1
             try:
                 for result in self.flatmap_transform(
```
jaheba: So, `num_idle_transforms` is increased, iff `flat_map_transform` returns an empty result?

Do you have an example when that is the case?

src/gluonts/transform.py
```diff
@@ -345,14 +347,25 @@ class FlatMapTransformation(Transformation):
...
                     data_entry.copy(), is_train
                 ):
+                    num_idle_transforms = 0
                     yield result
             except Exception as e:
                 raise e
+            if num_idle_transforms > GLUONTS_MAX_IDLE_TRANSFORMS:
```
jaheba: I think what I don't like, is that there is no way to turn this check off.

src/gluonts/transform.py
vafl: yes, sorry. Forgot to remove this.

src/gluonts/transform.py
vafl: Yes, I ran into this. 

If you have a dataset with time series that all have length 100 and you set prediction length to something larger than 100. The transformations loop forever. The `InstanceSplitter` skips over time series that are too short. This is the right thing when there are time series with varying length.

src/gluonts/transform.py
vafl: Why would you want to switch it off? If you really want to you can set it to a very large number.

src/gluonts/transform.py
jaheba: I think I understand. Since we loop over the dataset to sample enough data, this leads to problems if there are no applicable time-series.

I wonder if we can filter first and then check if any data is availabe.

vafl: Mmh, maybe. One problem is that the filtering is stochastic during training. Every time you loop over the dataset you may get a different number of results. 

We could perhaps do one loop over the dataset (without repetition) and make sure we get at least one output or something like that. But there are also edge cases her. You could have a dataset with just one time series. Because of the stochastic filtering, you may not yield anything in this test execution and so we would need more complicated logic.

szha:
Job PR-139/2 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-139/2/index.html

src/gluonts/transform.py
```diff
@@ -345,14 +346,25 @@ class FlatMapTransformation(Transformation):
...
+                raise Exception(
+                    f'Reached maximum number of idle transformation calls.\n'
+                    f'This means the transformation looped over '
+                    f'GLUONTS_MAX_IDLE_TRANSFORMS={GLUONTS_MAX_IDLE_TRANSFORMS} '
+                    f'inputs without returning any output.\n'
+                    f'The transformation this occurred in is:\n{self}'
```
aalexandrov: Grammar

> The transformation **this** occurred ...

lostella:
```suggestion
                    f'This occurred in the following transformation:\n{self}'
```

szha:
Job PR-139/3 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-139/3/index.html

jaheba:
Although I have some feelings about this, I would say we go ahead and merge this.

If we get false positives, we can still amend the behaviour.

PR #154, opened 2019-06-26
Moved requirement files to a dedicated folder.

jaheba:


szha:
Job PR-154/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-154/1/index.html

szha:
Job PR-154/2 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-154/2/index.html

dev_setup.sh
```diff
@@ -56,6 +56,6 @@ fi
 git config core.hooksPath ".devtools/githooks"
 
 # install project requirements
-python -m pip install -r requirements-setup.txt
-python -m pip install -r requirements-test.txt
-python -m pip install -r requirements.txt
+python -m pip install -r requirements/requirements-setup.txt
```
aalexandrov: Maybe use the occasion to also add:

```bash
python -m pip install -r requirements/requirements-prophet.txt
python -m pip install -r requirements/requirements-r.txt
python -m pip install -r requirements/requirements-docs.txt
``` 

to the list

jaheba: Are r and prophet really required for a dev-setup?

Say, if I want to build a new model with GluonTS, why would I need these?

aalexandrov: My rationale is that it is probably better that an execution of `dev_setup.sh` guarantees that the development environment is setup completely, as opposed to setup in a minimal state where the extras don't work. 

Alternatively we might have a question asked in the loop that guides which set of dependencies we install.

Issue #147, opened 2019-06-25
Provide SageMaker compatible docker container


jaheba:
## Description

We should think about creating a GluonTS container, which can be used in Amazon SageMaker. It could behave similar to SageMaker DeepAR w.r.t data-loading (`train` and `test` channels) and evaluation of models.

Having these containers could make it a lot easier for customers to try out GluonTS models or even use them in a production setting.

This would require two things:

- implementing a “shell” which can interacts with the SageMaker system
- building and distribution of the docker images

## References
- [SageMaker DeepAR](https://docs.aws.amazon.com/sagemaker/latest/dg/deepar.html)

jaheba:
One thing I’m unsure about is how we dispatch between different models.

Also, do we want to support training of more than one model per run?

/cc @aalexandrov @lostella

lostella:
> Also, do we want to support training of more than one model per run?

Not sure why we would want to do that: the way I see it, each job would be one training, and should produce one model artifact that should be servable

jaheba:
Although I think there is a point to be made that it could be simpler to use, SageMaker has the concept of one training-configuration is one job.

aalexandrov:
> One thing I’m unsure about is how we dispatch between different models.

The approach used in #151 is to dispatch the `Estimator` based on an environment variable `SWIST_ESTIMATOR_CLASS` (should really be `GLUONTS_ESTIMATOR_CLASS`) or, if that variable is not set, through a special `estimator_class` hyperparamter.

jaheba:
> through a special estimator_class hyperparamter

Which is not available during inference. Except, if we store it as part of the model. And that would imply that all models have to go through training. Not sure I like that.

jaheba:
Another thought:

Since we have models which don't require offline training, we can make the shell accept `Forecaster`s, where:

```py
Forecaster = Union[Predictor, Estimator]
```

That way we would skip training if a `Predictor` was specified and just start with the `test`-step.

lostella:
> > through a special estimator_class hyperparamter
> 
> Which is not available during inference. Except, if we store it as part of the model. And that would imply that all models have to go through training. Not sure I like that.

[...]

PR #156, opened 2019-06-28
Started shell refactoring.

jaheba:
Issue: #147

---

The idea of these changes is to adapt the shell so it can be used both locally *and* on SageMaker.

Example usage:

    python -m gluonts.shell train R --data-path ~/tmp/exp/optml


src/gluonts/shell/__main__.py
```diff
@@ -0,0 +1,52 @@
...
+    'deepar': 'gluonts.model.deepar.DeepAREstimator',
+    'r': 'gluonts.model.r_forecast.RForecastPredictor',
+}
+
+
+@cli.command()
+@click.option(
+    "--data-path",
+    type=click.Path(exists=True),
+    envvar="SAGEMAKER_DATA_PATH",
```
aalexandrov: If the goal is to make the shell SageMaker-agnostic, maybe the envvar should be called `GLUONTS_DATA_PATH`.

src/gluonts/shell/train.py
```diff
@@ -12,233 +12,118 @@
...
-def main() -> None:
-    try:
-        run()
-    except GluonTSException as error:
-        logging.error(error)
-        exit(-1)
+    # forecaster_path = env.hyperparameters.get("forecaster_class")
 
+    Forecaster = get_forecaster(forecaster_path)
+    forecaster = from_hyperparameters(Forecaster, **env.hyperparameters)
```
aalexandrov: `from_hyperparameters` is only defined on `Estimator` classes, but not on `Predictor`... In general I am not sure whether there is a clean solution of the `Forecaster` union type approach. This will remove the need of a need a noop `train` call but might create other problems elsewhere.

jaheba: The method `from_hyperparameters` on `Estimator` classes is just calling the function I'm using here directly. At the moment it works quite well, and you can use the `Predictor` classes directly.

src/gluonts/shell/__main__.py
jaheba: This is the variable name used by official 1P algorithms, so I wanted to be compliant here.

src/gluonts/shell/train.py
jaheba: Oh, I think I now get your point. I would say we leave it as it is for now and maybe rethink the `from_hyperparameters` approach in the future when we are more sure about things.

src/gluonts/shell/__main__.py
aalexandrov: I understand.

src/gluonts/shell/serve/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+# Third-party imports
+import flask
+from gunicorn.app.base import BaseApplication
+from pydantic import BaseSettings
+
+# First-party imports
+from gluonts.core.component import check_gpu_support
+
+from gluonts.model.predictor import Predictor
+from gluonts.shell.env import SageMakerEnv
```
aalexandrov: I cannot locate the `SageMakerEnv` definition in this diff. Shouldn't this be part of the same PR?

src/gluonts/shell/train.py
aalexandrov: We can still remove the need on a per-predictor subclass of a `DummyEstimator` by dispatching on the type of the `predictor_type` after we do `locate(predictor_class)` in the shell. If the type is a subtype of `Estimator`, we call `from_hyperparameters` directly. If it is a subtype of `Predictor`, we can directly call `DummyEstimator` with an extended `config` that also includes the `predictor_type`. I will try to make a concrete proposal in a separate PR.

jaheba: Are you fine with leaving as is for now?

src/gluonts/shell/__main__.py
```diff
@@ -0,0 +1,62 @@
...
+
+    if forecaster is not None:
+        run_dynamic(forecaster)
+    else:
+        run(data_path)
+
+
+FORECASTER_BY_NAME = {
+    'deepar': 'gluonts.model.deepar.DeepAREstimator',
+    'r': 'gluonts.model.r_forecast.RForecastPredictor',
```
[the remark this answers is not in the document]
jaheba: And all others ;)

src/gluonts/shell/serve/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+            "bind": "0.0.0.0:8080",
+            "workers": workers,
+            # "post_worker_init": ScoringService.post_worker_init,
+            "timeout": 100,
+        },
+    ).run()
+
+    # NOTE: Stop Flask application when SIGTERM is received as a result
+    # of "docker stop" command.
+    # signal.signal(signal.SIGTERM, self.stop)
```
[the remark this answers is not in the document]
jaheba: These lines were a reminder to ensure that the container stops, when CTL+C is sent.

However, this can also be achieved when `STOPSIGNAL SIGINT` is set when building the container.

src/gluonts/shell/serve/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+from gluonts.model.predictor import Predictor
+from gluonts.shell.env import SageMakerEnv
+
+from gluonts.shell.serve.util import number_of_workers
+
+
+MB = 1024 * 1024
+
+
+class Env(BaseSettings):
```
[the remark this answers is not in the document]
jaheba: Will do. Also, I rename this `env` to `settings` to not be confused with the `SageMakerEnv`

setup.py
```diff
@@ -226,9 +226,9 @@ def run(self):
         ]
     ),
     cmdclass=dict(
-        type_check=TypeCheckCommand,
-        style_check=StyleCheckCommand,
-        build_py=BuildPyCommand,
+        # type_check=TypeCheckCommand,
```
jaheba: I guess this needs to be removed.

src/gluonts/dataset/jsonl.py
```diff
@@ -17,7 +17,8 @@
 from typing import NamedTuple
 
 # Third-party imports
-import ujson as json
```
jaheba: This also.

src/gluonts/shell/__init__.py
```diff
@@ -1,17 +0,0 @@
...
-#
-# or in the "license" file accompanying this file. This file is distributed
-# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
-# express or implied. See the License for the specific language governing
-# permissions and limitations under the License.
-
-# Relative imports
-from ._base import PathsEnvironment
-
-__all__ = ['PathsEnvironment']
```
jaheba: I need to re-add the file.

But do we want to expose anything here?

src/gluonts/shell/__main__.py
```diff
@@ -0,0 +1,69 @@
...
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+import click
+
+FORECASTER_BY_NAME = {
+    'deepar': 'gluonts.model.deepar.DeepAREstimator',
```
jaheba: Should I add the remaining existing ones?

setup.py
aalexandrov: Unless the code in the current PR is affected by this, can we handle this a separate PR? I can prepare something that removes these checks from the build chain and adds them to the Jenkins pipeline instead.

src/gluonts/dataset/jsonl.py
aalexandrov: Why do we want to remove this? If I remember correctly the data with `ujson` was much faster than with the stdlib `json` module.

src/gluonts/model/forecast.py
```diff
@@ -516,7 +516,10 @@ class OutputType(str, Enum):
 
 
 class Config(pydantic.BaseModel):
-    num_eval_samples: int
+    class Config:
+        allow_population_by_alias = True
+
+    num_eval_samples: int = pydantic.Schema(..., alias='num_samples')
```
aalexandrov: Nice, I didn't know about this feature! Can we have multiple aliases (e.g., so that both `num_samples` and `num_eval samples` work correctly)?

src/gluonts/shell/__init__.py
aalexandrov: I don't think there is a need for that for now.

src/gluonts/shell/__main__.py
aalexandrov: Not for now. In a next step we could rework this by introducing some plugin or registry mechanism which allows third-party packages to register their estimators with the shell. I remember that something along those lines was possible with `setup.py`.

src/gluonts/dataset/jsonl.py
jaheba: Sorry, what I meant is that I want to use it again, as it was replaced here with normal `json`.

I used a dataset for testing, which encoded nan-values as `NaN` instead of `"NaN"`, which caused ujson to be throw errors.

src/gluonts/model/forecast.py
jaheba: That's what it's doing in this case. This requires `allow_population_by_alias` to be set as well.

src/gluonts/dataset/jsonl.py
aalexandrov: > I used a dataset for testing, which encoded nan-values as `NaN` instead of `"NaN"`, which caused ujson to be throw errors.

I think I hit that issue at once in the past as well.

szha:
Job PR-156/13 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-156/13/index.html

PR #166, opened 2019-07-02
Added process method to forecast.Config.

jaheba:

### DIFF
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -13,8 +13,8 @@
 
 # Standard library imports
 import re
-import typing
-from typing import Dict, List, NamedTuple, Optional, Union
+from enum import Enum
+from typing import Dict, List, NamedTuple, Optional, Set, Union
 
 # Third-party imports
 import mxnet as mx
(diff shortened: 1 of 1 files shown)


src/gluonts/model/forecast.py
```diff
@@ -508,42 +511,33 @@ def __repr__(self):
...
 
-    @classmethod
-    def validate(cls, v):
-        if not isinstance(v, str):
-            raise ValueError(
-                f'strict string: expected string value, got {type(v)}'
+        if OutputType.samples in self.output_types:
+            result['samples'] = (
+                forecast.samples.tolist()
+                if hasattr(forecast, 'samples')
```
lostella: I thought the `get_samples` method introduced above was meant to avoid this `if`?

jaheba: Right. So, do you prefer having a samples method on every Forecast object, or do the logic elsewhere?

lostella: I don't have strong preferences, I just wanted to make sure this was intended and the `get_samples` is not some left-over code. The code can go as-is

szha:
Job PR-166/4 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-166/4/index.html

PR #169, opened 2019-07-03
Move `forecast.Config.process` method to `Forecast` class

aalexandrov:
The solution in #166 does not allow to override the logic for a specific `Forecast` subclass. To do that, the processing logic should be attached to the `Forecast` class instead of `Config`.

The method is now also renamed to the more descriptive `as_json_dict`.

*Description of changes:*

### DIFF
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -158,6 +158,19 @@ class Forecast:
         """
         raise NotImplementedError()
 
+    def as_json_dict(self, config: 'Config') -> dict:
+        result = {}
+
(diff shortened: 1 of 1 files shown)


szha:
Job PR-169/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-169/1/index.html

jaheba:
Actually, this breaks the current behaviour. Here `samples` is not set for non-sampling forecasts, instead of setting an empty list.

aalexandrov:
I will prepare a fix.

Issue #134, opened 2019-06-20
Streamline "freq", "time_granularity" naming


mbohlkeschneider:
Find a naming that is consistent and change it throughout the code for freq/time_granularity.

geoalgo:
Given that Pandas use 'freq'  ([link](http://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html)), this is probably the best option.

mbohlkeschneider:
I agree with that. We should make this distinction clear somewhere in the documentation, though, that this is not confused with seasonality frequency.

jaheba:
I'm currently looking into the shell (essentially integrating with SageMaker) and noticed this too.

In [SageMaker DeepAR](https://docs.aws.amazon.com/sagemaker/latest/dg/deepar_hyperparameters.html) we use `time_freq` and I think there is an argument to be made that we should aim for consistency.

mbohlkeschneider:
Consistency is a good argument. But consistency with SageMaker might not necessarily be the right thing in a library. Pandas uses `freq` for generating time series Series objects and I think it would be good to stick to that.

jaheba:
I'm fine with either, but at least we should be consistent within GluonTS.

Do we have a set of required hyper-parameters? Besides `freq` `prediction_length` comes to mind.

geoalgo:
`freq` and `prediction_length` are the only two required hyper-parameters. Pandas is much more frequent than Sagemaker, so I also `freq` will be nicer.

lostella:
addressed in #172 and #173

PR #172, opened 2019-07-04
time_granularity -> freq

lostella:
*Issue #, if available:* #134 

*Description of changes:* s/time_granularity/freq/g

**NOTE:** this is a breaking change.

### DIFF
--- a/src/gluonts/dataset/artificial/_base.py
+++ b/src/gluonts/dataset/artificial/_base.py
@@ -55,8 +55,8 @@ class ArtificialDataset:
     Parent class of a dataset that can be generated from code.
     """
 
-    def __init__(self, time_granularity) -> None:
-        self.time_granularity = time_granularity
+    def __init__(self, freq) -> None:
+        self.freq = freq
 
     @property
     def metadata(self) -> MetaData:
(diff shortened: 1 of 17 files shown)


lostella:
> Can you just check that 'time_granularity' is gone from every folder?

I did. However, it looks like the host producing the docs is now failing, because in the notebook there is `"regenerate=False"`, so probably the datasets are kept on the machine and the metadata file still has `time_granularity`...

szha:
Job PR-172/2 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-172/2/index.html

src/gluonts/dataset/artificial/_base.py
```diff
@@ -84,7 +84,7 @@ def __init__(
         self,
         num_timeseries: int = 10,
         num_steps: int = 30,
-        time_granularity: str = "1H",
+        freq: str = "1H",
```
aalexandrov: This needs to be kept backwards compatible in the json files for now. Use

```
freq: str = pydantic.Schema(..., alias='time_granularity', schema: '1H')
```

in order to allow use with existing `metadata.json` files.

aalexandrov: FYI you can use `# type: ignore` to silence mypy for this line.

jaheba:
Can you check the env module?

lostella:
@jaheba I'll rebase this once #173 is merged

aalexandrov:
> @jaheba I'll rebase this once #173 is merged

@lostella let's do it the other way around.

PR #173, opened 2019-07-04
Further changes to the new `shell` package

aalexandrov:
- Use entry_points-based name -> type resolution of the forecaster type in the CLI.
- Mark the module as `.typesafe`.

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #174, opened 2019-07-04
Renamed env to sagemaker.

jaheba:


aalexandrov:
This will cause conflicts in #173. If it's just a renaming I can do that in #173 and push a revision.

PR #194, opened 2019-07-11
Fixed path requirements for shell serve.

jaheba:

When we use the dynamic forecaster, we don't require the `/opt/ml` path to exist.


szha:
Job PR-194/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-194/1/index.html

Issue #197, opened 2019-07-16
Remove warnings


aalexandrov:
## Description

Remove warnings occurring during standard build whenever this is possible.

Care has to be taken in order to preserve backwards-compatibility with the declared minimum versions of MXNet and numpy.

## References

[Sample Jenkins build](http://ci.mxnet.io/blue/organizations/jenkins/GluonTS-py3-master-cpu-unittest/detail/PR-196/1/pipeline)

lostella:
Warnings are down to 29 after #198, see [sample build](http://ci.mxnet.io/blue/organizations/jenkins/GluonTS-py3-cpu-unittest/detail/PR-198/1/pipeline)

lostella:
Relevant for the numpy's nan-related warnings: https://stackoverflow.com/a/25354869/387656

PR #201, opened 2019-07-16
Fix docstrings and numpy warnings

lostella:
*Issue #, if available:* #197 (partially)

*Description of changes:* Some warnings are raised by numpy when

1. masked arrays with masked elements are compared (e.g. using <, >=, and so on)
2. the `np.ma.masked` value is converted to `nan` with `float`

This PR addresses 1 by wrapping metrics computation around a context manager (as suggested [here](https://stackoverflow.com/a/25354869/387656)), and 2 by using an `if` instead of `float` for casting.

### DIFF
--- a/src/gluonts/evaluation/_base.py
+++ b/src/gluonts/evaluation/_base.py
@@ -55,6 +55,27 @@ def get_seasonality(freq: str) -> int:
 
 
 class Evaluator:
+    """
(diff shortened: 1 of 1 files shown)


szha:
Job PR-201/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-201/1/index.html

PR #209, opened 2019-07-19
Bump up `setuptools_scm` version

aalexandrov:
*Description of changes:*

### DIFF
--- a/requirements/requirements-setup.txt
+++ b/requirements/requirements-setup.txt
@@ -1,4 +1,4 @@
 black==19.3b0
 mypy==0.630
 setuptools>=40.1.0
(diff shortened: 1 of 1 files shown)


szha:
Job PR-209/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-209/1/index.html

jaheba:
Why is this needed?

aalexandrov:
Contains a bugfix which allows resolution of the version (can be tested with `python setup.py --version`) to work even if there is not a `.git` folder or the client git repository has not pulled the tags from the origin.

PR #215, opened 2019-07-22
Exclude MXNet 1.5.* from allowed requirements

aalexandrov:
Seems that GluonTS has an issue with the new MXNet version.

*Description of changes:*

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -1,7 +1,7 @@
 boto3==1.*
 holidays==0.9.*
 matplotlib==3.*
-mxnet>=1.3.1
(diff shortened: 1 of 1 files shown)


lostella:
MXNet 1.5.0 is not out yet, so I guess the problem is with installing pre-relases? That should not be recommended anyway.

aalexandrov:
> MXNet 1.5.0 is not out yet, [...]

[PyPi seems to disagree on that.](https://pypi.org/project/mxnet/#history)

lostella:
@aalexandrov true, I was looking at MXNet’s homepage, which is out of date by almost a couple of months it seems

Issue #218, opened 2019-07-22
SageMaker dynamic code


jaheba:
With `shell` we already provide a mechanism to run GluonTS on SageMaker.

However, this requires that a docker container is uploaded to ECR each time code has changed. For developing purposes it would be nice to just run algorithms hosted on S3.

One can then sync their current workspace to s3 and then start a new job.

PR #221, opened 2019-07-23
Started work on algorithm channel.

jaheba:

**WIP**

---
Addresses: #218 

The idea is that customers can add a new data channel named `algorithm`.

In it, there is a `algorithm.json` file which can be used to setup a new algorithm on the fly.

### DIFF
--- a/src/gluonts/shell/__main__.py
+++ b/src/gluonts/shell/__main__.py
@@ -144,18 +144,16 @@ def train_command(data_path: str, forecaster: Optional[str]) -> None:
 
     env = TrainEnv(Path(data_path))
 
-    if forecaster is None:
-        try:
-            forecaster = env.hyperparameters["forecaster_name"]
-        except KeyError:
-            msg = (
-                "Forecaster shell parameter is `None`, but "
-                "the `forecaster_name` key is not defined in the "
-                "hyperparameters.json dictionary."
-            )
-            raise GluonTSForecasterNotFoundError(msg)
+    forecaster = forecaster if forecaster is not None else env.forecaster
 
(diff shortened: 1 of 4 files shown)


szha:
Job PR-221/2 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-221/2/index.html

aalexandrov:
I like the general idea, but I am not sure whether we should allow the execution of arbitrary bash commands through our codebase. 

A bit more robust is to have an explicit `entrypoint.sh` that 

- checks whether `/opt/ml/input/data/algorithm/setup.py` exists,
- if the above is true, runs `pip install .` from the enclosing folder, and
- delegates execution to `python -m gluonts.shell`.

Then we don't need a lot of extra Python code (clients can configure an `algorithm` channel and pass the `forecaster` value as part of the `hyperparameters.json`).

jaheba:
Once thing I thought about is to add `algorithm` to ``sys.path``. That way, if you have a simple setup (like `my_algo.py`), you could just add `my_algo.MyEstimator` as the forecaster and things should just work.

I think if you actually want to add more complicated dependencies, you would create a new base-image first (e.g. if you were to add the R-forecaster that way).

aalexandrov:
> Once thing I thought about is to add algorithm to `sys.path`.

I like that idea.

szha:
Job PR-221/3 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-221/3/index.html

PR #229, opened 2019-08-06
Bring deep state to gluon-ts

rshyamsundar:
Implementation for the paper: https://papers.nips.cc/paper/8004-deep-state-space-models-for-time-series-forecasting

### DIFF
--- a/src/gluonts/distribution/lds.py
+++ b/src/gluonts/distribution/lds.py
@@ -316,6 +316,10 @@ class LDS(Distribution):
 
         # Sample the prior state.
         # samples_lat_state: (num_samples, batch_size, latent_dim, 1)
+        # The prior covariance is observed to be slightly negative definite whenever there is
+        # excessive zero padding at the beginning of the time series.
+        # We add positive tolerance to the diagonal to avoid numerical issues.
+        # Note that `jitter_cholesky` adds positive tolerance only if the decomposition without jitter fails.
         state = MultivariateGaussian(
             self.prior_mean,
             jitter_cholesky(
--- a/src/gluonts/distribution/lds.py
+++ b/src/gluonts/distribution/lds.py
@@ -462,8 +462,8 @@ class LDSArgsProj(mx.gluon.HybridBlock):
     def __init__(
         self,
         output_dim: int,
-        noise_std_ub: float = 20,
-        innovation_ub: float = 50,
+        noise_std_ub: float = 1.0,
+        innovation_ub: float = 0.01,
     ) -> None:
         super().__init__()
         self.output_dim = output_dim
--- a/src/gluonts/distribution/lds.py
+++ b/src/gluonts/distribution/lds.py
@@ -14,11 +14,13 @@
 from typing import Tuple, Optional
 
 import mxnet as mx
+import numpy as np
 
 from gluonts.distribution import Distribution, Gaussian, MultivariateGaussian
 from gluonts.distribution.distribution import getF
 from gluonts.model.common import Tensor
 from gluonts.support.util import make_nd_diag, _broadcast_param
+from gluonts.support.linalg_util import jitter_cholesky
 
 
 class LDS(Distribution):
@@ -315,7 +317,10 @@ class LDS(Distribution):
         # Sample the prior state.
         # samples_lat_state: (num_samples, batch_size, latent_dim, 1)
         state = MultivariateGaussian(
-            self.prior_mean, F.linalg_potrf(self.prior_cov)
+            self.prior_mean,
+            jitter_cholesky(
+                F, self.prior_cov, self.latent_dim, float_type=np.float32
+            ),
         )
         samples_lat_state = state.sample(num_samples).expand_dims(axis=-1)
 
--- a/src/gluonts/model/deepstate/_estimator.py
+++ b/src/gluonts/model/deepstate/_estimator.py
@@ -1,0 +1,261 @@
+# Standard library imports
+from typing import List, Optional
+
+# Third-party imports
+from mxnet.gluon import HybridBlock
+
+# First-party imports
+from gluonts.core.component import validated
+from gluonts.model.deepstate.issm import ISSM, CompositeISSM
+from gluonts.time_feature.lag import (
+    time_features_from_frequency_str,
+    longest_period_from_frequency_str,
+)
+from gluonts.model.estimator import GluonEstimator
+from gluonts.model.predictor import Predictor, RepresentableBlockPredictor
+from gluonts.support.util import copy_parameters
+from gluonts.time_feature.lag import (
+    TimeFeature,
+    time_features_from_frequency_str,
+)
+from gluonts.trainer import Trainer
+from gluonts.transform import (
+    FieldName,
+    AddObservedValuesIndicator,
+    AddTimeFeatures,
+    AsNumpyArray,
+    Chain,
+    ExpandDimArray,
+    ExpectedNumInstanceSampler,
+    TestSplitSampler,
+    CanonicalInstanceSplitter,
+    Transformation,
+    SetFieldIfNotPresent,
+)
+
+# Relative imports
+from ._network import DeepStatePredictionNetwork, DeepStateTrainingNetwork
+
+SEASON_INDICATORS_FIELD = "seasonal_indicators"
+
+
+class DeepStateEstimator(GluonEstimator):
+    """
+    Construct a DeepState estimator.
+    
+    This implements the deep state space model described in
+    [Rangapuram et. al. 2018]
+
+    .. [Rangapuram et. al. 2018] Syama S. Rangapuram, Matthias W. Seeger,
+    Jan Gasthaus, Lorenzo Stella, Yuyang Wang, Tim Januschowski.
+    Deep State Space Models for Time Series Forecasting.
+    In NeurIPS 31, pages 7785–7794, 2018.
+    https://papers.nips.cc/paper/8004-deep-state-space-models-for-time-series-forecasting
+
+    Parameters
+    ----------
+    freq
+        Frequency of the data to train on and predict
+    prediction_length
+        Length of the prediction horizon
+    add_trend
+        Flag to indicate whether to include trend component in the
+        state space model
+    past_length
+        Number of steps to unroll the RNN for before computing predictions.
+        Set this to (at most) the length of the shortest time series in the dataset.
+    trainer
+        Trainer object to be used (default: Trainer())
+    num_layers
+        Number of RNN layers (default: 2)
+    num_cells
+        Number of RNN cells for each layer (default: 40)
+    cell_type
+        Type of recurrent cells to use (available: 'lstm' or 'gru';
+        default: 'lstm')
+    num_eval_samples
+        Number of samples paths to draw when computing predictions
+        (default: 100)
+    dropout_rate
+        Dropout regularization parameter (default: 0.1)
+    cardinality
+        Number of values of the each categorical feature (default: [1])
+    embedding_dimension
+        Dimension of the embeddings for categorical features (the same
+        dimension is used for all embeddings, default: 5)
+    scaling
+        Whether to automatically scale the target values (default: true)
+    lags_seq
+        Indices of the lagged target values to use as inputs of the RNN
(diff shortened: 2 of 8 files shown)


szha:
Job PR-229/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-229/1/index.html

lostella:
Looks like the following files are missing the required header:
```
File src/gluonts/model/deepstate/_estimator.py does not have correct header
File src/gluonts/model/deepstate/_network.py does not have correct header
File src/gluonts/model/deepstate/issm.py does not have correct header
File src/gluonts/model/deepstate/__init__.py does not have correct header
```

See the header here: https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/transform.py

src/gluonts/time_feature/lag.py
```diff
@@ -51,6 +51,28 @@ def get_granularity(freq_str: str) -> Tuple[int, str]:
...
+    "B": 365,  # yearly seasonality
+    "H": 168,  # weekly seasonality
+    "T": 1440,  # daily seasonality
+    "min": 1440,  # daily seasonality
+}
+
+
+def longest_period_from_frequency_str(freq_str: str) -> int:
+    multiple, granularity = get_granularity(freq_str)
+    return FREQ_LONGEST_PERIOD_DICT[granularity] // multiple
```
lostella: This may be minor, but: this line will round down the period length. For example, with `freq_str="3W"` it will yield `52//3` which is 17, but to cover a year one would need to use 18.

src/gluonts/time_feature/lag.py
```diff
@@ -51,6 +51,28 @@ def get_granularity(freq_str: str) -> Tuple[int, str]:
...
+
```
lostella: Also: similar code to the one here appears in the `evaluation` package, [see here](https://github.com/awslabs/gluon-ts/blob/d6bc5468c28fa665b2d3d0e57e1a62221231f481/src/gluonts/evaluation/_base.py#L30-L54).

These two are of course different: in the `evaluation` package, for example, a seasonality of length 1 is used for `"D"` or `"W"`, and a seasonality of length 24 is used for `"H"`.

I'm not sure why that is the case, but I think that having similar pieces of code lying around can create confusion: I suggest keeping these definition here local to the `model.deepstate` package, since this is only used there. Then we will see about consolidating these.

szha:
Job PR-229/2 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-229/2/index.html

src/gluonts/time_feature/lag.py
lostella: I mean, if a model looks over multiple periods, then this may be almost irrelevant; it's more relevant if one looks at a single period of course.

rshyamsundar: yes, year has slightly more than 52 weeks. Since this is minor (only used to determine the training length) I went with 52 which is also the recommended frequency value in R forecast package.

src/gluonts/model/deepstate/_estimator.py
```diff
@@ -0,0 +1,312 @@
...
+        Flag to indicate whether to include trend component in the
+        state space model
+    past_length
+        This is the length of the training time series;
+        i.e., number of steps to unroll the RNN for before computing predictions.
+        Set this to (at most) the length of the shortest time series in the dataset.
+        (default: None, in which case the training length is set such that at least
+        `num_seasons_to_train` seasons are included in the training.
+        See `num_seasons_to_train`)
+    num_seasons_to_train
```
lostella: Maybe a better name for this is `num_periods_to_train`? In my idea of naming things, a period is a full cycle of seasons

src/gluonts/time_feature/lag.py
lostella: I think 52 is fine, I was referring to `52//3` rounding down to something shorter than 52 weeks.

But again, I think this is minor, feel free to resolve this conversation.

src/gluonts/model/deepstate/_estimator.py
```diff
@@ -0,0 +1,312 @@
...
+
+SEASON_INDICATORS_FIELD = "seasonal_indicators"
+
+
+class DeepStateEstimator(GluonEstimator):
+    """
+    Construct a DeepState estimator.
+    
+    This implements the deep state space model described in
+    [Rangapuram et. al. 2018]
```
benidis: et al.

src/gluonts/model/deepstate/_estimator.py
```diff
@@ -0,0 +1,312 @@
...
+
+    .. [Rangapuram et. al. 2018] Syama S. Rangapuram, Matthias W. Seeger,
```
benidis: et al.

src/gluonts/model/deepstate/_estimator.py
```diff
@@ -0,0 +1,312 @@
...
+
+    @validated()
+    def __init__(
+        self,
+        freq: str,
+        prediction_length: int,
+        add_trend: bool = False,
+        past_length: Optional[int] = None,
+        num_seasons_to_train: int = 4,
+        trainer: Trainer = Trainer(epochs=25, hybridize=False),
```
benidis: Is there a reason to not use the default `Trainer()` like in the other models?

src/gluonts/model/deepstate/_estimator.py
```diff
@@ -0,0 +1,312 @@
...
+        ), "The value of `prediction_length` should be > 0"
+        assert (
+            past_length is None or past_length > 0
+        ), "The value of `past_length` should be > 0"
+        assert num_layers > 0, "The value of `num_layers` should be > 0"
+        assert num_cells > 0, "The value of `num_cells` should be > 0"
+        assert (
+            num_eval_samples > 0
+        ), "The value of `num_eval_samples` should be > 0"
+        assert dropout_rate > 0, "The value of `dropout_rate` should be > 0"
```
benidis: I think you should include equality here, i.e., `dropout_rate >= 0`

src/gluonts/model/deepstate/issm.py
```diff
@@ -0,0 +1,323 @@
...
+        block_diagonal = _make_2_block_diagonal(
+            F=F, left=block_diagonal, right=blocks[i]
+        )
+
+    return block_diagonal
+
+
+def _make_2_block_diagonal(F, left: Tensor, right: Tensor) -> Tensor:
+    """
+    Creates a block diagonal matrix of shape (batch_size, m+n, m+n) where m and n are the sizes of
```
benidis: fix documentation style

src/gluonts/model/deepstate/issm.py
```diff
@@ -0,0 +1,323 @@
...
+    _block_diagonal = F.concat(
+        F.concat(left, zeros_off_diag, dim=3),
+        F.concat(zeros_off_diag_tr, right, dim=3),
+        dim=2,
+    )
+
+    return _block_diagonal
+
+
+class ISSM:
```
benidis: would be nice to have some documentation here describing what the class does.

src/gluonts/distribution/lds.py
```diff
@@ -315,7 +317,10 @@ def sample(
         # Sample the prior state.
         # samples_lat_state: (num_samples, batch_size, latent_dim, 1)
         state = MultivariateGaussian(
-            self.prior_mean, F.linalg_potrf(self.prior_cov)
+            self.prior_mean,
+            jitter_cholesky(
+                F, self.prior_cov, self.latent_dim, float_type=np.float32
+            ),
```
lostella: Could this be made optional? Like having `sampling_jitter=False` in the constructor, and set `sampling_jitter=True` when constructing the distribution in `DeepStateEstimator`?

src/gluonts/model/deepstate/_estimator.py
```diff
@@ -0,0 +1,312 @@
...
+        use_feat_static_cat: bool = False,
+        cardinality: Optional[List[int]] = None,
+        embedding_dimension: int = 20,
+        issm: Optional[ISSM] = None,
+        scaling: bool = True,
+        time_features: Optional[List[TimeFeature]] = None,
+    ) -> None:
+        super().__init__(trainer=trainer)
+
+        assert (
```
jaheba: You can use `prediction_length: pydantic.PositiveInt` and then remove the assert statements.

src/gluonts/model/deepstate/_estimator.py
```diff
@@ -0,0 +1,312 @@
...
+            prediction_length > 0
+        ), "The value of `prediction_length` should be > 0"
+        assert (
+            past_length is None or past_length > 0
```
jaheba: `past_length: pydantic.PositiveInt`

src/gluonts/model/deepstate/_estimator.py
rshyamsundar: yes..

* **epochs:** The total number of effective epochs by default is: `epochs * num_batches_per_epoch = 2500`. This is too much and wasteful for deep state model given that the structure of the time series is hard-coded (via SSM) and deep state is basically learning the parameters of this structure.
 
* **hybridize:** To avoid numerical issues when there is excess padding, I am using a jittering  method which does not work in symbolic mode.

src/gluonts/distribution/lds.py
rshyamsundar: Ah.. forgot to add a comment about this. The `jitter_cholesky` method applies jitter only if the Cholesky decomposition fails for the given matrix and returns normal result without jitter if there is no problem. So this is essentially having the jitter only for problematic cases (btw, this happened only for the constant dataset in the evaluations).

I think `jitter_cholesky` was written this way to avoid having further option/flag like `sampling_jitter` on the caller side.

src/gluonts/model/deepstate/issm.py
rshyamsundar: Thanks

src/gluonts/model/deepstate/_estimator.py
rshyamsundar: Right. Thanks.

src/gluonts/model/deepstate/_estimator.py
rshyamsundar: I think we had a problem with this earlier so we used asserts in other estimators as well. So to be consistent with other estimators I will leave it like this for now.

src/gluonts/model/deepstate/issm.py
rshyamsundar: Added, thanks.

src/gluonts/model/deepstate/_estimator.py
jaheba: Right, I think we should address this at a different time.

Do you know what the problems where?

src/gluonts/distribution/lds.py
dcmaddix: Yes exactly. It only applies the jitter when it is needed and otherwise does the standard Cholesky decomposition.

szha:
Job PR-229/6 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-229/6/index.html

lixixibj:
> Job PR-229/1 is complete.
> Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-229/1/index.html

Hi, the URL of the docs cannot be accessed. Please give me a valid URL.

lixixibj:
Hi, in the paper deep state space, the emission_coeff transition_coeff is time-varying, however, in the code in gluon-ts, these coefficients seem to be time-invariant and fixed. I don't know which actually is true. Could you please update the code of the paper?

rshyamsundar:
For seasonal models the emission coefficient is time/season dependent and other are constants because of the way we implement the innovation state space models.

Check the Appendix A.1 in the [long version](https://papers.nips.cc/paper/8004-deep-state-space-models-for-time-series-forecasting-supplemental.zip) to see the specific instantiations (that we used in the experiments) of the more general state space model discussed in the paper.

PR #232, opened 2019-08-08
Fix num_eval_samples

benidis:
With this PR I am trying to clear the ambiguity around `num_eval_samples`.

Currently, `num_eval_samples` can be defined in two places: In the estimator and in the `make_evaluation_predictions`. This is confusing to an external user since it is not clear what is the difference between these two, if one overwrites the other, or if they have a different usage.

The actual place that defines how many samples we get for prediction is `num_eval_samples` in the `make_evaluation_predictions`. We introduce `num_eval_samples` in an estimator just to increase the output dimension of the prediction network for computational efficiency. This suggests that we should distinguish between these two `num_eval_samples` hyperparameters. It also suggests that `num_eval_samples` in an estimator should be a bit "hidden" or something that the user shouldn't really try to define (unless he knows what he is doing).

For this, I changed `num_eval_samples` to `_num_eval_samples_per_ts` in all models. 

As a side, I fixed some documentation too.

### DIFF
--- a/src/gluonts/model/deepar/_estimator.py
+++ b/src/gluonts/model/deepar/_estimator.py
@@ -80,9 +80,6 @@ class DeepAREstimator(GluonEstimator):
     cell_type
         Type of recurrent cells to use (available: 'lstm' or 'gru';
         default: 'lstm')
-    num_eval_samples
-        Number of samples paths to draw when computing predictions
-        (default: 100)
     dropout_rate
         Dropout regularization parameter (default: 0.1)
     use_feat_dynamic_real
@@ -109,6 +106,9 @@ class DeepAREstimator(GluonEstimator):
     time_features
         Time features to use as inputs of the RNN (default: None, in which
         case these are automatically determined based on freq)
+    _num_eval_samples_per_ts
+        Number of evaluation samples per time series to increase parallelism during inference
+        (default: 100)
     """
 
     @validated()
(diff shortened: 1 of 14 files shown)


szha:
Job PR-232/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-232/1/index.html

jaheba:
I'm actually not sure using a variable with a leading `_` works with validated.

Pydantic ignores those variables, and thus I think the info will not be stored as part of the model.

src/gluonts/model/deepar/_estimator.py
```diff
@@ -109,6 +106,9 @@ class DeepAREstimator(GluonEstimator):
     time_features
         Time features to use as inputs of the RNN (default: None, in which
         case these are automatically determined based on freq)
+    _num_eval_samples_per_ts
```
vafl: Looks good overall, thanks!

Perhaps a different name like `num_parallel_samples` is clearer? If the leading `_` doesn't work we can comment that this is just an optimization that does not influence the accuracy.

szha:
Job PR-232/2 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-232/2/index.html

PR #235, opened 2019-08-12
Fixed backtest test.

jaheba:
The test implied that we wanted to test the output of `backtest_metrics`. However, we relied on an already checked-in file, which was never changed.

Also, I've removed the `__str__` method on `DatasetStatistics` since we otherwise can't evaluate the output.


szha:
Job PR-235/3 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-235/3/index.html

PR #236, opened 2019-08-12
fix #234: Added method to fixup non json-spec compliant floats to make the resp…

### LINKED ISSUE #234: Produce Json-compliant results in serve (Nan)
When the forecast contain NaN values, Python's `json` library encodes this as `Nan` and not as a the string `"Nan"`. This is not compliant to the json spec and some parsers don't know how to handle these values.

dotgc:
Added method to fixup non json-spec compliant floats to make the response json compliant

*Issue #, if available:*
https://github.com/awslabs/gluon-ts/issues/234

*Description of changes:*
Calls a method to sanitize and fix non json-spec compliant floats before returning the response from `/invocations` endpoint

### DIFF
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -188,32 +188,7 @@ class Forecast:
         if OutputType.samples in config.output_types:
             result["samples"] = []
 
-        return Forecast.jsonify_floats(result)
-
-    @staticmethod
-    def jsonify_floats(json_object):
-        """
-        Traverses through the JSON object and converts non JSON-spec compliant
-        floats(nan, -inf, inf) to their string representations.
-
-        Parameters
-        ----------
-        json_object
-            JSON object
-        """
-        if isinstance(json_object, dict):
-            return {k: Forecast.jsonify_floats(v) for k, v in json_object.items()}
-        elif isinstance(json_object, list):
-            return [Forecast.jsonify_floats(item) for item in json_object]
-        elif isinstance(json_object, float):
-            if np.isnan(json_object):
-                return "NaN"
-            elif np.isposinf(json_object):
-                return "Infinity"
-            elif np.isneginf(json_object):
-                return "-Infinity"
-            return json_object
-        return json_object
+        return result
 
 
 class SampleForecast(Forecast):
@@ -343,7 +318,7 @@ class SampleForecast(Forecast):
         result = super().as_json_dict(config)
 
         if OutputType.samples in config.output_types:
-            result["samples"] = Forecast.jsonify_floats(self.samples.tolist())
+            result["samples"] = self.samples.tolist()
 
         return result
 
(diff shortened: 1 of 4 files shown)


src/gluonts/model/forecast.py
```diff
@@ -318,7 +343,7 @@ def as_json_dict(self, config: "Config") -> dict:
         result = super().as_json_dict(config)
 
         if OutputType.samples in config.output_types:
-            result["samples"] = self.samples.tolist()
+            result["samples"] = Forecast.jsonify_floats(self.samples.tolist())
```
jaheba: I'm not sure what the right place for this call is. Maybe we want to add it to the `shell` module instead to ensure that it is always called on the result?

src/gluonts/model/forecast.py
```diff
@@ -188,7 +188,32 @@ def as_json_dict(self, config: "Config") -> dict:
         if OutputType.samples in config.output_types:
             result["samples"] = []
 
-        return result
+        return Forecast.jsonify_floats(result)
+
+    @staticmethod
+    def jsonify_floats(json_object):
```
jaheba: I think this is not the right place for the method. Maybe, we could add it to some `util` module. And maybe it's also better to move it to `shell`.

test/model/test_forecast.py
```diff
@@ -56,3 +57,26 @@ def percentile(value):
     assert forecast.prediction_length == 1
     assert len(forecast.index) == pred_length
     assert forecast.index[0] == pd.Timestamp(START_DATE)
+
+
+JSON_OBJECT_WITH_NON_COMPLIANT_FLOATS = {
```
jaheba: Maybe move this inside the test-method.

szha:
Job PR-236/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-236/1/index.html

src/gluonts/model/forecast.py
dotgc: My initial thought was that `as_json_dict` method - given the method name, should be outputting well formed JSON and hence added it there, but it makes more sense to do it right before sending the response.

src/gluonts/model/forecast.py
dotgc: Done.

test/model/test_forecast.py
```diff
@@ -17,7 +17,7 @@
 import pytest
 
 # First-party imports
-from gluonts.model.forecast import QuantileForecast, SampleForecast
+from gluonts.model.forecast import Forecast, QuantileForecast, SampleForecast
```
jaheba:
```suggestion
from gluonts.model.forecast import QuantileForecast, SampleForecast
```

This is not needed anymore.

test/model/test_forecast.py
dotgc: Done.

test/model/test_forecast.py
dotgc: Yep. Removed.

szha:
Job PR-236/2 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-236/2/index.html

PR #240, opened 2019-08-13
Added `predict_item` to RepresentablePredictor and adapted subclasses.

jaheba:
Addresses #238.

Also added a way to create a MeanPredictor based on an existing Predictor.

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #238, opened 2019-08-13
API: Single Item Prediction


jaheba:
## Background
Prediction happens on an item to item bases. That means, during prediction all items are independent and thus we also forecast each item individually.

However, in some algorithms (NN based) we group items together during prediction for performance reasons.

Our `predict`-API always takes a whole `Dataset` and not an individual item.

This is a problem, since we loose a one-to-one relationship between input and output time-series.

## My Problem

I'm currently thinking about a way to incorporate "fallback"-forecasts, where a model isn't able to produce a result and thus want to use another method to fall back on. Our current API design makes it difficult to handle the items individually, since I always have to handle a whole set of items.

What I want is to have something like this:

```python

predictor = FallbackPredictor(
    PredictorA(...), # can't handle NaN values
    PredictorB(...)
)
```

Essentially, we would try `A` first on each item and fallback on `B` if no result could be produced on the first try.

## Possible Solutions

At the moment I see two ways to solve this:

 1.  Split predictors in single-item and multi-item predictors with different behaviour and capabilities.

 2. Require that all steps need to be capable of handling multiple items at once.

The obvious drawback of 1. being that not all predictors can be treated the same, and the drawback of 2. being the added complexity and hacks to make it work.

vafl:
Yes. Essentially, the batching is a performance optimization (for throughput).
The main benefit is for neural network models which can parallelize the batch. 

For predictions behind an endpoint we usually use mxnet's naive engine, which does not parallelize. I don't know how much speedup the batching gives in that case.

We could measure this and if there is no big speedup we could switch to the single item predict method. Alternatively, could we maybe have an additional single item predict method or pass the batch size as a parameter when calling predict?

lostella:
I think I like the idea (naming is arbitrary) of a `predict_item` method implemented in specific predictors, and a default

```
def predict(self, dataset: Dataset) -> Generator[Forecast]:
    for item in dataset:
        yield self.predict_item(item)
```

defined in `Predictor`.

How to do this in the `GluonPredictor` though? There could have a `predict_batch` instead, a different implementation of `predict` that uses the `InferenceDataLoader` and calls `predict_batch`, and a `predict_item` which maybe wraps the single item in a `Dataset` and calls `predict`?

lostella:
I think #240 only partially solved this, and this API should be enforced (and tested) on all predictor types. Also, the naming there is not ideal in my opinion: if the method predicts a single `DataEntry` then it should probably be called `predict_entry`.

PR #241, opened 2019-08-13
Fallback predictor

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/model/predictor.py
+++ b/src/gluonts/model/predictor.py
@@ -20,7 +20,15 @@ import traceback
 from pathlib import Path
 from pydoc import locate
 from tempfile import TemporaryDirectory
-from typing import TYPE_CHECKING, Callable, Dict, Iterator, List, Optional
+from typing import (
+    TYPE_CHECKING,
+    Callable,
+    Dict,
+    Iterator,
+    List,
+    Optional,
+    Type,
+)
 
 # Third-party imports
 import mxnet as mx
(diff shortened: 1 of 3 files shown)


szha:
Job PR-241/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-241/1/index.html

src/gluonts/model/predictor.py
```diff
@@ -729,3 +737,26 @@ def predict(self, dataset: Dataset, **kwargs) -> Iterator[Forecast]:
...
+            except Exception as error:
+                logging.warning(
+                    f"Base predictor didn't return result: {error}"
+                )
+                fallback_predictor = fallback_cls.from_predictor(self)
+                return fallback_predictor.predict_item(item)
+
+        return fallback_predict
+
+    return decorator
```
lostella: What’s the connection between this (`FallbackPredictor`, `fallback`) and the mechanism added to `MeanPredictor` in #240? It’s a bit involved and not apparent, maybe an example would clarify

jaheba: Here is a small example:

```python
class MyPredictor(RepresentablePredictor):
    ....
    @fallback(MeanPredictor)
    def predict_item(item):
        ...
```

The idea is that `FallbackPredictor`s know to be constructed from other predictors. Essentially, this requires that they only have to required arguments `freq` and `predictions_length`, with all others being optional.

Granted, it's a bit ad-hoc at the moment, but I think I like the overall design approach.

lostella: I see. But this way the fallback mechanism is hardcoded in the `MyPredictor` type, isn't it? Correct me if I'm wrong: one can also do
```
base_predictor = WhateverPredictor(...)
predictor = MeanPredictor.from_predictor(base_predictor)
```
or something similar? Because that seems more like the example you mentioned in #238.

I mean, ultimately just wrapping a sequence of predictors in a `FallbackPredictor` with its own `predict_item` that applies base predictors in sequence until one succeeds seems simple and flexible (and exactly what you suggest in #238). But maybe this is what's happening here under the decorators magic :-)

PR #243, opened 2019-08-14
GluonTSException bypass fallback.

jaheba:

---

Not sure if catching all GluonTS errors is the right approach. However, one could argue that throwing a GluonTS error is deliberate and thus should be signaled to the user.


szha:
Job PR-243/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-243/1/index.html

Issue #131, opened 2019-06-19
Reduce the Conda dependencies listed in `env`


aalexandrov:
Packages that are duplicated between a `requirements*.txt` and a `*.yml` file in `env` should be removed from the `*.yml` file.

AaronSpieler:
The problem here is that Anaconda is not able to resolve the dependencies correctly if they are installed through pip, so we had to add some dependencies back.

See issue: https://github.com/awslabs/gluon-ts/pull/661
And issue: https://github.com/awslabs/gluon-ts/pull/658

And finally a potential solution: https://github.com/awslabs/gluon-ts/issues/662

PR #245, opened 2019-08-14
mxnet 1.5.0 compatibility

lostella:
*Description of changes:* MXNet requirement was set to `<1.5.*` in #215 due to some incompatibility with `mxnet-1.5.0`. By running tests, it seems like the only problem was in the code for WaveNet. This PR updates the problematic lines to make our WaveNet implementation compatible with `mxnet-1.5.0`.


lostella:
Errata: it looks like there's more issues than the one I addressed in the initial commit (despite the fact all tests are passing locally :-/). Diving deeper.

lostella:
I'm not sure what's running in the tests: [these logs](http://ci.mxnet.io/blue/rest/organizations/jenkins/pipelines/GluonTS-py3-master-cpu-unittest/branches/PR-245/runs/1/nodes/81/steps/126/log/?start=0) mention at some point

```
# packages in environment at /var/lib/jenkins/workspace/workspace/gluon-ts-cpu-py3/conda/cpu/py3-master:
#
# Name                    Version                   Build  Channel
[...]
mxnet-mkl                 1.4.1                     <pip>
[...]
numpy                     1.14.6                    <pip>
```

But if I locally install it the same way as in CI (using `pip install -v -e ".[shell,Prophet]"`, see [here](https://github.com/lostella/gluon-ts/blob/mxnet-1.5.0/ci/prepare_clean_env.sh)) then I get `mxnet-1.5.0` and `numpy-1.17.0`.

Could it be some caching messing up with this? @szha

lostella:
Nevermind: I think this problem is caused by #131

lostella:
There seem to be an issue with sampling on mxnet-mkl, see here: http://ci.mxnet.io/blue/rest/organizations/jenkins/pipelines/GluonTS-py3-cpu-unittest/branches/PR-245/runs/7/nodes/81/steps/126/log/?start=0

szha:
@reminisce

lostella:
Related mxnet issue: https://github.com/apache/incubator-mxnet/issues/16135

strawberrypie:
The fix for the issue mentioned above is merged into mxnet: https://github.com/apache/incubator-mxnet/pull/16139
Sadly, it was done after the release of 1.5.1: https://github.com/apache/incubator-mxnet/releases/tag/1.5.1
What is the plan moving forward regarding mxnet version? Updating to mxnet 1.5 will allow the project to remove `Trainer` class and replace it with `gluon.contrib.Estimator` — which supports things like early stopping.

lostella:
@strawberrypie good point, that’s something which is worth looking into. Although I think that’s kind of an experimental API of MXNet’s? Another thing that we’ll look into is the numpy-compatible arrays API, see https://github.com/apache/incubator-mxnet/issues/14253

lostella:
Btw I think this PR can be closed in favor of a new one whenever the mxnet version can be bumped

PR #259, opened 2019-08-27
deep factor [WIP]

lovvge:
*Issue #, if available:* 219

*Description of changes:*

Initial commits for the deep factor, still WIP. One of the blockers to make it working as desired is the inefficiency of the data loader that prevents it running for a large number of iterations. 

### DIFF
--- a/src/gluonts/model/deep_factor/_estimator.py
+++ b/src/gluonts/model/deep_factor/_estimator.py
@@ -1,0 +1,181 @@
+# Standard library imports
+from typing import List
+
+# First-party imports
+from dataset.repository.datasets import get_dataset
+from evaluation import Evaluator
+from gluonts import transform
+from gluonts.block.feature import FeatureEmbedder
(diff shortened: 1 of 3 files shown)


src/gluonts/model/deep_factor/_estimator.py
```diff
@@ -0,0 +1,181 @@
+# Standard library imports
+from typing import List
+
+# First-party imports
+from dataset.repository.datasets import get_dataset
```
lostella:
```suggestion
from gluonts.dataset.repository.datasets import get_dataset
```

src/gluonts/model/deep_factor/_estimator.py
```diff
@@ -0,0 +1,181 @@
...
+from evaluation import Evaluator
```
lostella:
```suggestion
from gluonts.evaluation import Evaluator
```

Issue #283, opened 2019-09-05
The calculation of dataset statistics reads wrong feature fields


benidis:
Given a dataset `ds` that includes features, `ds.calc_stats()` searches for `cat` and `dynamic_feat` to compute feature statistics, which are outdated fields. This always givas as ouput `cats=[]` and `num_dymanic_feat=0`.

This issue appeared in #280 as a side problem.

benidis:
Fixed in #292 .

PR #292, opened 2019-09-06
fixes fieldname issues

benidis:
The field names of a dataset are not checked for consistency across the different modules. If a field changes in the `FieldName` class, there is no error raised, the old fields are never there in the new datasets and therefore ignored, and the new fields are not processed.

This caused issue #283 in the dataset statistics calculation `dataset.calc_stats()`.

This PR includes the following changes:
- fixes the `calc_stats` issue (fixes issue #283 )
- moves the `FieldName` class to a separate module to avoid circular dependencies and to increase visibility for potential future changes (it was in the `transform` package with all the transformations and it did not really belong there)
- adds tests for potential change on the `FieldName` class so people are warned at least that they should not touch it without a larger refactoring
- adds tests for the `calc_stats` method

### DIFF
--- a/docs/examples/basic_forecasting_tutorial/tutorial.md
+++ b/docs/examples/basic_forecasting_tutorial/tutorial.md
@@ -351,7 +351,8 @@ from gluonts.model.estimator import GluonEstimator
 from gluonts.model.predictor import Predictor, RepresentableBlockPredictor
 from gluonts.core.component import validated
 from gluonts.support.util import copy_parameters
-from gluonts.transform import ExpectedNumInstanceSampler, Transformation, InstanceSplitter, FieldName
+from gluonts.transform import ExpectedNumInstanceSampler, Transformation, InstanceSplitter
+from gluonts.dataset.field_names import FieldName
 from mxnet.gluon import HybridBlock
 ```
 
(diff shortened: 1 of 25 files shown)


szha:
Job PR-292/3 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-292/3/index.html

lostella:
Could you point to the changes in the tests that would catch #283 without this fix? I assume these are in https://github.com/benidis/gluon-ts/blob/af9c7386a4ed1eaf66c44aa04a0b3f7c77f177f9/test/dataset/test_stat.py

lostella:
Looks good, thanks!

benidis:
@lostella: The issue with #283 was that the fields in `calc_stats` were hard coded so when we changed them  in the `FieldName` class the function kept the old ones. It was not possible to use `FieldName.{field}` in `calc_stat` due to circular dependency. The fix basically removed the circular dependency and uses `ts[FieldName.{field}]` to reference a field instead of  the previous hardcoded `ts["field"]` that can be obsolete. This would prevent an issue like #283 . 

The tests you referred to were obsolete since they were for the older fields and the fixes there reflect that. I also added a test that hardcodes the current fields. This test aims to ensure that a change in the fields will throw an error and should only be done as a part of a bigger refactoring.

PR #314, opened 2019-09-16
include quantile function in distribution

vafl:
*Issue #, if available:*

*Description of changes:*

Include a `quantile` function for distributions.

### DIFF
--- a/src/gluonts/distribution/distribution.py
+++ b/src/gluonts/distribution/distribution.py
@@ -170,6 +170,13 @@ class Distribution:
         """
         return len(self.batch_shape)
 
+    @property
+    def all_dim(self) -> int:
+        r"""
+        Number of overall dimensions.
+        """
+        return self.batch_dim + self.event_dim
+
     def sample(self, num_samples: Optional[int] = None) -> Tensor:
         r"""
         Draw samples from the distribution.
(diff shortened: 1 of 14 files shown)


szha:
Job PR-314/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-314/1/index.html

szha:
Job PR-314/2 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-314/2/index.html

src/gluonts/distribution/distribution.py
```diff
@@ -223,6 +230,12 @@ def cdf(self, x: Tensor) -> Tensor:
         """
         raise NotImplementedError()
 
+    def quantile(self, level: Tensor) -> Tensor:
+        r"""
+        Returns the quantile corresponding to the passed level
```
lostella: It would be better to document `level` here as well as the return value (shape and so on)

src/gluonts/distribution/gaussian.py
```diff
@@ -102,6 +102,20 @@ def s(mu: Tensor, sigma: Tensor) -> Tensor:
             s, mu=self.mu, sigma=self.sigma, num_samples=num_samples
         )
 
+    def quantile(self, level: Tensor) -> Tensor:
+        F = self.F
+        # we consider level to be an independent axis and so expand it
+        # to shape (num_levels, 1, 1, ...)
```
lostella: For example, this expansion here is a bit counterintuitive if one reasons in terms of numpy’s broadcasting

lostella: But yeah, I guess we should be looking at how mxnet does broadcasting :-)

vafl: Yeah, we may need to iterate a bit on this. For now I assume the simplest case where level is a 1d array and the method returns a tensor with `(num_levels, ...DISTRIBUTION SHAPE...)`. I documented this in the base class.

szha:
Job PR-314/3 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-314/3/index.html

src/gluonts/distribution/gaussian.py
lostella: makes sense, thanks

Issue #313, opened 2019-09-15
Can't load a serialized predictor and utilize the GPU


lotsoffood:
## Description
I trained a model with `ctx='GPU'`, and saved the resulting `predictor`. Using directions from the documentation, I get the error described in https://github.com/awslabs/gluon-ts/issues/213. 

If I run the code mentioned in the above issue, the `predictor` loads, but only predicts using the CPU. Setting `ctx='gpu'` in the code below results in the same error, and therefore I'm unable to save a trained model and load it back while being able to retain the ability to predict with the GPU.

```
predictor = RepresentableBlockPredictor.deserialize(Path("models/"))
predictor.ctx = mx.Context('cpu')
```

## To Reproduce
Same as described here: https://github.com/awslabs/gluon-ts/issues/213

## Error Message
`RuntimeError: Parameter 'deeparpredictionnetwork0_featureembedder0_cat_0_embedding_weight' was not initialized on context gpu(0). It was only initialized on [cpu(0)].`

Any thoughts on being able to save out a `predictor` object, and load it back in and still be able to utilize `GPU` acceleration for inference?

vafl:
Can you try

```python
with mx.Context(mx.gpu(0)):
    predictor = Predictor.deserialize(Path(model_dir))
```

lotsoffood:
This works - `INFO:root:Using GPU` when the model is loaded, and it looks like it is using the GPU for predictions. Thank you!

lostella:
Closing since this was also properly addressed in #318 -- @lotsoffood feel free to ask further question, or reopen the issue if the problem persists!

PR #318, opened 2019-09-16
Manage mxnet context when deserializing predictors

vafl:
*Issue #, if available:*
#313 

*Description of changes:*
Manage mxnet context when deserializing predictors and don't swallow error messages.

### DIFF
--- a/src/gluonts/model/predictor.py
+++ b/src/gluonts/model/predictor.py
@@ -106,25 +106,28 @@ class Predictor:
         raise NotImplementedError
 
     def serialize(self, path: Path) -> None:
-        try:
-            # serialize Predictor type
-            with (path / "type.txt").open("w") as fp:
-                fp.write(fqname_for(self.__class__))
-        except Exception as e:
-            raise IOError(
-                f"Cannot serialize {fqname_for(self.__class__)} in {path}"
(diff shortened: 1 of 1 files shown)


szha:
Job PR-318/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-318/1/index.html

src/gluonts/model/predictor.py
```diff
@@ -106,25 +106,28 @@ def predict(self, dataset: Dataset, **kwargs) -> Iterator[Forecast]:
         raise NotImplementedError
 
     def serialize(self, path: Path) -> None:
-        try:
-            # serialize Predictor type
-            with (path / "type.txt").open("w") as fp:
-                fp.write(fqname_for(self.__class__))
-        except Exception as e:
```
lostella: Why do all the try-catch need to be removed?

vafl: They are swallowing actual error messages and stack traces and replace them by some generic error message.

vafl: Maybe we can add a test, since the build also runs on a GPU instance...

Issue #144, opened 2019-06-21
Use categoical and dynamc features by default in DeepAR


rshyamsundar:
Currently DeepAR does not use categorical and dynamic features by default even though they are present in the dataset. The flags `use_feat_dynamic_real` and `use_feat_static_cat` are set to `False` by default in `DeepAREstimator`. This is very bad for us in terms of results since people usually run methods with their default options or miss setting these flags explicitly. 

There were a couple of arguments against not using them by default but there are better remedies for them than not silently running DeepAR with incorrect options and returning less accurate results:

* Issue with not setting `cardinality` and still using `feat_static_cat`: we can make `cardinality` argument compulsory (or ideally derive this from data)

* If all time series do not have same features: there is no harm in failing with suitable error when the data is not consistent. 

Which case is important/priority for us right now: Running smoothly even if the data is not properly formatted or running with correct options that succeeds only when the data is consistent?

mbohlkeschneider:
> Which case is important/priority for us right now: Running smoothly even if the data is not properly formatted or running with correct options that succeeds only when the data is consistent?

I think in case of GluonTS, we aspire to make a _scientific_ library. Thus, I think the algorithms should fail if there are issues in the data. That informs the user that something is not right. Otherwise, you are left wondering why your results are not as good as you are expecting, especially if something is silently not used/discarded/filtered. I think this behavior should be avoided throughout the code base.

benidis:
I have started looking at this issue the last two days and it is a combination of addressing the input format question and defining the correct transformations behaviour. I agree with Michael that we should not do things silently and if something is wrong we should throw an error instead of trying to filter it internally.  However, this opens more questions: 
1. Should we check if all the fields are correct in a dataset (probably while creating windows) and throw an error if not? This adds some complexity since it needs to be applied at each created window.
2. What should we do with custom fields in a dataset or with fields that are not used by the model, especially with the ones that can break the code, e.g. #94 (note that the fix is not global but only for deepar - any other estimator can fail with the same issue)?
3. Setting aside deepar and looking at the bigger picture, what should be the behaviour of all estimators regarding the input data? Should they always take into account (or at least have the option to do so) a field that appears in the dataset or should they use only prespecified fields regardless of the input data as we were doing up to now? 

[...]

PR #319, opened 2019-09-16
Assert that static cat features are used in case cardinality is explicitly set

rshyamsundar:
This is extra safeguard to make sure that `use_feat_static_cat` is explicitly set if  cardinality is set. I saw people leaving the defaults for `use_feat_static_cat` and believing the static features are used by deepar since they have set `cardinality`.

There is already issue/enhancement #144 opened to change the default behavior of setting `use_feat_static_cat=False`. We at least need the current fix until #144 is resolved. 

*Description of changes:*
Currently it asserts one way implication:  `use_feat_static_cat=True` \implies `cardinality is not None`.
We need the converse as well: `cardinality is not None` \implies `use_feat_static_cat=True`.


szha:
Job PR-319/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-319/1/index.html

PR #286, opened 2019-09-05
More operators for synthetic data generation

jgasthaus:
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #321, opened 2019-09-16
Changed item key to item_id in write_csv_row to reflect changes to th…

dcmaddix:
…e ArtificalDataset in PR #286

*Issue #, if available:*

- Changed `item` key to `item_id` key in the `write_csv_row` for the synthetic dataset generation to reflect the changes to the `ArtificialDataset` class in PR #286

### DIFF
--- a/src/gluonts/dataset/artificial/generate_synthetic.py
+++ b/src/gluonts/dataset/artificial/generate_synthetic.py
@@ -56,7 +56,7 @@ def write_csv_row(
         if freq_week_start == "W":
             freq_week_start = f"W-{week_dict[timestamp.weekday()]}"
         timestamp = pd.Timestamp(data["start"], freq=freq_week_start)
-        item_id = int(data["item"])
+        item_id = int(data["item_id"])
         for j, target in enumerate(data["target"]):
             # Using convention that there are no missing values before the start date
             if is_missing and j != 0 and j % num_missing == 0:


szha:
Job PR-321/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-321/1/index.html

src/gluonts/dataset/artificial/generate_synthetic.py
```diff
@@ -56,7 +56,7 @@ def write_csv_row(
         if freq_week_start == "W":
             freq_week_start = f"W-{week_dict[timestamp.weekday()]}"
         timestamp = pd.Timestamp(data["start"], freq=freq_week_start)
-        item_id = int(data["item"])
+        item_id = int(data["item_id"])
```
lostella: The best thing would be to use [gluonts.dataset.field_names.FieldName](https://github.com/awslabs/gluon-ts/blob/ff22722cb3c19779dfefdd5b6d2c0e38ad93e605/src/gluonts/dataset/field_names.py#L15) (not just here, but all over the module, in case other field names are hardcoded)

dcmaddix: Yes, there are other hardcodings of `item_id` in `recipe.py` and `_base.py` that we should update as well.

Issue #303, opened 2019-09-12
DeepAR for multivariate time series


FadhelA:
Hello all,

I am trying to use DeepAR for multivariate time series forecasting. I am generating a bivariate time series, where each coordinate is independently sampled from a normal distribution. The first coordinate have mean 0 and variance 1, the second mean zero and variance 25. The algorithm seems to restrict the covariance matrix to lambda*Identity. I also tried the same code with the SimpleFeedForward model, and it works fine.

```py
import pandas as pd
import numpy as np


from gluonts.dataset.common import ListDataset
from gluonts.model.deepar import DeepAREstimator
from gluonts.trainer import Trainer

from gluonts.distribution.multivariate_gaussian import MultivariateGaussianOutput

from gluonts.evaluation.backtest import make_evaluation_predictions
    
        
# Generate data 

N = 20  # number of time series
T = 1000  # number of timesteps
dim = 2 # dimension of the observations
prediction_length = 25
freq = '1H'

custom_datasetx = np.random.normal(size=(N, dim, T))
custom_datasetx[:,1,:] = 5*custom_datasetx[:,1,:]
start = pd.Timestamp("01-01-2019", freq=freq)

train_ds = ListDataset(
    [
        {'target': x, 'start': start}
        for x in custom_datasetx[:, :, :-prediction_length]
    ],
    freq=freq,
    one_dim_target=False,
)



test_ds = ListDataset(
    [
        {'target': x, 'start': start}
        for x in custom_datasetx[:, :, :]
    ],
    freq=freq,
    one_dim_target=False,
)

# Deep AR 

# Trainer parameters
epochs = 10
learning_rate = 1E-3
batch_size = 5
num_batches_per_epoch = 100

# create estimator
estimator = DeepAREstimator(
    prediction_length=prediction_length,
    context_length=prediction_length,
    freq=freq,
    trainer=Trainer(
        ctx="cpu",
        epochs=epochs,
        learning_rate=learning_rate,
        hybridize=True,
        batch_size=batch_size,
        num_batches_per_epoch=num_batches_per_epoch,
    ),
    distr_output=MultivariateGaussianOutput(dim=dim)
)

predictor = estimator.train(train_ds)


forecast_it, ts_it = make_evaluation_predictions(
    dataset=test_ds,  # test dataset
    predictor=predictor,  # predictor
    num_eval_samples=100,  # number of sample paths we want for evaluation
)


forecasts = list(forecast_it)
tss = list(ts_it)
```

[...]

PR #330, opened 2019-09-19
Fix sample reshaping in DeepAR

lostella:
*Issue #, if available:* #303 

*Description of changes:* This fixes some bad reshaping of the samples generated in DeepAR. The issue does not affect the univariate case, but in the multivariate case it confuses the sample dimensions which result in the weird behaviour highlighted in #303.

### DIFF
--- a/src/gluonts/model/deepar/_network.py
+++ b/src/gluonts/model/deepar/_network.py
@@ -530,12 +530,12 @@ class DeepARPredictionNetwork(DeepARNetwork):
         # (batch_size * num_samples, prediction_length, *target_shape)
         samples = F.concat(*future_samples, dim=1)
 
-        # (batch_size, num_samples, *target_shape, prediction_length)
+        # (batch_size, num_samples, prediction_length, *target_shape)
         return samples.reshape(
             shape=(
                 (-1, self.num_parallel_samples)
-                + self.target_shape
                 + (self.prediction_length,)
+                + self.target_shape
             )
         )
 


szha:
Job PR-330/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-330/1/index.html

jaheba:
Should we have a test for that?

lostella:
> Should we have a test for that?

Good point: I think we have no tests for models outputting `SampleForecast` object created from multivariate distributions, essentially because it's undocumented territory there. For example, I'm not sure I like much the (N, T, C) layout of the samples, but that happens to be what all models using `MultivariateGaussian` produce as of now (including DeepAREstimator, with this fix). The reason I don't like it is that for the input data we have (C, T) layout, and in my opinion it should either be

> data: (C, T), forecast samples: (N, C, T)

or

> data: (T, C), forecast samples: (N, T, C)

but not a mixture of the two.

lostella:
To be clear: this PR fixes a reshaping operation that just *made no sense* before. This is regardless of whatever layout we want to have our samples in: the reshaping before this PR was just mixing two axes for no reason.

dai-ichiro:
I think 'copy_dim' also should be changed.

lostella:
@dai-ichiro can you provide more details? Where is `copy_dim` and why should it be changed?

dai-ichiro:
```
    dataset=test_data,  # test dataset
    num_eval_samples=100,  # number of sample paths we want for evaluation
)

forecasts = list(forecast_it)[0]
```
After fixing, the shape of forecasts.samples is (number of samples, length of prediction, number of targes).
'copy_dim' returns forecasts.samples[:, dim, :].
I think [:, :, dim] is better.

dai-ichiro:
'copy_dim' in 'gluon-ts/src/gluonts/model/forecast.py'

My sample code is
```
    dataset=test_data,  # test dataset
    num_eval_samples=100,  # number of sample paths we want for evaluation
)

plot_length = 36
legend = ["observations", "median prediction"] + [f"{k}% prediction interval" for k in prediction_intervals][::-1]

target_num = 3

for x, y in zip(ts_it, forecast_it):
    for i in range(target_num):
        plt.subplot(target_num,1,i+1)
        x.iloc[-plot_length:, i].dropna().plot()
        y.copy_dim(i).plot(color='g', prediction_intervals=(50.0,90.0))
        plt.grid(which='both')
        plt.legend(legend, loc = 'upper left')

plt.show()
```

lostella:
@dai-ichiro thanks for pointing this out. Can you open an issue with the code snippet reproducing the problem? Or even better, if you feel like opening a pull request to solve, that would be very welcome :-)

Issue #337, opened 2019-09-24
DeepState plotting


mharvan:
## Description
Plotting predictions made with a DeepState model fails.
The same plotting code works with a SimpleFeedForwardEstimator.

## To Reproduce
Please see the attached file
[benchmark_m4.py.txt](https://github.com/awslabs/gluon-ts/files/3647682/benchmark_m4.py.txt)
It is based on examples/benchmark_m4.py.

## Error Message
Length of passed values is 1, index implies 48

## Environment
Ubuntu 16.04
conda 4.7.10
Python 3.7.4

$ pip list | egrep 'gluon|mxnet'
gluonts 0.3.4.dev22+g515c45c /home/matus/src/gluon-ts/src
mxnet-cu101mkl 1.4.1

$ git log | head -n 1
commit 29ab758c9ddebdbba39a12597a1d5fcaaf906a97

lostella:
#340 should solve this, let me know if this works and feel free to reopen if it doesn't

mharvan:
Plotting works with #340, thanks for the fix!

PR #340, opened 2019-09-25
Fix DeepState samples tensor shape

lostella:
*Issue #, if available:* #337 

*Description of changes:* The shape of the samples tensor is fixed, and some room is made for potentially handling linear dynamical systems with multivariate output.

### DIFF
--- a/src/gluonts/model/deepstate/_network.py
+++ b/src/gluonts/model/deepstate/_network.py
@@ -68,7 +68,7 @@ class DeepStateNetwork(mx.gluon.HybridBlock):
                 flatten=False,
             )
             self.lstm = mx.gluon.rnn.HybridSequentialRNNCell()
-            self.lds_proj = LDSArgsProj(output_dim=1)
+            self.lds_proj = LDSArgsProj(output_dim=self.issm.output_dim())
             for k in range(num_layers):
                 cell = mx.gluon.rnn.LSTMCell(hidden_size=num_cells)
                 cell = mx.gluon.rnn.ResidualCell(cell) if k > 0 else cell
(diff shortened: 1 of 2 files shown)


src/gluonts/model/deepstate/issm.py
```diff
@@ -186,6 +192,9 @@ class LevelTrendISSM(LevelISSM):
     def latent_dim(self) -> int:
         return 2
 
+    def output_dim(self) -> int:
+        return 1
```
rshyamsundar: We need to update ISSM to handle multi-variate case; this can be set from outside then.

(I will open a new pull request for this.)

src/gluonts/model/deepstate/_network.py
```diff
@@ -248,5 +250,9 @@ def hybrid_forward(
         # convert samples from
         # (num_samples, batch_size, prediction_length, target_dim)
         # to
-        # (batch_size, num_samples, target_dim, prediction_length)
-        return samples.transpose(axes=(1, 0, 3, 2))
+        # (batch_size, num_samples, prediction_length, target_dim)
+        # and last axis in the univariate case
```
rshyamsundar: and _squeeze_ last axis

src/gluonts/model/deepstate/_network.py
```diff
@@ -138,7 +138,7 @@ def compute_lds(
             prior_mean=prior_mean,
             prior_cov=prior_cov,
             latent_dim=self.issm.latent_dim(),
-            output_dim=1,
+            output_dim=self.issm.output_dim(),
```
rshyamsundar: There is one more place it is hardcoded: https://github.com/awslabs/gluon-ts/blob/29ab758c9ddebdbba39a12597a1d5fcaaf906a97/src/gluonts/model/deepstate/_network.py#L69

PR #342, opened 2019-09-26
Adding gpu support to the cholesky jitter and eig tests.

dcmaddix:
*Issue #, if available:*
- Results in the error with Deep State described here https://github.com/awslabs/gluon-ts/issues/336

*Description of changes:*
- Enabled gpu tests for the cholesky jitter iterative and eigenvalue methods for numerically non-positive definite matrices
- On the gpu, if the cholesky factor `L` is not positive definite, it stores `nan` in the entries rather than throwing an error like on the cpu
- Updated assert error on gpu that compares `L.nansum()` and `L.sum()` to remove incorrect end parenthesis and to add an additional absolute value due to numerical issues.
- Moved `test/gp/test_jitter.py` to `test/support/test_jitter.py` since the jitter methods are under `src/support/linalg_util.py`

### DIFF
--- a/test/support/test_jitter.py
+++ b/test/support/test_jitter.py
@@ -29,12 +29,10 @@ import pytest
 # This test verifies that both eigenvalue decomposition and iterative jitter method
 # make a non-positive definite matrix positive definite to be able to compute the cholesky.
 # Both gpu and cpu as well as single and double precision are tested.
-# FIXME: Add gpu support for the tests in brazil gets error Check failed: e == cudaSuccess || e ==
-# FIXME: cudaErrorCudartUnloading CUDA: CUDA driver version is insufficient for CUDA runtime version
-# FIXME: @pytest.mark.parametrize('ctx', [mx.Context('gpu'), mx.Context('cpu')])
+@pytest.mark.parametrize("ctx", [mx.Context("gpu"), mx.Context("cpu")])
 @pytest.mark.parametrize("jitter_method", ["iter", "eig"])
 @pytest.mark.parametrize("float_type", [np.float32, np.float64])
-def test_jitter_unit(jitter_method, float_type, ctx=mx.Context("cpu")):
+def test_jitter_unit(jitter_method, float_type, ctx):
     matrix = nd.array(
         [[[1, 2], [3, 4]], [[10, 100], [-21.5, 41]]], ctx=ctx, dtype=float_type
     )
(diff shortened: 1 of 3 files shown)


szha:
Job PR-342/4 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-342/4/index.html

test/support/test_jitter.py
```diff
@@ -25,16 +25,25 @@
 import numpy as np
 import pytest
 
+# This function is used to see if the gpu is available.
+def gpu_device(gpu_number=0):
```
vafl: Cool. We have some similar functionality in https://github.com/awslabs/gluon-ts/blob/7ef55e4d6824b77703bb15201b180f00617e1abd/src/gluonts/core/component.py#L484 .
Perhaps it makes sense to move this function there and consolidate things.

test/support/test_jitter.py
```diff
@@ -51,14 +60,13 @@ def test_jitter_unit(jitter_method, float_type, ctx=mx.Context("cpu")):
...
-# FIXME: @pytest.mark.parametrize('ctx', [mx.Context('gpu'), mx.Context('cpu')])
+@pytest.mark.parametrize("ctx", [mx.Context("gpu"), mx.Context("cpu")])
 @pytest.mark.parametrize("jitter_method", ["iter", "eig"])
 @pytest.mark.parametrize("float_type", [np.float32, np.float64])
-def test_jitter_synthetic(
-    jitter_method, float_type, ctx=mx.Context("cpu")
-) -> None:
+def test_jitter_synthetic(jitter_method, float_type, ctx) -> None:
+    # TODO: Enable GPU tests on Jenkins
+    if ctx == mx.Context("gpu") and not gpu_device():
```
vafl: Perhaps turn this around, so `gpu_device()` guards the creation of the context?

test/support/test_jitter.py
dcmaddix: Great, it looks like the `check_gpu_support()` function already checks for GPU support.  I will update to call this instead.  Thanks!

szha:
Job PR-342/5 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-342/5/index.html

szha:
Job PR-342/6 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-342/6/index.html

PR #343, opened 2019-09-27
Adding GP example on synthetic dataset with built-in plotting

dcmaddix:
*Issue #, if available:*

*Description of changes:*
- Added Gaussian Process (GP) synthetic example
- Shows how to call the `GaussianProcess` with the `exact_inference` and `plot` methods as well as how to select the kernel
- The plot shows the predictive mean and +/- 2 predictive stds

### DIFF
--- a/examples/gp_synthetic_example.py
+++ b/examples/gp_synthetic_example.py
@@ -1,0 +1,92 @@
+# Standard library imports
+import math
+
+# First-party imports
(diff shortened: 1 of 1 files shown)


lostella:
I’m not sure what’s going on with the docs build, but it seems to be failing for a different reason than before. The other test failing is one of the shaky inference ones.

So it looks good, thanks :-)

PR #347, opened 2019-09-27
Fix WaveNet prediction length during training

steverab:
*Issue #, if available:* -

*Description of changes:*

When training a `WaveNetEstimator` the prediction length used during training is currently determined by the mean length of all observed time series. As the mean is not robust to outliers, few very long time series (or few very short time series) can have a large influence on this value. This can lead to some of the selected time series selected for training being shorter than the prediction length, which in turn can cause the used `InstanceSplitter` to not return any instances if `GLUONTS_MAX_IDLE_TRANSFORMS`-many time series in sequence happen to be too short. While this issue can be circumvented by simply setting the `train_window_length` parameter during estimator initialization, I believe that changing the mean to the median is a saner default, as it ensures that at least half of all time series will lie above this threshold and the other half below it. Furthermore, the prediction length during training should also account for the fact that we need to partition the entire time series into a context range and a prediction range, meaning that we need to subtract the context window length from the median time series length to result in the final prediction length.

This problem can be easily reproduced by running the [GluonTS quick start tutorial](https://gluon-ts.mxnet.io/examples/basic_forecasting_tutorial/tutorial.html) using the `m4_hourly` dataset (which contains a variety of differently sized time series) in conjunction with the `WaveNetEstimator`.


szha:
Job PR-347/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-347/1/index.html

vafl:
Great, thanks for the contribution!

src/gluonts/model/wavenet/_estimator.py
```diff
@@ -269,9 +271,11 @@ def train(self, training_data: Dataset) -> Predictor:
         # Here we override the prediction length for training.
         # This computes the loss over longer windows and makes the convolutions more
         # efficient, since calculations are reused.
-        pred_length = min(mean_length, self.train_window_length)
+        pred_length = min(
+            median_length - self.context_length, self.train_window_length
+        )
 
-        logging.info(f"mean series length = {mean_length}")
+        logging.info(f"mean series length = {median_length}")
```
jaheba: The log message still says mean.

steverab:
Good catch! Just fixed this.

szha:
Job PR-347/2 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-347/2/index.html

src/gluonts/model/wavenet/_estimator.py
```diff
@@ -269,9 +271,11 @@ def train(self, training_data: Dataset) -> Predictor:
         # Here we override the prediction length for training.
         # This computes the loss over longer windows and makes the convolutions more
         # efficient, since calculations are reused.
-        pred_length = min(mean_length, self.train_window_length)
+        pred_length = min(
+            median_length - self.context_length, self.train_window_length
```
jaheba: This can cause issues when `context_length` is larger than `median_length`.

steverab: Added a guard to prevent negative prediction length.

szha:
Job PR-347/3 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-347/3/index.html

szha:
Job PR-347/4 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-347/4/index.html

steverab:
After even more testing, I ran into some more problems on various datasets while using the median. I would therefore opt for using the standard `prediction_length` (the one which is used during the evaluation stage) also for the training stage by default. However, we should definitely still allow users to easily override this default value by explicitly specifying the desired `training_window_length` during initialization.

lostella:
> After even more testing, I ran into some more problems on various datasets while using the median.

@steverab Which problems exactly?

steverab:
@lostella The same that caused me to open this pull request in the first place, see first comment in the PR. Again, the point is to have a safe default value which works well across datasets without crashing the `InstanceSplitter`, but to still offer the ability to improve training performance by specifying a larger `training_window_length` during initialization.

lostella:
> @lostella The same that caused me to open this pull request in the first place, see first comment in the PR.

I see, thank you. Then maybe a lower quantile (not sure which: 0.2? 0.1?) would be more appropriate? That we we can directly bound the probability of selecting time series which are too short for the InstanceSplitter to process.

src/gluonts/model/wavenet/_estimator.py
```diff
@@ -258,20 +258,22 @@ def __init__(
...
-        # Here we override the prediction length for training.
+        # Here we override the prediction length for training if explicitly specified via train_window_length.
         # This computes the loss over longer windows and makes the convolutions more
         # efficient, since calculations are reused.
-        pred_length = min(mean_length, self.train_window_length)
+        pred_length = (
+            self.train_window_length
+            if self.train_window_length is not None
+            else self.prediction_length
+        )
```
lostella: This definitely mitigates the `InstanceSampler` problem (it's reasonable to have time series of at least length `prediction_length` if you plan on training a model that makes `prediction_length`-long predictions). However, to make this more in line with the `DeepAREstimator` (as well as to take advantage of WaveNet being more efficient at training over longer windows) we could make this `2*self.prediction_length` if not specified otherwise?

Also, `pred_length` is a weird name for this variable here, and this piece of code could really go in the constructor as
```
self.training_window_length = training_window_length if training_window_length is not None else 2*prediction_length
```
so that, from there on, `self.training_window_length` should be used to configure the training network.

src/gluonts/model/wavenet/_estimator.py
```diff
@@ -212,7 +212,11 @@ def __init__(
         self.n_residue = n_residue
         self.n_skip = n_skip
         self.n_stacks = n_stacks
-        self.train_window_length = train_window_length
+        self.train_window_length = (
+            train_window_length
+            if train_window_length is not None
+            else 2 * prediction_length
```
lostella:
```suggestion
            else prediction_length
```

src/gluonts/model/wavenet/_estimator.py
```diff
@@ -258,24 +262,20 @@ def __init__(
...
-        logging.info(f"mean series length = {mean_length}")
-        logging.info(f"using training windows of length = {pred_length}")
+        logging.info(
+            f"using training windows of length = {self.train_window_length}"
+        )
 
         transformation = self.create_transformation(
-            bin_edges, pred_length=pred_length
+            bin_edges,
+            pred_length=self.train_window_length - self.prediction_length,
```
lostella:
```suggestion
            pred_length=self.train_window_length,
```

src/gluonts/model/wavenet/_estimator.py
```diff
@@ -292,7 +292,9 @@ def train(self, training_data: Dataset) -> Predictor:
         # context as the one that will be used during training
         with self.trainer.ctx:
             params = self._get_wavenet_args(bin_centers)
-            params.update(pred_length=pred_length)
+            params.update(
+                pred_length=self.train_window_length - self.prediction_length
```
lostella:
```suggestion
                pred_length=self.train_window_length
```

PR #348, opened 2019-09-29
Introduce ForecastGenerator to wrap mxnet output into forecast object

vafl:
*Issue #, if available:*

*Description of changes:*

Make predictors and outputs more modular by introducing a `ForecastGenerator` that is used to bring the output of an mxnet block into a forecast.

### DIFF
--- a/src/gluonts/model/forecast_generator.py
+++ b/src/gluonts/model/forecast_generator.py
@@ -71,7 +71,7 @@ def _extract_instances(x: Any) -> Any:
         assert False
 
 
-class ForecastWrapper:
+class ForecastGenerator:
     """
     Classes used to bring the output of a network into a class.
     """
@@ -89,7 +89,7 @@ class ForecastWrapper:
         raise NotImplementedError()
 
 
-class DistributionForecastWrapper(ForecastWrapper):
+class DistributionForecastGenerator(ForecastGenerator):
     @validated()
     def __init__(self, distr_output: DistributionOutput) -> None:
         self.distr_output = distr_output
@@ -133,7 +133,7 @@ class DistributionForecastWrapper(ForecastWrapper):
             assert i + 1 == len(batch["forecast_start"])
 
 
-class QuantileForecastWrapper(ForecastWrapper):
+class QuantileForecastGenerator(ForecastGenerator):
     @validated()
     def __init__(self, quantiles: List[str]) -> None:
         self.quantiles = quantiles
@@ -174,7 +174,7 @@ class QuantileForecastWrapper(ForecastWrapper):
             assert i + 1 == len(batch["forecast_start"])
 
 
-class SampleForecastWrapper(ForecastWrapper):
+class SampleForecastGenerator(ForecastGenerator):
     @validated()
     def __init__(self):
         pass
--- a/src/gluonts/model/predictor.py
+++ b/src/gluonts/model/predictor.py
@@ -51,7 +51,7 @@ from gluonts.core.component import (
 from gluonts.core.exception import GluonTSException
 from gluonts.core.serde import dump_json, fqname_for, load_json
 from gluonts.dataset.common import DataEntry, Dataset, ListDataset
-from .forecast_wrapper import ForecastWrapper, SampleForecastWrapper
+from .forecast_generator import ForecastGenerator, SampleForecastGenerator
 from gluonts.dataset.loader import DataBatch, InferenceDataLoader
 from gluonts.model.forecast import Forecast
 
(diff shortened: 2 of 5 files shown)


szha:
Job PR-348/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-348/1/index.html

szha:
Job PR-348/2 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-348/2/index.html

src/gluonts/model/predictor.py
```diff
@@ -501,10 +477,9 @@ def as_symbol_block_predictor(
             freq=self.freq,
             ctx=self.ctx,
             input_transform=self.input_transform,
+            forecast_wrapper=self.forecast_wrapper,
             output_transform=self.output_transform,
             float_type=self.float_type,
-            forecast_cls_name=self.forecast_cls_name,
-            forecast_kwargs=self.forecast_kwargs,
```
lostella: The `_kwargs` also need to be forwarded

vafl: Thanks, good point. Renamed it to `ForecastGenerator`.

Also I just realized that the `forecast_kwargs` is not needed anymore, since such fields can be passed to the constructor of the specific `ForecastGenerator` as is done e.g. for the `QuantileForecast`.

szha:
Job PR-348/3 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-348/3/index.html

src/gluonts/model/forecast_generator.py
```diff
@@ -0,0 +1,222 @@
...
+    if isinstance(x, (np.ndarray, mx.nd.NDArray)):
+        for i in range(x.shape[0]):
+            # yield x[i: i + 1]
+            yield x[i]
+    elif isinstance(x, tuple):
+        for m in zip(*[_extract_instances(y) for y in x]):
+            yield tuple([r for r in m])
+    elif x is None:
+        while True:
+            yield None
```
lostella: What is the purpose of this? Shouldn't the `x is None` also throw an exception?

src/gluonts/model/forecast_generator.py
```diff
@@ -0,0 +1,222 @@
...
+                    freq=freq,
+                    item_id=batch[FieldName.ITEM_ID][i]
+                    if FieldName.ITEM_ID in batch
+                    else None,
+                    info=batch["info"][i] if "info" in batch else None,
+                )
+            assert i + 1 == len(batch["forecast_start"])
+
+
+class QuantileForecastGenerator(ForecastGenerator):
```
lostella: This class looks a lot like `DistributionForecastGenerator`, except for minor differences in the way forecast objects are constructed. We can merge it this way for the time being, I guess, but in the future it would be nice to consolidate `QuantileForecast[Generator]` into `DistributionForecast[Generator]` through an appropriate `Distribution` subtype. (The same could apply to `SampleForecast[Generator]` but I see there's a bit more logic there so that might require some thinking)

src/gluonts/model/forecast_generator.py
vafl: This handles the case where one of the return tensors is `None`, which mxnet supports. Then the function extracts things correctly, with a `None` value in the corresponding entry of each instance. In the specific case of the `DistributionForecast` the `scale` can be none.

src/gluonts/model/forecast_generator.py
vafl: yes, that might be nice

src/gluonts/model/forecast_generator.py
```diff
@@ -0,0 +1,222 @@
...
+        **kwargs
+    ) -> Iterator[DistributionForecast]:
+        for batch in inference_data_loader:
+            inputs = [batch[k] for k in input_names]
+            outputs = prediction_net(*inputs)
+            if output_transform is not None:
+                outputs = output_transform(batch, outputs)
+            if num_eval_samples:
+                log_once(
+                    "Forecast is not sample based. Ignoring parameter `num_eval_samples` from predict method."
```
lostella: We've been calling this `num_eval_samples` everywhere, but I think this is confusing: if anything, I would call this `num_samples` in this context (and in many other places, but outside the scope of this PR). Also, the `"predict"` method is out of context.

My opinion is that this should be hidden in the `**kwargs`, and only be explicitly exposed in `SampleForecastGenerator`, but it's not a big deal.

src/gluonts/model/forecast_generator.py
lostella: I see, so this is primarily for the case of tuples having `None` among the elements -- let's make sure `_extract_instances` is never invoked with `None` at the top level :-)

PR #349, opened 2019-09-29
Small fixes to box cox and `TransformedDistribution`

vafl:
*Issue #, if available:*

*Description of changes:*

Some small fixes to boxcox:
- `lb_obs` was used with the wrong sign in boxcox. I.e. when the smallest observation is `-0.1` you had to set `lb_obs=0.1`
- fix broadcasting issue in the condition

### DIFF
--- a/src/gluonts/distribution/box_cox_tranform.py
+++ b/src/gluonts/distribution/box_cox_tranform.py
@@ -197,7 +197,7 @@ class BoxCoxTranform(Bijection):
         base = F.relu(y * lambda_1 + 1.0)
 
         return F.where(
-            condition=F.abs(lambda_1).__ge__(tol_lambda_1),
+            condition=(F.abs(lambda_1).__ge__(tol_lambda_1)).broadcast_like(y),
             x=_power(base, 1.0 / lambda_1) - lambda_2,
             y=F.exp(y) - lambda_2,
             name="Box_Cox_inverse_trans",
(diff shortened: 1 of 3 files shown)


src/gluonts/distribution/box_cox_tranform.py
```diff
@@ -254,10 +254,10 @@ def __init__(self, lb_obs: float = 0.0, fix_lambda_2: bool = True) -> None:
     def domain_map(self, F, *args: Tensor) -> Tuple[Tensor, ...]:
         lambda_1, lambda_2 = args
         if self.fix_lambda_2:
-            lambda_2 = self.lb_obs * F.ones_like(lambda_2)
+            lambda_2 = -self.lb_obs * F.ones_like(lambda_2)
         else:
             # This makes sure that :math:`z +  \lambda_2 > 0`, where :math:`z > lb_obs`
-            lambda_2 = softplus(F, lambda_2) - self.lb_obs * F.ones_like(
+            lambda_2 = softplus(F, lambda_2) + self.lb_obs * F.ones_like(
```
rshyamsundar: I think it should be `-` here. For example if observations are negative say with minimum value -100, then `lambda_2` with your change becomes `x - 100` where `x > 0`. Then the expression `z + \lambda2` is not guaranteed to be positive anymore and the transformation fails because it applies logarithm to this expression (when `\lambda_1 = 0`).

vafl: Yes, thanks. Only the fixed case had the bug.

PR #363, opened 2019-10-02
Datatype

rshyamsundar:
*Description of changes:*

This PR adds support for changing precision for gluon estimators. This is useful for example to speedup models by switching to lower precision or make the them (e.g., `gp_forecaster`) more robust by using double precision.

The changes are done only for `DeepAREstimator` on the computation side i.e., only changing estimator, network and distribution output. Changes on the dataset loading part as well as for other estimators will be sent separately to make the PR manageable. 
 
* Need new mxnet version; in older versions backward fails in symbolic mode if you use `np.float64`: one of the inputs of `grad_sum` is not `np.float32`. It maybe because  `float32` is hardcoded somewhere.
* sampling operators always return `np.float32` values, so had to set `dtype` explicitly here. 
* Also made the terminology consistent across by replacing `float_type` to `dtype`.

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -1,7 +1,7 @@
 boto3==1.*
 holidays==0.9.*
 matplotlib==3.*
-mxnet>=1.3.1,<1.5.*
+mxnet>=1.4.1,<1.5.*
 numpy==1.14.*
 pandas>=0.25.0
 pydantic==0.28.*
--- a/src/gluonts/distribution/binned.py
+++ b/src/gluonts/distribution/binned.py
@@ -17,6 +17,7 @@ from typing import Tuple, List
 # Third-party imports
 import mxnet as mx
 from mxnet import gluon
+import numpy as np
 
 # First-party imports
 from gluonts.core.component import validated
@@ -157,7 +158,7 @@ class Binned(Distribution):
         a = centers_expanded.pick(idx, axis=-1)
         return a.swapaxes(0, 1)
 
-    def sample(self, num_samples=None):
+    def sample(self, num_samples=None, dtype=np.float32):
         def s(bin_probs):
             F = self.F
             indices = F.sample_multinomial(bin_probs)
--- a/src/gluonts/distribution/distribution_output.py
+++ b/src/gluonts/distribution/distribution_output.py
@@ -50,18 +50,18 @@ class ArgProj(gluon.HybridBlock):
         self,
         args_dim: Dict[str, int],
         domain_map: Callable[..., Tuple[Tensor]],
-        float_type: DType = np.float32,
+        dtype: DType = np.float32,
         prefix: Optional[str] = None,
         **kwargs,
     ) -> None:
         super().__init__(**kwargs)
         self.args_dim = args_dim
-        self.float_type = float_type
+        self.dtype = dtype
         self.proj = [
             gluon.nn.Dense(
                 dim,
                 flatten=False,
-                dtype=self.float_type,
+                dtype=self.dtype,
                 prefix=f"{prefix}_distr_{name}_",
             )
             for name, dim in args_dim.items()
@@ -83,22 +83,22 @@ class Output:
     """
 
     args_dim: Dict[str, int]
-    _float_type: DType = np.float32
+    _dtype: DType = np.float32
 
     @property
-    def float_type(self):
-        return self._float_type
+    def dtype(self):
+        return self._dtype
 
-    @float_type.setter
-    def float_type(self, float_type: DType):
-        self._float_type = float_type
+    @dtype.setter
+    def dtype(self, dtype: DType):
+        self._dtype = dtype
 
     def get_args_proj(self, prefix: Optional[str] = None) -> ArgProj:
         return ArgProj(
             args_dim=self.args_dim,
             domain_map=gluon.nn.HybridLambda(self.domain_map),
             prefix=prefix,
-            float_type=self._float_type,
+            dtype=self.dtype,
         )
 
     def domain_map(self, F, *args: Tensor):
--- a/src/gluonts/distribution/distribution_output.py
+++ b/src/gluonts/distribution/distribution_output.py
@@ -83,12 +83,22 @@ class Output:
     """
 
     args_dim: Dict[str, int]
+    _float_type: DType = np.float32
+
+    @property
+    def float_type(self):
+        return self._float_type
+
+    @float_type.setter
+    def float_type(self, float_type: DType):
+        self._float_type = float_type
 
     def get_args_proj(self, prefix: Optional[str] = None) -> ArgProj:
         return ArgProj(
             args_dim=self.args_dim,
             domain_map=gluon.nn.HybridLambda(self.domain_map),
             prefix=prefix,
+            float_type=self._float_type,
         )
 
     def domain_map(self, F, *args: Tensor):
--- a/src/gluonts/gp/gaussian_process.py
+++ b/src/gluonts/gp/gaussian_process.py
@@ -250,7 +250,7 @@ class GaussianProcess:
             self._compute_cholesky_gp(
                 covariance, self.prediction_length, self.sample_noise
             ),
-            float_type=self.float_type,
+            dtype=self.float_type,
         ).sample_rep(
             self.num_samples
         )  # Shape (num_samples, batch_size, prediction_length)
--- a/src/gluonts/gp/gaussian_process.py
+++ b/src/gluonts/gp/gaussian_process.py
@@ -250,9 +250,8 @@ class GaussianProcess:
             self._compute_cholesky_gp(
                 covariance, self.prediction_length, self.sample_noise
             ),
-            dtype=self.float_type,
         ).sample_rep(
-            self.num_samples
+            self.num_samples, dtype=self.float_type
         )  # Shape (num_samples, batch_size, prediction_length)
         return self.F.transpose(samples, axes=(1, 2, 0))
 
--- a/src/gluonts/kernels/_kernel_output.py
+++ b/src/gluonts/kernels/_kernel_output.py
@@ -97,7 +97,7 @@ class KernelOutputDict(KernelOutput):
         return ArgProj(
             args_dim=self.args_dim,
             domain_map=gluon.nn.HybridLambda(self.domain_map),
-            float_type=float_type,
+            dtype=float_type,
         )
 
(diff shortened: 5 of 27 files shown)


src/gluonts/model/deepar/_network.py
```diff
@@ -274,6 +279,13 @@ def unroll_encoder(
             length=subsequences_length,
             layout="NTC",
             merge_outputs=True,
+            begin_state=self.rnn.begin_state(
```
rshyamsundar: The default `begin_state` still uses `np.float32` even if the RNN cell is of type `np.float64`.

szha:
Job PR-363/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-363/1/index.html

szha:
Job PR-363/2 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-363/2/index.html

src/gluonts/distribution/distribution_output.py
```diff
@@ -82,12 +82,22 @@ class Output:
...
     args_dim: Dict[str, int]
+    _dtype: DType = np.float32
+
+    @property
+    def dtype(self):
+        return self._dtype
+
+    @dtype.setter
+    def dtype(self, dtype: DType):
+        self._dtype = dtype
```
lostella: I’m wondering why using this getter/setter pair here? Anywhere else this is just stored as an attribute

src/gluonts/gp/gaussian_process.py
```diff
@@ -250,7 +250,7 @@ def sample(self, mean: Tensor, covariance: Tensor) -> Tensor:
             self._compute_cholesky_gp(
                 covariance, self.prediction_length, self.sample_noise
             ),
-            float_type=self.float_type,
+            dtype=self.float_type,
```
lostella: `self.dtype`?

src/gluonts/model/deepar/_estimator.py
```diff
@@ -116,7 +117,9 @@ def __init__(
         self,
         freq: str,
         prediction_length: int,
-        trainer: Trainer = Trainer(),
+        trainer: Trainer = Trainer(
+            hybridize=True, num_batches_per_epoch=1, epochs=1
```
lostella: These are not supposed to be in the PR I guess

src/gluonts/model/deepar/_network.py
```diff
@@ -96,9 +99,11 @@ def __init__(
                     else cell
                 )
                 self.rnn.add(cell)
+            self.rnn.cast(dtype=dtype)
```
lostella: Couldn’t (shouldn’t?) the single blocks in the `HybridSequentialRNNCell` be set to the right dtype?

src/gluonts/model/deepar/_network.py
lostella: Is this for hybrid sequential only, or also for elementary cells?

src/gluonts/model/deepar/_network.py
```diff
@@ -274,6 +279,13 @@ def unroll_encoder(
             length=subsequences_length,
             layout="NTC",
             merge_outputs=True,
+            begin_state=self.rnn.begin_state(
+                func=F.zeros,
```
lostella: Is it necessary to specify this function?

src/gluonts/model/gp_forecaster/_estimator.py
```diff
@@ -104,15 +104,15 @@ def __init__(
...
         max_iter_jitter: int = 10,
         jitter_method: str = "iter",
         sample_noise: bool = True,
         time_features: Optional[List[TimeFeature]] = None,
         num_parallel_samples: int = 100,
     ) -> None:
-        self.float_type = float_type
-        super().__init__(trainer=trainer, float_type=self.float_type)
+        self.float_type = dtype
+        super().__init__(trainer=trainer, dtype=self.float_type)
```
lostella: `self.dtype`?

src/gluonts/kernels/_kernel_output.py
```diff
@@ -97,7 +97,7 @@ def get_args_proj(self, float_type: DType = np.float32) -> ArgProj:
         return ArgProj(
             args_dim=self.args_dim,
             domain_map=gluon.nn.HybridLambda(self.domain_map),
-            float_type=float_type,
+            dtype=float_type,
```
lostella: `dtype=dtype`?

src/gluonts/model/gp_forecaster/_estimator.py
```diff
@@ -104,15 +104,15 @@ def __init__(
...
+        dtype: DType = np.float64,
         max_iter_jitter: int = 10,
         jitter_method: str = "iter",
         sample_noise: bool = True,
         time_features: Optional[List[TimeFeature]] = None,
         num_parallel_samples: int = 100,
     ) -> None:
-        self.float_type = float_type
-        super().__init__(trainer=trainer, float_type=self.float_type)
+        self.float_type = dtype
```
lostella: `self.dtype`?

src/gluonts/model/deepar/_network.py
lostella: My guess is that any dtype a HybridSequential(RNN) block deals with (dtype of output, dtype of state) depends on the dtypes its component blocks deal with.

dcmaddix: Yes, I remember similar issues with RNNs when doing this for the GP forecaster.

rshyamsundar:
> Jenkins is complaining a lot: do tests pass locally?

Sorry was testing this live! 
Sampling was failing because we use prediction network in hybrid mode as well which means I cannot get `dtype` from the parameters of the distribution during sampling. I will update the PR soon.

src/gluonts/model/deepar/_network.py
rshyamsundar: I tested by casting each cell to proper `dtype` but that did not help either (see the comment below).

src/gluonts/distribution/distribution_output.py
rshyamsundar: This is for setting the `dtype` of the distribution output correctly in case the default used [here](https://github.com/awslabs/gluon-ts/blob/e021fa4e20cb562f836211f716602b359c79af36/src/gluonts/model/deepar/_estimator.py#L133) is not consistent with the  `dtype` passed to the estimator.

src/gluonts/gp/gaussian_process.py
rshyamsundar: As mentioned, this would trigger changes on the `gp_forecaster` as well which I would like to do separately. The `gp_forecaster` also needs to be updated to use proper `dtype`. Simply passing `dtype` to `GluonEstimator` does not suffice.

src/gluonts/model/deepar/_estimator.py
rshyamsundar: Ooops.

src/gluonts/model/deepar/_network.py
rshyamsundar: I tried that as well, i.e., casting each cell individually to the right type before adding to the `rnn`. Even that does not force it to use the right dtype for the begin state. I ended up with this version for no particular reason.

src/gluonts/model/gp_forecaster/_estimator.py
rshyamsundar: Ideally I should replace `float_type` with `dtype` here and remove this line (why did we have this line here anyway? this is not needed here); but refactoring the name triggers lot of changes on the `gp_forecaster` so postponing it.

src/gluonts/kernels/_kernel_output.py
rshyamsundar: This is for the next PR as it triggers many changes.

src/gluonts/gp/gaussian_process.py
dcmaddix: Yes, I think the `gp_forecaster` will work will either `dtype`.  We may just need to do some renaming of the variables.

szha:
Job PR-363/3 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-363/3/index.html

requirements/requirements.txt
```diff
@@ -1,7 +1,7 @@
 boto3==1.*
 holidays==0.9.*
 matplotlib==3.*
-mxnet>=1.3.1,<1.5.*
+mxnet>=1.4.1,<1.5.*
```
jaheba: Do we need to update the version somewhere else for the CI?

src/gluonts/distribution/binned.py
```diff
@@ -157,7 +158,7 @@ def step(p, state):
         a = centers_expanded.pick(idx, axis=-1)
         return a.swapaxes(0, 1)
 
-    def sample(self, num_samples=None):
+    def sample(self, num_samples=None, dtype=np.float32):
```
jaheba: `dtype` isn't used in the function?

src/gluonts/distribution/distribution_output.py
jaheba: I'm not sure I understand. This looks like you just use `dtype` as an alias for `_dtype`.

src/gluonts/distribution/binned.py
rshyamsundar: This distribution does not need the `dtype` because it uses sampling from multinomial distribution which returns integers (`int32`). I added here to be consistent with the signature of `sample` from the abstract class.

We should remove this once [this issue](https://github.com/apache/incubator-mxnet/issues/16407) is fixed on the mxnet side.

src/gluonts/distribution/distribution_output.py
rshyamsundar: just defining `dtype` as a property to avoid ugly getter and setter methods and also not to access private fields directly from outside.

rshyamsundar:
> How would you go about changing the dtype? I.e. how would a different dtype propagate through the system.

> 
> And, could we have tests checking for a different dtype?

One can set the `dtype` while creating the estimator if the corresponding model needs to do the operations in higher or lower precision.

Ya, maybe we should add tests to check things work with different `dtype`s.

requirements/requirements.txt
rshyamsundar: I am not sure; let me check.

PR #378, opened 2019-10-10
Use validation to prevent overfitting 

### LINKED ISSUE #258: How to check validation loss?
## Description
I am currently using the deepar package. I am able to see the training loss. But, I am not able to see the validation loss. My model is converging on training loss. But, how will I know when to stop it? If I continue the training by monitoring the training loss, it would result in an over fitted model. What do you suggest

## References
- list reference and related literature
- list known implementations

strawberrypie:
Fixes #258 

*Description of changes:*
  * `Trainer` can accept a validation set
  * `GluonEstimator` and `WaveNetEstimator` pass a validation set to the `Trainer`
  * `"test"` channel datasets in SageMaker are passed to the Estimator as validation


szha:
Job PR-378/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-378/1/index.html

szha:
Job PR-378/2 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-378/2/index.html

src/gluonts/dataset/loader.py
```diff
@@ -135,6 +132,9 @@ class TrainDataLoader(DataLoader):
         Number of batches to return in one complete iteration over this object.
     float_type
         Floating point type to use.
+    use_training_transform
+        Enables training mode for transformations, i.e. with the flag
+    `is_train = True`.
```
lostella: Why is this required? It seems unused in what follows

src/gluonts/model/estimator.py
```diff
@@ -178,6 +186,19 @@ def train_model(self, training_data: Dataset) -> TrainOutput:
...
+                dataset=validation_data,
+                transform=transformation,
+                batch_size=self.trainer.batch_size,
+                num_batches_per_epoch=self.trainer.num_batches_per_epoch,
+                ctx=self.trainer.ctx,
+                float_type=self.float_type,
+            )
+            if validation_data is not None
+            else None
+        )
```
lostella: This could be fine in case the `validation_data` consists of a different set of time series wrt training data. But even in that case, I guess one should consider batches in which the prediction time window falls entirely *after* the training dataset. This is probably tricky to obtain though...

cc @geoalgo @jgasthaus

lostella:
Thank you! This is really some impressive contribution. I'm just not sure about the way the validation trainer is defined, so let's wait also for comments from others.

lostella:
@strawberrypie also, there seem to be some conflicts with some upstream changes, which are hopefully easy to resolve

src/gluonts/shell/train.py
```diff
@@ -58,18 +58,24 @@ def run_train_and_test(
     if isinstance(forecaster, Predictor):
         predictor = forecaster
     else:
-        predictor = run_train(forecaster, env.datasets["train"])
+        predictor = run_train(
+            forecaster, env.datasets["train"], env.datasets.get("test")
```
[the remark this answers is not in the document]
strawberrypie: Yes, this is a better name. I guess that it means that three channel are going to be used for training — `train`, `validation` (the new one — to estimate metrics during training), and `test` (to estimate metrics in the end). 
Is there documentation that should be updated with this change?

src/gluonts/dataset/loader.py
strawberrypie: I've missed a place where is should have been located: `model/estimator.py`, in the constructor for validation data loader. We need to perform the correct transformation on validation, so that's why this option is here.
What do you think about the name for this DataLoader. Should it be changed due to the fact that it is also used for validation?

src/gluonts/model/estimator.py
strawberrypie: Yes, I've wrote it the most general way — you need to prepare correct validation yourself. 

I guess that we can implement several strategies to generate validation on the fly from the training set, but I am not sure about all of the possible options. (Cutting the most recent windows from the training data might not cover all the cases.)

szha:
Job PR-378/3 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-378/3/index.html

src/gluonts/dataset/loader.py
lostella: I think this can lead to confusion, because of how things are currently. As of now, the `InferenceDataLoader` is available that has this flag hardcoded to `False`. What this implies is that iterating through the `InferenceDataLoader` yields batches sliced from the trailing part of all time series (in particular, having the "prediction" interval at the very end), and each time series is considered only once: this is used to construct the "test" data loader, but I guess it should be the same for the validation one. This way a user can simply slice the train/validation/test dataset appropriately, e.g. (suppose prediction length is `-----`)

```
train:        +---------------------------------------+
validation:   +---------------------------------------|-----+
test:         +---------------------------------------------|-----+
```

and the corresponding data loaders will yield batches with the prediction interval lying in the right place (not overlapping with training data, nor with validation in the test case).

Does this make any sense? In summary, I would remove this option here, and set the `validation_data_loader` to be `InferenceDataLoader` in case validation data is provided.

src/gluonts/model/estimator.py
```diff
@@ -83,7 +87,9 @@ class DummyEstimator(Estimator):
     def __init__(self, predictor_cls: type, **kwargs) -> None:
         self.predictor = predictor_cls(**kwargs)
 
-    def train(self, training_data: Dataset) -> Predictor:
+    def train(
+        self, training_data: Dataset, valid_dataset: Optional[Dataset] = None
```
jaheba: This is a breaking change, and the more that I think about it, the more I dislike our overall design of estimators and predictors. Essentially, such a change shouldn't be breaking.

We have models that don't require training data, i.e. local models which "train" online. Still, if we want an estimator, we would have to implement a train method for them which takes a training dataset as input.

I think this approach is another step in this direction. Now we add another dataset to `train`, and some models use it and others don't. 

Our design also inhabits the "class with only one method" code-smell, albeit there might be some good reasons to do it this way.

I'm also wondering if our current approach limits our future design space. E.g., if we want to develop models for anomaly detection, we might have different inputs than for normal forecasting models.

This is not to take anything away from your great work @strawberrypie, but I think I would like to have this discussion rather now then later.

/cc @vafl @lostella @jgasthaus

lostella: @jaheba I feel your pain and agree that the interaction between components here is a bit weird

However, before we dive into the details of this larger issue:
- How is this change breaking? Note that the validation dataset here is optional
- It seems to me that with this change all estimators would support validation data during training

jaheba: There are estimators outside of GluonTS ;)

These would also need to be adapted, since they now need to take an additional argument.

jaheba: How about we merge this PR and we discuss that topic at a later point?

jaheba:
@strawberrypie I think disagree on the use of abbreviations for "validation". Instead I would propose to always use `validation` instead of `val` or `valid`, since in my view it is a bit cleaner.

src/gluonts/model/estimator.py
```diff
@@ -178,6 +186,18 @@ def train_model(self, training_data: Dataset) -> TrainOutput:
             float_type=self.float_type,
         )
 
+        validation_data_loader = (
+            InferenceDataLoader(
```
vafl: I'm not sure whether the `InferenceDataLoader` does the right thing here. It runs the transformations in non-training mode, which does not provide the prediction target that is needed to calculate the loss. I'm surprised this doesn't throw an error, since the field should be missing.

Also the signature in the trainer is given as `Optional[TrainDataLoader]`.

strawberrypie: Fixed the signature and missing prediction target by using training transforms in `ValidationDataLoader`. It turns out that with the current state of transformations it is impossible to choose the non-training mode for the InstanceSplitter transformation — so, the `ValidationDataLoader` provides the way to estimate validation loss on a separate dataset that the user provides. I think that validation scheme that @lostella provided [here](https://github.com/awslabs/gluon-ts/pull/378#discussion_r337042900) depends on the algorithms — there is no common `prediction_length` field that can be used to handle all of the algorithms that GluonTS can use. 

So, the user is forced to provide a correct validation set in accordance with the algorithms and problems that they need to solve. 

cc @jaheba

PR #381, opened 2019-10-10
Removed dataset/split.

jaheba:
AFAIK, this is not used and thus can be safely removed.


szha:
Job PR-381/1 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-381/1/index.html

jaheba:
> We might need the train/test splitter if we allow for computing validation loss during training.

I think this was meant as a tool to generally split datasets into train/test-sets. 

Maybe we can remove this for now and re-add it later when we need it again?

jaheba:
Oh, It's actually used in the example notebooks :|

lostella:
> Oh, It's actually used in the example notebooks :|

Did CI catch this? Anyway, I think something like this is needed and should be exposed (eg documented) better: computing train/validation/test splits is what gets people confused, especially if it involves some rolling-slicing of time series.

PR #383, opened 2019-10-10
Print network parameters after the forward pass to have the correct number of parameters

timoschowski:
*Issue #, if available:*
https://github.com/awslabs/gluon-ts/issues/310

*Description of changes:*
Trying to fix the above issue. My local build is not working, so I'm relying on the CI here.


timoschowski:
Ok, I need to fix my build first...

src/gluonts/trainer/_base.py
```diff
@@ -260,6 +254,14 @@ def base_path() -> str:
                                 else:
                                     loss = output
 
+                            # only invoke the following during the first forward pass
+                            if it == 1:
```
geoalgo: I think this would fail as 'it' is not comparable and that you want to make this test on 'batch_no'. Otherwise the change looks good to me.

Issue #278, opened 2019-09-02
Store GluonTS Version with stored Model


jaheba:
When we serialise a model, we should also store the current GluonTS version with it.

That way we can track backwards compatible or breaking changes.

timoschowski:
PR https://github.com/awslabs/gluon-ts/pull/388 is a rather minimal answer to the issue. I was thinking whether it makes sense to check it during deserialization, but this didn't lead anywhere.

jaheba:
As a follow-up, we should also read-out the version when we deserialise the model and do some checks on the version.

PR #388, opened 2019-10-12
address issue #278

timoschowski:
*Issue #, if available:*
#278
*Description of changes:*
This adds the version number of gluonts to the type.txt file during serialization, but doesn't do anything with this during deserialization. I tested this locally, but didn't add any unit tests. If I should, please point me to where I'd best do this.

### DIFF
--- a/src/gluonts/model/predictor.py
+++ b/src/gluonts/model/predictor.py
@@ -119,7 +119,9 @@ class Predictor:
         with (path / "type.txt").open("w") as fp:
             fp.write(fqname_for(self.__class__))
         with (path / "version.json").open("w") as fp:
-            json.dump({"model": self.__version__, "gluonts": gluonts.__version__}, fp)
+            json.dump(
+                {"model": self.__version__, "gluonts": gluonts.__version__}, fp
+            )
 
     @classmethod
     def deserialize(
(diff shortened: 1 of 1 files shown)


jaheba:
Thanks!

I think it would be better to create a new file to store the version into. At the moment you store it into `type.txt` which should just contain the class-path of the predictor.

Maybe we can add a `version.json` which could store two fields:

* the model's version (via `self.__version__`)
* the version of gluon-ts (`gluonts.__version__`)

For gluon-ts based algorithms, these are the same, but external algorithms can have independent ones. Thus, we should store both versions.

src/gluonts/model/predictor.py
```diff
@@ -21,6 +21,7 @@
 from pathlib import Path
 from pydoc import locate
 from tempfile import TemporaryDirectory
+import json
```
jaheba: I think normally we would want imports sorted alphabetically, but I don't mind too much.

szha:
Job PR-388/2 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-388/2/index.html

szha:
Job PR-388/3 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-388/3/index.html

szha:
Job PR-388/4 is complete. 
Docs are uploaded to http://gluon-ts-staging.s3-accelerate.dualstack.amazonaws.com/PR-388/4/index.html

PR #391, opened 2019-10-14
Stopped CI from posting doc-build links to PRs.

jaheba:
*Issue #, if available:*

*Description of changes:*


lostella:
This is actually useful when changes to docstrings or tutorial notebooks are made. Is there a way we can trigger this step on-demand?

jaheba:
The build is still made, but you have to get it from the logs which is not as easy as having it posted directly. 

I’m in favour of removing it because most of the time we don’t want to look at the logs. Maybe we could move it to GitHub Actions in the future and have it linked as a build step.

lostella:
@jaheba could you open an issue about this? We probably want to be able to trigger this in case a PR is concerned with docs formatting and all

PR #403, opened 2019-10-21
Dockerfile for GPU container. Fix for installing GPU version of MXNet

strawberrypie:
*Description of changes:*
  * fixes a bug in `setup.py` that didn't work with the current version of requirements
  * `Dockerfile.gpu` for building GPU-enabled Docker image


setup.py
```diff
@@ -46,10 +46,15 @@ def read(*names, encoding="utf8"):
...
+        mxnet_old = r"mxnet"
+        mxnet_new = "mxnet-cu92mkl" if GPU_SUPPORT else mxnet_old
         return [
-            line.rstrip().replace(mxnet_old, mxnet_new, 1)
+            re.subn(
+                pattern=mxnet_old,
+                repl=mxnet_new,
+                string=line.rstrip(),
+                count=1,
+            )[0]
```
jaheba: How does `re.subn` help over just `str.replace` here? We should maybe think about making this more robust.

But more importantly, I don't really like what we are doing here. Are there always compatible releases between `mxnet` and `mxnet-cu92mkl`? And, should we be more explicit with which version we install? However, we should discuss this probably in another issue.

strawberrypie: Well, I was using the substitution with regex like `mxnet[><=]?=`, but it looks like simple substitution in enough. Will change it to `str.replace`.

Regarding the choice of `mxnet` vs `mxnet-cu92mkl` — I think that we should do the same thing as with MXNet releases — separate MXNet, MXNet + CUDA, MXNet + MKL versions and Docker images.  
Personally, I would prefer to have a GPU version that could seamlessly switch to CPU if there are 0 GPUs found — that's how PyTorch works by default. @jaheba do you know if MXNet can work the same way? Currently, the GPU image fails to work on my device without the Nvidia GPU (Macbook Pro), throws the following error: 
```
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.7/dist-packages/gluonts/shell/__main__.py", line 27, in <module>
    from gluonts.model.estimator import Estimator
  File "/usr/local/lib/python3.7/dist-packages/gluonts/model/estimator.py", line 19, in <module>
    from mxnet.gluon import HybridBlock
  File "/usr/local/lib/python3.7/dist-packages/mxnet/__init__.py", line 24, in <module>
    from .context import Context, current_context, cpu, gpu, cpu_pinned
  File "/usr/local/lib/python3.7/dist-packages/mxnet/context.py", line 24, in <module>
    from .base import classproperty, with_metaclass, _MXClassPropertyMetaClass
  File "/usr/local/lib/python3.7/dist-packages/mxnet/base.py", line 213, in <module>
    _LIB = _load_lib()
  File "/usr/local/lib/python3.7/dist-packages/mxnet/base.py", line 204, in _load_lib
    lib = ctypes.CDLL(lib_path[0], ctypes.RTLD_LOCAL)
  File "/usr/lib/python3.7/ctypes/__init__.py", line 356, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libcuda.so.1: cannot open shared object file: No such file or directory
```

Dockerfile.gpu
```diff
@@ -0,0 +1,19 @@
...
+
+# We're emulating nvidia-smi to be able to build GPU container on devices without GPU
+RUN echo '#!/bin/bash\necho Building container with GPU support' > /usr/bin/nvidia-smi && \
+    chmod +x /usr/bin/nvidia-smi
+
+ADD . /gluonts
+
+RUN pip3.7 install /gluonts[shell]
+
+ENTRYPOINT ["python3.7", "-m", "gluonts.shell"]
```
jaheba:
```suggestion
ENTRYPOINT ["python3.7", "-m", "gluonts.shell"]

```

Can we have a new-line at the end?

setup.py
```diff
@@ -46,8 +44,8 @@ def read(*names, encoding="utf8"):
 
 def find_requirements(filename):
     with (ROOT / "requirements" / filename).open() as f:
-        mxnet_old = "mxnet=="
-        mxnet_new = "mxnet-cu92mkl==" if GPU_SUPPORT else mxnet_old
+        mxnet_old = r"mxnet"
```
jaheba:
```suggestion
        mxnet_old = "mxnet"
```

strawberrypie:
I've just started to experiment with GPU instances. I'm using DeepAR in my project and using `p2.xlarge` with this change is just a bit faster than `c5.4xlarge`. Sagemaker shows that only 20% of GPU is used. @jaheba can you suggest something to speed it up? Or should I create an issue to investigate GPU usage?

setup.py
jaheba: > Personally, I would prefer to have a GPU version that could seamlessly switch to CPU if there are 0 GPUs found — that's how PyTorch works by default. @jaheba do you know if MXNet can work the same way? Currently, the GPU image fails to work on my device without the Nvidia GPU (Macbook Pro), throws the following error:

Yes, I agree. Maybe @szha can help us out here.

jaheba:
> I've just started to experiment with GPU instances. I'm using DeepAR in my project and using `p2.xlarge` with this change is just a bit faster than `c5.4xlarge`. Sagemaker shows that only 20% of GPU is used. @jaheba can you suggest something to speed it up? Or should I create an issue to investigate GPU usage?

We've also seen no real performance benefit using GPUs with SageMaker DeepAR as well. There might be some performance increase when large batch sizes are used.

However, other models (e.g. Wavenet) should benefit much more from using GPUs than DeepAR does.

/cc @vafl

jaheba:
@strawberrypie oh, I think having a dedicated issues regarding GPUs would be great, thanks.

PR #407, opened 2019-10-23
bug multivariate gaussian

FadhelA:
There was a Validation error in the LowrankMultivariateGaussianOutput class: "transforms value is not a valid list"

*Description of changes:*
Write transformation as a list

### DIFF
--- a/src/gluonts/distribution/binned.py
+++ b/src/gluonts/distribution/binned.py
@@ -87,7 +87,7 @@ class Binned(Distribution):
 
     @property
     def batch_shape(self) -> Tuple:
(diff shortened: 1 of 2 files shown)


mbohlkeschneider:
Don't we have a test that should catch this? If no, I propose to add one with this PR.

lostella:
[This test](https://github.com/awslabs/gluon-ts/blob/4226aa3276240b28459e4d182c53001843e73495/test/distribution/test_distribution_output_shapes.py#L106) seems to be missing the `LowrankMultivariateGaussian` case @FadhelA you could add it there I guess

PR #413, opened 2019-10-28
 DeepStateEstimator: Make cardinality argument compulsory 

rshyamsundar:
*Description of changes:*

This PR tries to make sure that `DeepStateEstimator` is not misused :) The assumption is that `DeepStateEstimator` is usually applied to dataset of multiple time series that have at least one categorical feature (e.g., an *id* for each time series). This way comparison against classical state space models that fit a separate model to each time series in the dataset makes sense.

* Added missing smoke test for `DeepStateEstimator`.
* Added test to make sure that invalid settings with respect static categorical features raise exceptions.

### DIFF
--- a/src/gluonts/model/deepstate/_estimator.py
+++ b/src/gluonts/model/deepstate/_estimator.py
@@ -12,6 +12,7 @@
 # permissions and limitations under the License.
 
 # Standard library imports
+import numpy as np
 from typing import List, Optional
 
 # Third-party imports
@@ -82,6 +83,10 @@ class DeepStateEstimator(GluonEstimator):
         Frequency of the data to train on and predict
     prediction_length
         Length of the prediction horizon
+    cardinality
+        Number of values of each categorical feature.
+        This must be set by default unless ``use_feat_static_cat``
+        is set to `False` explicitly (which is NOT recommended).
     add_trend
         Flag to indicate whether to include trend component in the
         state space model
@@ -117,9 +122,6 @@ class DeepStateEstimator(GluonEstimator):
     use_feat_static_cat
         Whether to use the ``feat_static_cat`` field from the data
         (default: False)
-    cardinality
-        Number of values of each categorical feature.
-        This must be set if ``use_feat_static_cat == True`` (default: None)
     embedding_dimension
         Dimension of the embeddings for categorical features
         (default: [min(50, (cat+1)//2) for cat in cardinality])
@@ -135,6 +137,7 @@ class DeepStateEstimator(GluonEstimator):
         self,
         freq: str,
         prediction_length: int,
+        cardinality: List[int],
         add_trend: bool = False,
         past_length: Optional[int] = None,
         num_periods_to_train: int = 4,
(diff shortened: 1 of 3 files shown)


src/gluonts/model/deepstate/_estimator.py
```diff
@@ -166,15 +168,13 @@ def __init__(
...
-        ], "Elements of `cardinality` should be > 0"
-        assert embedding_dimension is None or [
-            e > 0 for e in embedding_dimension
-        ], "Elements of `embedding_dimension` should be > 0"
+        assert not use_feat_static_cat or np.prod(cardinality) > 1, (
+            f"Number of distinct values across static categorical features (i.e., prod(`cardinality`)) "
+            f"must be larger than 1 if `use_feat_static_cat=True`. But cardinality provided is: {cardinality}"
+        )
+        assert embedding_dimension is None or all(
+            [e > 0 for e in embedding_dimension]
```
lostella: no need to make it a list

src/gluonts/model/deepstate/_estimator.py
```diff
@@ -166,15 +168,13 @@ def __init__(
...
-        assert (cardinality is not None and use_feat_static_cat) or (
-            cardinality is None and not use_feat_static_cat
-        ), "You should set `cardinality` if and only if `use_feat_static_cat=True`"
-        assert cardinality is None or [
-            c > 0 for c in cardinality
-        ], "Elements of `cardinality` should be > 0"
-        assert embedding_dimension is None or [
-            e > 0 for e in embedding_dimension
-        ], "Elements of `embedding_dimension` should be > 0"
+        assert not use_feat_static_cat or np.prod(cardinality) > 1, (
```
lostella: I like `all(c > 1 for c in cardinality)` better since it's more explicit. With this change the `import numpy as np` also goes

src/gluonts/model/deepstate/_estimator.py
```diff
@@ -135,6 +137,7 @@ def __init__(
         self,
         freq: str,
         prediction_length: int,
+        cardinality: List[int],
```
lostella: So in case one wants to turn off `use_feat_static_cat` (even if this is not recommended) what should they pass as `cardinality`? Empty list? Would it make sense to keep this `cardinality: Optional[List[int]]` with no default, but with `None` as valid value?

src/gluonts/model/deepstate/_estimator.py
rshyamsundar: But that is not equivalent to the condition we want: we should still allow one of  the cardinalities to be equal to 1 if there are more than one categorical feature (e.g., see smoke test). This, for example, is useful if one wants to experiment on some slice of data by fixing one of the categorical features without changing the dataset.

I should have written this with `any` as you mentioned offline.

src/gluonts/model/deepstate/_estimator.py
rshyamsundar: I actually tried making it optional first but somehow I was able to instantiate the estimator without passing `cardinality` at all: 
`DeepStateEstimator(prediction_length=1, freq='H')`
 (although I did not explicitly make the  default equal to `None`).

It is bit strange but it looks like the `validated` decorator is passing None as default. This is not what we want because the the above default instantiation throws an exception, so I made it non-optional.

If `use_feat_static_cat == False`, then `cardinality` must be set to `[1]` explicitly, meaning all the time series belong to a single category.

PR #414, opened 2019-10-28
Bug fix: assert all elements in the list are True.

rshyamsundar:
*Issue #, if available:*

Fix for assertions regarding cardinalities and embedding dimensions.


jaheba:
If you use `all` you don't need to use the `[]`:

```python
        assert cardinality is None or all(
            c > 0 for c in cardinality
        ), "Elements of `cardinality` should be > 0"
```

PR #415, opened 2019-10-28
DeepStateEstimator: Some adjustments to hyperparameter settings

rshyamsundar:
*Description of changes:*

* Made the variance parameters configurable. Apart from upper bounds (which can be tuned in principle), I added lower bounds as well not for tuning but to ensure numerical stability (e.g., bounding these parameters away from zero might resolve some of the numerical issues with Cholesky factorization)

* Made the default number of epochs equal to that of `DeepAREstimator`

* Adjusted the default `past_length` for daily time series since the current value is too high; now it  trains on 4 years of daily data by default.


src/gluonts/distribution/lds.py
```diff
@@ -24,6 +27,11 @@
 from gluonts.support.linalg_util import jitter_cholesky
 
 
+class BoundedParam(NamedTuple):
+    lower_bound: float
+    upper_bound: float
+
+
```
benidis: Trivial but shouldn't we move this to a general utils module for general usage?

src/gluonts/model/deepstate/_estimator.py
```diff
@@ -57,8 +59,8 @@
 FREQ_LONGEST_PERIOD_DICT = {
     "M": 12,  # yearly seasonality
     "W-SUN": 52,  # yearly seasonality
-    "D": 365,  # yearly seasonality
-    "B": 365,  # yearly seasonality
+    "D": 31,  # monthly seasonality
+    "B": 31,  # monthly seasonality
```
benidis: `"B": 22`

src/gluonts/distribution/lds.py
rshyamsundar: yes, but earlier Lorenzo insisted that these things be kept local until we find an occasion to use them somewhere else.

src/gluonts/model/deepstate/_estimator.py
rshyamsundar: True..

src/gluonts/model/deepstate/_estimator.py
```diff
@@ -174,6 +187,42 @@ def __init__(
...
+            np.isfinite(
+                [
+                    noise_std.lower_bound,
+                    prior_cov.lower_bound,
+                    innovation.lower_bound,
+                    noise_std.upper_bound,
+                    prior_cov.upper_bound,
+                    innovation.upper_bound,
+                ]
+            )
```
lostella: All this could be made more compact with a list comprehension similar to a few lines below

src/gluonts/distribution/lds.py
```diff
@@ -473,40 +485,38 @@ class LDSArgsProj(mx.gluon.HybridBlock):
...
         )
 
-        self.innovation_factor = (
-            1.0 if innovation_ub is float("inf") else innovation_ub
-        )
-        self.noise_factor = (
-            1.0 if noise_std_ub is float("inf") else noise_std_ub
-        )
+        self.innovation = innovation_bounds
+        self.noise_std = noise_std_bounds
```
rshyamsundar: minor: change the names of the member variables as well to include "bounds" in the name.

lostella: Thanks! I missed that!

PR #417, opened 2019-10-29
Dirichlet for pr

FadhelA:
*Description of changes:*
Adding the Dirichlet distribution (multivariate distribution) and corresponding tests. For now it only works with simplefeedforward, to make it work with other models, the flag pick_imcomplete has to be exposed and set to false.


.gitignore
```diff
@@ -163,3 +163,4 @@ conda
 # not to be tracked
 src/gluonts/version.py
 
+/debug/
```
lostella: This should be removed (it's an absolute path?)

src/gluonts/model/deepar/_estimator.py
```diff
@@ -197,6 +197,7 @@ def __init__(
 
         self.num_parallel_samples = num_parallel_samples
 
+
```
lostella: This should go

test/distribution/test_distribution_inference.py
```diff
@@ -253,6 +262,39 @@ def test_multivariate_gaussian() -> None:
     ), f"Sigma did not match: sigma = {Sigma}, sigma_hat = {Sigma_hat}"
 
 
+@pytest.mark.timeout(30)
```
lostella: Why is this needed?

test/distribution/test_distribution_inference.py
```diff
@@ -253,6 +262,39 @@ def test_multivariate_gaussian() -> None:
...
+def test_dirichlet() -> None:
+    num_samples = 2000
+    dim = 3
+
+    alpha = np.random.gamma(shape=1, size=dim) + 0.5
```
lostella: It would maybe be better to use `@pytest.mark.parametrize` to provide `alpha` as an argument to this function

test/distribution/test_distribution_inference.py
```diff
@@ -253,6 +262,39 @@ def test_multivariate_gaussian() -> None:
...
+    distr = Dirichlet(alpha=mx.nd.array(alpha))
+    cov = distr.variance.asnumpy()
+
+    samples = distr.sample(num_samples)
+
+    alpha_hat = maximum_likelihood_estimate_sgd(
+        DirichletOutput(dim=dim),
+        samples,
+        init_biases=None,
+        hybridize=False,
```
lostella: Does this not work with `hybridize=True`? If it does work, then you can add `hybridize` to the test arguments and try both `False` and `True`.

test/distribution/test_distribution_sampling.py
```diff
@@ -161,6 +162,36 @@ def test_multivariate_sampling(distr, params, serialize_fn) -> None:
     )
 
 
+test_cases_dirichlet = [(Dirichlet, {"alpha": mx.nd.array([0.2, 0.4, 0.9])})]
+
+
+@pytest.mark.parametrize("distr, params", test_cases_dirichlet)
+@pytest.mark.parametrize("serialize_fn", serialize_fn_list)
+def test_dirichlet_sampling(distr, params, serialize_fn) -> None:
```
lostella: Could we have this and `test_multivariate_sampling` merged into a single test function with cases for both `MultivariateGaussian` and `Dirichlet`? I see there are some differences in how the covariance is tested, but maybe these can be unified, not sure

test/distribution/test_distribution_inference.py
lostella: This would also be nice for multivariate Gaussian ones

Issue #301, opened 2019-09-11
API: number of samples


lostella:
## Summary

All predictors currently produce probabilistic forecasts using sampling and the `SampleForecast` class. The number of samples that a predictor produces (and that are used for evaluating forecasts) can be set in a number of places with inconsistent argument names and is a bit confusing in my opinion.

## Current status of things

* `Estimator` types accept a `num_parallel_samples` constructor argument ([see for example DeepAR](https://github.com/awslabs/gluon-ts/blob/6ca5d5d67565c24b72899792b6325a25b1341051/src/gluonts/model/deepar/_estimator.py#L130)) which is used to configure the resulting predictor (DeepAR does it [here](https://github.com/awslabs/gluon-ts/blob/6ca5d5d67565c24b72899792b6325a25b1341051/src/gluonts/model/deepar/_estimator.py#L270)). This is purely for optimization purposes, in that the `GluonPredictor` that such estimators produce allow for customizing the number of samples at inference time, see [here](https://github.com/awslabs/gluon-ts/blob/6ca5d5d67565c24b72899792b6325a25b1341051/src/gluonts/model/predictor.py#L314-L326).
However, some estimators call this constructor argument `num_eval_samples` (see e.g. [DeepState](https://github.com/awslabs/gluon-ts/blob/6ca5d5d67565c24b72899792b6325a25b1341051/src/gluonts/model/deepstate/_estimator.py#L148) and [DeepFactor](https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/model/deep_factor/_estimator.py#L102)).
* `make_evaluation_predictions` accepts a `num_eval_samples` and forwards it to the predict method. However, this is sometimes hidden in `**kwargs` and silently ignored, [for example by NPTS](https://github.com/awslabs/gluon-ts/blob/6ca5d5d67565c24b72899792b6325a25b1341051/src/gluonts/model/npts/_predictor.py#L204-L228)

## Proposed API

The way the number of samples is managed should be made uniform. To me, it makes sense to follow the following convention:

* Gluon-based `Estimator` types **do** have the `num_parallel_samples` option specified at construction time, in case their forecasts are sample-based.
* `Predictor`s **do not** have any `num_XXX_samples` option in the constructor.
* `Predictor.predict` method accepts the `num_samples` argument (optional, should have a default value e.g. 100).
* Note that I'm using `num_samples` and not `num_eval_samples`, since the `eval` part will be implied by the context (in case evaluations are being done).
[...]

PR #421, opened 2019-11-05
num_sample variation fixes

benidis:
Fixes issue #301 (following #232).

### DIFF
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -613,7 +613,7 @@ class OutputType(str, Enum):
 
 
 class Config(pydantic.BaseModel):
-    num_eval_samples: int = pydantic.Schema(100, alias="num_samples")
+    num_samples: int = pydantic.Schema(100, alias="num_samples")
     output_types: Set[OutputType] = {"quantiles", "mean"}
     # FIXME: validate list elements
     quantiles: List[str] = ["0.1", "0.5", "0.9"]
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -613,7 +613,7 @@ class OutputType(str, Enum):
 
 
 class Config(pydantic.BaseModel):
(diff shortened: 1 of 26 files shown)


lostella:
Looks like this broke the tests for `shell`: looking at the logs it seems that the `num_samples` configuration is not catched and 100 samples (the default somewhere?) get produced instead of 4

src/gluonts/model/forecast.py
```diff
@@ -613,7 +613,7 @@ class OutputType(str, Enum):
 
 
 class Config(pydantic.BaseModel):
-    num_eval_samples: int = pydantic.Schema(100, alias="num_samples")
+    num_samples: int = pydantic.Schema(100, alias="num_samples")
```
jaheba: To ensure backwards compatibility, this should be:

```suggestion
    num_samples: int = pydantic.Schema(100, alias="num_eval_samples")
```

src/gluonts/model/npts/_predictor.py
```diff
@@ -169,11 +168,11 @@ def __init__(
         use_default_time_features: bool = True,
         num_default_time_features: int = 1,
         feature_scale: float = 1000.0,
-        num_parallel_samples: int = 100,
+        num_samples: int = 100,
```
lostella: My proposal was to remove this from the constructor of predictors, and keep it only as an argument to the `predict` method, see #301

src/gluonts/model/prophet/_predictor.py
```diff
@@ -140,34 +137,37 @@ def __init__(
 
         assert "uncertainty_samples" not in prophet_params, (
             "Parameter 'uncertainty_samples' should not be set directly. "
-            "Please use 'num_eval_samples' instead."
+            "Please use 'num_samples' instead."
```
lostella:
```suggestion
            "Please use 'num_samples' in the `predict` method instead."
```

PR #432, opened 2019-11-08
Added multivariate grouper and tests

mbohlkeschneider:
*Description of changes:*
This is the first pull request of bringing the GP Copula model to GluonTS master. 

This PR implements the multivariate grouper which allows the use of univariate datasets in the multivariate setting. It groups the univariate dataset into a multivariate dataset consisting of one multivariate time series (and performs alignments where necessary). It also supports rolling evaluation scenarios in the test dataset. 


src/gluonts/dataset/multivariate_grouper.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Rules for padding for training and test datasets can be specified by the
+    user.
+    """
+
+    @validated()
+    def __init__(
+        self,
+        max_target_dim: int = None,
+        num_test_dates: int = None,
```
rshyamsundar: Should not these be `Optional[int]`

src/gluonts/dataset/multivariate_grouper.py
```diff
@@ -0,0 +1,191 @@
...
+        train_fill_rule: Callable = np.mean,
+        test_fill_rule: Callable = lambda x: 0.0,
+    ) -> None:
+        self.num_test_dates = num_test_dates
+        self.max_target_dimension = max_target_dim
+        self.train_fill_function = train_fill_rule
+        self.test_fill_rule = test_fill_rule
+
+        self.first_timestamp = pd.Timestamp(2200, 1, 1, 12)
+        self.last_timestamp = pd.Timestamp(1800, 1, 1, 12)
```
rshyamsundar: It is better to define these constants in a separate file or outside of the class and call them `LATEST_FIRST_TIMESTAMP` and `OLDEST_LAST_TIMESTAMP`.

src/gluonts/dataset/multivariate_grouper.py
```diff
@@ -0,0 +1,191 @@
...
+          the test dataset mimics a rolling evaluation scenario). In this case,
+          the univariate dataset will be split into n multivariate time series,
+          where n is the number of evaluation dates. Again, the
+          time series will be grouped but only left padded. Note that the
+          padded value will influence the prediction if the context length is
+          longer than the length of the time series.
+
+    Rules for padding for training and test datasets can be specified by the
+    user.
+    """
```
rshyamsundar: please add annotations for constructor parameters here.

src/gluonts/dataset/multivariate_grouper.py
mbohlkeschneider: Yes, thanks!

src/gluonts/dataset/multivariate_grouper.py
mbohlkeschneider: Done.

src/gluonts/dataset/multivariate_grouper.py
mbohlkeschneider: Makes sense. Not sure what the best (aka not confusing) naming is, but this would be my suggestion.

src/gluonts/dataset/multivariate_grouper.py
```diff
@@ -44,13 +45,32 @@ class MultivariateGrouper:
...
+
+    Parameters
+    ----------
+    max_target_dim
+        Set maximum dimensionality (for faster testing or when hitting
+        constraints of multivariate model). Takes the last max_target_dim
+        time series and groups them to multivariate time series.
+    num_test_dates
+        Number of test dates in the test set. This can be more than one if
+        the test set contains more then one forecast start dates (often the
```
rshyamsundar: more then -> more than

mbohlkeschneider: oops.

PR #433, opened 2019-11-08
Fixes to setup.py to make it work on Windows

### LINKED ISSUE #85: Investigate build process on Windows
The package build seems to be failing on Windows, apparently once again due to some hard-coded path.

Context is in #75

lostella:
*Issue #, if available:* Fixes #85


jaheba:
I think we should also start to use windows in our CI build.

Issue #454, opened 2019-11-16
Move unit tests from jenkins to travis-ci


tomzx:
It appears that currently CI is done on a private jenkins instance somewhere. I haven't found any documentation leading to an instance where it would be possible to verify that a PR does not break the build.

I've also observed that the project has a configured `.travis.yml` for the task of deploying to PyPI.

Given that GluonTS is open source project, it would be great if contributors could benefits from CI features.

As such, it would be great if it was possible to access the jenkins instance used to do CI, that it supported testing PR from forks, or to use travis and add cpu unit tests. This would also take care of verifying whether various versions of python and/or the latest package dependencies will work with GluonTS.

tomzx:
I just realized you guys use github actions and that your jenkins CI is available. This request is unlikely to be necessary.

PR #455, opened 2019-11-16
Test on travis-ci

tomzx:
*Issue #:* #454

*Description of changes:* This adds running the tests on travis for python 3.6, 3.7 and 3.8.

It currently fails to build under 3.8 because of errors such as

```
AttributeError: type object 'pandas._libs.tslibs.conversion._TSObject' has no attribute '__reduce_cython__'
```

`mypy` was bumped to 0.740 in order to fix an issue where `typed-ast` 1.1.2 would be installed, and would build from source under python 3.8 (no wheels available) and it would fail the build.

Example of a build: https://travis-ci.org/TomzxForks/gluon-ts/builds/612933608

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,15 @@
 dist: xenial
 language: python
 python:
-- '3.7'
-install: true
-script: true
+  - 3.6
+  - 3.7
+  - 3.8
+cache: pip
+install:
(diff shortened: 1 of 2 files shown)


tomzx:
I just realized you guys use github actions and that your jenkins CI is available. This PR is unlikely to be necessary.

PR #456, opened 2019-11-16
Relax requirements constraints

tomzx:
This allows gluon-ts to be installed in a project that requires
similar dependencies. Because gluon-ts dependencies constraints
are too strict, it may prevent it from being installed side by
side with those other libraries if the constraints cannot resolve
to a common version.

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -2,11 +2,11 @@
 # see: https://github.com/boto/botocore/commit/e87e7a745fd972815b235a9ee685232745aa94f9
 python-dateutil==2.8.0
 
-boto3==1.*
-holidays==0.9.*
-matplotlib==3.*
-numpy==1.14.*
-pandas>=0.25.0
-pydantic==0.28.*
-tqdm>=4.23.0
-ujson>=1.35
+boto3>=1.0,<2
+holidays>=0.9,<0.10
+matplotlib>=3.0,<4
+numpy>=1.14,<2
+pandas>=0.25,<0.26
+pydantic>=0.28,<2
+tqdm>=4.23,<5
+ujson>=1.3,<2


requirements/requirements.txt
```diff
@@ -2,11 +2,11 @@
...
-pandas>=0.25.0
-pydantic==0.28.*
-tqdm>=4.23.0
-ujson>=1.35
+boto3>=1.0,<2
+holidays>=0.9,<0.10
+matplotlib>=3.0,<4
+numpy>=1.14,<2
+pandas>=0.25,<0.26
+pydantic>=0.28,<2
```
tomzx: The constraint is not `pydantic>=0.28,<1` here because `pydantic` now has versions 1.x which it would be nice to support. Furthermore, I'd like in a subsequent PR to bump `pydantic` to support one of the stable versions.

At the moment, with `pydantic` 1.1, the existing code still works but emits a few warnings that I'll fix in the PR associated with the `pydantic` version bump.

requirements/requirements.txt
```diff
@@ -2,11 +2,11 @@
...
 
-boto3==1.*
-holidays==0.9.*
-matplotlib==3.*
-numpy==1.14.*
-pandas>=0.25.0
-pydantic==0.28.*
-tqdm>=4.23.0
-ujson>=1.35
+boto3>=1.0,<2
```
jaheba:
```suggestion
boto3~=1.0
```

requirements/requirements.txt
```diff
@@ -2,11 +2,11 @@
...
+holidays>=0.9,<0.10
+matplotlib>=3.0,<4
```
jaheba:
```suggestion
matplotlib~=3.0
```

requirements/requirements.txt
```diff
@@ -2,11 +2,11 @@
...
+numpy>=1.14,<2
```
jaheba:
```suggestion
numpy~=1.14
```

requirements/requirements.txt
```diff
@@ -2,11 +2,11 @@
...
+pandas>=0.25,<0.26
```
jaheba: Do we have issues with 0.26?

lostella: I think this is OK for now: pandas 0.26 does not exist yet, and we can bump this manually when it comes out

PR #459, opened 2019-11-18
Added aggregation functionality to MultivariateEvaluator

mbohlkeschneider:
*Description of changes:*

Second pull request for the multivariate model. This ports over the ability of the `MultivariateEvaluator` to evaluate on aggregates of the target and forecast (the aggregates should usually be more consistent in a multivariate model than in an univariate model). 

### DIFF
--- a/src/gluonts/evaluation/_base.py
+++ b/src/gluonts/evaluation/_base.py
@@ -16,7 +16,17 @@ import logging
 import re
 from functools import lru_cache
 from itertools import chain, tee
-from typing import Any, Dict, Iterable, Iterator, List, Optional, Tuple, Union
+from typing import (
+    Any,
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Tuple,
+    Union,
+    Callable,
+)
 
 # Third-party imports
 import numpy as np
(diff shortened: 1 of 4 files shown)


src/gluonts/evaluation/_base.py
```diff
@@ -554,4 +625,21 @@ def __call__(
...
         )
 
+        if len(self.target_agg_funcs) != 0:
+            multivariate_metrics = {
+                key_value[0]: self.calculate_aggregate_multivariate_metrics(
+                    ts_iterator_set[-(index + 1)],
+                    fcst_iterator_set[-(index + 1)],
+                    key_value[1],
+                )
+                for index, key_value in enumerate(
```
rshyamsundar: we can write `key_value` as a pair like this:
```
for index, (key, value) in enumerate(
```

src/gluonts/evaluation/_base.py
```diff
@@ -554,4 +625,21 @@ def __call__(
             all_agg_metrics, all_metrics_per_ts
         )
 
+        if len(self.target_agg_funcs) != 0:
```
rshyamsundar: minor: one can check if the `dict` is not empty as follows (maybe also better readable):
```
if self.target_agg_funcs:
```

src/gluonts/evaluation/_base.py
rshyamsundar: Also maybe rename `(key, value)` as (`agg_fun_name`, `agg_fun`).

src/gluonts/evaluation/_base.py
mbohlkeschneider: Nice! Thank you!

src/gluonts/evaluation/_base.py
mbohlkeschneider: Thanks! Good suggestion!

src/gluonts/model/forecast.py
```diff
@@ -266,6 +266,19 @@ def copy_dim(self, dim: int):
         """
         raise NotImplementedError()
 
+    def copy_aggregate(self, agg_fun: Callable):
```
lostella: You could annotate the return type here

src/gluonts/model/forecast.py
```diff
@@ -399,6 +413,20 @@ def copy_dim(self, dim: int):
             info=self.info,
         )
 
+    def copy_aggregate(self, agg_fun: Callable):
```
lostella: And here

src/gluonts/evaluation/_base.py
```diff
@@ -488,29 +527,57 @@ def get_eval_dims(self, target_dimensionality: int) -> List[int]:
...
         all_agg_metrics
             dictionary with aggregate metrics of individual dimensions
         all_metrics_per_ts
-            DataFrame containing metrics for all time series of all evaluated dimensions
+            DataFrame containing metrics for all time series of all evaluated
+            dimensions
 
         Returns
         -------
-        Dict[int, Dict[str, float]]
```
lostella: I think the type annotation can stay here, see https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

src/gluonts/evaluation/_base.py
```diff
@@ -488,29 +527,57 @@ def get_eval_dims(self, target_dimensionality: int) -> List[int]:
...
+            self.extract_aggregate_forecast(forecast_iterator, agg_fun),
+        )
+        return agg_metrics
+
     def calculate_aggregate_vector_metrics(
         self,
         all_agg_metrics: Dict[int, Dict[str, float]],
         all_metrics_per_ts: pd.DataFrame,
-    ):
+    ) -> Dict[int, Dict[str, float]]:
```
lostella: I’m wondering if the gluonts.evaluation makes mypy happy? You could run mypy against it, and if everything is fine mark it as typesafe by adding the `.typesafe` file into it

mbohlkeschneider: Yeah, the typing of this module is a bit messy. I'll refactor a bit and clean it up. Thank you for the comment!

PR #462, opened 2019-11-19
remove constraints on numpy version

### LINKED ISSUE #461: Upgrade required Numpy version
Is there a reason that Numpy is pinned to version 1.14.* (released mid. 2018)? This creates difficulties for resolving dependencies in larger projects when it is so strictly constrained.

lostella:
*Issue #, if available:* fixes #461 

*Description of changes:*

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -5,7 +5,7 @@ python-dateutil==2.8.0
 boto3==1.*
 holidays==0.9.*
 matplotlib==3.*
-numpy==1.14.*
+numpy>=1.14.0
 pandas>=0.25.0
 pydantic==0.28.*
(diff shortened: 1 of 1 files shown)


requirements/requirements.txt
```diff
@@ -5,7 +5,7 @@ python-dateutil==2.8.0
 boto3==1.*
 holidays==0.9.*
 matplotlib==3.*
-numpy==1.14.*
+numpy>=1.14.0
```
jaheba:
```suggestion
numpy>=1.14
```

Should we do this in combination with #456?

lostella: Sure, let's get that one through maybe (somehow I hadn't seen it)

PR #465, opened 2019-11-20
Updated pydantic to version 1.

jaheba:
Related to #456.

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -7,6 +7,6 @@ holidays==0.9.*
 matplotlib==3.*
 numpy==1.14.*
 pandas>=0.25.0
-pydantic==0.28.*
+pydantic~=1
 tqdm>=4.23.0
 ujson>=1.35
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -7,6 +7,6 @@ holidays==0.9.*
 matplotlib==3.*
 numpy==1.14.*
 pandas>=0.25.0
-pydantic~=1
+pydantic~=1.1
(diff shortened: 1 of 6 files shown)


requirements/requirements.txt
```diff
@@ -2,11 +2,11 @@
...
-matplotlib>=3.0,<4
-numpy>=1.14,<2
+matplotlib~=3.0
+numpy~=1.14
 pandas>=0.25,<0.26
-pydantic>=0.28,<2
-tqdm>=4.23,<5
-ujson>=1.3,<2
+pydantic~=1.1
+tqdm~=4.23,<5
```
lostella:
```suggestion
tqdm~=4.23
```

requirements/requirements.txt
```diff
@@ -2,11 +2,11 @@
...
+ujson~=1.35,<2
```
lostella:
```suggestion
ujson~=1.35
```

requirements/requirements.txt
lostella: (aren't these equivalent?)

lostella:
The following needs to go in `gluonts.core.component`, lines 344-346, in order to get rid of the doctest complaint:
```python
    pydantic.error_wrappers.ValidationError: 1 validation error for ComplexNumberModel
    y
      none is not an allowed value (type=type_error.none.not_allowed)
```

requirements/requirements.txt
jaheba: Yes, but IMO `~=` looks a bit nicer.

PR #480, opened 2019-11-27
Specify --upgrade in README.md install example

leezu:
Try to avoid user problems such as https://github.com/apache/incubator-mxnet/issues/16924


jaheba:
@leezu Do you disagree?

leezu:
@jaheba agree about mxnet, there is no version specifier for `gluonts`, so `gluonts` wouldn't update.

PR #482, opened 2019-11-27
Dirichlet Multinomial distribution

FadhelA:
*Description of changes:*
Implementation of the Dirichlet Multinomial distribution, with corresponding tests.


lostella:
The inference test seems to be failing, you can try locally how frequently this occurs if you change the random seed, and maybe play with the stepsize to try to make this test more robust.

The same could be done for Dirichlet, I also saw that test failing every now and then

Issue #470, opened 2019-11-21
Update numpy in requirements.txt to avoid pip (warning) conflicts 


lcoulet:
## Description
GluonTS currently states a depency towards Numpy 1.14.x (`~=1.14`)

numpy 1.14 is quite outdated, so when used with other libraries it generates a conflict in pip.
Can numpy be updated to latest (e.g. >=1.17.3), or simply to `>=1.14`?

mbohlkeschneider:
I don't see an issue in doing so. Any concerns, @lostella ?

lostella:
Sure, this makes perfect sense, and was already addressed in #456 (see also #465). The relaxed requirements will be part of the next bugfix release, likely to be issued tomorrow. For now, you will find the changes in the master branch if you want to try it out.

lostella:
Sorry, I misread the issue: to be clear, I think `~=1.14` is fine in that it is equivalent to `>= 1.4, == 1.*` (where the comma means "and").

See: https://www.python.org/dev/peps/pep-0440/#compatible-release

lostella:
@lcoulet `mxnet==1.4.1` has a dependency on `numpy<1.15`, which is what may be causing trouble when installing the package from scratch. You can upgrade to `mxnet>=1.5.0` if it works for you, we saw that it had some issues on Linux, so it's not our recommended version. But if you're on macOS or Windows than it should be fine.

jaheba:
I've tried installing installing numpy to `1.17` and all tests pass. I guess, you should be fine with just upgrading numpy without the need for mxnet.

mbohlkeschneider:
I've also not seen any issues with upgrading numpy. In fact, we have run all our experiments for the GluonTS port of the Gaussian Copula paper on numpy 1.16 and it was fine.

mbohlkeschneider:
There will be error messages from mxnet with newer numpy versions, so we decided to keep it as it is until we can upgrade mxnet.

PR #483, opened 2019-11-28
Bump up numpy version

mbohlkeschneider:
*Issue #, if available:*
Related to #470 . Bumping up numpy because: 

1. Users ask for it. 
2. We will need the numpy `take_along_axis` operator for the marginal CDF transformation of the Gaussian Copula model that comes in numpy version `1.16)

*Description of changes:*
Bumping up numpy version. 

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -5,7 +5,7 @@ python-dateutil==2.8.0
 boto3~=1.0
 holidays>=0.9,<0.10
 matplotlib~=3.0
-numpy~=1.14
+numpy~=1.16
 pandas>=0.25,<0.26
 pydantic~=1.1
 tqdm~=4.23


jaheba:
I think the issue is that mxnet has a lower version requirement for numpy and thus pip will complain. Can we wait for mxnet 1.5.2?

lostella:
I think this change is not necessary, since `~=1.14` is equivalent to `>=1.14 and ==1.*`

More here: https://www.python.org/dev/peps/pep-0440/#compatible-release

mbohlkeschneider:
It looks that the builds use numpy 1.14, no (as evidenced here: https://github.com/awslabs/gluon-ts/pull/483/checks?check_run_id=324352451) Thus, they would fail if you would have something with higher dependency?

jaheba:
No, I think it's using 1.16. But it logs an error: https://github.com/awslabs/gluon-ts/pull/483/checks?check_run_id=324352451#step:4:89

mbohlkeschneider:
I guess I don't understand the interactions then. [This build](https://github.com/awslabs/gluon-ts/pull/476/checks?check_run_id=322821447) (without the change) does suggest that it would take numpy `1.14` (unless I'm miskaten again). So it DOES matter?

mbohlkeschneider:
> I think the issue is that mxnet has a lower version requirement for numpy and thus pip will complain. Can we wait for mxnet 1.5.2?

I'd like to have the gaussian copula model merged to mainline until the end of next week for NeurIPS... :-(.

mbohlkeschneider:
I'll monkey patch this for now. I can implement the operator differently by monkey patching, although it will be quite a bit (~5 times) slower than the native numpy operator that comes with numpy `1.16`. We should bump this up as soon as mxnet let us do it...

jaheba:
Maybe we should track issues depending on mxnet 1.5.2?

PR #491, opened 2019-12-03
Introduces DeepVAR model

mbohlkeschneider:
*Description of changes:*

Introduces the DeepVAR model for multivariate forecasts. This is the basic abstraction leading up to the GP Copula model (which comes in the next PR). 

### DIFF
--- a/src/gluonts/model/deepvar/_estimator.py
+++ b/src/gluonts/model/deepvar/_estimator.py
@@ -1,3 +1,16 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
 # Standard library imports
 from typing import List, Optional, Callable, Tuple
 
--- a/src/gluonts/model/deepvar/_estimator.py
+++ b/src/gluonts/model/deepvar/_estimator.py
@@ -9,7 +9,11 @@ from mxnet.gluon import HybridBlock
 
 # First-party imports
 from gluonts.core.component import validated
-from gluonts.distribution import DistributionOutput, StudentTOutput
+from gluonts.distribution import (
+    DistributionOutput,
+    StudentTOutput,
+    LowrankMultivariateGaussianOutput,
+)
 from gluonts.model.estimator import GluonEstimator
 from gluonts.model.predictor import Predictor, RepresentableBlockPredictor
 from gluonts.support.util import copy_parameters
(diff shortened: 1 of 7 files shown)


src/gluonts/model/deepvar/_estimator.py
```diff
@@ -0,0 +1,338 @@
...
+        self.freq = freq
+
+    def __call__(self, index: pd.DatetimeIndex) -> np.ndarray:
+        values = getattr(index, self.freq)
+        num_values = max(values) + 1
+        steps = [x * 2.0 * np.pi / num_values for x in values]
+        return np.vstack([np.cos(steps), np.sin(steps)])
+
+
+def get_granularity(freq_str: str) -> Tuple[int, str]:
```
jaheba: Don't we have this code elsewhere as well?

src/gluonts/model/deepvar/_network.py
```diff
@@ -0,0 +1,822 @@
...
+            *lagged_values, num_args=len(indices), dim=1
+        ).transpose(axes=(0, 2, 3, 1))
+
+    def unroll(
+        self,
+        F,
+        lags: Tensor,
+        scale: Tensor,
+        time_feat: Tensor,
+        target_dimensions: Tensor,
```
lcallot: It's kind of confusing that target_dim and target_dimension refer to two fairly different things.

mbohlkeschneider: Good point. I'll try to make that clearer. I think the confusion stems mostly from not being really relevant to this model - maybe I'll factor this over to the GP model once, which I will work on next.

mbohlkeschneider: I refactored a bit and put some description to the estimator constructor. Hope that is clearer.

src/gluonts/model/deepvar/_estimator.py
mbohlkeschneider: Not exactly, AFAIK. 

One can use the `to_offset` function from pandas, though, to produce the same thing.

mbohlkeschneider: Not exactly. One can use the `to_offset` function from pandas to do the same thing, though.

test/model/test_deepvar.py
```diff
@@ -102,9 +102,9 @@ def test_deepvar(
         freq=metadata.freq,
         distr_output=distr_output,
         scaling=False,
-        use_copula=use_copula,
+        use_marginal_transformation=use_copula,
```
lcallot: you should probably update use_copula here for consistency.

test/model/test_deepvar.py
```diff
@@ -46,7 +46,7 @@ def load_multivariate_constant_dataset():
 estimator = DeepVAREstimator
 
 
-@pytest.mark.timeout(20000)
+@pytest.mark.timeout(10)
 @pytest.mark.parametrize(
     "distr_output, num_batches_per_epoch, Estimator, hybridize, use_copula",
```
lcallot: copula -> marginal_transform

test/model/test_deepvar.py
mbohlkeschneider: Missed to update that file. Fixed. Thank you!

test/model/test_deepvar.py
mbohlkeschneider: See above.

PR #500, opened 2019-12-06
Added strategies for timezone information.

jaheba:
Currently, we throw an error if we encounter tz-information in a start-field.

The idea is to have different strategies dealing with such time-information:

* error -- current behaviour
* utc -- get time in UTC, e.g. if time zone is eastern time, we add 5 hours to the time and then remove the tz information
* ignore -- removes the timezone information

Maybe we could use `global` for `utc` and `local` for `ignore` instead.

### DIFF
--- a/src/gluonts/dataset/common.py
+++ b/src/gluonts/dataset/common.py
@@ -13,6 +13,7 @@
 
 # Standard library imports
 import shutil
+from enum import Enum
 from functools import lru_cache
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


jaheba:
Todos:

* tests
* add a way one can specify that information for a dataset

jaheba:
Somehow the notebook failed to build on jenkins, but I will merge this now anyhow.

PR #502, opened 2019-12-08
Gamma distribution implemented.

AaronSpieler:
This is related to Issue https://github.com/awslabs/gluon-ts/issues/467.


src/gluonts/distribution/gamma.py
```diff
@@ -0,0 +1,144 @@
...
+    def stddev(self) -> Tensor:
+        return self.F.sqrt(self.alpha) / self.beta
+
+    def sample(
+        self, num_samples: Optional[int] = None, dtype=np.float32
+    ) -> Tensor:
+        return _sample_multiple(
+            partial(self.F.sample_gamma, dtype=dtype),
+            alpha=self.alpha,
+            beta=1.0 / self.beta,  # because MXNet doesnt make sense
```
lostella:
```suggestion
            beta=1.0 / self.beta,
```

src/gluonts/distribution/gamma.py
```diff
@@ -0,0 +1,144 @@
...
+        self, num_samples: Optional[int] = None, dtype=np.float32
+    ) -> Tensor:
+        def s(alpha: Tensor, beta: Tensor) -> Tensor:
+            return self.F.sample_gamma(alpha=alpha, beta=beta, dtype=dtype)
+
+        return _sample_multiple(
+            s,
+            alpha=self.alpha,
+            beta=1.0 / self.beta,
+            num_samples=num_samples,  # because MXNet doesnt make sense
```
lostella:
```suggestion
            num_samples=num_samples,
```

src/gluonts/distribution/gamma.py
```diff
@@ -0,0 +1,144 @@
...
+    def sample_rep(
+        self, num_samples: Optional[int] = None, dtype=np.float32
+    ) -> Tensor:
+        def s(alpha: Tensor, beta: Tensor) -> Tensor:
+            return self.F.sample_gamma(alpha=alpha, beta=beta, dtype=dtype)
+
+        return _sample_multiple(
+            s,
+            alpha=self.alpha,
+            beta=1.0 / self.beta,
```
lostella: I would pass `self.beta` here, and invert it inside of `s` above

lostella: Nevermind, disregard my comment

src/gluonts/distribution/gamma.py
```diff
@@ -0,0 +1,144 @@
...
+            num_samples=num_samples,  # because MXNet doesnt make sense
+        )
```
lostella: I think you can remove this definition for now, since it's the same as `sample`. The idea of having `sample_rep` is to be able to differentiate through samples.

In the case of Gaussian, you can express samples from a generic Gaussian distribution by scaling and translating samples from a standard normal (see [here](https://github.com/awslabs/gluon-ts/blob/fbcb244da611767ecf961955a27ad0f39b029133/src/gluonts/distribution/gaussian.py#L103-L107)): this allows you e.g. to get the derivative of the samples wrt mu and sigma.

I'm not sure this can be done for Gamma (it can be done wrt the beta parameter, not sure about alpha).

src/gluonts/distribution/gamma.py
```diff
@@ -0,0 +1,144 @@
...
+    Parameters
+    ----------
+    alpha
+        Tensor containing the shape parameters, of shape `(*batch_shape, *event_shape)`.
+    beta
+        Tensor containing the rate parameters, of shape `(*batch_shape, *event_shape)`.
+    F
+    """
+
+    is_reparameterizable = True
```
lostella: As a side effect of my other comment below, this can be set to `False`

src/gluonts/distribution/gamma.py
AaronSpieler: For large Alpha there is a convenient way to sample based on the Normal Distribution. 
In general however, sampling from a Gamma distribution is more involved(see: https://www.hongliangjie.com/2012/12/19/how-to-generate-gamma-random-variables/ or http://home.iitk.ac.in/~kundu/paper120.pdf) which entails loops and ifs.
So I could implement one of the methods, but it would be probably painfully inefficient in python.

So ok, I"ll remove sample_rep.

src/gluonts/distribution/gamma.py
```diff
@@ -0,0 +1,144 @@
...
+            Tensor of shape `(*batch_shape, 1)`
+
+        Returns
+        -------
+        Tuple[Tensor, Tensor]
+            Two squeezed tensors, of shape `(*batch_shape)`: both have entries mapped to the
+            positive orthant.
+        """
+        alpha = softplus(F, alpha)
+        beta = softplus(F, beta)
```
lostella: These values can get arbitrarily close to zero, which can create problems when computing the `log_prob`. We had similar issues with `NegativeBinomial`, and there we add some small constant to make sure we're away from zero: https://github.com/awslabs/gluon-ts/blob/4bc1f36a076ba44f12568036221c75506b0cd5b4/src/gluonts/distribution/neg_binomial.py#L112-L113

We have a test for `NegativeBinomial` that triggers that problem, maybe you can check if something similar occurs here: https://github.com/awslabs/gluon-ts/blob/master/test/distribution/test_negative_binomial.py

src/gluonts/distribution/gamma.py
lostella: Yeah, I think it's OK for now to just remove it and set `is_reparameterizable = False`

src/gluonts/distribution/gamma.py
AaronSpieler: You were right, the test failed for the Gamma Distribution as well.
We probably should do the tests for the other cases of using `softplus`.

lostella:
Nice, thanks! One thing you could do, just to see it working, is to try out some model using GammaOutput. For example, you can take the quick start example from the README and make sure that runs through

PR #505, opened 2019-12-09
Added operator to allow for easier chaining of transformations.

jaheba:
Makes it easier to use multiple transformations.

```python
chain = SetField("foo", 42) + SetField("bar", 99)
```

We could think about using a different operator than `+`, but `+` is probably most fitting.


jaheba:
I guess something like `<<` could also work, but looks probably cooler than it actually is.

PR #510, opened 2019-12-10
Added shift operator to Transformation to apply it to a dataset.

jaheba:
I think from the point of view of transformations, we be more generic and handle every stream of `DataEntry` (which is just a dict):

```python
# Dictionary used for data flowing through the transformations.
# A Dataset is an iterable over such dictionaries.
DataEntry = Dict[str, Any]
```

For example, `[{"foo": "bar"}]` is a valid dataset in the eyes of transformations.

---

Also, I like to use `{dataset} >> {transformation}` to apply a transformation to a dataset:

```python
dataset = [{}]
t = SetField("foo", 42) + SetField("bar", 99)
list(dataset >> t) == [{'foo': 42, 'bar': 99}]

# as a one liner
list([{}]>> SetField("foo", 42) + SetField("bar", 99)) == [{'foo': 42, 'bar': 99}]
```

---

Further, I propose to use `>` to apply a simple transformation:

```python
{} > SetField("foo" 42) == {'foo': 42}
```

---

### DIFF
--- a/src/gluonts/transform/_base.py
+++ b/src/gluonts/transform/_base.py
@@ -14,7 +14,7 @@
 # Standard library imports
 import abc
 from functools import reduce
-from typing import Callable, Iterator, List
+from typing import Callable, Iterable, Iterator, List
 
 # First-party imports
 from gluonts.core.component import validated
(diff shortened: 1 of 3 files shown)


src/gluonts/transform/_base.py
```diff
@@ -39,11 +39,46 @@ def estimate(self, data_it: Iterator[DataEntry]) -> Iterator[DataEntry]:
...
+    """
+
+    def __init__(
+        self, dataset: Iterator[DataEntry], transformation: Transformation
+    ) -> None:
+        self.dataset = dataset
+        self.transformation = transformation
+
+    def __iter__(self) -> Iterator[DataEntry]:
+        yield from self.transformation(self.dataset, is_train=True)
```
vafl: Fixing the `is_train` here seems like a problem.

vafl:
I'm ok with the syntactic sugar and I also think that we need to rethink the transformations in general.

lostella:
> Further, I propose to use `>` to apply a simple transformation:
> 
> ```python
> {} > SetField("foo" 42) == {'foo': 42}
> ```

My understanding is that here the difference between `>>` and `>` is not in the transformation which is being applied (simple or not), but in it being applied to a collection vs single data entry.

Essentially, `>> = map(>, .)`. Is my understanding correct?

src/gluonts/transform/_base.py
```diff
@@ -39,11 +39,46 @@ def estimate(self, data_it: Iterator[DataEntry]) -> Iterator[DataEntry]:
...
+        return Chain([self, other])
 
     def __add__(self, other: "Transformation") -> "Chain":
         return self.chain(other)
 
+    def __lshift__(self, dataset: Iterator[DataEntry]) -> Iterator[DataEntry]:
+        return TransformedDataset(dataset, self)
+
+    def __rrshift__(self, dataset: Iterator[DataEntry]) -> Iterator[DataEntry]:
+        return self << dataset
```
lostella: I'm getting a headache understanding this

lostella: Forget it, I got it

lostella: Seems like `<<` is used only here. Do you think it could be used elsewhere? Otherwise I would avoid defining `__lshift__` at all, and just use `TransformedDataset(dataset, self)` here

jaheba:
The issue is that the generic Transformation works on a stream of entries, and that MapTransform (and SimpleTransformation) is a special case.

Only the latter can be applied to a single item, whilst others (e.g. a filter) can only be used with the entire stream. That’s why I’ve defined > on them.

src/gluonts/transform/_base.py
jaheba: The idea was to define >> in terms of <<. 

So that you can do `data >> t` and `t << data`.

lostella:
> That’s why I’ve defined > on them.

I see. You actually defined `__lt__`: does `left.__gt__(right)` default to `right.__lt__(left)` I suppose?

jaheba:
Yes, that one works out of the box.

PR #511, opened 2019-12-10
Added Transform.contextualize.

jaheba:
This is a replacement for `is_train`, which we currently pass.

The idea is that one can "contextualize" a transformation, i.e. adding a context to it:

```python
transformation_with_context = transformation.contextualize(is_train=True)
```

Within the transformation, this sets the `ctx` attribute:

```python
self.ctx["is_train"]
```

When one tries to use the ctx when none exists, an error gets thrown:

> Use of `.ctx` of uncontextualized Transformation.


lostella:
I’m not sure I like this.

As far as I can see, this mechanism (just like the current one based on the `is_train` option) is used to switch the behavior of a very limited set of transformations, which should behave differently in two cases: it seems to me that the most straightforward way to get that is to have two different transformations.

vafl:
The original idea of having `is_train` is that you build up your pipeline from blocks and the pipeline works for training as well as inference. If each block is tested to be consistent between training and inference then the whole pipeline will work consistently, too. It, arguably, helps avoiding these bugs. If we feel it is making things too complicated, we can rethink this.

alexw91:
# [Codecov](https://codecov.io/gh/awslabs/gluon-ts/pull/511?src=pr&el=h1) Report
> Merging [#511](https://codecov.io/gh/awslabs/gluon-ts/pull/511?src=pr&el=desc) into [master](https://codecov.io/gh/awslabs/gluon-ts/commit/357c8638194a68f8c03d5a09e93f4d69655f27d8&el=desc) will **decrease** coverage by `0.06%`.
> The diff coverage is `n/a`.


[Impacted file tree graph](https://codecov.io/gh/awslabs/gluon-ts/pull/511?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #511      +/-   ##
==========================================
- Coverage   83.72%   83.65%   -0.07%     
==========================================
  Files         181      181              
  Lines       10326    10339      +13     
==========================================
+ Hits         8645     8649       +4     
- Misses       1681     1690       +9     
```

alexw91:
# [Codecov](https://codecov.io/gh/awslabs/gluon-ts/pull/511?src=pr&el=h1) Report
> Merging [#511](https://codecov.io/gh/awslabs/gluon-ts/pull/511?src=pr&el=desc) into [master](https://codecov.io/gh/awslabs/gluon-ts/commit/bc24d1cef69421275efe2e6cb83e2f15be5280f8&el=desc) will **decrease** coverage by `0.06%`.
> The diff coverage is `n/a`.


[Impacted file tree graph](https://codecov.io/gh/awslabs/gluon-ts/pull/511?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #511      +/-   ##
==========================================
- Coverage   83.73%   83.66%   -0.07%     
==========================================
  Files         181      181              
  Lines       10327    10340      +13     
==========================================
+ Hits         8647     8651       +4     
- Misses       1680     1689       +9     
```

PR #512, opened 2019-12-10
Beta distribution implemented.

AaronSpieler:
This is related to Issue https://github.com/awslabs/gluon-ts/issues/467.

### DIFF
--- a/test/distribution/test_issue_287.py
+++ b/test/distribution/test_issue_287.py
@@ -17,8 +17,9 @@ import pytest
 from gluonts.distribution import DistributionOutput
 from gluonts.distribution.neg_binomial import NegativeBinomialOutput
 from gluonts.distribution.gamma import GammaOutput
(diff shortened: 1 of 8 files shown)


test/distribution/test_issue_287.py
```diff
@@ -29,6 +30,6 @@ def test_issue_287(distr_out_class: DistributionOutput):
     args_proj.initialize(init=mx.init.Constant(-1e2))
     distr_args = args_proj(network_output)
     distr = distr_output.distribution(distr_args)
-    x = mx.nd.array([1.0])
+    x = mx.nd.array([1.0 - 1e-7])  # open interval (0,1)
```
lostella: There should be no need for this to be close to 1.0. You can verify that the following still creates problem if the regularizations are removed from `NegativeBinomialOutput`, `GammaOutput`, `BetaOutput`

```suggestion
    x = mx.nd.array([0.5])
```

PR #515, opened 2019-12-11
Created notebooks from the docs tutorials

jiwidi:
*Issue #, if available:*

*Description of changes:*
I discovered that the examples under docs/examples/ do not have any py/notebooks files so I thought it would be useful to add a notebook version of them so the user can also execute cells and see the output :) 

I created notebooks for these tutorials under examples/ where the rest of the code tutorials are. I'm not sure if this is the best location so this is a good point for discussion.


jaheba:
Thanks for the work!

As far as I know, you can download the notebooks from the webpage, e.g clicking the "Download" button on https://gluon-ts.mxnet.io/examples/basic_forecasting_tutorial/tutorial.html

I'm currently not sure how they are generated / stored.

But I agree that it would make sense to also store them in the repository for easier access.

jiwidi:
@jaheba No problem! I just thought some people may encounter the same as me and want to access the notebooks right after cloning the repository.

I see some test are failing, are those test failing on master as well? I just want to know if there is any work I should do towards this PR even tho is a super basic thing.

:)

jgasthaus:
The original reasoning for not having the notebooks in the repo, but storing the markdown versions and compiling the notebooks during the website build process was that the notebook format doesn't play nice with diffs and code reviews, and this still makes sense to me. 

I also don't think we should store the same content in two formats (.md and .ipynb), as that would inevitably lead to inconsistencies between the two versions. If it is not obvious or too tedious to download the compiled versions after cloning the repo, maybe we can include a script to download them from the website (we already include a script to build them from the markdown files in ``docs/md2ipynb.py``)?

BTW, the notebooks get converted (and executed) during build here:

https://github.com/awslabs/gluon-ts/blob/5c15eb65dedee0b644bc85fb5195536d5fd0a4a1/Makefile#L48-L56

lostella:
> If it is not obvious or too tedious to download the compiled versions after cloning the repo, maybe we can include a script to download them from the website (we already include a script to build them from the markdown files in `docs/md2ipynb.py`)?

I think that's a good idea. Perhaps such a script could download & unzip the notebook into a new `notebooks` folder, which is `.gitignore`'d. @jiwidi do you want to take a shot at it?

jiwidi:
> > If it is not obvious or too tedious to download the compiled versions after cloning the repo, maybe we can include a script to download them from the website (we already include a script to build them from the markdown files in `docs/md2ipynb.py`)?
> 
> I think that's a good idea. Perhaps such a script could download & unzip the notebook into a new `notebooks` folder, which is `.gitignore`'d. @jiwidi do you want to take a shot at it?

Hi! sorry for the super late reply! Yes I would love to take a shot at it if its still of relevance. I can see your point @jgasthaus that is a bad idea to store them in two different formats. Do you think a script inside docs to download them into a notebooks folder could be useful? I would work into it right away.

PR #524, opened 2019-12-13
Made constant_dataset prediction length consistent

kashif:
perhaps it doesn't matter but since the test_ds has target length of 30 it might be more consistent to make the prediction length match up?

*Issue #, if available:*

*Description of changes:*


lostella:
Hi @kashif I think one could argue that it's actually better *not* to have the time ranges of train/test data aligned in any way. In short: it's a more general test case.

In general, models should be trainable on some data and then applied to a completely unrelated time range. Moreover, short test instances should be compatible with models that contemplate padding. Hard-coding that `len(train_data) = len(test_data) + prediction_length` in some test cases may hide issues, in case someone checks in a model that *does* assume such relationship.

I hope that's OK with you!

kashif:
makes sense thanks!

PR #526, opened 2019-12-14
Fixed distribution sampling issues.

AaronSpieler:
Fixed distribution invalid sampling issues. Added machine epsilon as minimal delta.

### DIFF
--- a/src/gluonts/distribution/gamma.py
+++ b/src/gluonts/distribution/gamma.py
@@ -127,7 +127,7 @@ class GammaOutput(DistributionOutput):
             Two squeezed tensors, of shape `(*batch_shape)`: both have entries mapped to the
             positive orthant.
         """
-        epsilon = np.finfo(np.float32).eps  # machine epsilon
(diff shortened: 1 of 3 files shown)


src/gluonts/distribution/gamma.py
```diff
@@ -122,8 +127,10 @@ def domain_map(cls, F, alpha, beta):
             Two squeezed tensors, of shape `(*batch_shape)`: both have entries mapped to the
             positive orthant.
         """
-        alpha = softplus(F, alpha) + 1e-8
-        beta = softplus(F, beta) + 1e-8
+        epsilon = np.finfo(cls._dtype).eps  # machine epsilon
```
lostella: Similarly to my comments in the other PR: I don’t like much this static _dtype property. I think we could use the dtype of alpha (we can assume this is the same as beta’s), what do you think?

The same applies to the other similar cases

lostella: Resolving after offline discussion

PR #532, opened 2019-12-17
add poisson distribution

### LINKED ISSUE #529: Implement Poisson distribution
The Poisson distribution is currently missing.

michiboo:
*Issue #, if available:* fixes #529

*Description of changes:*
Implement poisson distribution based on scipy.stats.poisson, Please correct me if there is any mistakes!
Also I can't seem to find a way to set all 0 to 1 in the x tensor for log_prob method, as the x seem to change to symbol object as training went on.

### DIFF
--- a/src/gluonts/distribution/poisson.py
+++ b/src/gluonts/distribution/poisson.py
@@ -1,0 +1,115 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+# Standard library imports
+from typing import Dict, Optional, Tuple, List
+
+# Third-party imports
+import numpy as np
+
+# First-party imports
+from gluonts.model.common import Tensor
+from gluonts.core.component import validated
+
+# Relative imports
+from .distribution import Distribution, _sample_multiple, getF, softplus
+from .distribution_output import DistributionOutput
+
+
+class Poisson(Distribution):
+    r"""
+    Poisson distribution, i.e. the distribution of the number of
+    successes in a specified region.
+
+    Parameters
+    ----------
+    rate
+        Tensor containing the means, of shape `(*batch_shape, *event_shape)`.
+    F
+    """
+
+    is_reparameterizable = False
+
+    @validated()
+    def __init__(self, rate: Tensor, F=None) -> None:
+        self.rate = rate
+        self.F = F if F else getF(rate)
+
+    @property
+    def batch_shape(self) -> Tuple:
+        return self.rate.shape
+
+    @property
+    def event_shape(self) -> Tuple:
+        return ()
+
+    @property
+    def event_dim(self) -> int:
+        return 0
+
+    def log_prob(self, x: Tensor) -> Tensor:
+        F = self.F
+        ll = x * F.log(self.rate) - F.gammaln(x + 1.0) - self.rate
+        return ll
+
+    @property
+    def mean(self) -> Tensor:
+        return self.rate
+
+    @property
+    def stddev(self) -> Tensor:
+        return self.F.sqrt(self.rate)
+
+    def sample(
+        self, num_samples: Optional[int] = None, dtype=np.float32
+    ) -> Tensor:
+        def s(rate: Tensor) -> Tensor:
+            return self.F.random.poisson(lam=rate, dtype=dtype)
+
+        return _sample_multiple(s, rate=self.rate, num_samples=num_samples)
+
+    @property
+    def args(self) -> List:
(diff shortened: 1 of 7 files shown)


lostella:
> Also I can't seem to find a way to set all 0 to 1 in the x tensor for log_prob method, as the x seem to change to symbol object as training went on.

What do you mean?

lostella:
@michiboo just checking: are you planning on adding tests to this? no problem if you aren't :-) someone else can add commits to this PR in that case

michiboo:
@lostella  Hi I was just working on it! Could be push the changes today!

lostella:
> @lostella Hi I was just working on it! Could be push the changes today!

oh wow, thanks! no rush, feel free to push whenever you're comfortable ;-)

test/distribution/test_distribution_inference.py
```diff
@@ -805,3 +807,35 @@ def test_binned_likelihood(
     assert all(
         mx.nd.abs(mx.nd.array(bin_prob_hat) - bin_prob) < TOL * bin_prob
     ), f"bin_prob did not match: bin_prob = {bin_prob}, bin_prob_hat = {bin_prob_hat}"
+
+
+@pytest.mark.parametrize("mu", [(3.75, 1.25)])
+@pytest.mark.parametrize("hybridize", [True, False])
+def test_beta_likelihood(mu: float, hybridize: bool) -> None:
```
lostella:
```suggestion
def test_poisson_likelihood(mu: float, hybridize: bool) -> None:
```

src/gluonts/distribution/poisson.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class Poisson(Distribution):
+    r"""
+    Poisson distribution, i.e. the distribution of the number of
+    successes in a specified region.
+
+    Parameters
+    ----------
+    mu
```
lostella: This parameter is commonly denoted as `lambda`. I think it would be better to stick to that naming, what do you think?

test/distribution/test_distribution_methods.py
```diff
@@ -83,6 +84,7 @@
             ).repeat(axis=0, repeats=2),
         },
     ),
+    (Poisson, {"mu": mx.nd.array([1000.0, -1000.0])},),
```
lostella: The parameter of a Poisson should be positive, I think `-1000.0` creates issues

test/distribution/test_distribution_sampling.py
```diff
@@ -88,6 +89,7 @@
             ).repeat(axis=0, repeats=2),
         },
     ),
+    (Poisson, {"mu": mx.nd.array([1000.0, -1000.0]),},),
```
lostella: Similar to the previous comment

test/distribution/test_distribution_shapes.py
```diff
@@ -217,6 +218,7 @@
             (3, 4, 5),
             (),
         ),
+        (Poisson(mu=mx.nd.zeros(shape=(3, 4, 5)),), (3, 4, 5), (),),
```
lostella: Similar to the previous comments: maybe you can use `mx.nd.ones` here

test/distribution/test_distribution_output_shapes.py
```diff
@@ -141,6 +142,13 @@
             (3, 4),
             (5,),
         ),
+        (
+            PoissonOutput(),
+            mx.nd.random.normal(shape=(3, 4, 5, 6)),
```
lostella: Since Poisson has the set of natural number as support, it's better to draw from something other than Normal here: you could use `mx.nd.random.poisson`, see https://mxnet.incubator.apache.org/api/python/docs/api/ndarray/random/index.html#mxnet.ndarray.random.poisson

src/gluonts/distribution/poisson.py
michiboo: Hi, I think lambda is better too, I will change it

michiboo: But lambda is a reserved word, should I just put it like _lambda?

lostella: oh right, you can call it `rate` then

test/distribution/test_distribution_output_shapes.py
```diff
@@ -141,6 +142,13 @@
             (3, 4),
             (5,),
         ),
+        (
+            PoissonOutput(),
+            mx.nd.random.poisson(shape=(3, 4, 5, 6)),
```
lostella: This was my bad, apologies: this tensor is supposed to be the output of the network, which will be automatically mapped to the parameter domain (positive reals) by the `PoissonOutput` class. I mistakenly though this was supposed to be data in the support of the distribution.

```suggestion
            mx.nd.random.normal(shape=(3, 4, 5, 6)),
```

michiboo:
> @michiboo looks good, thanks for the contribution! About the following comment:
> 
> > Also I can't seem to find a way to set all 0 to 1 in the x tensor for log_prob method, as the x seem to change to symbol object as training went on.
> 
> What do you mean exactly? Is there anything you think should be fixed? Or maybe that comment is now outdated...

Hi it's just I want to make sure that there is no 0s in the input because log0 is undefined, but since all input is > 0 then there is no need for this, I think.

lostella:
> Hi it's just I want to make sure that there is no 0s in the input because log0 is undefined, but since all input is > 0 then there is no need for this, I think.

Shouldn't be an issue, since `x` is supposed to be nonnegative, so the `log_prob` is well defined

PR #533, opened 2019-12-17
Fix instance splitter issue with short time series

lostella:
*Description of changes:* Some checks are added in the `InstanceSplitter` class for the case where `pick_incomplete=False`. This would otherwise cause problems downstream, when the sampler would to sample from an interval that has inverted bounds.

Further checks and typing cleanup are added.

Opening for feedback, trying to include a small test for this.

### DIFF
--- a/src/gluonts/transform/split.py
+++ b/src/gluonts/transform/split.py
@@ -179,7 +179,7 @@ class InstanceSplitter(FlatMapTransformation):
                 else self.train_sampler(target, *sampling_bounds)
             )
         else:
-            assert self.pick_incomplete or len_target >= minimum_length
+            assert self.pick_incomplete or len_target >= self.past_length
             sampled_indices = np.array([len_target], dtype=int)
         for i in sampled_indices:
             pad_length = max(self.past_length - i, 0)
--- a/src/gluonts/transform/split.py
+++ b/src/gluonts/transform/split.py
@@ -157,17 +157,20 @@ class InstanceSplitter(FlatMapTransformation):
         len_target = target.shape[-1]
 
         if is_train:
-            if len_target < self.future_length:
-                # We currently cannot handle time series that are shorter than
-                # the prediction length during training, so we just skip these.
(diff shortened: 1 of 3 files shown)


lostella:
Polished the code, updated the tests, added test cases with `pick_incomplete=True` (I verified that these raise the issue on `master`)

test/test_transform.py
```diff
@@ -204,10 +209,25 @@ def test_InstanceSplitter(start, target, is_train):
...
 
     if is_train:
-        assert len(out) == max(0, len(target) - pred_length + 1)
+        assert len(out) == max(
+            0,
+            len(target)
+            - pred_length
+            + 1
+            - (0 if pick_incomplete else train_length),
+        )
```
lostella: Sometimes I really don't understand black

src/gluonts/transform/split.py
```diff
@@ -156,30 +156,37 @@ def flatmap_transform(
...
+            # mask the loss.
+            sampled_indices = (
+                np.array([], dtype=int)
+                if len_target < minimum_length
+                else self.train_sampler(target, *sampling_bounds)
+            )
         else:
-            sampling_indices = [len_target]
-        for i in sampling_indices:
+            assert self.pick_incomplete or len_target >= minimum_length
```
vafl: For inference we don't need the future, so I think this should be

```python
asser self.pick_incomplete or len_target >= self.past_length
```

test/test_transform.py
```diff
@@ -204,10 +209,25 @@ def test_InstanceSplitter(start, target, is_train):
         "some_other_col": "ABC",
     }
 
-    out = list(t.flatmap_transform(data, is_train=is_train))
+    if (
+        not is_train
+        and not pick_incomplete
+        and len(target) < train_length + pred_length
```
vafl: and correspondingly here 

```
len(target) < train_length
```

lostella: good catch!

PR #534, opened 2019-12-17
Distribution dependent loss masking.

AaronSpieler:
Before sometimes it happened that `-inf` was multiplied with `0` which still led to `nan`s in the negative log loss.

### DIFF
--- a/src/gluonts/distribution/beta.py
+++ b/src/gluonts/distribution/beta.py
@@ -69,13 +69,9 @@ class Beta(Distribution):
         F = self.F
         alpha, beta = self.alpha, self.beta
 
-        # add machine epsilon to avoid infinities due to 0 padding which is masked,
-        # however, causes issues in backpropagation
-        epsilon = np.finfo(self._dtype).eps
-
         ll = (
-            (alpha - 1) * F.log(x + epsilon)
-            + (beta - 1) * F.log(1 - x + epsilon)
+            (alpha - 1) * F.log(x)
+            + (beta - 1) * F.log(1 - x)
             - F.gammaln(alpha)
             - F.gammaln(beta)
             + F.gammaln(alpha + beta)
--- a/src/gluonts/distribution/beta.py
+++ b/src/gluonts/distribution/beta.py
@@ -54,6 +54,10 @@ class Beta(Distribution):
         )  # assuming alpha and beta of same type
 
     @property
+    def point_in_support(self) -> Tensor:
+        return self.F.ones_like(self.alpha) * 0.5
+
+    @property
     def batch_shape(self) -> Tuple:
         return self.alpha.shape
 
--- a/src/gluonts/distribution/beta.py
+++ b/src/gluonts/distribution/beta.py
@@ -54,10 +54,6 @@ class Beta(Distribution):
         )  # assuming alpha and beta of same type
 
     @property
-    def point_in_support(self) -> Tensor:
-        return self.F.ones_like(self.alpha) * 0.5
-
-    @property
     def batch_shape(self) -> Tuple:
         return self.alpha.shape
 
@@ -73,15 +69,13 @@ class Beta(Distribution):
         F = self.F
         alpha, beta = self.alpha, self.beta
 
-        ll = (
+        return (
             (alpha - 1) * F.log(x)
             + (beta - 1) * F.log(1 - x)
             - F.gammaln(alpha)
             - F.gammaln(beta)
             + F.gammaln(alpha + beta)
         )
-
-        return ll
 
     @property
     def mean(self) -> Tensor:
@@ -157,3 +151,7 @@ class BetaOutput(DistributionOutput):
     @property
     def event_shape(self) -> Tuple:
         return ()
+
+    @property
+    def value_in_support(self) -> float:
+        return 0.5
--- a/src/gluonts/distribution/beta.py
+++ b/src/gluonts/distribution/beta.py
@@ -101,6 +101,8 @@ class Beta(Distribution):
     def sample(
         self, num_samples: Optional[int] = None, dtype=np.float32
     ) -> Tensor:
+        epsilon = np.finfo(dtype).eps  # machine epsilon
+
         def s(alpha: Tensor, beta: Tensor) -> Tensor:
             F = getF(alpha)
             samples_X = F.sample_gamma(
@@ -111,9 +113,11 @@ class Beta(Distribution):
             )
             return samples_X / (samples_X + samples_Y)
 
-        return _sample_multiple(
+        samples = _sample_multiple(
             s, alpha=self.alpha, beta=self.beta, num_samples=num_samples
         )
+
+        return self.F.clip(data=samples, a_min=epsilon, a_max=1 - epsilon)
 
     @property
     def args(self) -> List:
@@ -144,8 +148,10 @@ class BetaOutput(DistributionOutput):
             Two squeezed tensors, of shape `(*batch_shape)`: both have entries mapped to the
             positive orthant.
         """
-        alpha = softplus(F, alpha) + 1e-8
-        beta = softplus(F, beta) + 1e-8
+        epsilon = np.finfo(cls._dtype).eps  # machine epsilon
+
+        alpha = softplus(F, alpha) + epsilon
+        beta = softplus(F, beta) + epsilon
         return alpha.squeeze(axis=-1), beta.squeeze(axis=-1)
 
     @property
--- a/src/gluonts/distribution/beta.py
+++ b/src/gluonts/distribution/beta.py
@@ -69,13 +69,19 @@ class Beta(Distribution):
         F = self.F
         alpha, beta = self.alpha, self.beta
 
-        return (
-            (alpha - 1) * F.log(x)
-            + (beta - 1) * F.log(1 - x)
+        # add machine epsilon to avoid infinities due to 0 padding which is masked,
+        # however, causes issues in backpropagation
+        epsilon = np.finfo(self._dtype).eps
+
+        ll = (
+            (alpha - 1) * F.log(x + epsilon)
+            + (beta - 1) * F.log(1 - x + epsilon)
             - F.gammaln(alpha)
             - F.gammaln(beta)
             + F.gammaln(alpha + beta)
         )
+
+        return ll
 
     @property
     def mean(self) -> Tensor:
--- a/src/gluonts/distribution/dirichlet.py
+++ b/src/gluonts/distribution/dirichlet.py
@@ -136,14 +136,15 @@ class Dirichlet(Distribution):
 class DirichletOutput(DistributionOutput):
     @validated()
     def __init__(self, dim: int) -> None:
+        super().__init__(self)
         assert dim > 1, "Dimension should be larger than one."
         self.args_dim = {"alpha": dim}
         self.distr_cls = Dirichlet
         self.dim = dim
         self.mask = None
 
-    def distribution(self, distr_args, scale=None, **kwargs) -> Distribution:
-        # todo deal with scaling
+    def distribution(self, distr_args, loc=None, scale=None) -> Distribution:
+        assert loc is None and scale is None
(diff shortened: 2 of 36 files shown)


src/gluonts/distribution/beta.py
```diff
@@ -69,14 +69,20 @@ def log_prob(self, x: Tensor) -> Tensor:
         F = self.F
         alpha, beta = self.alpha, self.beta
 
-        return (
-            (alpha - 1) * F.log(x)
-            + (beta - 1) * F.log(1 - x)
+        # add machine epsilon to avoid infinities due to 0 padding which is masked,
+        # however, causes issues in backpropagation
+        epsilon = np.finfo(self._dtype).eps
```
lostella: Since you defined `dtype` as a `@property`, why not using that instead?

src/gluonts/distribution/distribution.py
```diff
@@ -48,6 +49,16 @@ class Distribution:
     arg_names: Tuple
     is_reparameterizable = False
 
+    _dtype: DType = np.float32
```
lostella: Is this property set by all subtypes of `Distribution` upon construction?

src/gluonts/distribution/beta.py
```diff
@@ -69,14 +69,20 @@ def log_prob(self, x: Tensor) -> Tensor:
...
+
+        ll = (
+            (alpha - 1) * F.log(x + epsilon)
+            + (beta - 1) * F.log(1 - x + epsilon)
```
lostella: I don’t get it: why is this needed?

If a value outside of the domain ends up in x due to padding, then masking should take care of the inf or nan that comes out of this. Otherwise I don’t think we should try to recover from 0s or 1s in the data (one should validate the data beforehand)

src/gluonts/support/util.py
```diff
@@ -390,11 +390,12 @@ def cumsum(
     return cumulative_sum.squeeze(axis=exp_dim)
 
 
-def weighted_average(
+def masked_weighted_average(
```
lostella: This function computes the weighted average, so its name can stay `weighted_average` :-)

Changing name would be a breaking change, so let’s avoid it unless it’s necessary

src/gluonts/model/deepar/_network.py
```diff
@@ -416,10 +416,13 @@ def hybrid_forward(
             else observed_values.min(axis=-1, keepdims=False)
         )
 
-        weighted_loss = weighted_average(
+        weighted_loss = masked_weighted_average(
             F=F, x=loss, weights=loss_weights, axis=1
         )
 
+        # need to mask possible nans and -inf
+        loss = F.where(condition=loss_weights, x=loss, y=F.zeros_like(loss))
```
lostella: Why masking again? Isn’t the weighted average supposed to be doing this already?

AaronSpieler: This is just loss, not weigthed loss. Idk why there are two here, however, this is how it was, I just changed it so it gets properly masked here too.

src/gluonts/support/util.py
AaronSpieler: Ok, fine... :(((

src/gluonts/distribution/beta.py
AaronSpieler: you mean calling that function?

lostella: No need to call it, since it’s a @property you can just use self.dtype

src/gluonts/model/deepar/_network.py
lostella: Hmmm I see. Yeah I’m not sure why there are two here. Anyway I would leave it as it is, without doing the masking in this case. This second output should not be used (in my opinion it shouldn’t be exposed at all) and it’s probably not used by the default training loop, for example.

src/gluonts/distribution/distribution.py
AaronSpieler: Yeah, its a class variable. I coded this analogous to DistributionOutput.

src/gluonts/distribution/beta.py
AaronSpieler: ok

src/gluonts/distribution/distribution.py
lostella: I don’t think this is ideal (neither here nor in DistributionOutput, don’t get me wrong): I think dtype should better be a property of the object, either explicitly set upon construction, or inferred by the dtype of the constructor arguments.

src/gluonts/model/deepar/_network.py
AaronSpieler: Ok, but then it doesn't make any sense not to change it? The main point of this PR is to mask the loss properly. So whatever is done with that loss, it cannot be done properly with '-inf's in there...

lostella: ok, sounds good

src/gluonts/distribution/distribution.py
```diff
@@ -83,26 +93,42 @@ def crps(self, x: Tensor) -> Tensor:
...
         Tensor
             Tensor of shape `batch_shape` containing the value of the loss
             for each event in `x`.
         """
-        return -self.log_prob(x)
+        F = getF(x)
+        if mask is not None:
+            x = F.where(condition=mask, x=x, y=self.point_in_support)
+            loss = self._loss(x)
+            loss = F.where(condition=mask, x=loss, y=F.zeros_like(x))
```
vafl: Why is `point_in_support` necessary here? Shouldn't

```
F.where(condition=mask, x=loss, y=F.zeros_like(x))
```
be sufficient? Can you maybe give a minimal example with some nd arrays?

AaronSpieler: TLDR: essentially the problem is that we get `nan`s upon back-propagation if we only use `F.where`:

If we have `0` due to padding in the data and we calculate the `-log_loss(x)` for the Beta or Gamma Distribution we get `-inf`, which we want to mask.

So in the forward pass everything works out with `F.where` and we get valid loss values, however, upon back-propagation we get `-inf * 0 = nan` because of the way the gradient is calculated: 

Screenshot 2019-12-17 at 21 49 48

Because `F.where` is basically just another network layer and because `-inf * 0` yields `nan` and not `0`.

Apparently this is a known issue: 
https://github.com/pytorch/pytorch/issues/23156

lostella: @vafl there are issues if some elements of `x` are imputed with values that are not in the support of the distribution. For example, in `Beta` or `Gamma` zero is not in the support, but `nan`s in the data are imputed with zeros. This is fine in the forward pass: you'll get some `inf` in the log-likelihood which will be masked in this line, as you say. But this creates problems in the backward pass, apparently, which can only be avoided by not producing the `inf` in the first place. Aaron found a similar issue in pytorch: https://github.com/pytorch/pytorch/issues/23156

A MWE for this issue is the following:

```python
import mxnet as mx
from mxnet import autograd

param = mx.nd.array([1.0])
y = mx.nd.array([0.0])

param.attach_grad()

with autograd.record():
    loss = mx.nd.log(y*param)
    mask = mx.nd.array([0.0])
    masked_loss = mx.nd.where(condition=mask, x=loss, y=mx.nd.zeros_like(loss))

print(masked_loss)

dparam = autograd.grad(masked_loss, [param])[0]

print(dparam)
```

lostella: The above prints out

```
[0.]
<NDArray 1 @cpu(0)>

[nan]
<NDArray 1 @cpu(0)>
```

vafl: I see, thanks for clarifying this! One alternative design would be to allow passing `nan`'s in the input and then do all the masking and logic of selecting a dummy value inside `log_prob`.

lostella: @vafl I think the problem with that is that nans need to be out of the way once the data reaches the network, or did I misunderstand?

vafl: AFAIK you can pass nans in via the data. I think we used this at some point. I'm not sure whether it is a common practice, but in the end nans should be just floats.

lostella: But if a nan goes as input to any block, it will likely result in nans coming out of it. Or do you mean substitute nans with zeros in the input (past) but leave nans in the label (future)?

src/gluonts/distribution/gamma.py
```diff
@@ -52,6 +52,10 @@ def __init__(self, alpha: Tensor, beta: Tensor, F=None) -> None:
             F if F else getF(alpha)
         )  # assuming alpha and beta of same type
 
+    @property
+    def point_in_support(self) -> Tensor:
+        return self.F.ones_like(self.alpha) * 0.5
```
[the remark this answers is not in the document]
AaronSpieler: yes, true

src/gluonts/distribution/transformed_distribution.py
```diff
@@ -49,6 +49,10 @@ def __init__(
         self._event_shape: Optional[Tuple] = None
         self._batch_shape: Optional[Tuple] = None
 
+    @property
+    def point_in_support(self) -> Tensor:
+        return self.base_distribution.point_in_support
```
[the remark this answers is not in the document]
AaronSpieler: transform how?

src/gluonts/model/deepvar/_network.py
```diff
@@ -529,6 +523,12 @@ def train_hybrid_forward(
 
         assert_shape(loss_weights, (-1, seq_len, 1))
 
+        # we sum the last axis to have the same shape for all likelihoods
+        # (batch_size, subseq_length, 1)
+        likelihoods = distr.loss(target).expand_dims(axis=-1)
```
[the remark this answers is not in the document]
AaronSpieler: ok

src/gluonts/distribution/dirichlet.py
```diff
@@ -58,6 +58,12 @@ def __init__(
         self.F = F if F else getF(alpha)
         self.float_type = float_type
 
+    @property
+    def point_in_support(self) -> Tensor:
+        point = self.F.ones_like(self.alpha)
+        sum_point = self.F.sum(point, axis=-1).expand_dims(axis=-1)
+        return self.F.broadcast_div(point, sum_point)
```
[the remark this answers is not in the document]
AaronSpieler: I see, yeah thats a problem.

src/gluonts/distribution/distribution.py
vafl: Ok, yeah. Handling nans inside the graph may be tricky with this gradient issue.

AaronSpieler:
> 
> 
> Since it seems like values compatible with the distributions support can be known statically (ie only by the distribution type), a maybe less invasive way of doing this is the following.
> 
>     * Expose a `value_in_support` (float) from the `DistributionOutput` classes (maybe with default `0.0` in the base `DistributionOutput` class).
> 
>     * Have models use such value, as fill-in for nans, as part of the data transformation chain: see [here](https://github.com/awslabs/gluon-ts/blob/5c15eb65dedee0b644bc85fb5195536d5fd0a4a1/src/gluonts/model/deepar/_estimator.py#L240-L244) for DeepAR, in particular changing [this setting of the transformation](https://github.com/awslabs/gluon-ts/blob/5c15eb65dedee0b644bc85fb5195536d5fd0a4a1/src/gluonts/transform/feature.py#L59).
> 
>     * Mask the loss in the network using the `weighted_average` function with your proposed changes.
> 
> 
> It seems to me like that should also work. @AaronSpieler what do you think?

I agree in that that would be a more elegant and less invasive step.
I think we should do that instead.
But what do you mean by masking 'nan's, because there shouldnt be any nans being produced?

AaronSpieler:
Funny how we went full circle on this, anyways, I think this solution is good because its non invasive and it works well with the beta and gamma distributions.
I also had to change the `InstanceSplitter` by adding `dummy_value` there too, because thats where the actual padding happend.

vafl:
lgtm, thanks @AaronSpieler

PR #540, opened 2019-12-20
Add `loc` argument to distribution output classes

lostella:
*Description of changes:* This PR adds the `loc` argument to the `distribution` method of `DistributionOutput` classes. Assertions are added in case the specific distributions don't support either `loc` or `scale`. A bug in the `TransformedPiecewiseLinear` distribution was also fixed: this went unnoticed since it only appears when the base distribution is composed with more than one affine transformation, which never happens when using `PiecewiseLinearOutput`.

### DIFF
--- a/src/gluonts/distribution/bijection.py
+++ b/src/gluonts/distribution/bijection.py
@@ -198,9 +198,9 @@ class AffineTransformation(Bijection):
     Parameters
     ----------
     loc
-        Translation parameter.
+        Translation parameter. If unspecified or `None`, this will be zero.
     scale
-        Scaling parameter.
+        Scaling parameter. If unspecified or `None`, this will be one.
     """
 
     @validated()
--- a/src/gluonts/distribution/bijection.py
+++ b/src/gluonts/distribution/bijection.py
@@ -229,8 +229,8 @@ class AffineTransformation(Bijection):
     def log_abs_det_jac(self, x: Tensor, y: Tensor) -> Tensor:
         if self.scale is not None:
             return self.scale.broadcast_like(x).abs().log()
-        else:
-            raise RuntimeError("scale is of type None in log_abs_det_jac")
+        F = getF(x)
+        return F.zeros_like(x)
 
     @property
     def sign(self):
--- a/src/gluonts/distribution/distribution_output.py
+++ b/src/gluonts/distribution/distribution_output.py
@@ -117,7 +117,10 @@ class DistributionOutput(Output):
         pass
 
     def distribution(
-        self, distr_args, scale: Optional[Tensor] = None
+        self,
+        distr_args,
+        loc: Optional[Tensor] = None,
+        scale: Optional[Tensor] = None,
     ) -> Distribution:
         r"""
         Construct the associated distribution, given the collection of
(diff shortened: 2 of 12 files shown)


src/gluonts/distribution/bijection.py
```diff
@@ -229,8 +229,8 @@ def f_inv(self, y: Tensor) -> Tensor:
     def log_abs_det_jac(self, x: Tensor, y: Tensor) -> Tensor:
         if self.scale is not None:
             return self.scale.broadcast_like(x).abs().log()
-        else:
-            raise RuntimeError("scale is of type None in log_abs_det_jac")
+        F = getF(x)
+        return F.zeros_like(x)
```
mbohlkeschneider: I think we should avoid silently doing something with undefined inputs. At the very least, I would issue a warning here.

src/gluonts/distribution/distribution_output.py
```diff
@@ -127,16 +130,19 @@ def distribution(
...
             batch_shape+event_shape of the resulting distribution.
         """
-        if scale is None:
+        if loc is None and scale is None:
             return self.distr_cls(*distr_args)
         else:
             distr = self.distr_cls(*distr_args)
             return TransformedDistribution(
-                distr, [AffineTransformation(scale=scale)]
+                distr, [AffineTransformation(loc=loc, scale=scale)]
```
mbohlkeschneider: `loc` is actually never passed in any of our models, right? See [DeepAR](https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/model/deepar/_network.py#L338) here for example.

lostella: It's not indeed, and will default to `None` so long as it's not used

lostella: To expand on this a bit: the reason for having `loc` alongside `scale` is to allow using a more general scaling, like min-max scaling or normalizing data (in the true sense of the word) instead of just dividing by some scaling factor. And if this is doable in a minimally invasive, 100% backwards compatible manner why not

src/gluonts/distribution/bijection.py
lostella: The reason for this change is that when `scale is None`, the `f` and `f_inv` methods act like if `scale == 1`: in this case the Jacobian is exactly `scale`, and if `scale == 1` then `log(abs(det(scale))) == 0`, hence the returned value.

To summarize, in my opinion `None` here is not an undefined input, but rather a way of encoding the default `1` (for `scale`, and `0` for `loc`). In fact, no warning is issued in `f` or `f_inv` when either `scale` or `loc` are `None`.

`F.zeros_like(x)` is exactly log|det(J)| in case `scale is None`, where the Jacobian J in this case is precisely `scale`: if you check what the implementations of `f` and `f_inv`, they act like if `scale = 1`, in which case `log(abs(det(scale))) == 0`

mbohlkeschneider: I see. Wouldn't it then be cleaner to just set the defaults in the constructor when `None` is passed and avoid the `if sth is None` logic throughout the class?

mbohlkeschneider: > I see. Wouldn't it then be cleaner to just set the defaults in the constructor when `None` is passed and avoid the `if sth is None` logic throughout the class?

Hmm, maybe not. It's tricky because we don't know the shape of the input tensor if the values are `None`.

mbohlkeschneider: > I see. Wouldn't it then be cleaner to just set the defaults in the constructor when `None` is passed and avoid the `if sth is None` logic throughout the class?

Hmm, maybe not. It's tricky because we don't know the shape of the input tensor if the values are `None`.

lostella: Yes, exactly. I think having `None` as default is not too bad: when `loc` (resp. `scale`) is `None`, it is the additive (multiplicative) identity. It just needs to be clear: maybe I'll specify this better in the docstring.

lostella:
> This PR reminds me that we actually use the wrong abstraction for the `AffineTransformation` transformed output. I think the right abstraction would be to have some kind of `ConditionalBijection` to model this (not for this PR, but we should think about this in the future).

@mbohlkeschneider yes, I think we could adjust a bit how distributions and transformations interact. Especially in the ubiquitous case of scaling, where we "manually" transform data before the model input layer, and use transformed distributions in the likelihood instead.

PR #544, opened 2019-12-27
fix quantile fn for transformed distribution

vafl:
*Description of changes:*
Fix quantile function for transformed distribution and minor other improvements.
This reduces memory consumption when calculating quantiles for transformed distribution.

Also minor improvements for forecast methods.

### DIFF
--- a/src/gluonts/distribution/transformed_distribution.py
+++ b/src/gluonts/distribution/transformed_distribution.py
@@ -129,14 +129,18 @@ class TransformedDistribution(Distribution):
             sign = sign * t.sign
 
         if not isinstance(sign, (mx.nd.NDArray, mx.sym.Symbol)):
-            sign = sign + level.zeros_like()
+            level = level if sign > 0 else (1.0 - level)
+            q = self.base_distribution.quantile(level)
+        else:
+            # level.shape = (#levels,)
+            # q_pos.shape = (#levels, batch_size, ...)
+            # sign.shape = (batch_size, ...)
+            q_pos = self.base_distribution.quantile(level)
+            q_neg = self.base_distribution.quantile(1.0 - level)
+            cond = F.broadcast_greater(sign, sign.zeros_like())
+            cond = F.broadcast_add(cond, q_pos.zeros_like())
+            q = F.where(cond, q_pos, q_neg)
 
-        cond = F.broadcast_greater(sign, sign.zeros_like())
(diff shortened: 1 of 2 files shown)


src/gluonts/distribution/transformed_distribution.py
```diff
@@ -129,14 +129,17 @@ def quantile(self, level: Tensor) -> Tensor:
...
+            level = level if sign > 0 else (1.0 - level)
+            q = self.base_distribution.quantile(level)
+        else:
+            q_pos = self.base_distribution.quantile(level)
+            q_neg = self.base_distribution.quantile(1.0 - level)
+            cond = F.broadcast_greater(sign, sign.zeros_like())
+            cond = F.broadcast_add(
+                cond.expand_dims(axis=1), q_pos.zeros_like()
+            )
+            q = F.where(cond, q_pos, q_neg)
```
lostella: This makes two calls to `.quantile` instead of one. Wouldn't something like

```python
cond = F.broadcast_greater(sign, sign.zeros_like())
signed_level = F.where(cond, level, 1.0 - level)
q = self.base_distribution.quantile(signed_level)
```

be more efficient?

vafl: This is what we had before. The problem is that in the current API for `.quantile` `level` is supposed to be a 1d tensor of level values, i.e., it doesn't know about the shape of the distribution. The previous code blew up the `level` to a larger tensor mixing in the shape of `sign`. It did work (not entirely sure how), but it resulted in very large tensors during the `quantile` computation that don't fit in GPU memory for e.g. `Binned` distribution when the distribution dimension was large. Maybe we can rethink the api for `quantile` to not only support 1d level values, but also level values per distribution dimension.

lostella: Makes sense, thanks for clarifying. If that’s the contract that `quantile` assumes (scalar or 1D tensor) then I agree that your solution is probably the best.

In a separate story, we may want to start shape assertions in order to avoid these weird situations where things accidentally work :-) (maybe because of some strange broadcasting coincidence)

jgasthaus:
Do we have a use case for these bijections with negative Jacobian determinant (i.e. monotonically decreasing)?

vafl:
Good point. Maybe we should just remove the sign stuff and document clearly that we expect positive determinant.

Issue #291, opened 2019-09-06
Item metrics dataframe incorrectly formatted


AaronSpieler:
## Description

The data-frame containing the per time step metrics has a column item_id and filled with NaNs.
![image](https://user-images.githubusercontent.com/25365592/64429596-a2796200-d0b6-11e9-89fd-a77bd31ffcc9.png)

## To Reproduce
For example in the extended tutorial:
```
evaluator = Evaluator(quantiles=[0.1, 0.5, 0.9])
agg_metrics, item_metrics = evaluator(iter(tss), iter(forecasts), num_series=len(test_ds))
item_metrics.head()
```

## Environment
- Operating system: MacOs
- Python version: 3.6.9
- GluonTS version: 0.3.3

lostella:
@AaronSpieler I believe that depends on the dataset you use: if the entries in the dataset have an “item_id” field, then that will be used to tag the forecasts and appropriately fill in the associated field in the metrics dataframe.

What dataset did that occur with? If it’s one of the provided datasets, then the field may be missing or have a different name, which should be fixed.

AaronSpieler:
Its the m4_hourly dataset (as used in the quick tutorial).

jaheba:
@AaronSpieler can you follow up on this?

AaronSpieler:
So all the provided datasets I tested contain a categorical feature called "feat_static_cat", which is equal to the corresponding index of the series.

<img width="1091" alt="image" src="https://user-images.githubusercontent.com/25365592/67270295-ab809000-f4b8-11e9-85aa-cc05ba93368c.png">

Is this supposed to be renamed to "item_id" @lostella ?

matsmaiwald:
This is probably related to the fact that, unlike in section 1.3 of the Extended Forecasting Tutorial ([here](https://gluon-ts.mxnet.io/examples/extended_forecasting_tutorial/extended_tutorial.html#1.3-Use-your-time-series-and-features)), in version 0.3.3 `FieldName` does not have the `ITEM_ID` attribute anymore. 

![Screenshot 2019-10-22 at 13 52 02](https://user-images.githubusercontent.com/32721837/67282999-2a82c200-f4d3-11e9-81df-7087232a0751.png)

## To Reproduce
```
from gluonts.transform import FieldName
[f"FieldName.{k} = '{v}'" for k, v in FieldName.__dict__.items() if not k.startswith('_')]
```

## Environment
- Operating system: CentOS
[...]

PR #546, opened 2020-01-01
Fixes NaNs in DataFrama item_id column issue #291

AaronSpieler:
Also the 'item_id' will be the new index.

### DIFF
--- a/src/gluonts/dataset/common.py
+++ b/src/gluonts/dataset/common.py
@@ -215,11 +215,12 @@ class FileDataset(Dataset):
 
     def __iter__(self) -> Iterator[DataEntry]:
         for path in self.files():
-            for line in jsonl.JsonLinesFile(path):
+            for i, line in enumerate(jsonl.JsonLinesFile(path)):
                 data = self.process(line.content)
                 data["source"] = SourceContext(
-                    source=line.span, row=line.span.line
+                    source=line.span.path, row=line.span.line
                 )
+                data["item_id"] = i
                 yield data
 
     def __len__(self):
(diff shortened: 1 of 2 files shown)


src/gluonts/dataset/common.py
```diff
@@ -215,11 +215,12 @@ def __init__(
 
     def __iter__(self) -> Iterator[DataEntry]:
         for path in self.files():
-            for line in jsonl.JsonLinesFile(path):
+            for i, line in enumerate(jsonl.JsonLinesFile(path)):
                 data = self.process(line.content)
                 data["source"] = SourceContext(
-                    source=line.span, row=line.span.line
+                    source=line.span.path, row=line.span.line
```
AaronSpieler: I think I accidentally leaked this from latest master branch, however, should be no problem when merging with master.

kashif:
@AaronSpieler it might be best to check if `data[item_id]` is empty before overwriting it with an index

vafl:
I think the index of the dataframe is already the row number, so we don't need to add it explicitly. 

The purpose of `item_id` was a bit different. It was supposed to be a unique id that the user can provide in the dataset with each time series, and which can then be used later for instance to join results with other information that the user has. We should probably update the documentation on this.

Was there a problem with the NaN column or is this more of a usability question?

kashif:
@vafl yes more of a usability issue since my data has unique `item_id`s and I its handy to see the metrics for specific `item_ids`...

vafl:
@kashif yes exactly. So I think we either do what you suggest and overwrite it only if it is not there. But perhaps it's best to keep the current behavior as is and update the docs instead.

lostella:
Note that the original issue #291 is really about the "item_id" field missing in m4_hourly (and maybe other provided datasets). So another option is to make sure that such field is filled-in in the scripts that generate the datasets.

lostella:
I think this can be closed in favor of a new one that adjusts the provided datasets, see my [comment](https://github.com/awslabs/gluon-ts/issues/291#issuecomment-576018487) in #291

PR #553, opened 2020-01-08
N-Beats implementation for GluonTS

AaronSpieler:
This addresses issue https://github.com/awslabs/gluon-ts/issues/551 .

Waiting for review, other than that extensive testing is up next.

### DIFF
--- a/.github/workflows/mxnet-nightly.yml
+++ b/.github/workflows/mxnet-nightly.yml
@@ -32,4 +32,4 @@ jobs:
         pip install -r requirements/requirements-test.txt
     - name: Test with pytest
       run: |
-        pytest -m 'not (gpu or serial)' --cov src/gluonts --cov-report=term --cov-report xml test
+        pytest --timeout=60 -m 'not (gpu or serial)' --cov src/gluonts --cov-report=term --cov-report xml test
--- a/.github/workflows/mxnet-nightly.yml
+++ b/.github/workflows/mxnet-nightly.yml
@@ -32,4 +32,4 @@ jobs:
         pip install -r requirements/requirements-test.txt
     - name: Test with pytest
       run: |
-        pytest --timeout=60 -m 'not (gpu or serial)' --cov src/gluonts --cov-report=term --cov-report xml test
+        pytest -m 'not (gpu or serial)' --cov src/gluonts --cov-report=term --cov-report xml test
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -28,4 +28,4 @@ jobs:
         pip install -r requirements/requirements-test.txt
     - name: Test with pytest
       run: |
-        pytest --timeout=60 -m 'not (gpu or serial)' --cov src/gluonts --cov-report=term --cov-report xml test
+        pytest -m 'not (gpu or serial)' --cov src/gluonts --cov-report=term --cov-report xml test
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -28,4 +28,4 @@ jobs:
         pip install -r requirements/requirements-test.txt
     - name: Test with pytest
       run: |
-        pytest -m 'not (gpu or serial)' --cov src/gluonts --cov-report=term --cov-report xml test
+        pytest --timeout=60 -m 'not (gpu or serial)' --cov src/gluonts --cov-report=term --cov-report xml test
--- a/src/gluonts/model/n_beats/_ensemble.py
+++ b/src/gluonts/model/n_beats/_ensemble.py
@@ -1,0 +1,12 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
--- a/src/gluonts/model/n_beats/_ensemble.py
+++ b/src/gluonts/model/n_beats/_ensemble.py
@@ -44,7 +44,7 @@ from ._estimator import NBEATSEstimator
 # None is also a valid parameter
 AGGREGATION_METHODS = "median", "mean", "none"
 
-
+# TODO: change parent class to RepresentablePredictor
 class NBEATSEnsemblePredictor(Predictor):
     """"
     An ensemble predictor for N-BEATS.
@@ -163,6 +163,10 @@ class NBEATSEnsemblePredictor(Predictor):
 
             for iterator in iterators:
                 prediction = next(iterator)
+
+                # on order to avoid mypys complaints
+                assert isinstance(prediction, SampleForecast)
+
                 output.append(prediction.samples)
 
                 # get the forecast start date
@@ -179,6 +183,9 @@ class NBEATSEnsemblePredictor(Predictor):
                 output = np.mean(output, axis=0)
             else:  # "none": do not aggregate
                 pass
+
+            # on order to avoid mypys complaints
+            assert start_date is not None
 
             yield SampleForecast(
                 output,
--- a/src/gluonts/model/n_beats/_ensemble.py
+++ b/src/gluonts/model/n_beats/_ensemble.py
@@ -22,9 +22,11 @@ import logging
 # Third-party imports
 import mxnet as mx
 import numpy as np
+from pydantic import ValidationError
 
 # First-party imports
-from gluonts.core.component import validated
+from gluonts.core.component import validated, from_hyperparameters
+from gluonts.core import fqname_for
 from gluonts.core.serde import dump_json, load_json
 from gluonts.dataset.common import Dataset
 from gluonts.dataset.field_names import FieldName
@@ -32,6 +34,7 @@ from gluonts.dataset.loader import DataBatch
 from gluonts.model.estimator import Estimator
 from gluonts.model.forecast import Forecast, SampleForecast
 from gluonts.model.predictor import Predictor, RepresentableBlockPredictor
+from gluonts.core.exception import GluonTSHyperparametersError
 from gluonts.trainer import Trainer
 
 # Relative imports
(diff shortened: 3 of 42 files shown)


src/gluonts/model/n_beats/_network.py
```diff
@@ -0,0 +1,320 @@
...
+    prediction_length,
+    is_forecast,
+):
+    p = expansion_coefficient_length
+    t = linear_space(
+        F, context_length, prediction_length, fwd_looking=is_forecast
+    )
+    T = F.transpose(F.stack(*[t ** i for i in range(p)], axis=0))
+    T = F.cast(T, np.float32)
+    return F.dot(thetas, F.transpose(T))
```
AaronSpieler: I have basically used the exact same approach as here: https://github.com/philipperemy/n-beats/blob/master/nbeats_pytorch/model.py
This implementation contradicts the paper in one point by always dividing by backcast_length (even for forecast).
However, the n-beats paper is unfortunately under-specifying how to implement this part regarding the backcast and this implementation seems reasonable.

AaronSpieler: I tryed couple of different ways of implementing this, they dont seem to drastically vary in performance. Also I reworked the whole seasonal and trend model implementation.

AaronSpieler:
Something thats weird tough is that on the m4 dataset with the sMAPE loss function the interpretable network doesnt seem to learn. It leans with MAPE, but not sMAPE. However, the generic mode learns with both.

src/gluonts/model/n_beats/_estimator.py
```diff
@@ -0,0 +1,287 @@
...
+        Length of the prediction horizon
+    context_length
+        Number of time units that condition the predictions
+        (default: None, in which case context_length = prediction_length)
+    trainer
+        Trainer object to be used (default: Trainer())
+    num_stacks:
+        The number of stacks the network should contain.
+        Default and recommended value for generic mode: 30
+        Recommended value for interpretable mode: 2
```
lostella: Are these the defaults/recommended values mentioned in the paper?

src/gluonts/model/n_beats/_estimator.py
```diff
@@ -0,0 +1,287 @@
...
+    widths:
+        Widths of the fully connected layers with ReLu activation.
+        A list of ints of length 1 or 'num_stacks'.
+        Default and recommended value for generic mode: [512]
+        Recommended value for interpretable mode: [256, 2048]
+    blocks:
+        The number of blocks blocks per stack.
```
lostella:
```suggestion
        The number of blocks per stack.
```

src/gluonts/model/n_beats/_estimator.py
```diff
@@ -0,0 +1,287 @@
...
+        Number of time units that condition the predictions
+        (default: None, in which case context_length = prediction_length)
+    trainer
+        Trainer object to be used (default: Trainer())
+    num_stacks:
+        The number of stacks the network should contain.
+        Default and recommended value for generic mode: 30
+        Recommended value for interpretable mode: 2
+    widths:
+        Widths of the fully connected layers with ReLu activation.
```
lostella: Check how a similar hyperparameter is named in other models: I believe this is `num_units`.

Also, to make the sequence of hyperparameters more logical, maybe this should go a bit more below next to the `block_layers`

src/gluonts/model/n_beats/_estimator.py
```diff
@@ -0,0 +1,287 @@
...
+        A list of ints of length 1 or 'num_stacks'.
+        Default and recommended value for generic mode: [512]
+        Recommended value for interpretable mode: [256, 2048]
+    blocks:
```
lostella: Maybe this is better

```suggestion
    num_blocks:
```

src/gluonts/model/n_beats/_estimator.py
```diff
@@ -0,0 +1,287 @@
...
+        The number of blocks blocks per stack.
+        A list of ints of length 1 or 'num_stacks'.
+        Default and recommended value for generic mode: [1]
+        Recommended value for interpretable mode: [3]
+    block_layers:
```
lostella:
```suggestion
    num_block_layers:
```

src/gluonts/model/n_beats/_estimator.py
```diff
@@ -0,0 +1,287 @@
...
+    freq
+        Time time granularity of the data
+    prediction_length
+        Length of the prediction horizon
+    context_length
+        Number of time units that condition the predictions
+        (default: None, in which case context_length = prediction_length)
+    trainer
+        Trainer object to be used (default: Trainer())
+    num_stacks:
```
lostella: Note that the colon is not needed (also in the other parameters)

```suggestion
    num_stacks
```

src/gluonts/model/n_beats/_network.py
```diff
@@ -0,0 +1,521 @@
...
+        past_target
+            Tensor with past observations.
+            Shape: (batch_size, context_length, target_dim).
+        future_target
+            Not used.
+
+        Returns
+        -------
+        Tensor
+            Prediction sample. Shape: (samples, batch_size, prediction_length).
```
lostella:
```suggestion
            Prediction sample. Shape: (1, batch_size, prediction_length).
```

src/gluonts/model/n_beats/_network.py
```diff
@@ -0,0 +1,521 @@
...
+        """
+        # future_target never used
+        forecasts = super().hybrid_forward(
+            F, past_target=past_target, future_target=past_target
+        )
+        # dimension collapsed previously because we only have one sample each:
+        forecasts = F.expand_dims(forecasts, axis=1)
```
lostella: Should it be `axis=0` here? I mean in order to get the shape as in the docstring (which I believe is the correct one that the `SampleForecast` class expects)

src/gluonts/model/n_beats/_estimator.py
AaronSpieler: Yes. They can be found on page 26 of the paper.

src/gluonts/model/n_beats/_estimator.py
AaronSpieler: I tried to follow the nomenclature of the paper as close as possible. Except for prediction_length and context_length. But I guess in this case it also makes sense.

src/gluonts/model/n_beats/_network.py
AaronSpieler: No, it has to be 1 here. Axis one is the num_samples axis. The axis is removed somewhere if there is only one sample, this leads to dimensionality errors down the line.

src/gluonts/model/n_beats/_network.py
AaronSpieler: I see, the documentation from where I copied this seems to be out of date. The first two are switched.

lostella: Oh I see: could you open PR to fix these doc issues as you find them? In any case, if `samples` is `1` in this case, it's better to put `1` directly, whatever the axis

src/gluonts/model/n_beats/_estimator.py
AaronSpieler: Actually I looked at the other models and the naming is heterogenous. There is num_cells, num_hidden_global, num_hidden_dimensions. Considering that widths is what the paper calls it, I would leave it like this.

AaronSpieler: I moved widths below block_layers.

src/gluonts/model/n_beats/_estimator.py
AaronSpieler: done

src/gluonts/model/n_beats/_estimator.py
AaronSpieler: done

src/gluonts/model/n_beats/_estimator.py
AaronSpieler: done

src/gluonts/model/n_beats/_estimator.py
AaronSpieler: done

src/gluonts/model/n_beats/_network.py
AaronSpieler: Yeah I changed it to 1. Somehow I missed to update the doc here. I'll open a PR for this.

AaronSpieler:
```        
        Tensor
            Prediction sample. Shape: (samples, batch_size, prediction_length).
        """
        distr = self.get_distr(F, past_target)

        # (num_samples, batch_size, prediction_length)
        samples = distr.sample(self.num_parallel_samples)

        # (batch_size, num_samples, prediction_length)
        return samples.swapaxes(0, 1)```

This is from simple feed forward. Interestingly the comment is correct, but the doc isnt.

File: src/gluonts/model/n_beats/_ensemble.py [hunk 10]
```diff
@@ -0,0 +1,366 @@
...
+
+            for iterator in iterators:
+                prediction = next(iterator)
+                output.append(prediction.samples)
+
+                # some weird transformation is going on with the start date,
+                # so I just take the already transformed one, otherwise
+                # evaluator will give Nans for the seasonal_error
+                if start_date is None:
+                    start_date = prediction.start_date
```
AaronSpieler: I checked and saw its the forecast start date, but idk whether there is an easier way to calculate it, then just grabbing it here.

src/gluonts/model/n_beats/_ensemble.py
```diff
@@ -0,0 +1,366 @@
...
+
+            # aggregating output of different models
+            # default according to paper is median,
+            # but we can also make use of not aggregating
+            if self.aggregation_method == "median":
+                output = np.median(output, axis=0)
+            elif self.aggregation_method == "mean":
+                output = np.mean(output, axis=0)
+            else:  # do not aggregate
+                pass
```
AaronSpieler: I looked at the unaggregated forecast results, and it is not trivial to aggregate them later, because the actual forecasts are in a list of dicts. Its just way easier and more convenient to specify the aggregation method as an argument that can be set right before the evaluation. Additionally setting the default the median would be according to the paper specification.

AaronSpieler: Bit mean seems to be doing good too.

src/gluonts/model/n_beats/_ensemble.py
```diff
@@ -0,0 +1,366 @@
...
+        # we always have to predict for one series in the dataset with
+        # all models and return it as a 'SampleForecast' so that it is
+        # clear that all these prediction concern the same series
+        for item in dataset:
+            output = []
+            start_date = None
+
+            for iterator in iterators:
+                prediction = next(iterator)
+                output.append(prediction.samples)
```
AaronSpieler: I get a weird typechecking error (or rather warning). The typechecker thinks this is a Forecast object, even though its a SampleForecast object. 
Can we change the signature to reflect that? Or will this cause problems? We would have to change a couple.

AaronSpieler:
I dont think the EnsemblePredictor predict method is quite there yet, not really elegant, so suggestions are appreciated.
I will start working on adding tests now.

src/gluonts/model/n_beats/_ensemble.py
lostella: You can add an assertion on the type of forecast, and that should remove mypy's (as well as pycharm's) complaints:

```python
assert isinstance(prediction, `SampleForecast`)
```

src/gluonts/model/n_beats/_network.py
```diff
@@ -0,0 +1,531 @@
...
+    t = linear_space(
+        F, context_length, prediction_length, fwd_looking=is_forecast
+    )
+    cosines = F.stack(
+        *[F.cos(2 * np.pi * i * t) for i in range(num_coefficients)]
+    )
+    sines = F.stack(
+        *[F.sin(2 * np.pi * i * t) for i in range(num_coefficients)]
+    )
+    S = F.concat(cosines, sines, dim=0)
```
lostella: All these "feature" tensors are allocated every time the `NBEATSBlock` is invoked, so this could be optimized (same for `trend_model`). In the end, in interpretable modes, computing the forecast and backcast amounts to doing the final `F.dot` operation between the `thetas` and the block-type-specific feature tensor (lines 63 and 81 here) which can be pre-allocated. In the generic mode this is instead a dense layer.

One option is the following:
* Define the `NBEATSBlock` base class, implementing the `hybrid_forward` pretty much as it is now.
* Extend the base class into `NBEATSGenericBlock`, `NBEATSTrendBlock`, `NBEATSSeasonalBlock`, each one allocating *at construction time* the appropriate feature tensors given the hyperparameters (for the interpretable modes) or constructing the dense layer (for the generic mode), in addition to constructing the dense layers used to compute theta.
* Have the network decide what blocks to fill in the stacks with.

This comes maybe at the cost of a bit of additional boilerplate code (one constructor for each block class), but I think it's better than using the allocating hybrid lambda blocks wrapping a lambda wrapping a function.

src/gluonts/model/n_beats/_network.py
```diff
@@ -0,0 +1,531 @@
...
+        self.context_length = context_length
+        self.has_backcast = has_backcast
+
+        with self.name_scope():
+            self.fc_stack = mx.gluon.nn.HybridSequential()
+            for i in range(num_block_layers):
+                self.fc_stack.add(
+                    mx.gluon.nn.Dense(units=self.width, activation="relu")
+                )
+            if block_type == "G":
```
lostella: As mentioned in the comment above, maybe instead of having this switching mechanism here you could have different block subtypes.

src/gluonts/model/n_beats/_network.py
```diff
@@ -0,0 +1,531 @@
...
+                )
+            elif block_type == "S":
+                num_coefficients = (
+                    int((prediction_length / 2) - 1) + 1
+                )  # according to paper, dont use floor because prediction_length=1 should be mapped to 0
+                if has_backcast:
+                    self.theta_backcast = mx.gluon.nn.Dense(
+                        units=2 * num_coefficients  # linear activation:
+                    )
+                    self.backcast = mx.gluon.nn.HybridLambda(
```
lostella: In my idea, I think this would just be a method of the subclass in the interpretable modes (which just does the `F.dot`) or a dense layer in the generic mode.

src/gluonts/model/n_beats/_ensemble.py
```diff
@@ -0,0 +1,427 @@
...
```
lostella: You can add an assertion here to get rid of mypy's complaints (see the bottom of the log https://ci.mxnet.io/blue/rest/organizations/jenkins/pipelines/GluonTS-py3-cpu-unittest/branches/PR-553/runs/9/nodes/42/steps/66/log/?start=0)

src/gluonts/model/n_beats/_ensemble.py
```diff
@@ -0,0 +1,427 @@
...
+                output = np.median(output, axis=0)
+            elif self.aggregation_method == "mean":
+                output = np.mean(output, axis=0)
+            else:  # "none": do not aggregate
+                pass
+
+            yield SampleForecast(
+                output,
+                start_date=start_date,
+                freq=start_date.freqstr,
```
lostella: You can assert that `start_date is not None` before the get rid of mypy's complaint (https://ci.mxnet.io/blue/rest/organizations/jenkins/pipelines/GluonTS-py3-cpu-unittest/branches/PR-553/runs/9/nodes/42/steps/66/log/?start=0)

src/gluonts/model/n_beats/_network.py
AaronSpieler: Very good catch! I agree this makes sense,  besides performance this should improve code readability, since the conceptual segmentation of the different block types is also present in the class structure then.

AaronSpieler:
By separating the block-types into their different classes I think its not so straight forward anymore to see what a block is doing overall.
Also, the base class is nor really usable on its own, it's more of an abstract class. It is essentially not usable unless it's subclassed, since the layers it is using in hybrid_forward have to be defined still in its subclass.

.github/workflows/mxnet-nightly.yml
```diff
@@ -32,4 +32,4 @@ jobs:
         pip install -r requirements/requirements-test.txt
     - name: Test with pytest
       run: |
-        pytest -m 'not (gpu or serial)' --cov src/gluonts --cov-report=term --cov-report xml test
+        pytest --timeout=60 -m 'not (gpu or serial)' --cov src/gluonts --cov-report=term --cov-report xml test
```
lostella:
```suggestion
        pytest -m 'not (gpu or serial)' --cov src/gluonts --cov-report=term --cov-report xml test
```

.github/workflows/pythonpackage.yml
```diff
@@ -28,4 +28,4 @@ jobs:
         pip install -r requirements/requirements-test.txt
     - name: Test with pytest
       run: |
-        pytest -m 'not (gpu or serial)' --cov src/gluonts --cov-report=term --cov-report xml test
+        pytest --timeout=60 -m 'not (gpu or serial)' --cov src/gluonts --cov-report=term --cov-report xml test
```
lostella:
```suggestion
        pytest -m 'not (gpu or serial)' --cov src/gluonts --cov-report=term --cov-report xml test
```

lostella: Let's just not add these I guess, I think this will prevent the windows builds from failing

src/gluonts/model/n_beats/_ensemble.py
```diff
@@ -0,0 +1,441 @@
...
+            predictor.serialize(composite_path)
+
+        # serialize all remaining constructor parameters
+        with (path / "parameters.json").open("w") as fp:
+            parameters = dict(
+                prediction_length=self.prediction_length,
+                freq=self.freq,
+                aggregation_method=self.aggregation_method,
+                num_predictors=len(self.predictors),
+            )
```
AaronSpieler: Upon doing the evaluation I noticed that it would be useful to include some information here about what kind of network is serialised here. At this point there is basically no way to retrieve this information other than looking at the weights.

AaronSpieler: Ok, I just missed the files, its already working, my bad.

PR #560, opened 2020-01-15
Remove Dataset class in favor of type alias

lostella:
*Description of changes:* This removes the `Dataset` class in favor of a type alias, allowing e.g. to just use a list with the appropriate fields as dataset.

```python
train_data = [
    "start": pd.Timestamp(...),
    "target": np.array([...])
]
```

### DIFF
--- a/src/gluonts/evaluation/backtest.py
+++ b/src/gluonts/evaluation/backtest.py
@@ -33,6 +33,7 @@ from gluonts.evaluation import Evaluator
 from gluonts.model.estimator import Estimator, GluonEstimator
 from gluonts.model.forecast import Forecast
 from gluonts.model.predictor import GluonPredictor, Predictor
+from gluonts.support.util import maybe_len
 from gluonts.transform import TransformedDataset
 
 
@@ -201,7 +202,7 @@ def backtest_metrics(
     )
 
     agg_metrics, item_metrics = evaluator(
-        ts_it, forecast_it, num_series=len(test_dataset)
+        ts_it, forecast_it, num_series=maybe_len(test_dataset)
     )
 
     # we only log aggregate metrics for now as item metrics may be very large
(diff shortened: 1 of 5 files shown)


src/gluonts/model/trivial/mean.py
```diff
@@ -112,12 +112,11 @@ def train(
         training_data: Dataset,
         validation_dataset: Optional[Dataset] = None,
     ) -> ConstantPredictor:
-        contexts = np.broadcast_to(
```
jaheba: This doesn't look like it belongs to this PR.

src/gluonts/evaluation/backtest.py
```diff
@@ -201,7 +202,7 @@ def backtest_metrics(
     )
 
     agg_metrics, item_metrics = evaluator(
-        ts_it, forecast_it, num_series=len(test_dataset)
+        ts_it, forecast_it, num_series=maybe_len(test_dataset)
```
jaheba: Why is this needed?

What happens if it returns `None`?

src/gluonts/model/trivial/mean.py
lostella: This is needed in the PR, since `broadcast_to` requires using `len` on the `Dataset` (when it's not strictly needed), but the method may not be available being `Dataset` an `Iterable`.

src/gluonts/evaluation/backtest.py
lostella: Same as above: `len` may not be available. In `Evaluator.__call__` the `num_series` argument is `Optional`, and is only used for displaying progress: if a dataset has no `len`, the progress will be shown only partially (no percentage wrt total, only number of processed items).

src/gluonts/model/trivial/mean.py
lostella: That's mypy who asked me to do this

PR #562, opened 2020-01-16
Removed timeouts from tests.

jaheba:
*Issue #, if available:*

*Description of changes:*


lostella:
Do you know which timeout in particular is causing issues? Just to go there and verify that there’s no regression

Issue #563, opened 2020-01-16
PiecewiseLinear gives wrong CRPS outside of distribution support


lostella:
It seems like the PiecewiseLinear's CRPS method does not compute correct results when the input falls outside of the distribution's support. The following example is for a uniform distribution between -1 and 1: I would expect CRPS(-2) = CRPS(2), but that's not the case.

```python
gamma = mx.nd.array([-1.0])
slopes = mx.nd.array([
    [2.0, 2.0]]
)
knot_spacings = mx.nd.array([[0.5, 0.5]])

distr = PiecewiseLinear(
    gamma=gamma,
    slopes=slopes,
    knot_spacings=knot_spacings,
)

print(distr.crps(mx.nd.array([-2.0])))
print(distr.crps(mx.nd.array([2.0])))
```

yields

```
[1.6666667]
<NDArray 1 @cpu(0)>

[2.1666667]
<NDArray 1 @cpu(0)>
```

With more complex example one can also get negative values of CRPS.

```python
gamma = mx.nd.array([-82.70746])
slopes = mx.nd.array([
    [1.8083544e+01, 2.9134894e-02, 6.3528619e+00, 4.8505863e-05, 4.9647682e+01, 3.9714784e-16, 1.3027676e+01, 1.8006723e-14]]
)
knot_spacings = mx.nd.array([[0.001, 0.009, 0.09, 0.4, 0.4, 0.09, 0.009, 0.001]])
x = mx.nd.array([-70.8735308])

distr = PiecewiseLinear(
    gamma=gamma,
    slopes=slopes,
    knot_spacings=knot_spacings,
)

print(distr.crps(x))
```

yields

```
[-79.25251]
<NDArray 1 @cpu(0)>
```

lostella:
We could consider implement the computation of CRPS in its more intuitive form, see first equation in https://journals.ametsoc.org/doi/pdf/10.1175/1520-0434%282000%29015%3C0559%3ADOTCRP%3E2.0.CO%3B2

If all slopes are strictly positive then the quantile function is invertible and maybe things are easier to compute in the CDF domain instead of the quantile level domain.

lostella:
Well, one should probably end up with the same formula for the integral no matter what (equivalent) formulation one starts from. It looks like some summands are missing when x falls off of the support.

lostella:
fixed by #564

PR #564, opened 2020-01-17
Fix bug in cdf method of piecewise linear distributions

lostella:
*Issue #, if available:* partially addressing #563 

*Description of changes:* This fixes a bug in the cdf method of PiecewiseLinear, which was returning values larger than 1.0: this addresses the simpler example in #563. The more complex example there has to do with numerical problems, which make the cdf evaluation fail there, and crps as well as a consequence I believe. That problem will need more care.

### DIFF
--- a/src/gluonts/distribution/piecewise_linear.py
+++ b/src/gluonts/distribution/piecewise_linear.py
@@ -336,15 +336,7 @@ class PiecewiseLinearOutput(DistributionOutput):
     @classmethod
     def domain_map(cls, F, gamma, slopes, knot_spacings):
         # slopes of the pieces are non-negative
-        thresh = F.zeros_like(slopes) + 20.0
-        I_grt_thresh = F.broadcast_greater_equal(slopes, thresh)
-        I_sml_thresh = F.broadcast_lesser(slopes, thresh)
-        slopes_proj = (
-            I_sml_thresh
-            * F.Activation(data=(I_sml_thresh * slopes), act_type="softrelu")
-            + I_grt_thresh * slopes
-        )
-        # slopes = F.Activation(slopes, 'softrelu')
+        slopes_proj = F.Activation(data=slopes, act_type="softrelu") + 1e-4
 
         # the spacing between the knots should be in [0, 1] and sum to 1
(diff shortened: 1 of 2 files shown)


test/distribution/test_piecewise_linear.py
```diff
@@ -136,3 +136,21 @@ def test_shapes(
...
+    distr = PiecewiseLinear(
+        gamma=gamma, slopes=slopes, knot_spacings=knot_spacings
+    )
+
+    assert distr.cdf(mx.nd.array([-2.0])).asnumpy().item() == 0.0
+    assert distr.cdf(mx.nd.array([+2.0])).asnumpy().item() == 1.0
+
+    assert np.all(
+        distr.crps(mx.nd.array([-2.0])).asnumpy()
+        == distr.crps(mx.nd.array([+2.0])).asnumpy()
```
lostella: I’ll actually put a check here to make sure these equal the actual hand-computed CRPS

src/gluonts/distribution/piecewise_linear.py
```diff
@@ -336,15 +336,7 @@ def __init__(self, num_pieces: int) -> None:
...
         # slopes of the pieces are non-negative
-        thresh = F.zeros_like(slopes) + 20.0
-        I_grt_thresh = F.broadcast_greater_equal(slopes, thresh)
-        I_sml_thresh = F.broadcast_lesser(slopes, thresh)
-        slopes_proj = (
-            I_sml_thresh
-            * F.Activation(data=(I_sml_thresh * slopes), act_type="softrelu")
-            + I_grt_thresh * slopes
-        )
-        # slopes = F.Activation(slopes, 'softrelu')
```
lostella: I'm not sure what all this logic was needed for: I checked and softrelu seems to behave fine with inputs much larger than 20 (when it's essentially the identity function). Maybe this was added in a tentative of fixing all other issues?

vafl: Yes, this is outdated. There used to be an mxnet bug but the fix should already be in 1.4 (https://github.com/apache/incubator-mxnet/pull/6834/files).

PR #585, opened 2020-01-30
Fixed minor bugs in GluonTSFramework.

AaronSpieler:
*Description of changes:*
Fixed bug regarding wait parameter in train function of GluonTSFramework.


jaheba:
Is that base_job_name handling the same as in other frameworks?

AaronSpieler:
> Is that base_job_name handling the same as in other frameworks?

Actually it does not seem to be that way, job_name will be the actual job name. However, I find that counter intuitive, since one already defines a base_job_name.

jaheba:
Then let’s not do that as well. The idea behind the base job is to specify the job name just once and then spawn of jobs from that.

src/gluonts/nursery/sagemaker_sdk/estimator.py
```diff
@@ -528,7 +528,7 @@ def train(
         wait: bool = True,
         logs: bool = True,
         job_name: str = None,
-    ) -> Tuple[Predictor, dict, pd.DataFrame, str]:
+    ) -> Optional[Tuple[Predictor, dict, pd.DataFrame, str]]:
```
jaheba: I think we always should return something. How do I know the training-job-name if I only the base_job_name?

Also, the return type looks off, shouldn't it be a `TrainResult`?

src/gluonts/nursery/sagemaker_sdk/utils.py
```diff
@@ -19,7 +19,7 @@
 def make_metrics(metrics_names):
     avg_epoch_loss_metric = {
         "Name": "training_loss",
-        "Regex": r"'avg_epoch_loss'=(\S+)",
+        "Regex": r"'epoch_loss'=(\S+)",
```
jaheba: There is no avg_epoch_loss?

AaronSpieler: If you look into the logs, there is one, but not in lines with "INFO:...", so apparently the metric extraction only works on info lines, thats why its not showing up in the charts.

src/gluonts/nursery/sagemaker_sdk/estimator.py
AaronSpieler: Could be. A lot of things changed after the refactoring so I will have too look that up.

AaronSpieler: Also you're right, we could always return the complete job name.

AaronSpieler:
> Then let’s not do that as well. The idea behind the base job is to specify the job name just once and then spawn of jobs from that.

ok

src/gluonts/nursery/sagemaker_sdk/utils.py
lostella: I verified this: indeed, `avg_epoch_loss` is displayed in the lines produced by tqdm, not in the logging lines

alexw91:
# [Codecov](https://codecov.io/gh/awslabs/gluon-ts/pull/585?src=pr&el=h1) Report
> Merging [#585](https://codecov.io/gh/awslabs/gluon-ts/pull/585?src=pr&el=desc) into [master](https://codecov.io/gh/awslabs/gluon-ts/commit/b6ec6a09c73c3daeeb68ae72fab5acaf5238d053&el=desc) will **decrease** coverage by `0.01%`.
> The diff coverage is `n/a`.


[Impacted file tree graph](https://codecov.io/gh/awslabs/gluon-ts/pull/585?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #585      +/-   ##
==========================================
- Coverage   83.73%   83.72%   -0.02%     
==========================================
  Files         181      181              
  Lines       10324    10326       +2     
==========================================
  Hits         8645     8645              
- Misses       1679     1681       +2     
```

PR #586, opened 2020-01-31
Change causal conv1d default activation to match the doc

ehsanmok:
*Description of changes:*

Default activation used in [CausalConv1d](https://gluon-ts.mxnet.io/api/gluonts/gluonts.block.cnn.html#) doesn't match the code. Changing it to`None` from `"relu"`.


lostella:
@jgasthaus @lovvge do you know what the intended default is for this option? Either the default value is changed to match the docstring (as per the proposed fix), or the other way around.

jgasthaus:
> @jgasthaus @lovvge do you know what the intended default is for this option? Either the default value is changed to match the docstring (as per the proposed fix), or the other way around.

Changing the default to None (as the doc states) makes sense to me. All callers do explicitly set it to "relu" when needed.

PR #588, opened 2020-01-31
Updated N-Beats defaults.

AaronSpieler:
*Description of changes:*
I found that the trend block expansion_coefficient_lengths default is off by one when compared to the paper. I checked again, and the paper doesn't specify the dimension of expansion_coefficient_lengths for the generic block. Before it was the same as for the trend block, however, I think it makes sense to increase it to 32, since for example the seasonality block has horizon/2 number of coefficients, so its reasonable to give the generic block more possibilities to learn a larger generic basis.
The changes marginally improve performance in some cases.

### DIFF
--- a/src/gluonts/model/n_beats/_ensemble.py
+++ b/src/gluonts/model/n_beats/_ensemble.py
@@ -296,7 +296,7 @@ class NBEATSEnsembleEstimator(Estimator):
         If type is "T" (trend), then it corresponds to the degree of the polynomial.
         If the type is "S" (seasonal) then its not used.
         A list of ints of length 1 or 'num_stacks'.
-        Default and recommended value for generic mode: [32]
+        Default value for generic mode: [32]
         Recommended value for interpretable mode: [3]
     stack_types
         One of the following values: "G" (generic), "S" (seasonal) or "T" (trend).
--- a/src/gluonts/model/n_beats/_ensemble.py
+++ b/src/gluonts/model/n_beats/_ensemble.py
@@ -296,8 +296,8 @@ class NBEATSEnsembleEstimator(Estimator):
         If type is "T" (trend), then it corresponds to the degree of the polynomial.
         If the type is "S" (seasonal) then its not used.
         A list of ints of length 1 or 'num_stacks'.
-        Default and recommended value for generic mode: [2]
-        Recommended value for interpretable mode: [2]
+        Default and recommended value for generic mode: [32]
+        Recommended value for interpretable mode: [3]
     stack_types
         One of the following values: "G" (generic), "S" (seasonal) or "T" (trend).
         A list of strings of length 1 or 'num_stacks'.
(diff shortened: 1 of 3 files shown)


src/gluonts/model/n_beats/_ensemble.py
```diff
@@ -296,8 +296,8 @@ class NBEATSEnsembleEstimator(Estimator):
         If type is "T" (trend), then it corresponds to the degree of the polynomial.
         If the type is "S" (seasonal) then its not used.
         A list of ints of length 1 or 'num_stacks'.
-        Default and recommended value for generic mode: [2]
-        Recommended value for interpretable mode: [2]
+        Default and recommended value for generic mode: [32]
+        Recommended value for interpretable mode: [3]
```
lostella: A minor comment here: 32 should mentioned just as “default” (not “recommended”), since it’s not mentioned in the paper, and 3 if I understand correctly is also “default” (other than “recommended”). Similarly in the other affected modules.

AaronSpieler: Actually, the polynomial degree of 2 is mentioned in the paper (to get that the parameter needs to be 3 though, not 2). The rest is True.

src/gluonts/model/n_beats/_ensemble.py
```diff
@@ -296,8 +296,8 @@ class NBEATSEnsembleEstimator(Estimator):
         If type is "T" (trend), then it corresponds to the degree of the polynomial.
         If the type is "S" (seasonal) then its not used.
         A list of ints of length 1 or 'num_stacks'.
-        Default and recommended value for generic mode: [2]
-        Recommended value for interpretable mode: [2]
+        Default value for generic mode: [32]
+        Recommended value for interpretable mode: [3]
```
lostella: Yes, my suggestion was to also do this change

```suggestion
        Default and recommended value for interpretable mode: [3]
```

src/gluonts/model/n_beats/_estimator.py
```diff
@@ -91,8 +91,8 @@ class NBEATSEstimator(GluonEstimator):
         If type is "T" (trend), then it corresponds to the degree of the polynomial.
         If the type is "S" (seasonal) then its not used.
         A list of ints of length 1 or 'num_stacks'.
-        Default and recommended value for generic mode: [2]
-        Recommended value for interpretable mode: [2]
+        Default value for generic mode: [32]
+        Recommended value for interpretable mode: [3]
```
lostella:
```suggestion
        Default and recommended value for interpretable mode: [3]
```

src/gluonts/model/n_beats/_network.py
```diff
@@ -396,8 +396,8 @@ class NBEATSNetwork(mx.gluon.HybridBlock):
         If type is "T" (trend), then it corresponds to the degree of the polynomial.
         If the type is "S" (seasonal) then its not used.
         A list of ints of length 1 or 'num_stacks'.
-        Default and recommended value for generic mode: [2]
-        Recommended value for interpretable mode: [2]
+        Default value for generic mode: [32]
+        Recommended value for interpretable mode: [3]
```
lostella:
```suggestion
        Default and recommended value for interpretable mode: [3]
```

AaronSpieler:
> Looks good! good that you spotted this

Thx! Yeah I was combing through everything again to see what could be the reason for some of the weird values I was getting when taking the mean.

PR #593, opened 2020-01-31
Fixed sMAPE related learning issues.

AaronSpieler:
*Issue #, if available:*

Using the sMAPE function in conjunction with a trainer with a learning rate above 1e-4 lead to the models not converging/ learning at all in many cases (i.e. different datasets and different model instantiation like generic vs interpretable etc.).

This is now fixed by stopping gradient flow in the denominator of sMAPE.

### DIFF
--- a/src/gluonts/model/n_beats/_network.py
+++ b/src/gluonts/model/n_beats/_network.py
@@ -509,7 +509,6 @@ class NBEATSNetwork(mx.gluon.HybridBlock):
             # connect last block
             return forecast + self.net_blocks[-1](backcast)
 
-    # TODO: somehow interpretable mode not learning?
     def smape_loss(self, F, forecast: Tensor, future_target: Tensor) -> Tensor:
         r"""
         .. math::
@@ -519,7 +518,8 @@ class NBEATSNetwork(mx.gluon.HybridBlock):
         According to paper: https://arxiv.org/abs/1905.10437.
         """
 
-        denominator = F.abs(future_target) + F.abs(forecast)
+        # Stop gradient required for stable learning
+        denominator = F.stop_gradient(F.abs(future_target) + F.abs(forecast))
         flag = denominator == 0
 
         smape = (200 / self.prediction_length) * F.mean(


alexw91:
# [Codecov](https://codecov.io/gh/awslabs/gluon-ts/pull/593?src=pr&el=h1) Report
> Merging [#593](https://codecov.io/gh/awslabs/gluon-ts/pull/593?src=pr&el=desc) into [master](https://codecov.io/gh/awslabs/gluon-ts/commit/8a62e002af536af69a6d9252477e047f370f38da&el=desc) will **decrease** coverage by `0.05%`.
> The diff coverage is `n/a`.


[Impacted file tree graph](https://codecov.io/gh/awslabs/gluon-ts/pull/593?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #593      +/-   ##
==========================================
- Coverage   83.73%   83.68%   -0.06%     
==========================================
  Files         181      181              
  Lines       10324    10340      +16     
==========================================
+ Hits         8645     8653       +8     
- Misses       1679     1687       +8     
```

src/gluonts/model/n_beats/_ensemble.py
```diff
@@ -363,9 +364,25 @@ def __init__(
...
+            if "sMAPE" in self.meta_loss_function:
+                trainer = Trainer(learning_rate=1e-4)
+            else:
+                trainer = Trainer()
+        assert trainer is not None
+        if trainer.learning_rate > 1e-4 and "sMAPE" in self.meta_loss_function:
+            logging.warning(
+                "You may experience issues when using 'sMAPE' as "
+                "loss-function and a learning rate above 1e-4."
+            )
```
lostella: I'm wondering if we want to keep this warning here. The thing is, with any loss you'll get problems if the learning rate is too large. Maybe having the ad-hoc default learning rate for "sMAPE" is fine for now.

Same thing applies to the _estimator.py below

AaronSpieler: Thats true, however, in this case it actually makes all the difference. So the second evaluation with a learning rate of 1e-4 is done, and now all the weird behavior is gone, so thats like a night and day difference. 

So if I was doing my experience as a customer and I wanted to modify some property of the trainer (but not changing the learning rate, aka keeping the default of 1e-3), I would suddenly get super weird behavior, I would very much appreciate to know why and not have to debug by myself. The warning would in this case save a lot of headache.

In fact, looking at the experiments, I think we could make the default learning rate for all loss functions 1e-4.

alexw91:
# [Codecov](https://codecov.io/gh/awslabs/gluon-ts/pull/593?src=pr&el=h1) Report
> Merging [#593](https://codecov.io/gh/awslabs/gluon-ts/pull/593?src=pr&el=desc) into [master](https://codecov.io/gh/awslabs/gluon-ts/commit/8a62e002af536af69a6d9252477e047f370f38da&el=desc) will **not change** coverage by `%`.
> The diff coverage is `n/a`.


[Impacted file tree graph](https://codecov.io/gh/awslabs/gluon-ts/pull/593?src=pr&el=tree)

```diff
@@           Coverage Diff           @@
##           master     #593   +/-   ##
=======================================
  Coverage   83.73%   83.73%           
=======================================
  Files         181      181           
  Lines       10324    10324           
=======================================
  Hits         8645     8645           
  Misses       1679     1679           
```

PR #594, opened 2020-02-03
fix `time_feat` in deepvar

kashif:
When creating `time_feat ` don't we need to begin the slice from `self.history_length-self.context_length`?

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/model/deepvar/_network.py
+++ b/src/gluonts/model/deepvar/_network.py
@@ -323,7 +323,7 @@ class DeepVARNetwork(mx.gluon.HybridBlock):
 
         if future_time_feat is None or future_target_cdf is None:
             time_feat = past_time_feat.slice_axis(
-                axis=1, begin=-self.context_length, end=None
+                axis=1, begin=self.history_length-self.context_length, end=None
             )
             sequence = past_target_cdf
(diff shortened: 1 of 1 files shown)


mbohlkeschneider:
Hi @kashif,

thank you for your PR. 

I think code is equivalent (but maybe I'm missing something. Check this snippet: 

```
import numpy as np
import mxnet as mx

history_length = 100
context_length = 20

past_target = np.array([np.arange(1, 101), np.arange(1, 101)])

past_target_mxnet = mx.nd.array(past_target)

past_target_hist = past_target_mxnet.slice_axis(axis=1, begin=history_length - context_length, end=None)
past_target_con = past_target_mxnet.slice_axis(axis=1, begin=-context_length, end=None)

print(past_target_hist)
print(past_target_con)
assert (past_target_hist.asnumpy() == past_target_con.asnumpy()).all()
```

kashif:
thanks @mbohlkeschneider for the quick reply! I will have a look.

kashif:
@mbohlkeschneider yes you are right... I was thinking i was in a situation where the size of the time series was bigger than `history_length`... closing it.

PR #602, opened 2020-02-05
Introducing baseline models. Includes new Naive 2 model, new OWA metric, and a lot of refactoring.

AaronSpieler:
Added the Naïve 2 model as well as the OWA (overall weighted average) metric as described in the m4 competition: [here](https://www.m4.unic.ac.cy/wp-content/uploads/2018/03/M4-Competitors-Guide.pdf)
to GluonTS.

The naive seasonal model has been moved to m_competitions as well, since it shares so many similarities with the naive 2 model, and is an integral part of the m competitions. Both model only consist of a predictor.

For the Naïve 2 model I followed the R implementation that can be found [here](https://github.com/Mcompetitions/M4-methods/blob/master/Benchmarks%20and%20Evaluation.R#L118).
However, I optimized performance in a few places (no model impact).
I also added a potential TODO optimization with marginal model impact.

There is an additional requirement: statsmodels~=0.11


test/evaluation/test_evaluator.py
```diff
@@ -590,16 +590,16 @@ def test_metrics_multivariate(
 
 
 def test_evaluation_with_QuantileForecast():
-    start = "2012-01-01"
-    target = [2.4, 1.0, 3.0, 4.4, 5.5, 4.9] * 10
```
AaronSpieler: I changed this so the OWA metric calculation doesn't fail here, which requires some training samples.

src/gluonts/model/seasonal_naive/_predictor.py
```diff
@@ -70,13 +70,13 @@ def __init__(
...
-            target = np.asarray(data["target"], np.float32)
-            yield self._predict_time_series(start_time=start, target=target)
+    def predict_item(self, item: DataEntry) -> Forecast:
+        start_time = pd.Timestamp(item["start"], freq=self.freq)
+        target = np.asarray(item["target"], np.float32)
+        return self.predict_time_series(start_time, target)
 
-    def _predict_time_series(
+    # Needed for OWA metric
+    def predict_time_series(
```
AaronSpieler: The SeasonalNaivePredictor was neglected I think, because it didn't correctly implement the RepresentablePredictor functions.

src/gluonts/evaluation/_base.py
```diff
@@ -160,7 +160,7 @@ def __call__(
     @staticmethod
     def extract_pred_target(
         time_series: Union[pd.Series, pd.DataFrame], forecast: Forecast
-    ) -> Union[pd.Series, pd.DataFrame]:
+    ) -> np.ndarray:
```
AaronSpieler: Was also incorrect.

AaronSpieler:
This still has some issues. Might be due to the seasonal naive forecaster. Will look into it.

src/gluonts/evaluation/_base.py
```diff
@@ -387,6 +437,40 @@ def smape(target, forecast):
         )
         return smape
 
+    @staticmethod
+    def owa(target, forecast, train_target, seasonal_error, freq, start_date):
```
lostella: Why is the argument here called `train_target`?

AaronSpieler: Because the naive predictor calls it target, but its the train sequences.

lostella: There’s no training going on here. Is it the “past data” that precedes the target?

AaronSpieler: yes

lostella: ok so maybe you can change it to `past_target` I would say

AaronSpieler:
The `_base` class is still as I had it before, however, we might want to think about removing the `OWA` metric from the estimator, since it has a `10%` slowdown impact (potentially more if the train dataset contains long time-series) on the metric calculation speed (tested on `traffic` and `m4_hourly`), and instead introduce a separate static `OWA` calculation method that takes the `per-time-series` metric results of the desired model to be evaluated and the same from the Naïve 2 model, as suggested by @lostella .

src/gluonts/model/naive_2/_predictor.py
```diff
@@ -0,0 +1,155 @@
...
+# First-party imports
+from gluonts.core.component import validated
+from gluonts.dataset.common import Dataset, DataEntry
+from gluonts.model.forecast import SampleForecast, Forecast
+from gluonts.model.predictor import RepresentablePredictor
+from gluonts.evaluation import get_seasonality
+import statsmodels.api as sm
+
+
+def seasonality_test(past_ts_data, season_length):
```
lostella: Can you maybe add type annotations to these functions?

src/gluonts/model/naive_2/_predictor.py
```diff
@@ -0,0 +1,155 @@
...
+            * np.sqrt(np.cumsum(auto_correlations))
+        )
+        is_seasonal = (
+            abs(auto_correlations[season_length]) > limit[season_length]
+        )
+
+    return is_seasonal
+
+
+def naive_02(past_ts_data, forecast_start_time, forecast_length):
```
lostella: I think this function could get past data, period length (integer), and forecast length, and just return the array with the prediction. The predictor below will take care of mapping the frequency of the data to the period length (eg using the get_seasonality utility function), and wrap/unwrap inputs and outputs of this function. What do you think? I think it would make it a more independent utility.

So for example, in computing OWA you don’t have to take the median of the forecast, but you get the array directly.

Also, if you add type annotations it is clear whether we’re dealing with lists or arrays here.

AaronSpieler: I think even though it might be more independent if we just require `period_length` as an argument, it will reduce the convenience of the method, because we usually only have the frequency at hand. The user would need to know how to convert that to `period_length`. Therefore I think it might be best to give `freq` as an argument, and maybe an optional `period_length` that would take precedence if specified. What do you think?

With the rest I'm 100% with you.

lostella: > Therefore I think it might be best to give `freq` as an argument, and maybe an optional `period_length` that would take precedence if specified. What do you think?

Yes, that's a good compromise. I'd say both optional, at least one should be specified and not both. If `freq` is specified then some default mapping `freq -> period` is applied, otherwise the provided `period` is used.

src/gluonts/model/naive_2/_predictor.py
```diff
@@ -0,0 +1,153 @@
...
+        )  # i.e. no seasonality component
+
+    # calculate naive 1 forecast: (last value forecast_length times)
+    naive_1_forecast = (
+        np.ones(forecast_length) * seasonality_normed_context[-1]
+    )
+
+    forecast = np.mean(naive_1_forecast) * multiplicative_seasonal_component
+
+    return forecast
```
AaronSpieler: Do you think you could maybe take a look at this code and give feedback whether its true to your definition of the Naive 2 model? @Mcompetitions

src/gluonts/model/naive_2/_predictor.py
```diff
@@ -0,0 +1,153 @@
...
+    else:
+        # calculate auto-correlation for lags up to season_length
+        auto_correlations = sm.tsa.stattools.acf(
+            past_ts_data, fft=False, nlags=season_length
+        )
+        auto_correlations[1:] = 2 * auto_correlations[1:] ** 2
+        limit = (
+            critical_z_score
+            / np.sqrt(len(past_ts_data))
+            * np.sqrt(np.cumsum(auto_correlations))
```
lcallot: I think you are including `autocorrelations[0]` in the cumsum here, whereas the equivalent entry is excluded in the R code.

AaronSpieler: Actually they do it too in: `clim <- tcrit/sqrt(length(input)) * sqrt(cumsum(c(1, 2 * xacf^2)))` specifically here: `c(1, 2 * xacf^2)` where they prepend a `1` before they do the `cumsum`, if I understand the code correctly.
I just don't apply the times two squared transformation to the first element, so it will stay `1`, which is always the value of  `autocorrelations[0]`.

lcallot: You are correct, I overlooked the `c(1,...)`.

AaronSpieler:
If there are no further objections then from my side this code is ready.

src/gluonts/evaluation/_base.py
```diff
@@ -185,6 +192,41 @@ def extract_pred_target(
...
+        ), (
+            "Index of forecast is outside the index of target\n"
+            f"Index of forecast: {forecast.index}\n Index of target: {time_series.index}"
+        )
+
+        # cut the time series using the dates of the forecast object
+        return np.atleast_1d(
+            np.squeeze(
+                time_series.loc[
+                    time_series.index.difference(forecast.index)
```
lostella: According to pandas’ documentation, `difference` here computed returns 

> a new Index with elements from the index that are not in other

What if the forecast does not fall at the end of the time series, but in the middle of it? The index of the extracted data should just be whatever is before the forecast start date I’d say

AaronSpieler:
> One thing to add, if it’s not too hard, is a test that the output of Naive2 is approx the same as the original R code. You can for example use (or generate) one short time series, compute the prediction using R naive2 and copy-paste the prediction in the test to assert they are close

Ahhh, the pain. But yes, that is also a great idea, I'll do it.

src/gluonts/evaluation/_base.py
AaronSpieler: I agree, great catch!

src/gluonts/model/naive_2/_estimator.py
```diff
@@ -0,0 +1,36 @@
...
+
+# First-party imports
+from gluonts.core.component import validated
+from gluonts.model.estimator import DummyEstimator
+
+# Relative imports
+from ._predictor import Naive2Predictor
+
+
+class Naive2Estimator(DummyEstimator):
```
jaheba: I think this class isn't needed. We can use the predictor as forecaster directly.

src/gluonts/model/naive_2/_predictor.py
```diff
@@ -0,0 +1,166 @@
...
+        self.freq = freq
+        self.prediction_length = prediction_length
+        self.season_length = (
+            season_length
+            if season_length is not None
+            else get_seasonality(freq)
+        )
+
+    def predict_item(self, item: DataEntry) -> Forecast:
+        start_time = pd.Timestamp(item["start"], freq=self.freq)
```
jaheba: I think `item["start"]` already is of `pd.Timestamp`.

src/gluonts/model/naive_2/_predictor.py
```diff
@@ -0,0 +1,166 @@
...
+        past_ts_data = np.asarray(item["target"], np.float32)
```
jaheba: Same.

src/gluonts/model/naive_2/_estimator.py
AaronSpieler: > Can we split the OWA metric and the model in two PRs?
> 
> And, I would prefer a different folder for this. Maybe `model/m4`?

But why? It's just another model like seasonal naive. This is also the reason why there is this `Naive2Estimator` class analogous to `SeasonalNaiveEstimator`.

jaheba: I think we should remove the `SeasonalNaiveEstimator` as well then.

Models which don't train should not have `Estimator`s.

src/gluonts/model/naive_2/_predictor.py
AaronSpieler: yeah you're right

AaronSpieler:
The model seems to perfectly replicate the R implementation, the tests pass `np.allclose(r_naive_02_outputs, predictions)`!

AaronSpieler:
> * Rename module
> * Do two separate PRs

What do you mean by rename module?
Whats the purpose of the seperate PRs?

src/gluonts/model/naive_2/_estimator.py
AaronSpieler: Overall I agree with not having an `Estimator` for models that are not trained. However, This can be left for another PR where these two, and possibly other models? are refactored accordingly?

jaheba: No.

src/gluonts/model/naive_2/_predictor.py
```diff
@@ -0,0 +1,167 @@
...
+import numpy as np
+import pandas as pd
+
+# First-party imports
+from gluonts.core.component import validated
+from gluonts.dataset.common import Dataset, DataEntry
+from gluonts.model.forecast import SampleForecast, Forecast
+from gluonts.model.predictor import RepresentablePredictor
+from gluonts.evaluation import get_seasonality
+import statsmodels.api as sm
```
jaheba: This should go under third-party imports.

src/gluonts/model/naive_2/_predictor.py
```diff
@@ -0,0 +1,167 @@
...
+def seasonality_test(past_ts_data: np.array, season_length: int):
+    """
+    Test the time-series for seasonal patterns by performing a 90% auto-correlation test:
+
+    As described here: https://www.m4.unic.ac.cy/wp-content/uploads/2018/03/M4-Competitors-Guide.pdf
+    Code based on: https://github.com/Mcompetitions/M4-methods/blob/master/Benchmarks%20and%20Evaluation.R
+    """
+    critical_z_score = 1.645  # corresponds to 90% confidence interval
+    if len(past_ts_data) < 3 * season_length:
+        is_seasonal = False
```
jaheba: Why not just an early return?

src/gluonts/model/naive_2/_predictor.py
```diff
@@ -0,0 +1,167 @@
...
+
+    return is_seasonal
+
+
+def naive_02(
+    past_ts_data: np.array,
+    prediction_length: int,
+    freq: Optional[str] = None,
+    season_length: Optional[int] = None,
+):
```
jaheba: What's the return type?

src/gluonts/model/seasonal_naive/_predictor.py
```diff
@@ -70,16 +70,11 @@ def __init__(
...
-        for data in dataset:
-            start = pd.Timestamp(data["start"], freq=self.freq)
-            target = np.asarray(data["target"], np.float32)
-            yield self._predict_time_series(start_time=start, target=target)
-
-    def _predict_time_series(
-        self, start_time: pd.Timestamp, target: np.ndarray
-    ) -> SampleForecast:
+    def predict_item(self, item: DataEntry) -> Forecast:
+        target = np.asarray(item["target"], np.float32)
```
jaheba: Target is already np.array

src/gluonts/model/seasonal_naive/_predictor.py
```diff
@@ -70,16 +70,11 @@ def __init__(
...
         len_ts = len(target)
+        forecast_start_time = item["start"] + len_ts * item["start"].freq
```
jaheba: I think there is a function called forecast_start in `gluonts.util.pandas`, which you can use.

requirements/requirements.txt
```diff
@@ -10,3 +10,4 @@ pandas~=1.0
 pydantic~=1.1
 tqdm~=4.23
 ujson~=1.35
+statsmodels~=0.11
```
jaheba: This seems to be quite a big dependency. Can we move this to an extra?

src/gluonts/model/naive_2/_estimator.py
AaronSpieler: Having the `Estimator` class streamlines the testing of the models though.

src/gluonts/model/naive_2/_predictor.py
AaronSpieler: The R code was like this, so no particular reason, I can change it of course.

src/gluonts/model/seasonal_naive/_predictor.py
AaronSpieler: Yes, but in this case the explicit casting to `np.float32` is necessary, so that when we calculate the OWA we get exactly `1.0` when testing the `Naive2Predictor`.

src/gluonts/model/seasonal_naive/_predictor.py
AaronSpieler: it was in `gluonts.support.pandas`, thx

requirements/requirements.txt
lostella: yes, good idea

AaronSpieler: It doesn't seem to have any conflicts when I installed it, and it has pretty basic dependencies.

```
Python >= 3.5
NumPy >= 1.14 # we require the same
SciPy >= 1.0
Pandas >= 0.21 # we require 1.0
Patsy >= 0.5.1
```

Additionally, not having it in there would make using OWA without extra not possible.
So, overall, I dislike the idea.

src/gluonts/model/naive_2/_estimator.py
lostella: I think it's fine to have this class there since other models have the same, @jaheba we can open a new PR to remove these if we think they are redundant. case closed!

jaheba: Two wrongs don't make a right.

Is there any other reason keeping these, other than that existing models also do this?

requirements/requirements.txt
lostella: I agree it's better to have it in extra

src/gluonts/model/naive_2/_predictor.py
AaronSpieler: `np.ndarray`

src/gluonts/model/naive_2/_estimator.py
AaronSpieler: Ok, im working on removing them.

requirements/requirements.txt
AaronSpieler: I guess we could put it into `requirements-extras-m-competition.py` if necessary.

src/gluonts/model/naive_2/_predictor.py
AaronSpieler: Ok, so we get the exact same results, so I think we're good.

src/gluonts/model/naive_2/_predictor.py
```diff
@@ -0,0 +1,167 @@
...
+            * np.sqrt(np.cumsum(auto_correlations))
+        )
+        is_seasonal = (
+            abs(auto_correlations[season_length]) > limit[season_length]
+        )
+
+    return is_seasonal
+
+
+def naive_02(
```
lostella: A minor thing: the model subpackage is called `naive_2`, you can call this function too `naive_2`, as well as other occurrences in the tests.

lostella: Actually, not so minor

lostella: Also in file names in the tests: it's better to be consistent with `naive_2`, consistency is good

src/gluonts/model/naive_2/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+# Relative imports
+from ._estimator import Naive2Estimator
+from ._predictor import Naive2Predictor
+from ._predictor import naive_02
+
+__all__ = ["Naive2Estimator", "Naive2Predictor", "naive_02"]
```
lostella: If renaming `naive_02` into `naive_2` is a problem because of this, then call the function `predict_naive_2`, for example

test/model/m_competitions/test_predictors.py
```diff
@@ -0,0 +1,155 @@
...
+NUM_TS = 10
+
+
+@pytest.mark.parametrize(
+    "RepresentablePredictor", [SeasonalNaivePredictor, Naive2Predictor]
+)
+@pytest.mark.parametrize(
+    "freq", ["1min", "15min", "30min", "1H", "2H", "12H", "7D", "1W", "1M"]
+)
+def test_seasonal_naive(RepresentablePredictor, freq: str):
```
lostella: The argument name should not be camel-case, as that can be confused for a class name (and notice how GitHub indeed highlights it).

I suggest renaming `RepresentablePredictor` as `predictor_cls`

test/model/m_competitions/test_predictors.py
```diff
@@ -0,0 +1,155 @@
...
+        dict(prediction_length=CONSTANT_DATASET_PREDICTION_LENGTH),
+    )
+
+
+@flaky(max_runs=3, min_passes=1)
+@pytest.mark.parametrize(
+    "RepresentablePredictor, parameters, accuracy",
+    [seasonal_naive_predictor() + (0.0,), naive_2_predictor() + (0.0,)],
+)
+def test_accuracy(RepresentablePredictor, parameters, accuracy):
```
lostella: Here same as above

test/model/m_competitions/test_predictors.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+# SERIALIZATION/DESERIALIZATION TESTS:
+
+
+@pytest.mark.parametrize(
+    "RepresentablePredictor, parameters",
+    [seasonal_naive_predictor(), naive_2_predictor()],
+)
+def test_seriali_predictors(RepresentablePredictor, parameters):
```
lostella: Here as well

src/gluonts/evaluation/_base.py
```diff
@@ -196,28 +237,21 @@ def seasonal_error(
...
             forecast_freq = seasonality
         else:
             # edge case: the seasonal freq is larger than the length of ts
             # revert to freq=1
             # logging.info('The seasonal frequency is larger than the length of the time series. Reverting to freq=1.')
             forecast_freq = 1
-        y_t = np.ma.masked_invalid(ts.values[:-forecast_freq])
-        y_tm = np.ma.masked_invalid(ts.values[forecast_freq:])
+        y_t = past_data[:-forecast_freq]
+        y_tm = past_data[forecast_freq:]
```
lostella: Why removing the masking here?

AaronSpieler: Becaus `past_data` is already masked, so it is now not necessary to mask again. Previously this was not the case.

AaronSpieler:
> 
> 
> I see that the two naive methods share some similarity, however I’m not sure how the seasonal naive is connected to the M competitions. Since moving stuff around is a breaking change, it’s better to get it right: would `baseline` be a better name for the subpackage? What do you think @jaheba?

I think we could even merge the trivial forecasting methods, since they all share the commonality of not learning: `/statistical_models` or `/statistical_methods`?

AaronSpieler:
Most post agree on just setting: 

```
import os
os.environ['KMP_DUPLICATE_LIB_OK']='True'
```

However, that feels like a hack to me.

AaronSpieler:
> Most post agree on just setting:
> 
> ```
> import os
> os.environ['KMP_DUPLICATE_LIB_OK']='True'
> ```
> 
> However, that feels like a hack to me.

I think it might be necessary though, since we install `mxnet-cu92mkl>=1.3.1,<1.5.*` though pip, and the `mkl` there is causing issues, with the anaconda environment, now that numpy is istalled through the anaconda environment.

alexw91:
# [Codecov](https://codecov.io/gh/awslabs/gluon-ts/pull/602?src=pr&el=h1) Report
> Merging [#602](https://codecov.io/gh/awslabs/gluon-ts/pull/602?src=pr&el=desc) into [master](https://codecov.io/gh/awslabs/gluon-ts/commit/64505926f10cc8864349cb4121577fe3ae850261&el=desc) will **not change** coverage by `%`.
> The diff coverage is `n/a`.


[Impacted file tree graph](https://codecov.io/gh/awslabs/gluon-ts/pull/602?src=pr&el=tree)

```diff
@@           Coverage Diff           @@
##           master     #602   +/-   ##
=======================================
  Coverage   83.98%   83.98%           
=======================================
  Files         178      178           
  Lines       10390    10390           
=======================================
  Hits         8726     8726           
  Misses       1664     1664           
```

AaronSpieler:
YEEES!

alexw91:
# [Codecov](https://codecov.io/gh/awslabs/gluon-ts/pull/602?src=pr&el=h1) Report
> Merging [#602](https://codecov.io/gh/awslabs/gluon-ts/pull/602?src=pr&el=desc) into [master](https://codecov.io/gh/awslabs/gluon-ts/commit/25755e7487f44b8f6fbdb38e0eaee0bea198ded2&el=desc) will **not change** coverage by `%`.
> The diff coverage is `n/a`.


[Impacted file tree graph](https://codecov.io/gh/awslabs/gluon-ts/pull/602?src=pr&el=tree)

```diff
@@           Coverage Diff           @@
##           master     #602   +/-   ##
=======================================
  Coverage   83.98%   83.98%           
=======================================
  Files         178      178           
  Lines       10390    10390           
=======================================
  Hits         8726     8726           
  Misses       1664     1664           
```

AaronSpieler:
Once https://github.com/awslabs/gluon-ts/pull/661 is merged, is should be possible to merge master conflict free into this, and then all tests should pass.

alexw91:
# [Codecov](https://codecov.io/gh/awslabs/gluon-ts/pull/602?src=pr&el=h1) Report
> Merging [#602](https://codecov.io/gh/awslabs/gluon-ts/pull/602?src=pr&el=desc) into [master](https://codecov.io/gh/awslabs/gluon-ts/commit/3aa9c2a78494a9c48af9a17d48e6f568af7308ad&el=desc) will **not change** coverage by `%`.
> The diff coverage is `n/a`.


[Impacted file tree graph](https://codecov.io/gh/awslabs/gluon-ts/pull/602?src=pr&el=tree)

```diff
@@           Coverage Diff           @@
##           master     #602   +/-   ##
=======================================
  Coverage   83.98%   83.98%           
=======================================
  Files         178      178           
  Lines       10390    10390           
=======================================
  Hits         8726     8726           
  Misses       1664     1664           
```

Issue #608, opened 2020-02-08
infinite weighted quantile loss


jaheba:
## Description
When the true-target values are zeros, our weighted quantile loss become `inf`. 

This is especially a problem for HPO jobs, where we rely on the `mean_wQuantileLoss` as our objective metric to access the models performance.

```
[2020-02-08 12:04:19 +0100] [59599] [INFO] #test_score (local, wQuantileLoss[0.1]): inf
[2020-02-08 12:04:19 +0100] [59599] [INFO] #test_score (local, wQuantileLoss[0.2]): inf
[2020-02-08 12:04:19 +0100] [59599] [INFO] #test_score (local, wQuantileLoss[0.3]): inf
[2020-02-08 12:04:19 +0100] [59599] [INFO] #test_score (local, wQuantileLoss[0.4]): inf
[2020-02-08 12:04:19 +0100] [59599] [INFO] #test_score (local, wQuantileLoss[0.5]): inf
[2020-02-08 12:04:19 +0100] [59599] [INFO] #test_score (local, wQuantileLoss[0.6]): inf
[2020-02-08 12:04:19 +0100] [59599] [INFO] #test_score (local, wQuantileLoss[0.7]): inf
[2020-02-08 12:04:19 +0100] [59599] [INFO] #test_score (local, wQuantileLoss[0.8]): inf
[2020-02-08 12:04:19 +0100] [59599] [INFO] #test_score (local, wQuantileLoss[0.9]): inf
[2020-02-08 12:04:19 +0100] [59599] [INFO] #test_score (local, mean_wQuantileLoss): inf
```

## To Reproduce

The test time-series needs to contain zeros in the prediction-range. The following example triggers the problem until a prediction_length of 4.

```json
{"start": "2020", "target": [2, 1, 0, 0, 0, 0]}
```

jaheba:
Here is the code triggering it:

https://github.com/awslabs/gluon-ts/blob/90587baff19a764aed02ffab752f70cf927f50e5/src/gluonts/evaluation/_base.py#L321-L323

jaheba:
Looking at tensorflow, they seem to assume a minimum value by which they [divide](https://github.com/tensorflow/tensorflow/blob/ca123bc4199a42c65dcc817b6c771fa613171e08/tensorflow/python/keras/losses.py#L796):


```python
diff = math_ops.abs(
      (y_true - y_pred) / K.clip(math_ops.abs(y_true), K.epsilon(), None))
```

[...]

PR #609, opened 2020-02-08
Fix: #608. Evaluation of zero-valued true target.

jaheba:
*Issue #, if available:*

*Description of changes:*


jaheba:
This is the intended behaviour.

If one needs to handle zero called, then other medics can be used.

PR #619, opened 2020-02-11
Refactored DataLoader.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/dataset/loader.py
+++ b/src/gluonts/dataset/loader.py
@@ -100,7 +100,7 @@ class DataLoader(Iterable[DataEntry]):
     @property
     def batches(self):
         return BatchStacker(
-            self.batch_size, stream=self.stream, stack_fn=self.stack,
+            self.batch_size, stream=self.stream, stack_fn=self.stack
         )
 
     def stack(self, xs):
--- a/src/gluonts/dataset/loader.py
+++ b/src/gluonts/dataset/loader.py
@@ -13,8 +13,10 @@
 
 # Standard library imports
 import itertools
+import random
 from collections import defaultdict
-from typing import Any, Dict, Iterable, Iterator, List, Optional  # noqa: F401
+from functools import partial
+from typing import Any, Collection, Dict, Iterable, Iterator, List, Optional
 
 # Third-party imports
 import mxnet as mx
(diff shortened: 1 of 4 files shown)


alexw91:
# [Codecov](https://codecov.io/gh/awslabs/gluon-ts/pull/619?src=pr&el=h1) Report
> Merging [#619](https://codecov.io/gh/awslabs/gluon-ts/pull/619?src=pr&el=desc) into [master](https://codecov.io/gh/awslabs/gluon-ts/commit/90587baff19a764aed02ffab752f70cf927f50e5&el=desc) will **increase** coverage by `0.01%`.
> The diff coverage is `n/a`.


[Impacted file tree graph](https://codecov.io/gh/awslabs/gluon-ts/pull/619?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #619      +/-   ##
==========================================
+ Coverage   83.72%   83.73%   +0.01%     
==========================================
  Files         181      181              
  Lines       10326    10287      -39     
==========================================
- Hits         8645     8614      -31     
+ Misses       1681     1673       -8     
```

src/gluonts/transform/_base.py
```diff
@@ -76,7 +76,7 @@ def estimate(self, data_it: Iterator[DataEntry]) -> Iterator[DataEntry]:
 
 class Identity(Transformation):
     def __call__(
-        self, data_it: Iterator[DataEntry], is_train: bool
+        self, data_it: Iterable[DataEntry], is_train: bool
     ) -> Iterator[DataEntry]:
```
lostella: Looks like the return type should agree with the argument type?

src/gluonts/dataset/loader.py
```diff
@@ -28,55 +30,37 @@
...
-        for key in self._buffers.keys():
-            li = self._buffers[key]
-            self._buffers[key] = [li[i] for i in perm]
+class Shuffler(Iterable[DataBatch]):
+    def __init__(self, batch_size, stream):
+        self.batch_size = batch_size
+        self.stream = stream
+
+    def __iter__(self) -> Iterator[DataBatch]:
+        for batch in batcher(self.stream, self.batch_size):
```
lostella: I quite like this class, but I was confused at first by the fact that the “batch” here is really a “batch of batches”, because of how the class is used later on in the TrainDataLoader (that is, stream here is a sequence of batches).

Could this class be a more generic Iterable, instead of just an Iterable of DataBatches?

src/gluonts/dataset/loader.py
```diff
@@ -28,55 +30,37 @@
...
-        else:
-            return xs  # stack all other types as list
 
-    def shuffle(self):
-        perm = np.random.permutation(self._size)
-        for key in self._buffers.keys():
-            li = self._buffers[key]
-            self._buffers[key] = [li[i] for i in perm]
+class Shuffler(Iterable[DataBatch]):
+    def __init__(self, batch_size, stream):
```
lostella: Type hinting could help read this a bit better maybe?

src/gluonts/dataset/loader.py
```diff
@@ -148,121 +160,64 @@ def __init__(
...
-                ):
-                    yield batch
-                    batch_count += 1
-                    if batch_count >= self.num_batches_per_epoch:
-                        return
+        return itertools.islice(self.batches, self.num_batches_per_epoch)
 
 
 class ValidationDataLoader(DataLoader):
-    """
```
lostella: Why not keeping the docstring? We can improve it, in case we believe it’s not good

src/gluonts/dataset/loader.py
```diff
@@ -28,55 +30,37 @@
...
-        self._buffers: Dict[Any, List[Any]] = defaultdict(list)
+def batcher(iterable, batch_size):
+    while True:
+        items = list(itertools.islice(iterable, batch_size))
+        if not items:
+            break
+        yield items
+
+
+class BatchStacker:
```
lostella: These classes are nice, but could use a docstring to briefly summarize what they do: “An Iterable type, that given a sequence of X yields a sequence of Y”

src/gluonts/dataset/loader.py
jaheba: I want to get rid of the class in the future. The only difference is whether `is_train` is set, and I think that should be handled outside.

src/gluonts/dataset/loader.py
jaheba: I'm also not too sure about the names. Let me improve these a bit.

src/gluonts/dataset/loader.py
```diff
@@ -101,16 +57,43 @@ class DataLoader(Iterable[DataEntry]):
     def __init__(
         self,
         dataset: Dataset,
+        *,
```
lostella: What is this?

jaheba: Enforces all parameters afterwards to be keyword-only.

src/gluonts/dataset/util.py
```diff
@@ -83,3 +85,54 @@ def to_pandas(instance: dict, freq: str = None) -> pd.Series:
...
+    [0, 1, 2, 3, 4]
+    """
+    return itertools.islice(iterable, n)
+
+
+def batcher(iterable: Iterable[T], batch_size: int) -> Iterator[List[T]]:
+    """Groups elements from `iterable` into batches of size `batch_size`.
+
+    >>> list(batcher("ABCDEFG", 3))
+    [['A', 'B', 'C'], ['D', 'E', 'F'], ['G']]
```
lostella: Apparently this docstring is breaking the docs build, we better fix this

alexw91:
# [Codecov](https://codecov.io/gh/awslabs/gluon-ts/pull/619?src=pr&el=h1) Report
> Merging [#619](https://codecov.io/gh/awslabs/gluon-ts/pull/619?src=pr&el=desc) into [master](https://codecov.io/gh/awslabs/gluon-ts/commit/bc24d1cef69421275efe2e6cb83e2f15be5280f8&el=desc) will **increase** coverage by `0.00%`.
> The diff coverage is `n/a`.


[Impacted file tree graph](https://codecov.io/gh/awslabs/gluon-ts/pull/619?src=pr&el=tree)

```diff
@@           Coverage Diff           @@
##           master     #619   +/-   ##
=======================================
  Coverage   83.73%   83.74%           
=======================================
  Files         181      181           
  Lines       10327    10284   -43     
=======================================
- Hits         8647     8612   -35     
+ Misses       1680     1672    -8     
```

PR #624, opened 2020-02-12
fix mean_ts method

### LINKED ISSUE #623: mean_ts() has bug in class DistributionForecast & SampleForecast in forecast.py
## Description
mean_ts()
return pd.Series(self.index, self.mean)

Should be:
mean_ts()
return pd.Series(data=self.mean, index=self.index)

lostella:
*Issue #, if available:* fixes #623 

### DIFF
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -390,7 +390,7 @@ class SampleForecast(Forecast):
         """
         Forecast mean, as a pandas.Series object.
         """
-        return pd.Series(index=self.index, data=self.mean)
+        return pd.Series(data=self.mean, index=self.index)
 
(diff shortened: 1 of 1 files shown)


src/gluonts/model/forecast.py
```diff
@@ -390,7 +390,7 @@ def mean_ts(self):
         """
         Forecast mean, as a pandas.Series object.
         """
-        return pd.Series(self.index, self.mean)
+        return pd.Series(index=self.index, data=self.mean)
```
jaheba:
```suggestion
        return pd.Series(self.mean, index=self.index)
```

lostella: addressed in 0.3 seconds

PR #644, opened 2020-02-16
fix forecast index bug, add test

### LINKED ISSUE #642: Index of forecast is wrong in multivariate Time Series
## Description
When forecasting multivariate Time Series the index has the length of the target dimension instead of the prediction length

## To Reproduce

```python
import numpy as np

from gluonts.distribution import MultivariateGaussianOutput


train_dataset = ListDataset(
    data_iter=[
        {
            "start": "2019-01-01 00:00:00",
            "target": np.ones(shape=(4, 4)),
        },
    ],
    freq="W",
    one_dim_target=False,
)

test_dataset = ListDataset(
    data_iter=[
        {
            "start": "2019-01-01 00:00:00",
            "target": np.ones(shape=(4, 5)),
        },
    ],
    freq="W",
    one_dim_target=False,
)

estimator = DeepAREstimator(
    'W', prediction_length=1, trainer=Trainer(epochs=3, hybridize=False),
    distr_output=MultivariateGaussianOutput(dim=4),
)

predictor = estimator.train(train_dataset)

forecast_it, ts_it = make_evaluation_predictions(dataset=test_dataset, predictor=predictor, num_samples=10)

forecast_list = list(forecast_it)
ts_list = list(ts_it)
```

## Error Message

DatetimeIndex(['2019-02-03', '2019-02-10', '2019-02-17', '2019-02-24'], dtype='datetime64[ns]', freq='W-SUN')

While it should only be ['2019-02-03']

## Environment
- Operating system: Amazon Linux
- Python version: 3.6
- GluonTS version: a96a0cc4 internal

lostella:
*Issue #, if available:* fixes #642 

### DIFF
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -373,7 +373,7 @@ class SampleForecast(Forecast):
         """
         Time length of the forecast.
         """
-        return self.samples.shape[-1]
+        return self.samples.shape[1]
 
     @property
     def mean(self) -> np.ndarray:
--- a/test/model/test_forecast.py
+++ b/test/model/test_forecast.py
@@ -94,3 +94,38 @@ def test_DistributionForecast():
     assert forecast.prediction_length == pred_length
     assert len(forecast.index) == pred_length
     assert forecast.index[0] == pd.Timestamp(START_DATE)
+
+
+@pytest.mark.parametrize(
+    "forecast, exp_index",
+    [
+        (
+            SampleForecast(
+                samples=np.random.normal(size=(100, 7, 3)),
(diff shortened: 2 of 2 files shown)


src/gluonts/model/forecast.py
```diff
@@ -373,7 +373,7 @@ def prediction_length(self):
         """
         Time length of the forecast.
         """
-        return self.samples.shape[-1]
+        return self.samples.shape[1]
```
mbohlkeschneider: Thank you!

One alternative to consider is to take the `prediction_length` from the `predictor` that construct the forecast. This way, we would be correct (at least in prediction length) even if we change the data layout.

Issue #162, opened 2019-07-01
multi gpu support


cyrusmvahid:
## Description
currently estimator functions accept a Context object. Context can support only one gpu. What we need is to be able to process a list of Context object. the same way that we run multi gpu on one machine. and then in train we need to use ```gluon.util.split_and_load``` and context is a list of context usually can be coded as ```ctx=[gpu(i] for i in mx.context.num_gpu()]```

PR #657, opened 2020-02-21
Multi-gpu support in training loop

ehsanmok:
*Issue #162 

*Description of changes:*

Adding the multi-gpu support. Besides local test, **all** examples are replicable and been tested on p3.8xlarge with 4 gpus.

### DIFF
--- a/src/gluonts/core/component.py
+++ b/src/gluonts/core/component.py
@@ -36,7 +36,7 @@ from . import fqname_for
 
 logger = logging.getLogger(__name__)
 
-ContextType = Optional[Union[mx.Context, List[mx.Context]]]
+ContextType = Optional[Union[str, mx.Context, List[mx.Context]]]
 
 A = TypeVar("A")
 
@@ -505,19 +505,20 @@ def get_mxnet_context(gpu_number=0) -> mx.Context:
         return mx.context.cpu()
 
 
-def normalize_ctx(
-    ctx: Optional[Union[str, mx.Context, List[mx.Context]]]
-) -> List[mx.Context]:
+def normalize_ctx(ctx: ContextType) -> List[mx.Context]:
+    if ctx is None:
+        if mx.context.num_gpus():
+            return [mx.gpu(i) for i in range(mx.context.num_gpus())]
+        else:
+            return [mx.cpu()]
+
     if isinstance(ctx, str):
         return [mx.Context(ctx)]
-    if ctx is not None:
-        return [ctx] if not isinstance(ctx, list) else ctx
-    else:
-        return (
-            [mx.gpu(i) for i in range(mx.context.num_gpus())]
-            if mx.context.num_gpus()
-            else [mx.cpu()]
-        )
+
+    if isinstance(ctx, list):
+        return ctx
+
+    return [ctx]
 
 
 def check_gpu_support() -> bool:
--- a/src/gluonts/core/component.py
+++ b/src/gluonts/core/component.py
@@ -20,7 +20,7 @@ import re
 from collections import OrderedDict
 from functools import singledispatch
 from pydoc import locate
-from typing import Any, Type, TypeVar, Union
+from typing import Any, Type, TypeVar, Union, List, Optional
 
 # Third-party imports
 import mxnet as mx
@@ -35,6 +35,8 @@ from gluonts.core.serde import dump_code
 from . import fqname_for
 
 logger = logging.getLogger(__name__)
+
+ContextType = Optional[Union[mx.Context, List[mx.Context]]]
 
 A = TypeVar("A")
 
@@ -501,6 +503,21 @@ def get_mxnet_context(gpu_number=0) -> mx.Context:
     else:
         logger.info("Using CPU")
         return mx.context.cpu()
+
+
+def normalize_ctx(
+    ctx: Optional[Union[str, mx.Context, List[mx.Context]]]
+) -> List[mx.Context]:
+    if isinstance(ctx, str):
+        return [mx.Context(ctx)]
+    if ctx is not None:
+        return [ctx] if not isinstance(ctx, list) else ctx
+    else:
+        return (
+            [mx.gpu(i) for i in range(mx.context.num_gpus())]
+            if mx.context.num_gpus()
+            else [mx.cpu()]
+        )
 
 
 def check_gpu_support() -> bool:
--- a/src/gluonts/dataset/loader.py
+++ b/src/gluonts/dataset/loader.py
@@ -22,7 +22,7 @@ import mxnet as mx
 import numpy as np
 
 # First-party imports
-from gluonts.core.component import DType
+from gluonts.core.component import DType, ContextType, normalize_ctx
 from gluonts.dataset.common import DataEntry, Dataset, DataBatch
 from gluonts.dataset.parallelized_loader import ParallelDataLoader
 from gluonts.transform import Transformation
@@ -69,14 +69,14 @@ class DataLoader(Iterable[DataEntry]):
         cyclic: bool,
         is_train: bool,
         batch_size: int,
-        ctx: mx.Context,
+        ctx: ContextType,
         dtype: DType = np.float32,
         num_workers: Optional[int] = None,
         num_prefetch: Optional[int] = None,
         **kwargs,
     ) -> None:
         self.batch_size = batch_size
-        self.ctx = ctx
+        self.ctx = normalize_ctx(ctx)
         self.dtype = dtype
         self.is_train = is_train
         self.transform = transform
@@ -96,7 +96,7 @@ class DataLoader(Iterable[DataEntry]):
             cyclic=self.cyclic,
             is_train=self.is_train,
             batch_size=self.batch_size,
-            ctx=self.ctx,
+            ctx=self.ctx[0],
             dtype=self.dtype,
             num_workers=self.num_workers,
             num_prefetch=self.num_prefetch,
@@ -147,7 +147,7 @@ class TrainDataLoader(DataLoader):
         dataset: Dataset,
         transform: Transformation,
         batch_size: int,
-        ctx: mx.Context,
+        ctx: ContextType,
         num_batches_per_epoch: int,
         num_workers: Optional[int] = None,
         num_prefetch: Optional[int] = None,
@@ -193,7 +193,7 @@ class ValidationDataLoader(DataLoader):
         *,
         transform: Transformation,
         batch_size: int,
-        ctx: mx.Context,
+        ctx: ContextType,
         num_workers: Optional[int] = None,
         num_prefetch: Optional[int] = None,
         dtype: DType = np.float32,
@@ -220,7 +220,7 @@ class InferenceDataLoader(DataLoader):
         *,
         transform: Transformation,
         batch_size: int,
-        ctx: mx.Context,
+        ctx: ContextType,
         num_workers: Optional[int] = None,
         num_prefetch: Optional[int] = None,
         dtype: DType = np.float32,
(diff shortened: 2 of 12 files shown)


test/model/test_models_synthetic.py
```diff
@@ -32,13 +33,18 @@
 batch_size = 32
 context_length = 2
 epochs = 1
+ctx = (
```
jaheba: Couldn't we use `normalize_ctx` here?

src/gluonts/core/component.py
```diff
@@ -506,6 +508,21 @@ def get_mxnet_context(gpu_number=0) -> mx.Context:
...
+) -> List[mx.Context]:
+    if isinstance(ctx, str):
+        return [mx.Context(ctx)]
+    if ctx is not None:
+        return [ctx] if not isinstance(ctx, list) else ctx
+    else:
+        return (
+            [mx.gpu(i) for i in range(mx.context.num_gpus())]
+            if mx.context.num_gpus()
+            else [mx.cpu()]
```
jaheba: I think I prefer this:

```python
def normalize_ctx(
    ctx: Optional[Union[str, mx.Context, List[mx.Context]]]
) -> List[mx.Context]:
    if ctx is None:
        if mx.context.num_gpus():
            return [mx.gpu(i) for i in range(mx.context.num_gpus())]
        else:
            return [mx.cpu()]

    if isinstance(ctx, str):
        return [mx.Context(ctx)]

    if isinstance(ctx, list):
        return ctx

    return [ctx]
```

src/gluonts/core/component.py
```diff
@@ -506,6 +508,21 @@ def get_mxnet_context(gpu_number=0) -> mx.Context:
         return mx.context.cpu()
 
 
+def normalize_ctx(
+    ctx: Optional[Union[str, mx.Context, List[mx.Context]]]
```
jaheba:
```suggestion
    ctx: Union[str, ContextType]
```

src/gluonts/dataset/loader.py
```diff
@@ -14,19 +14,14 @@
 # Standard library imports
 import functools
 import itertools
-from typing import (
-    Any,
-    Dict,
-    Iterable,
-    Iterator,
-)
+from typing import Any, Dict, Iterable, Iterator, List, Union
```
jaheba: Are you using `List` and `Union`?

src/gluonts/dataset/loader.py
```diff
@@ -79,7 +74,11 @@ def stack(self, xs):
             data = np.asarray(xs)
             if data.dtype.kind == "f":
                 data = data.astype(self.dtype)
-            return mx.nd.array(data, dtype=data.dtype, ctx=self.ctx)
+            return mx.nd.array(
+                data,
+                dtype=data.dtype,
+                ctx=self.ctx[0] if isinstance(self.ctx, list) else self.ctx,
```
jaheba: Shouldn't this just be `ctx[0]`?

test/model/test_models_synthetic.py
ehsanmok: Can make it to `ctx=normalize_ctx(None)` instead but it's less clear I guess, so being explicit here.

src/gluonts/core/component.py
ehsanmok: No sure the difference! would it make it more readable?

jaheba: Yes, handling `None` first, then the other cases and all in one level.

src/gluonts/core/component.py
ehsanmok: The idea is to keep the `Optional` and `str` in `ContextType` to make it compatible with mxnet.

jaheba: Since `ContextType` already includes `Optional`, we can shorten this.

src/gluonts/dataset/loader.py
ehsanmok: Thanks! removing them. Was before the `ContextType` refactoring.

ehsanmok:
@jaheba thanks for the review! all the comments have been addressed. I avoid making incompatible names changes to mxnet so keeping the `ctx` name.

alexw91:
# [Codecov](https://codecov.io/gh/awslabs/gluon-ts/pull/657?src=pr&el=h1) Report
> Merging [#657](https://codecov.io/gh/awslabs/gluon-ts/pull/657?src=pr&el=desc) into [master](https://codecov.io/gh/awslabs/gluon-ts/commit/f6c12a4a3902a6f2137a0f6b98001c64b9cdb9a7&el=desc) will **increase** coverage by `0.00%`.
> The diff coverage is `n/a`.


[Impacted file tree graph](https://codecov.io/gh/awslabs/gluon-ts/pull/657?src=pr&el=tree)

```diff
@@           Coverage Diff           @@
##           master     #657   +/-   ##
=======================================
  Coverage   84.63%   84.64%           
=======================================
  Files         178      178           
  Lines       10401    10425   +24     
=======================================
+ Hits         8803     8824   +21     
- Misses       1598     1601    +3     
```

| [Impacted Files](https://codecov.io/gh/awslabs/gluon-ts/pull/657?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [src/gluonts/core/component.py](https://codecov.io/gh/awslabs/gluon-ts/pull/657/diff?src=pr&el=tree#diff-c3JjL2dsdW9udHMvY29yZS9jb21wb25lbnQucHk=) | `86.22% <0.00%> (-0.96%)` | :arrow_down: |

alexw91:
# [Codecov](https://codecov.io/gh/awslabs/gluon-ts/pull/657?src=pr&el=h1) Report
> Merging [#657](https://codecov.io/gh/awslabs/gluon-ts/pull/657?src=pr&el=desc) into [master](https://codecov.io/gh/awslabs/gluon-ts/commit/f6c12a4a3902a6f2137a0f6b98001c64b9cdb9a7&el=desc) will **increase** coverage by `0.00%`.
> The diff coverage is `n/a`.


[Impacted file tree graph](https://codecov.io/gh/awslabs/gluon-ts/pull/657?src=pr&el=tree)

```diff
@@           Coverage Diff           @@
##           master     #657   +/-   ##
=======================================
  Coverage   84.63%   84.64%           
=======================================
  Files         178      178           
  Lines       10401    10425   +24     
=======================================
+ Hits         8803     8824   +21     
- Misses       1598     1601    +3     
```

| [Impacted Files](https://codecov.io/gh/awslabs/gluon-ts/pull/657?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [src/gluonts/core/component.py](https://codecov.io/gh/awslabs/gluon-ts/pull/657/diff?src=pr&el=tree#diff-c3JjL2dsdW9udHMvY29yZS9jb21wb25lbnQucHk=) | `86.22% <0.00%> (-0.96%)` | :arrow_down: |

istorch:
@ehsanmok @jaheba Sorry to just randomly comment here (this is my first comment on this project), but when can we expect to see this released? I am a new user of GluonTS and it would be nice to have multi GPU support (my specific application uses DeepAR). Thanks!

lostella:
@istorch the changes look good to me, so this could be merged into master soon and be included in the next release (either this or next week)

src/gluonts/dataset/loader.py
```diff
@@ -96,7 +96,7 @@ def __init__(
             cyclic=self.cyclic,
             is_train=self.is_train,
             batch_size=self.batch_size,
-            ctx=self.ctx,
+            ctx=self.ctx[0],
```
lostella, line 99: I'm not sure I understand this: one can specify a list of contexts, but then the data is loaded in the first one? But I'm really ignorant about the argument, so maybe I'm missing something.

src/gluonts/model/deepar/_network.py
```diff
@@ -266,19 +266,27 @@ def unroll_encoder(
...
+                else inputs[0].context
+            )
+            with ctx:
+                begin_state = self.rnn.begin_state(
+                    func=F.zeros, dtype=self.dtype, batch_size=inputs.shape[0]
+                )
+        else:
+            begin_state = self.rnn.begin_state(
+                func=F.zeros, dtype=self.dtype, batch_size=0
+            )
```
lostella, line 282: It looks to me like this context logic should not be in the network code. Maybe it would be better to control the context once and for all from the outside somehow? Otherwise I can see how the code of every model could be filled with similar context logic, and lose readability. What do you think?

istorch:
@ehsanmok @lostella Sorry to bug you guys again, but has there been any progress on this PR?

src/gluonts/support/util.py
```diff
@@ -158,9 +160,16 @@ def copy_parameters(
...
+        for p in params:
+            ctx = params[p].list_ctx()
+            break
+        # force init otherwise may not load params
+        # since not all params have the same ctx in net_source
+        net_dest.initialize(ctx=ctx, force_reinit=True)
         net_dest.load_parameters(
             model_dir_path,
-            ctx=mx.current_context(),
+            ctx=ctx,
```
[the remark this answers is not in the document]
ehsanmok: If I remember, during testing, when not all params are in a single context, forcing to reinit with a list of contexts solved the issue. Please see my general comment at the end.

src/gluonts/model/deepar/_network.py
ehsanmok: Agree and in fact this has been repeated. Though I was expecting to have this in upstream mxnet to not hack this into.

ehsanmok:
@lostella Thanks for your comments! in general, I think, since gluonts hasn't taken multi-gpu into considerations and has been designed for single gpu (it felt like that at least at the time of creating this PR), this PR initiations a sub-optimal solution at least with rooms to improvements of course. 

As for the MVE, I've run all the existing examples and observed their improved sub-optimal speed ups which I can report later here.

ehsanmok:
Hi @istorch, thanks for your patience! As @lostella pointed out earlier, this PR offers a solution and it'd be helpful to have some external validations too, so if you can, please run your application to test things us further. You can install this patch via 

```
python -m pip install git+git://github.com/ehsanmok/gluon-ts@multi-gpu
```

src/gluonts/dataset/loader.py
ehsanmok: > is the idea to put everything in the first context, and then have split_and_load in the training loop take care of that?

Yes, indeed!

istorch:
@ehsanmok Thanks for picking this up again! I have installed your branch and passed the context in as a list, like `ctx = [mx.gpu(i) for i in range(8)]`. Watching nvidia-smi, I see that all 8 GPUs are utilized during training. However, when I go to generate predictions, it still only uses one GPU. Is there anything special I need to do to make it use all GPUs for prediction as well?

ehsanmok:
@istorch thanks a lot for the report! I'll look into the prediction part.

istorch:
@ehsanmok also, I did a little benchmarking and I'm not seeing any improvement in using multiple GPUs. With the same settings, my training ran in 7min 40s on a single GPU, and 8min 51s on 8 GPUs. I set epochs to 50, batch_size to 512, and num_batches_per_epoch to 16 (this is on a p2.8xlarge instance).

ehsanmok:
@istorch Just curious, did you multiply the batch size by the number of GPUs, like if you've used 512 for single GPU, then you'd need to use 512 * 8 for your 8 GPU context? Also it'd be helpful if you could share more context about what your code is doing and what model are you training.

ehsanmok:
@lostella @istorch after an investigation it turns out the adding the complete multi-gpu feature is harder than what I initially thought. This PR add the multi-gpu for training loop only. The current data loader seems to be a major bottleneck as it wasn't written for single context only and the other parts of GluonTS estimator is hard to decouple and extend. This is one more reason to consider #829.

PR #665, opened 2020-02-24
Rework LRS.

jaheba:
*Issue #, if available:*

*Description of changes:*


src/gluonts/ty.py
```diff
@@ -0,0 +1,38 @@
...
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+__all__ = [
+    "PositiveInt",
+    "PositiveFloat",
+    "NonNegativeInt",
+    "NonNegativeFloat",
```
lostella: Interval01 missing? Why do you need to explicitly list these here?

jaheba: Thanks!

>  Why do you need to explicitly list these here?

So that there is no complaint from flake8 about line 22.

src/gluonts/trainer/_base.py
```diff
@@ -147,19 +151,18 @@ def __init__(
         self.init = init
         self.hybridize = hybridize
         self.ctx = ctx if ctx is not None else get_mxnet_context()
-        self.halt = False
```
lostella: It’s fine to remove this, since it’s only used for interactively breaking the loops; however, we probably want to have some other early stopping mechanism (maybe through the learning rate scheduler, as soon as the minimum lr is hit or exceeded?) so something will need to be there

src/gluonts/trainer/_base.py
```diff
@@ -194,21 +196,15 @@ def base_path() -> str:
...
-                    patience=self.patience,
-                    decay_factor=self.learning_rate_decay_factor,
-                    min_lr=self.minimum_learning_rate,
-                )
+                self.lrs.reset()
 
                 optimizer = mx.optimizer.Adam(
-                    learning_rate=self.learning_rate,
-                    lr_scheduler=lr_scheduler,
+                    lr_scheduler=self.lrs,
```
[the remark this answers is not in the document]
jaheba: Looking at the code, the idea is that you just pass the lr_scheduler, and the optimizer then refers to the lrs for the learning rate:

```python
        self.lr = learning_rate
        if self.lr_scheduler is not None and learning_rate is not None:
            if self.lr_scheduler.base_lr != learning_rate:
                print(UserWarning("learning rate from ``lr_scheduler`` has been "
                                  "overwritten by ``learning_rate`` in optimizer."))
                self.lr_scheduler.base_lr = learning_rate

```

```
    @property
    def learning_rate(self):
        if self.lr_scheduler is not None:
            return self.lr_scheduler(self.num_update)
        else:
            return self.lr
```

If you don't pass a lrs, it does something itself.

PR #668, opened 2020-02-25
Mqcnn rts

lovvge:
*Issue #, if available:*

*Description of changes:*

## Todo

- [x] add support for categorial features

- [x]  option to ignore dynamical-features (like `ignore` in SageMaker DeepAR)

- [x] fix instance splitter

- [x] add observed value indicator (see #704)


src/gluonts/block/encoder.py
```diff
@@ -224,13 +224,15 @@ def hybrid_forward(
...
+            inputs = self._assemble_inputs(
                 F,
                 target=target,
                 static_features=static_features,
                 dynamic_features=dynamic_features,
             )
+        elif self.use_dynamic_feat:
+            inputs = F.concat(target, dynamic_features, dim=2)  # (N, T, C)
         else:
             inputs = target
```
AaronSpieler, line 237: Isn't this already handling the possibility of disabling the use of dynamic features?

AaronSpieler:
> _Issue #, if available:_
> 
> _Description of changes:_
> 
> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.
> 
> ## Todo
> * [ ]  add support for categorial features
> * [ ]   option to ignore dynamical-features (like `ignore` in SageMaker DeepAR)
> * [ ]  fix instance splitter (see #704)

How is #704 related to instance splitter?

src/gluonts/model/deepar/_estimator.py
```diff
@@ -197,6 +198,16 @@ def __init__(
...
+    @classmethod
+    def derive_auto_fields(cls, train_iter):
+        stats = calculate_dataset_statistics(train_iter)
+
+        return {
+            "use_feat_dynamic_real": stats.num_feat_dynamic_real > 0,
+            "use_feat_static_cat": bool(stats.feat_static_cat),
+            "cardinality": [len(cats) for cats in stats.feat_static_cat],
+        }
+
```
AaronSpieler, line 210: Isn't this adding support for categorical features?

src/gluonts/model/seq2seq/_forking_estimator.py
```diff
@@ -108,47 +115,107 @@ def __init__(
...
+                    output_field=FieldName.FEAT_DYNAMIC_REAL,
+                    input_fields=dynamic_feat_fields,
+                )
+            )
+        elif len(dynamic_feat_fields) == 1:
+            chain.append(
+                RenameFields(
+                    {dynamic_feat_fields[0]: FieldName.FEAT_DYNAMIC_REAL}
+                )  # TODO: find out why this is done
+            )
```
AaronSpieler, line 175: What is the purpose of this.?

jaheba: `FEAT_DYNAMIC_REAL` can consist of different features. If there is only one (for example `feat_age`, we rename the input field to be `feat_dynamic_real`.

AaronSpieler: I see, thx!

AaronSpieler:
The option to ignore dynamical-features will need some refactoring in the `_forking_network.py`.

AaronSpieler:
I have manually merged https://github.com/awslabs/gluon-ts/pull/704 into this pull request, since the differences between the versions are so stark that no auto merge could have worked.

src/gluonts/model/deepar/_estimator.py
AaronSpieler, line 210: I dont know why this is in this pull request.

AaronSpieler: Why DeepAR is modified at all in fact.

AaronSpieler: Ok it seems this pull request introduced the "from_inputs method."

src/gluonts/model/seq2seq/_seq2seq_estimator.py
```diff
@@ -278,7 +284,7 @@ def __init__(
             kernel_size_seq=([3] * len([30, 30, 30])),
             channels_seq=[30, 30, 30],
             use_residual=True,
-            use_covariates=True,
+            use_dynamic_feat=True,
```
AaronSpieler, line 287: maybe also use_dynamic_cat

test/model/seq2seq/test_model.py
```diff
@@ -41,11 +41,50 @@ def Estimator(request):
...
+    dataset_train, dataset_test = make_dummy_datasets_with_features(
+        cardinality=[3, 10],
+        num_feat_dynamic_real=2,
+        freq=hps["freq"],
+        prediction_length=hps["prediction_length"],
+    )
+
+    estimator = MQCNNEstimator.from_hyperparameters(**hps)
 
-    accuracy_test(Estimator, hyperparameters, accuracy=0.2)
```
AaronSpieler, line 48: I will check whether the models now pass this too, must have been removed at some point.

src/gluonts/model/deepar/_estimator.py
```diff
@@ -146,8 +147,8 @@ def __init__(
         assert num_layers > 0, "The value of `num_layers` should be > 0"
         assert num_cells > 0, "The value of `num_cells` should be > 0"
         assert dropout_rate >= 0, "The value of `dropout_rate` should be >= 0"
-        assert (cardinality is not None and use_feat_static_cat) or (
-            cardinality is None and not use_feat_static_cat
+        assert (cardinality and use_feat_static_cat) or (
+            not (cardinality or use_feat_static_cat)
```
jaheba, line 151: I think it's ok to set `cardinality`, even when `use_feat_static_cat` is `False`.

Then, I can deselect categorical features, by only setting `use_feat_static_cat`, and `cardinality` is still derived from the auto_fields.

src/gluonts/model/estimator.py
```diff
@@ -69,6 +69,21 @@ def train(
...
+    def derive_auto_fields(cls, train_iter):
+        return {}
+
+    @classmethod
+    def from_inputs(cls, train_iter, **params):
+        # auto_params usually include `use_feat_dynamic_real`, `use_feat_static_cat` and `cardinality`
+        auto_params = cls.derive_auto_fields(train_iter)
+        # FIXME: probably params should take precedence over auto_params, since they were deliberately set,
+        #   however, on that case this method does not make sense, since if params says `use_feat_dynamic_real`=True
+        #   but `auto_params`=False, then this will lead to an error, since the appropriate data does not exist.
```
jaheba, line 82: I think this is fine, if the user sets incorrect parameters, there is not much we can do.

src/gluonts/model/estimator.py
```diff
@@ -69,6 +69,21 @@ def train(
...
+        #   This the only context in which this method makes sense is when auto_params take precedence, which could
+        #   lead to overwriting of explicit parameters. In this case a warning should be issued.
+        return cls.from_hyperparameters(**auto_params, **params)
```
jaheba, line 85:
```suggestion
        params = {**auto_params, **params}
        return cls.from_hyperparameters(**params)
```

This prevents name-conflicts, otherwise we cannot override settings.

src/gluonts/model/estimator.py
```diff
@@ -126,7 +141,10 @@ def from_hyperparameters(cls, **hyperparameters) -> "GluonEstimator":
             )
 
         try:
-            trainer = from_hyperparameters(Trainer, **hyperparameters)
+            trainer = hyperparameters.get("trainer")
+            if not isinstance(trainer, Trainer):
+                trainer = from_hyperparameters(Trainer, **hyperparameters)
+
```
jaheba, line 147: Why is this part of this PR?

src/gluonts/model/seq2seq/_mq_dnn_estimator.py
```diff
@@ -14,112 +14,151 @@
...
         ), "Elements of `mlp_hidden_dimension_seq` should be > 0"
+        assert channels_seq is None or all(
+            [d > 0 for d in channels_seq]
+        ), "Elements of `channels_seq` should be > 0"
+        assert dilation_seq is None or all(
+            [d > 0 for d in dilation_seq]
+        ), "Elements of `dilation_seq` should be > 0"
+        # TODO: add support for kernel size=1
+        assert kernel_size_seq is None or all(
+            [d > 1 for d in kernel_size_seq]
```
jaheba, line 73: Just an FYI, there is not need for `[ ]` here.

src/gluonts/model/seq2seq/_mq_dnn_estimator.py
```diff
@@ -14,112 +14,151 @@
...
+            dilation_seq if dilation_seq is not None else [1, 3, 9]
+        )
+        self.kernel_size_seq = (
+            kernel_size_seq if kernel_size_seq is not None else [3, 3, 3]
+        )
+        self.quantiles = (
+            quantiles
+            if quantiles is not None
+            else [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9]
+        )
```
jaheba, line 95: Can't we just set this directly in the method signature?

src/gluonts/model/seq2seq/_mq_dnn_estimator.py
```diff
@@ -14,112 +14,151 @@
...
+            == len(self.dilation_seq)
+            == len(self.kernel_size_seq)
+        ), (
+            f"mismatch CNN configurations: {len(self.channels_seq)} vs. "
+            f"{len(self.dilation_seq)} vs. {len(self.kernel_size_seq)}"
+        )
+
+        if seed:
+            np.random.seed(seed)
+            mx.random.seed(seed)
```
jaheba, line 108: Not sure we want that here. But I guess it's ok for now and we think about how to integrate this into gluon-ts generally.

src/gluonts/model/deepar/_estimator.py
AaronSpieler: Yeah, good Idea.

src/gluonts/model/seq2seq/_mq_dnn_estimator.py
lostella: I got in trouble once for having used a mutable default value for an argument, and started agreeing with PyCharm's warning here. I don't remember what exactly the problem was, but the problem is discussed, see e.g. https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

jaheba: Yes, this is generally an issue, but since we rely on `pydantic` magic here, we actually do not.

src/gluonts/model/estimator.py
AaronSpieler: Well, we could still set the Warning?

AaronSpieler: Or rather in this case I think the program will fail and say that a parameter has been specified twice, so yeah, we don't need to warn, you're right.

src/gluonts/model/estimator.py
AaronSpieler: I did not add this, and I don't know.

src/gluonts/model/seq2seq/_mq_dnn_estimator.py
AaronSpieler: thx!

src/gluonts/model/seq2seq/_mq_dnn_estimator.py
AaronSpieler: Bernie wanted some reproducibility, I guess thats why it was in here, but I agree that it does not really belong there long term.

src/gluonts/model/estimator.py
```diff
@@ -69,6 +69,18 @@ def train(
...
+    def derive_auto_fields(cls, train_iter):
+        return {}
+
+    @classmethod
+    def from_inputs(cls, train_iter, **params):
+        # auto_params usually include `use_feat_dynamic_real`, `use_feat_static_cat` and `cardinality`
+        auto_params = cls.derive_auto_fields(train_iter)
+        # user specified 'params' will take precedence:
+        params = {**auto_params, **params}
+        return cls.from_hyperparameters(params)
```
jaheba, line 82:
```suggestion
        return cls.from_hyperparameters(**params)
```

AaronSpieler:
So from master there was a change that removed shuffling all-together except on a single batch basis.

src/gluonts/model/seq2seq/_seq2seq_network.py
```diff
@@ -89,7 +89,7 @@ def compute_decoder_outputs(
         encoder_output_static, encoder_output_dynamic = self.encoder(
             scaled_target, embedded_cat, past_feat_dynamic_real
         )
-        decoder_input_static, _, decoder_input_dynamic = self.enc2dec(
+        decoder_input_static, decoder_input_dynamic = self.enc2dec(
```
AaronSpieler, line 92: This must have been a pretty big bug in the `seq2seq` library right? If you look at the default `enc2dec` the third second output is the decoder_input_dynamic, not the third, thus decoder_input_dynamic has been dropped thus far and future_dynamic_feat used in its place

AaronSpieler: I also removed the third output of `enc2dec` because it was never used and I did not see the point, but I'm open to putting that back.

src/gluonts/transform/field.py
```diff
@@ -39,23 +39,31 @@ def __init__(self, mapping: Dict[str, str]) -> None:
...
+
     @validated()
     def __init__(self, field_names: List[str]) -> None:
         self.field_names = field_names
 
     def transform(self, data: DataEntry) -> DataEntry:
         for k in self.field_names:
-            if k in data.keys():
-                del data[k]
+            data.pop(k, None)
```
lostella, line 66: This change is in contrast with the docstring I think ("if present"): if the field is absent, `pop` will raise a `KeyError`

src/gluonts/block/encoder.py
```diff
@@ -246,27 +232,29 @@ def hybrid_forward(
...
         static_code = F.slice_axis(ct, axis=1, begin=-1, end=None)
         static_code = F.squeeze(static_code, axis=1)
+
         return static_code, ct
 
 
 class RNNEncoder(Seq2SeqEncoder):
     """
-    Defines an RNN as the encoder.
+     Defines RNN encoder that uses covariates and target as input to the RNN if desired.
```
lostella, line 241:
```suggestion
    Defines RNN encoder that uses covariates and target as input to the RNN if desired.
```

src/gluonts/transform/field.py
AaronSpieler: If it's not present it will return `None` and should not fail. We don't use the return of the `pop` method, so the implementation should be equivalent?

src/gluonts/model/seq2seq/_seq2seq_network.py
AaronSpieler: @lostella do you have any preference here regarding my last comment?

src/gluonts/block/encoder.py
AaronSpieler: @lostella We could expose the scaler which is set to no operation scaling by default. Also I would remove the train method and from_inputs methods of the MQCNN for the GluonTS merge.

src/gluonts/model/seq2seq/_forking_estimator.py
```diff
@@ -92,52 +124,200 @@ def __init__(
...
+                    pred_length=self.prediction_length,
+                    dtype=self.dtype,
+                ),
+            )
+            dynamic_feat_fields.append(FieldName.FEAT_AGE)
+
+        if self.use_feat_dynamic_real:
+            # Backwards compatibility:
+            chain.append(
+                RenameFields({"dynamic_feat": FieldName.FEAT_DYNAMIC_REAL})
```
AaronSpieler, line 238: dynamic_feat could be replaced by the corresponding FieldName

PR #689, opened 2020-03-03
Multiprocessing data loader.

AaronSpieler:
*Issue, if available:*
With a multiprocessing data loader we should overcome data loading bottlenecks.
Will fix issue https://github.com/awslabs/gluon-ts/issues/682.

*Description of changes:*
* Datasets use the class attributes of `MPWorkerInfo` to get information about their multiprocessing environment.
* Datasets are replicated among workers (only Object reference though, not the physical dataset), this happens exactly one in the beginning of the training
* Datasets are not cached by default (caching not implemented so far)
* Data loading can now be done in a multiprocessing fashion by specifying the number of workers, this works for training set and validation set (for inference there is some bug for now, but that has the least impact on performance of all datasets)
* **Parallelisation for datasets:** modulo based, i.e. every `num_worker`th ts will be assigned to the corresponding worker. // however, this does not guarantee that the batches will always be sampled from equidistant locations for training, since some workers could potentially be slower or faster
* The data loaders return batches of transformed samples of `batch_size` in the requested context. The transformation is done according to the provided transformation.
* There is no threading support (wouldn't make sense since we are also doing computation heavy transformations), there is no memory pinning support (not necessary since we load the batches into the right context right away)
* Which exact batches and samples one gets is nondeterministic if num_workers > 0

**Future extensibility:**
* the main functions to modify will be the batching function and the stacking function, the transformation can already be replaced to any that produces a list of samples if applied to a dataset
* any dataset that makes use of the `MPWorkerInfo` class can be effectively parallelelized

**Missing functionality:**
* Shuffling (beyond a single batch)
* Dataset caching
* Correct documentation

**Current bugs:**
* No mp support for windows due to pickling error.
* No mp support for InferenceDataLoader due to pickling error.

**Possible improvements**
* Create named tuple for all the different data the worker processes use
* Only pass subset of dataset to worker
* Switch away from `Pool` to something that allows for more fine-grained control, like manually creating Processes as seen in [Pytorch's data-loader](https://pytorch.org/docs/stable/_modules/torch/utils/data/dataloader.html#DataLoader) or make use of libraries like Ray using [Actors]( https://ray.readthedocs.io/en/latest/async_api.html)


AaronSpieler:
So I ran some preliminary tests, and it seem we could get the 3x performance improvement @vafl described, especially for preprocessing heavy models like WaveNetEstimator

There are some things to be done though:
* There is a bug with the InferenceDataLoader where the transformed data cannot be pickled when in multiprocessing (thus mp disabled by default there)
** Not such a big problem for now, would only affect inference performance
* I also think the data should be transformed to the proper context in the data-loader directly, and thus we don't need memory pinning, the appropriate places have been marked with TODO (also the superfluous transformation to correct context outside of data loader can then be removed)
* Finally, I didn't get to reimplement shuffling
* Additionally, reintroducing explicit caching mechanisms could be of interest (I removed the caching bug)
* I made a couple of other TODO annotation, however, many can be delayed for another pull request

Also, I didn't get to thoroughly and rigorously test everything, so that is to be done too.

vafl:
This is great!
So none of the TODOs are blocking, right?

AaronSpieler:
I supplemented the proper context handling (i.e. using the provided one in the data loader). Additionally fixed a nondeterminism in the Iterator resetting mechanism for non cyclic datasets, now its deterministic.

So from my side, there is no blocking issue I'm aware of. Pytests pass on my machine with multiple threads and single threads reliably.

Overall I think there are many small improvements to make, which I did not have the time for to do so far, but they are documented in the TODO. I also left some useful debugging prints commented out for the person who will finish this PR.

src/gluonts/trainer/_base.py
```diff
@@ -252,9 +252,8 @@ def loop(
                             epoch_loss.update(None, preds=loss)
                             it.set_postfix(
                                 ordered_dict={
-                                    "epoch": f"{epoch_no + 1}/{self.epochs}",
                                     ("" if is_training else "validation_")
-                                    + "avg_epoch_loss": loss_value(epoch_loss),
+                                    + "avg_epoch_loss": loss_value(epoch_loss)
                                 },
                                 refresh=False,
                             )
```
AaronSpieler: I think these changes were accidental.

src/gluonts/dataset/loader.py
```diff
@@ -58,52 +66,53 @@ def __init__(
...
+        # TODO: think about what a good value is, probably 0, and if multiprocessing=True, then what is below
+        if num_workers is None:
+            self.num_workers = min(len(list(dataset)), int(cpu_count() / 2))
+        else:
+            assert num_workers <= len(
+                list(dataset)
+            ), "Cannot have more workers than dataset entries currently."
+            self.num_workers = num_workers
+
+        self.num_workers = 0
```
vafl: This is probably left over from testing and should be removed?

AaronSpieler: yes

AaronSpieler: I alway ran pytest with what I have above, and with 0 to cover both cases

AaronSpieler: > Thanks @AaronSpieler, this is great.
> 
> I think we could write one small test to check that the `TrainDataLoader` gives (up to permutation) the same batches for `num_workers > 1` and `num_workers = 0` when running some simple deterministic transformation and then merge this change.

I can tell you already that we will get probably slightly different results: The instance splitter probabilistically samples the number of split points in a single time series, if we have split the dataset by time series, and run a transformation over each subset, the sampling will most probably give different results:
The only thing that can be realistically tested is: is evert time series transformed only once?

vafl: We can use the `UniformSplitSampler` and set the probability to `1.0` for the test. Then every possible split is selected.

AaronSpieler: I see, that should of course yield the exact same result.

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -0,0 +1,613 @@
...
+    if transformed_data:
+        success = True
+        batch = batchify_fn(
+            data=transformed_data, dtype=dtype, parallel_processing=True
+        )
+    else:
+        # the second time without being able to provide a batch we want to delay calling them again
+        # on fist exhaustion they should not be delayed, since they need to indicate depletion
+        if _worker_iterator_exhausted_indicator:
+            time.sleep(0.05)
```
AaronSpieler: If im not missing something I think we could bump the penalty significantly here, since after each Validation pass through there is a Training phase, so this shouldn't interfere with the separate TrainingData loader. Basically same case for InferenceDataLoader, after which there is no second Inference using the same loader.
Some values to test could be '1' or '5'.

The main idea here is that if a depleted worker is requested again, we should prevent it from being called again so many times. Due to the current implementation there is no way to explicitly tell which worker should fetch a new batch.

Additionally, we don't want to terminate the worker, since for example the same ValidationDataLoader is called at every epoch. If we terminate the process one its depleted we would have to set up a new pool at each Validation iteration, which could come with costly copying etc.

AaronSpieler: On the other hand, epochs can be very fast, so maybe .1-.2?

AaronSpieler: It only needs to be so long that potential differences in processing time of different subsets of data are evened out a bit.

AaronSpieler:
**What is left to do now:**
* figure out the correct multiprocessing dependency

**What needs improvement:**
* I had to change some of the burn in logic handling (not the idea of what it's supposed to do) and I don't like it, it's not really elegant.
* Same situation with how requesting batches is handled when for example Iterators are depleted when using a ValidationDataLoader.

**Regarding the pytests:**
* They should be good enough to disclose braking changes, and also cover some soft constraints that should reveal if the iteration mechanics change drastically

AaronSpieler:
I see one of the tests timed out: "Python package / build (3.7, ubuntu-latest) (pull_request)"
Might want to set `num_batches_per_epoch=int(5 * exp_num_batches)` to just `num_batches_per_epoch=int(2.5 * exp_num_batches),` or something. Or put it in an extra test for more time.

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -0,0 +1,626 @@
...
+        self.shuffle = shuffle
+
+        self.num_workers = num_workers if num_workers >= 0 else 0
+        self.worker_pool = None
+        self.prefetch = max(
+            0, int(prefetch) if prefetch is not None else 2 * self.num_workers
+        )
+
+        self.dataset = dataset
+        self.dataset_len = len(list(dataset))
```
vafl: This loads the whole dataset into memory. 
Let's maybe remove this and switch to the modulo approach, where we don't need the dataset_len.

vafl:
I got the following error after ~1h of training:

```
INFO:gluonts.trainer:Epoch[58] Evaluation metric 'epoch_loss'=8.910754
Worker timed out after 120 seconds. This might be caused by 

                - Slow transform. Please increase timeout to allow slower data loading in each worker.
                - Insufficient shared_memory if `timeout` is large enough.
                Please consider reduce `num_workers` or increase shared_memory in system.
                
#0150it [00:00, ?it/s]#0150it [02:03, ?it/s, validation_avg_epoch_loss=8.88]
  File "sm_entry_point.py", line 56, in <module>
    train_and_eval_model.main(parser.parse_args())
  File "/opt/ml/code/havel_anomaly_scoring/train_and_eval_model.py", line 330, in main
    model_dir=model_dir
  File "/opt/ml/code/havel_anomaly_scoring/train_and_eval_model.py", line 116, in get_model
    rnn_predictor = rnn.train(train_ds, validation_ds)
  File "/opt/ml/code/havel_anomaly_scoring/gluonts_models/rnn/_estimator.py", line 546, in train
    train_output = self.train_model(training_data, validation_data)
  File "/opt/ml/code/gluonts/model/estimator.py", line 213, in train_model
    validation_iter=validation_data_loader,
  File "/opt/ml/code/gluonts/trainer/_base.py", line 305, in __call__
    epoch_no, validation_iter, is_training=False
  File "/opt/ml/code/gluonts/trainer/_base.py", line 231, in loop
    for batch_no, data_entry in enumerate(it, start=1):
  File "/usr/local/lib/python3.6/dist-packages/tqdm/std.py", line 1108, in __iter__
    for obj in iterable:
  File "/opt/ml/code/gluonts/dataset/loader.py", line 103, in __iter__
    for batch in self.parallel_data_loader:
  File "/opt/ml/code/gluonts/dataset/parallelized_loader.py", line 465, in __iter__
    next_batch = next(self)
  File "/opt/ml/code/gluonts/dataset/parallelized_loader.py", line 421, in __next__
    got = ret.get(self._timeout)
  File "/usr/lib/python3.6/multiprocessing/pool.py", line 640, in get
    raise TimeoutError
multiprocessing.context.TimeoutError
```

There was only 50% memory usage, so it wasn't an OOM issue. 
@AaronSpieler  Do you have an idea what the cause could be?

AaronSpieler:
> Slow transform. Please increase timeout to allow slower data loading in each worker.

Since our Tranformations are somewhat unpredictable, it could really just be that the worker caught a super long time series and then sampled a lot of points from it which ultimately led to the timeout: i.e. `Slow transform. Please increase timeout to allow slower data loading in each worker.`

Would be interesting to see how long the average batch processing takes in your case:
* to check that just measure the time in `_worker_fn` from the very beginning to right before the return. If the values are close to 120 sec my hypothesis is very much possible.

AaronSpieler:
> > Slow transform. Please increase timeout to allow slower data loading in each worker.
> 
> Since our Tranformations are somewhat unpredictable, it could really just be that the worker caught a super long time series and then sampled a lot of points from it which ultimately led to the timeout: i.e. `Slow transform. Please increase timeout to allow slower data loading in each worker.`
> 
> Would be interesting to see how long the average batch processing takes in your case:
> 
> * to check that just measure the time in `_worker_fn` from the very beginning to right before the return. If the values are close to 120 sec my hypothesis is very much possible.

To combat this we could try batch size proportional timeouts. Or keep track of average batch processing time and set the timeout to 3x.

On the other hand, im not quite sure where shared_memory is defined, so could be that `shared_memory` has a lower bound than the total memory available.

src/gluonts/dataset/common.py
```diff
@@ -184,27 +186,66 @@ class FileDataset(Dataset):
...
+                            burnt_in=False,
+                        )
+                    )
+                    for path in self.files()
+                ]
+            )
+            self._len = len_sum
+            return len_sum
+        else:
+            return self._len
```
jaheba:
```suggestion

        return self._len
```

src/gluonts/dataset/common.py
```diff
@@ -237,25 +290,75 @@ class ListDataset(Dataset):
...
+            #         row_number,
+            #     )
+
+            data = self.process(data)
             data["source"] = SourceContext(source=source_name, row=row_number)
             yield data
 
     def __len__(self):
-        return len(self.list_data)
+        return len(list(self.list_data))
```
jaheba: Why use `list` here?

src/gluonts/dataset/jsonl.py
```diff
@@ -15,12 +15,16 @@
 import functools
 from pathlib import Path
 from typing import NamedTuple
+import linecache
+from queue import Queue
+import multiprocessing as mp
```
jaheba:
```suggestion
```

src/gluonts/dataset/loader.py
```diff
@@ -57,52 +62,51 @@ def __init__(
...
-
-        return xs
+        self.cyclic = cyclic
+
+        # TODO: think about what a good value is, probably 0, and if multiprocessing=True, then what is below
+        if num_workers is None:
+            self.num_workers = min(len(list(dataset)), int(cpu_count() / 2))
+        else:
+            assert num_workers <= len(
+                list(dataset)
```
jaheba: Why `list` here? Don't we have special implementation for `len` on `Dataset` that are much faster?

src/gluonts/dataset/loader.py
```diff
@@ -57,52 +62,51 @@ def __init__(
...
+            dtype=self.dtype,
+            num_workers=self.num_workers,
+            **kwargs,
+        )
 
     def __iter__(self) -> Iterator[DataBatch]:
-        return self.make_batch_iter()
+        # Will take all batches, so that all data is sampled exactly once
+        for batch in self.parallel_data_loader:
+            yield batch
```
jaheba:
```suggestion
        yield from self.parallel_data_loader
```

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -0,0 +1,665 @@
...
+from typing import Callable, Iterable
+
+import multiprocessing
+import multiprocessing.queues
+from multiprocessing.reduction import ForkingPickler
+from multiprocessing.pool import Pool
+from multiprocessing import Queue
+
+try:
+    import multiprocessing.resource_sharer
```
jaheba: When does this happen?

src/gluonts/trainer/_base.py
```diff
@@ -76,11 +76,11 @@ class Trainer:
     ----------
     ctx
     epochs
-        Number of epochs that the network will train (default: 100).
+        Number of epochs that the network will train (default: 1).
```
jaheba: This shouldn't be part of this PR, should it?

test/dataset/test_multiprocessing_loader.py
```diff
@@ -0,0 +1,243 @@
...
+    # multi-processed validation dataset
+    mp_training_data_loader_result_01 = list(train_dataset_loader_01)
+
+    # should contain an entry for every time series id
+    transformation_counts_01 = get_transformation_counts(
+        mp_training_data_loader_result_01
+    )
+
+    assert all(
+        [k in transformation_counts_01 for k in range(CD_NUM_TIME_SERIES)]
```
jaheba: No need for `[ ]` here.

vafl:
I ran some more tests. 

There are a few problems:
The json lines dataset is too slow, so even with multiprocessing it doesn’t speed things up. I then tried parquet files, which were much faster and gave the expected speedup. 
However, with longer training runs at some point workers time out. Not sure why this is happening (memory usage is only ~10% on the machine).

src/gluonts/dataset/parallelized_loader.py
AaronSpieler: This is copied from MXNet, so I'm not exactly sure why they decided to do it this way.

src/gluonts/dataset/parallelized_loader.py
AaronSpieler: yeah that was just there for debugging originally, I removed it.

vafl:
Thanks @AaronSpieler this is really great. It works like a charm now.

Before we merge it, I have a few more points I would like to discuss:
- The `set_replica_info` method is a bit invasive for the dataset. Should we just go with the pytorch approach and have a method to get the `WorkerInfo` that contains the number of workers and the worker id. Then each dataset can decide on its own how to use this (i.e.  burn-in,  modulo, or split by file etc).
- For the `ListDataset` (and also *raw lists*) we could do something special: instead of passing the full list to each worker, we cut the list up before and pass separate chunks to each worker. That way we would support multiprocessing for pure lists directly and it would also drastically reduce the memory consumption for multiprocessing with lists.
- To make it clearer to the user, can we make the `num_workers` an explicit parameter for the `.train` method instead of passing this via kwargs?

What do you think?

AaronSpieler:
> Thanks @AaronSpieler this is really great. It works like a charm now.
> 
> Before we merge it, I have a few more points I would like to discuss:
> 
> * The `set_replica_info` method is a bit invasive for the dataset. Should we just go with the pytorch approach and have a method to get the `WorkerInfo` that contains the number of workers and the worker id? Then each dataset can decide on its own how to use this (i.e.  burn-in,  modulo, or split by file etc).
> * For the `ListDataset` (and also raw lists) we could do something special: instead of passing the full list to each worker, we cut the list up before and pass separate chunks to each worker. That way we would support multiprocessing for pure lists directly and it would also drastically reduce the memory consumption for multiprocessing with lists.
> * To make it clearer to the user, can we make the `num_workers` an explicit parameter for the `.train` method instead of passing this via kwargs?
> 
> What do you think?

* A method `get_worker_info` should be possible and is less invasive, so I think it's a great idea.
* Giving each worker a subset of the data from the beginning is unfortunately not possible  at the moment since we use the worker Pool paradigm, where each worker gets the exact same data to begin with, however, if we manually manage processes at some point this should be possible, and we could also request specific processes for batches, but this would require a major refactoring.
* `num_workers` can of course be an explicit argument, but then the question is how many arguments do we want to expose explicitly? For example `num_prefetch` could also be an interesting argument.

vafl:
>     * Giving each worker a subset of the data from the beginning is unfortunately not possible  at the moment since we use the worker Pool paradigm, where each worker gets the exact same data to begin with, however, if we manually manage processes at some point this should be possible, and we could also request specific processes for batches, but this would require a major refactoring.

Ah, of course. We would have to start each worker separately (or send the data chunk first thing via the queue).

>     * `num_workers` can of course be an explicit argument, but then the question is how many arguments do we want to expose explicitly? For example `num_prefetch` could also be an interesting argument.

Yes, sure. BTW, if we think we need to experiment more to stabilize it, we can merge it with `num_workers=0`. Then people can start trying it if they want.

AaronSpieler:
> Ah, of course. We would have to start each worker separately (or send the data chunk first thing via the queue).

The problem with a queue, even if feasable, would be that we don't get the desired speedup, because the queue is blocking. 

> Yes, sure. BTW, if we think we need to experiment more to stabilize it, we can merge it with `num_workers=0`. Then people can start trying it if they want.

I think that is a good idea.

vafl:
> > Ah, of course. We would have to start each worker separately (or send the data chunk first thing via the queue).
> 
> The problem with a queue, even if feasable, would be that we don't get the desired speedup, because the queue is blocking.

The idea would be to only use it to send the whole dataset initially. It is currently pickled and send aswell.

AaronSpieler:
> > > Ah, of course. We would have to start each worker separately (or send the data chunk first thing via the queue).
> > 
> > 
> > The problem with a queue, even if feasable, would be that we don't get the desired speedup, because the queue is blocking.
> 
> The idea would be to only use it to send the whole dataset initially. It is currently pickled and send aswell.

I'm not sure whether when the processes retrieve the subsets of data from the queue they wouldn't block each other until the respective data part is de-pickled?

src/gluonts/dataset/jsonl.py
```diff
@@ -57,55 +54,22 @@ class JsonLinesFile:
...
-                #  metadata. What is done to return an entry should be the only dataset type specific thing.
-
-                # skip until start_index on first pass, aka do burn_in
-                # in case of cyclic data sets always do burn in, in case of non cyclic ones, only once
-                if line_number < self.replica_info.start_index and (
-                    self.replica_info.end_index is not None
-                    or not self._burnt_in
+                # The dataset is equally distributed among the workers
+                if not (
+                    line_number % MPWorkerInfo.num_workers
```
vafl, line 69: Would be more efficient to do this inside the jsonl_file, since we would only read the line but not process it furtherl. But let's change this later, I think we wanted to cleanup the datasets anyway.

src/gluonts/trainer/_base.py
```diff
@@ -27,16 +27,15 @@
...
-from gluonts.core.exception import GluonTSDataError, GluonTSUserError
+from gluonts.core.exception import GluonTSDataError
 from gluonts.dataset.loader import TrainDataLoader, ValidationDataLoader
 from gluonts.support.util import HybridContext
 from gluonts.gluonts_tqdm import tqdm
 
 # Relative imports
 from . import learning_rate_scheduler as lrs
 
-logger = logging.getLogger("gluonts").getChild("trainer")
```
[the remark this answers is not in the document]
AaronSpieler: must be accidental

src/gluonts/dataset/loader.py
```diff
@@ -134,22 +124,26 @@ def __init__(
         batch_size: int,
         ctx: mx.Context,
         num_batches_per_epoch: int,
+        num_mp_workers: Optional[int] = None,
```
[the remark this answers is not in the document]
AaronSpieler: yes we did, but as long as Im testing, its easyer to leave the default non-zero so I see if & when tests fail due to the dataloader

src/gluonts/dataset/jsonl.py
AaronSpieler: What do you mean, I though this just reads the line files put doesn't process them further, since it just opens the file like `with open(self.path) as jsonl_file:`?

vafl: yeah, the only thing done is `json.loads`, which shouldn't be too bad.

src/gluonts/trainer/_base.py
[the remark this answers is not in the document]
AaronSpieler: I think this comes from me merging master? I will merge master again, then these things should be gone.

AaronSpieler:
Thanks to @jaheba, @lostella and especially @vafl for the help!

PR #699, opened 2020-03-09
Add M5 template

steverab:
*Description of changes:*

This is an initial draft of a template for using GluonTS-based tooling to participate in the M5 forecasting competition.

### DIFF
--- a/examples/m5_gluonts_template.ipynb
+++ b/examples/m5_gluonts_template.ipynb
@@ -1,0 +1,505 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# M5 Forecasting Competition GluonTS Template\n",
+    "\n",
+    "This notebook can be used as a starting point for participating in the M5 forecasting competition using GluonTS-based tooling."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Standard imports\n",
+    "\n",
+    "First we import standard data manipulation libraries."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%matplotlib inline\n",
+    "import mxnet as mx\n",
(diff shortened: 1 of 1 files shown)


geoalgo:
Wow this is so cool!

geoalgo:
I had a look and the notebook is very clear to me: great work!
We have this "policy" of not checking in notebooks in the format ".ipynb" but rather ".md" that are automatically converted (as the tutorial example). Maybe it would make sense to convert it this way (in a second iteration) as code reviewing and maintenance would be easier.
Also, it would be cool to add "m5" in the get_dataset so that we can get it in one line (also independent of this PR).

steverab:
Thanks @geoalgo! I already talked to @lostella prior to submitting this PR about the `.ipynb` vs `.md` thingy and we agree that in a second iteration we should make sure that this is part of the docs and therefore under constant testing.
As for including the M5 data into the `get_dataset()` method, I am not sure whether this is possible as the data is technically only accessible on Kaggle and you need to agree to the terms of the competition before being able to download the data. This is why we currently rely on the user to just download the data from there and supply it to us in the specified directory.

geoalgo:
Makes sense, I missed this point, thanks!

examples/m5_gluonts_template.ipynb
```diff
@@ -0,0 +1,499 @@
...
+   "source": [
+    "cal_features = calendar.drop(['date', \n",
+    "                              'wm_yr_wk', \n",
+    "                              'weekday', \n",
+    "                              'wday', \n",
+    "                              'month', \n",
+    "                              'year', \n",
+    "                              'event_name_1', \n",
+    "                              'event_name_2', \n",
+    "                              'd'], axis=1)\n",
```
lostella: Other minor comment: I think we use the following newline policy elsewhere in the codebase:

```
function(
    argument1, argument2, ...
)

[
    element1, element2, ...
]
```

(essentially, going to a newline right after parentheses, and not after the first element)

steverab: Both comments fixed! :)

steverab:
I agree, I'll follow up with you on this soon! :)

Issue #677, opened 2020-02-26
Introduce a lead_time attribute for predictors


lostella:
## Description

Currently, the nature of the forecasts of a predictor is specified by its `freq` and `prediction_length` attributes only. That is to say, if a predictor is given a series until time `t`, a forecast will be produced for the time steps from `t+1` to `t+prediction_length`, each spaced `freq` away from the next.

More generally, we could have a `lead_time` attribute so that forecasts span `t+1+lead_time` to `t+1+lead_time+prediction_length`. If in the base `Predictor` class we default `lead_time=0`, then existing models will be unaffected, so this change is backward compatible. In other cases however (for example in LSTNet, see #596) one could configure the model to produce forecasts with a lead time greater than zero.

~~Note that the proposed change will affect `ForecastGenerator` classes as well, since that’s what many predictors ultimately use.~~ (this may not be needed in the end, since the forecast start is written by the instance splitter)

A natural test that every predictor should pass, then, would be checking that forecasts start at the expected point in time (given the data and the `lead_time`) and that their length equals `prediction_length`.

**Checklist**

- [x] add `lead_time` attribute to base predictor and estimator classes (default: 0)
- [x] add `lead_time` option to instance splitter (default: 0)
- [x] add tests for instance splitter with `lead_time > 0`
- [x] modify LSTNet
- [x] add standard tests for predictors

PR #700, opened 2020-03-09
Add lead_time to predictors and estimators, update LSTNet

lostella:
*Issue #, if available:* #677 

*Description of changes:* This PR adds a lead_time attribute to both predictor and estimators, and modifies the InstanceSplitter class to take into account for lead_time. By default, lead_time is everywhere 0, so existing predictors and estimators are backwards compatible (provided that they invoke the super constructor upon construction).

Some evaluation code needed to be adjusted, and the LSTNet implementation was also updated accordingly (and simplified in some aspects).

### DIFF
--- a/src/gluonts/model/estimator.py
+++ b/src/gluonts/model/estimator.py
@@ -44,6 +44,14 @@ class Estimator:
 
     prediction_length: int
     freq: str
+    lead_time: int
+
+    def __init__(self, lead_time: int = 0, **kwargs) -> None:
+        # TODO validation of prediction_length and freq could also
+        # TODO be bubbled-up here from subclasses classes
+        assert lead_time >= 0, "The value of `lead_time` should be >= 0"
+
+        self.lead_time = lead_time
 
     def train(
         self, training_data: Dataset, validation_data: Optional[Dataset] = None
@@ -85,6 +93,7 @@ class DummyEstimator(Estimator):
 
     @validated()
     def __init__(self, predictor_cls: type, **kwargs) -> None:
+        super().__init__(**kwargs)
         self.predictor = predictor_cls(**kwargs)
 
     def train(
@@ -110,7 +119,10 @@ class GluonEstimator(Estimator):
     """
 
     @validated()
-    def __init__(self, trainer: Trainer, dtype: DType = np.float32) -> None:
+    def __init__(
+        self, trainer: Trainer, lead_time: int = 0, dtype: DType = np.float32
+    ) -> None:
+        super().__init__(lead_time=lead_time)
         self.trainer = trainer
         self.dtype = dtype
 
--- a/src/gluonts/model/predictor.py
+++ b/src/gluonts/model/predictor.py
@@ -180,7 +180,9 @@ class RepresentablePredictor(Predictor):
     def __init__(
         self, prediction_length: int, freq: str, lead_time: int = 0
     ) -> None:
-        super().__init__(prediction_length, freq, lead_time)
+        super().__init__(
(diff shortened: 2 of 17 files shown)


src/gluonts/model/estimator.py
```diff
@@ -44,6 +44,14 @@ class Estimator:
 
     prediction_length: int
     freq: str
+    lead_time: int
+
+    def __init__(self, lead_time: int = 0, **kwargs) -> None:
+        # TODO validation of prediction_length and freq could also
+        # TODO be bubbled-up here from subclasses classes
+        assert lead_time >= 0, "The value of `lead_time` should be >= 0"
```
ehsanmok: I think having asserts like this can also be done for `prediction_length` and `freq` where through inheritance it avoids all the redundant checks for each estimator implementation.

test/model/test_lstnet.py
```diff
@@ -109,3 +107,9 @@ def test_lstnet(
         iter(tss), iter(forecasts), num_series=len(dataset.test)
     )
     assert agg_metrics["ND"] < 1.5
+
+
+if __name__ == "__main__":
+    test_lstnet(
```
ehsanmok: This test cast is already included in the parametrized pytest so is redundant.

lostella: Whoops! Thanks, yes, that must be leftovers from my debugging

src/gluonts/model/estimator.py
lostella: Yes, I agree

src/gluonts/model/predictor.py
```diff
@@ -241,11 +247,12 @@ def __init__(
         freq: str,
         ctx: mx.Context,
         input_transform: Transformation,
+        lead_time: int = 0,
         forecast_generator: ForecastGenerator = SampleForecastGenerator(),
         output_transform: Optional[OutputTransform] = None,
         dtype: DType = np.float32,
     ) -> None:
-        super().__init__(prediction_length, freq)
+        super().__init__(prediction_length, freq, lead_time)
```
jaheba, line 255: Looking at this, I don't know if I like the order. We could force named arguments to make things more explicit.

There could be other ways, for example we have an object defining the forecast horizon:

```python
class ForecastHorizon:
    prediction_length: int
    lead_time: int = 0
```

Overall, I think I would like to be able to change the API if we found a way which we like.

What if an algorithm can predict any range of future values; i.e. why do these values need to be fixed? Also, why do we have to pass freq?

lostella: > What if an algorithm can predict any range of future values; i.e. why do these values need to be fixed? Also, why do we have to pass freq?

Sure, one could think of having models that predict, say, 10, 12, and 19 time steps from now. For the moment, this proposal makes the way the prediction range is specified more general than the status quo by allowing shifting the prediction range ahead. To me, this is already very general, but in the future we can decide to change the way this information is encoded, and potentially make it even more general.

jaheba: I think I would opt for using named_arguments now, and be ready to make some breaking changes later.

lostella: agreed, named argument it is

D3V4N5H:
Hi. Is lead_time available on GluonTS 0.5.2?

lostella:
@D3V4N5H this PR made it into 0.5.0 (see [here](https://github.com/awslabs/gluon-ts/releases/tag/v0.5.0)) so yes.

However, not all models expose this setting. In fact I think only LSTNet does, all others defaulting to `lead_time=0` (i.e. forecast starting from the next time step).

Issue #555, opened 2020-01-09
Does Gluon-ts support early stopping?


kmosby1992:
Does Gluon-ts support early stopping (for example: stop training after the loss fails to reduce after 50 epochs)? I see that we can set the learning rate to reduce if the loss fails to go down but could not find any options for early stopping

lostella:
@kmosby1992 you're right, there is no such an option currently. Maybe one natural way of modifying the training loop to get that, is to have the learning rate reduction mechanism to actually halt training once the minimum learning rate is hit, and no progress is done withing the prescribed patience. What do you think?

jackdent:
How about add an optional additional argument to the `Trainer` which is a function that gets called at the end of each epoch, and returns either `True` or `False` (whether to continue training). This function could use a closure to keep track of the historical losses for each epoch, which would also be useful if you wanted to plot the loss curves after training e.g.

```
train_losses = []
validation_losses = []

def cb(training_loss, validation_loss):
    train_losses.append(training_loss)
    validation_losses.append(validation_losses)

    if len(train_losses) >= 50:
        return abs(train_losses[-50] - train_losses[-1]) > epsilon
    else:
        return True

trainer = Trainer(..., cb=cb)
```

This would also

lostella:
@jackdent I think a callback mechanism like that could be useful in general, I agree we could have that option in the trainer. But even before that, maybe the current mechanism of learning rate scheduling should be adjusted so that once the patience is exceeded and the minimum stepsize is reached, the iterations stop.

MaximilianProll:
I'd love to see a callback mechanism being implemented for the `Trainer` e.g. for external logging.

lostella:
@kmosby1992 the changes in #701 now allow for early stopping: essentially, the learning rate scheduler stop the training loop once the loss stops decreasing and the learning rate has reached the minimum

@jackdent @MaximilianProll I'll open a separate issue for the callback mechanism

MaximilianProll:
Thanks a lot, @lostella, can you link the issue here once it's created?

lostella:
@MaximilianProll see #706

Arfea:
I'd also love to see an implementation of callbacks. Would you suggest where to start if I want to implement a callback method?

kaijennissen:
[...]

PR #701, opened 2020-03-10
Improve trainer handling of learning rate scheduling and logging

vafl:

### DIFF
--- a/src/gluonts/trainer/_base.py
+++ b/src/gluonts/trainer/_base.py
@@ -27,7 +27,7 @@ import numpy as np
 
 # First-party imports
 from gluonts.core.component import get_mxnet_context, validated
-from gluonts.core.exception import GluonTSDataError
+from gluonts.core.exception import GluonTSDataError, GluonTSUserError
 from gluonts.dataset.loader import TrainDataLoader, ValidationDataLoader
 from gluonts.support.util import HybridContext
 from gluonts.gluonts_tqdm import tqdm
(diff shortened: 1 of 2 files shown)


lostella:
When merged, this solves #555

src/gluonts/trainer/_base.py
```diff
@@ -315,26 +323,31 @@ def loop(
                         )  # TODO: handle possible exception
 
                     if not trainer.learning_rate == curr_lr:
-                        logging.info(
+                        if best_epoch_info.epoch_no == -1:
+                            raise GluonTSUserError(
+                                "Got NaN in first epoch. Try reducing initial learning rate."
+                            )
```
lostella: I'm not a fan of this, but maybe we can improve at a later stage

vafl: I tried to save the model before the training start and then use that serialized model in this case. The challenge is that we cannot serialize the model before the first batch ran through. So this would complicate things a bit (although maybe not that much).

DayanSiddiquiNXD:
hey are there any user docs for this? I searched in these pages on the gluonTS documentation (https://ts.gluon.ai/api/gluonts/gluonts.trainer.html, https://ts.gluon.ai/api/gluonts/gluonts.model.deepar.html, https://ts.gluon.ai/api/gluonts/gluonts.model.estimator.html) as well as the extensive examples page but couldn't find anything about using learning rate based early stopping

PR #704, opened 2020-03-11
Add missing value support for ForkingSeq2Seq Estimators (MQCNN, MQRNN)

rshyamsundar:
Add missing value support for MQRNN, MQCNN similar to what is done in Wavenet:

* Use a dummy value in place of missing entry and add an indicator feature for observed values on the encoder side.
* On the decoder side, mask the loss corresponding to missing entries.


lostella:
This probably requires rebasing on top of the changes in #709, which also affected the forking network

src/gluonts/model/seq2seq/_forking_network.py
```diff
@@ -66,26 +67,36 @@ def __init__(
...
         -------
-        loss with shape (FIXME, FIXME)
+        loss with shape (num_ts, decoder_length)
         """
 
         # FIXME: can we factor out a common prefix in the base network?
+
         feat_static_real = F.zeros(shape=(1,))
-        past_feat_dynamic_real = F.zeros(shape=(1,))
+        past_feat_dynamic_real = past_observed_values
```
AaronSpieler, line 99: Why is `past_feat_dynamic_real` set to  `past_observed_values`? Shouldn't `past_observed_values` be used for masking, it's not a covariate itself is it?

rshyamsundar: Since `past_target` is also an input feature to the network and we are masking some entries in it, network should be able to distinguish between the actual values and the masked values. Passing `past_observed_values` as a feature would allow the network to make that distinction and possibly ignore the masked values.

AaronSpieler:
As stated, this pull hast been manually merged in https://github.com/awslabs/gluon-ts/pull/668.

PR #716, opened 2020-03-20
M5 data loader

steverab:
*Description of changes:*

Following our notebooks for the the M5 accuracy and uncertainty forecasting competitions, this PR now adds a convenience data loading method for the M5 data. This method will load the Kaggle data and add the relevant static and dynamic features. Unfortunately, since the data is only available once users agree to the terms and conditions on Kaggle, we have to ask the user to supply the data to us which we can then process into a GluonTS-readable format. A respective runtime error will alert the user to first download the data from Kaggle and paste the data into our dataset directory in case these files are not present.

### DIFF
--- a/src/gluonts/dataset/repository/_m5.py
+++ b/src/gluonts/dataset/repository/_m5.py
@@ -1,0 +1,156 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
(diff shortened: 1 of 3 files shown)


steverab:
Personally, I think that for the notebooks it is more valuable to explicitly go through the data processing step by step such that users can understand how the GluonTS data format works and (perhaps even more importantly) users can easily add more features or can perform custom pre-processing steps. For users who are not deeply familiar with GluonTS data transformations, this might be easier to do in numpy before dataset creation takes place.

geoalgo:
Sure, makes sense.
Maybe you can just mention the command that allow to do this processing in one line in the notebook?

steverab:
That's definitely an option. I can add this as a comment once a new version of GluonTS is released. :)

Issue #718, opened 2020-03-21
Negative binomial alpha parameter when `scale=1`


kashif:
## Description
The negative binomial distribution's scaling seems a bit counter intuitive to me e.g. when `scale=1` it should not change the parameters of the distribution... however it seems that the `alpha` parameter gets multiplied by `sqrt(2)`.  I would have thought it would be something like: `alpha /= sqrt(scale)` but need to go through the calculation...

See: https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/distribution/neg_binomial.py#L133

Am I missing something? Thanks!

Issue #636, opened 2020-02-14
Predictions are way too high when modeling an intermittent count series with DeepAR and NegBin distribution?


SkanderHn:
I'm trying to model a simulated series of weekly seasonal intermittent sales, with values between 0 and 4. I generated  5 years of simulated data:

![Screen Shot 2020-02-13 at 1 30 59 PM](https://user-images.githubusercontent.com/10626476/74480894-99ab6a00-4e66-11ea-9610-44ad8db54896.png)

I trained a DeepAR model with the output distribution set to Negative Binomial (all other settings were the default settings), on 3 years, and generated predictions for the next two. I got the following results (plotting the [70.0, 80.0, 95.0] predictions intervals):

![Screen Shot 2020-02-13 at 1 31 50 PM](https://user-images.githubusercontent.com/10626476/74481159-0e7ea400-4e67-11ea-89e6-514067e20f80.png)

Increasing number of training epochs doesn't change anything, the loss falls to its lowest value around the 8th to 10th epoch and hovers more or less around there, whether I train for 10 or 100 epochs. 
I thought training on 3 years and testing on 2 might be too ambitious, so I tried 4y/1y split instead, and the results got much worse - and downright strange - this time with values climbing into the 100s, even though the largest historical value the series ever reaches is 4 (I'm using the same input series, but is seems flat now because the scale is completely skewed by how large the predictions are): 

![Screen Shot 2020-02-13 at 3 55 13 PM](https://user-images.githubusercontent.com/10626476/74489153-47277900-4e79-11ea-8f2a-16cdfe606f70.png)

I'm wondering if I am doing anything wrong? Are there any special settings for DeepAR when applied to intermittent series? 

For comparison, the DeepAREstimator worked pretty well out of the box for more traditional series (using Student's distribution), for example: 

![Screen Shot 2020-02-12 at 4 49 47 PM](https://user-images.githubusercontent.com/10626476/74489885-9b335d00-4e7b-11ea-91df-ee8e2deb696d.png)

Details:

Train data:
[{'start': Timestamp('2014-01-05 00:00:00', freq='W-SUN'),
  'target': array([1., 0., 1., 0., 0., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
         0., 0., 1., 0., 0., 0., 0., 0., 1., 0., 0., 0., 0., 1., 0., 0., 1.,
[...]

PR #719, opened 2020-03-22
Fix negative binomial's scaling

kashif:
Fixes issue #718

Also might explain issue #636

*Issue #, if available:*

*Description of changes:*


canerturkmen:
Hi @kashif, thank you so much for surfacing this. I can confirm that indeed, it seems like a bug. However I couldn't validate your fix as well. Do you have a proof / argument / test anywhere that I can check?

As a counterargument here, this is the "mean-dispersion" parameterization of the negative binomial. The previous version "overdisperses" the distribution while the fix does the inverse. I don't think either are correct.

kashif:
thanks @canerturkmen So let see... in the current parametrisation, of `mu` and `alpha` the NB's mean is `mu` and variance is: `mu + alpha * mu**2`. 

So if we now want to sample from a NB scaled by `nu`, the mean has to become `mu = mu*nu` and the variance has to scale by `nu**2` in which case the transformation to alpha has to transform by:

```alpha =  alpha - (nu -1)/(nu*mu)```

The transformation I have here is not really correct either, but it mirrors the transformation of the paper and doesn't have the extra `+1`. I am happy to change it to the above version. What do you say?

**Comment by lostella** [comment]
> I am happy to change it to the above version. What do you say?

I agree that's how it should be. This also agrees with @canerturkmen 's intuition. I'd say we give it try it and check whether it solves the related issues? Maybe we they can be reproduced

**Comment by kashif** [comment]
ok cool! In our experiments this proposed transformation gave better neg. log likelihoods... I'll update this PR

**Comment by lostella** [comment]
@kashif maybe it's `alpha' = alpha + (nu-1)/(nu*mu)` with a plus?

**Comment by kashif** [comment]
@lostella yes just typo writing from my notes...

**Comment by canerturkmen** [comment]
Yup. That should be the one.

@kashif, out of interest, could you share which proposal gave better results and which paper does your initial reply refer to?

File: src/gluonts/distribution/neg_binomial.py [hunk 1]
```diff
@@ -130,7 +130,7 @@ def distribution(
         else:
             F = getF(mu)
             mu = F.broadcast_mul(mu, scale)
-            alpha = F.broadcast_div(alpha, F.sqrt(scale))
+            alpha = F.broadcast_add(alpha, F.broadcast_div(scale - 1, mu))
```
canerturkmen, line 133: Is this  `F.broadcast_div(scale - 1, scale*mu)` ?

kashif:
@canerturkmen  This current change gave the best neg. log likelihood and the DeepAR paper section 4.3 ( https://www.sciencedirect.com/science/article/pii/S0169207019301888 ) has the initial change...

kashif:
@lostella not sure about mxnet broadcast functions... perhaps I did something wrong, kindly change it if so, thanks!

canerturkmen:
> @lostella not sure about mxnet broadcast functions... perhaps I did something wrong, kindly change it if so, thanks!

Got it now, thanks also for the pointers.

kashif:
from my side once this issue is done, it would be a good time for a release a point release...

kashif:
@canerturkmen would you be able to run some tests on your side with this scaling?

canerturkmen:
> @canerturkmen would you be able to run some tests on your side with this scaling?

Yes, we'll give this another look @kashif. Thank you for your contribution.

PR #728, opened 2020-03-26
fix binned distribution for mxnet 1.6

vafl:
*Issue #, if available:*

*Description of changes:*

For some reason using `*` and `+` instead of the verbose `broadcast_mul` and `broadcast_add` with hybridize gives wrong results on mxnet 1.6 on a GPU for the binned distribution. I think it only occurs when using `get_constant`. I'm filing a bug report for mxnet for this.

### DIFF
--- a/src/gluonts/distribution/binned.py
+++ b/src/gluonts/distribution/binned.py
@@ -84,9 +84,12 @@ class Binned(Distribution):
         )
 
         means = (
-            bin_centers.slice_axis(axis=-1, begin=1, end=None)
-            + bin_centers.slice_axis(axis=-1, begin=0, end=-1)
-        ) / 2.0
+            F.broadcast_add(
+                bin_centers.slice_axis(axis=-1, begin=1, end=None),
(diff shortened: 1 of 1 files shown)


src/gluonts/distribution/binned.py
```diff
@@ -110,21 +113,25 @@ def event_dim(self) -> int:
...
+        return F.broadcast_mul(self.bin_probs, self.bin_centers).sum(axis=-1)
 
     @property
     def stddev(self):
-        ex2 = (self.bin_probs * self.bin_centers.square()).sum(axis=-1)
-        return (ex2 - self.mean.square()).sqrt()
+        ex2 = self.F.broadcast_mul(
+            self.bin_probs, self.bin_centers.square()
+        ).sum(axis=-1)
+        return F.broadcast_minus(ex2, self.mean.square()).sqrt()
```
lostella:
```suggestion
        return self.F.broadcast_minus(ex2, self.mean.square()).sqrt()
```

vafl:
Bug report: https://github.com/apache/incubator-mxnet/issues/17914

PR #731, opened 2020-03-28
Label smoothing binned distribution

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Introduces a version of the `Binned` distribution that features [label smoothing](https://arxiv.org/abs/1906.02629), a technique formerly found to promote better calibration, faster training, and better generalization with binned type losses.

The implementation subclasses `Binned` and `BinnedOutput`, overriding the `loss()` method of the former. Edits to `binned.py` are to prevent code duplication.

### DIFF
--- a/docs/bibliography.rst
+++ b/docs/bibliography.rst
@@ -11,6 +11,8 @@ Bibliography
 
 .. [LCY+18] Lai, Guokun, et al. "Modeling long-and short-term temporal patterns with deep neural networks." The 41st International ACM SIGIR Conference on Research & Development in Information Retrieval. ACM, 2018.
 
+.. [MKH19] Muller, Rafael, Simon Kornblith, and Geoffrey E. Hinton. "When does label smoothing help?." Advances in Neural Information Processing Systems. 2019.
+
 .. [ODZ+16] Oord, Aaron van den, et al. "Wavenet: A generative model for raw audio." arXiv preprint arXiv:1609.03499 (2016).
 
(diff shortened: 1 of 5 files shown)


lostella:
> Looks good to me! One general question maybe: do you think it would be possible to obtain the same without introducing `SmoothBinned`, but only a `SmoothBinnedOutput` which regularizes (== smoothens) the probabilities when constructing a `Binned` distribution in the `SmoothBinnedOutput.distribution` method? I can look in detail and find that out, but you know better :-)

Nevermind, I think this is not possible, since the smoothing is applied to `x`, not to the bin probabilities

canerturkmen:
> > Looks good to me! One general question maybe: do you think it would be possible to obtain the same without introducing `SmoothBinned`, but only a `SmoothBinnedOutput` which regularizes (== smoothens) the probabilities when constructing a `Binned` distribution in the `SmoothBinnedOutput.distribution` method? I can look in detail and find that out, but you know better :-)
> 
> Nevermind, I think this is not possible, since the smoothing is applied to `x`, not to the bin probabilities

I think it could be possible. Actually, I can try merging it all in, and possibly this could be handled in the constructor arguments of both `Binned` and `BinnedOutput`. It's a matter of style, you tell me.

the `loss` of `Binned` would be something like
```python
def loss(self, x: Tensor) -> Tensor:
    return self.smooth_ce_loss(x) if self.label_smoothing else -self.log_prob(x)
```

lostella:
> I think it could be possible. Actually, I can try merging it all in, and possibly this could be handled in the constructor arguments of both `Binned` and `BinnedOutput`.

Yeah, that way it's definitely possible. In that case we would need to have `label_smoothing=None` by default, for backward compatibility.

PR #736, opened 2020-03-31
Reduce amount of warnings.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/evaluation/_base.py
+++ b/src/gluonts/evaluation/_base.py
@@ -35,6 +35,7 @@ import pandas as pd
 # First-party imports
 from gluonts.model.forecast import Forecast, Quantile
 from gluonts.gluonts_tqdm import tqdm
+from gluonts.support.util import zdivide
 
 
 @lru_cache()
@@ -370,7 +371,7 @@ class Evaluator:
         ]
         for quantile in self.quantiles:
             totals[quantile.weighted_loss_name] = np.divide(
-                totals[quantile.loss_name], totals["abs_target_sum"]
+                totals[quantile.loss_name], totals["abs_target_sum"] + flag
             )
 
         totals["mean_wQuantileLoss"] = np.array(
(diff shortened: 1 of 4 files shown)


src/gluonts/evaluation/_base.py
```diff
@@ -370,7 +371,7 @@ def get_aggregate_metrics(
         ]
         for quantile in self.quantiles:
             totals[quantile.weighted_loss_name] = np.divide(
-                totals[quantile.loss_name], totals["abs_target_sum"]
+                totals[quantile.loss_name], totals["abs_target_sum"] + flag
```
lostella: I'm not sure this is what we want

jaheba: Right, but it doesn't throw a warning!!

(I'm not sure just ignoring the division by zero is a good idea either.)

lostella: When does the division by zero occur? Is it in tests? If I understand correctly, `totals["abs_target_sum"]` is the sum of all (absolute) values from all the time series in the dataset, so I guess in practice one shouldn't be getting zero there?

lostella: It actually contains the sums *per time series*, so division by zero can happen in case a constantly zero time series is in the dataset: again, I don't think this is a problem in practice. Maybe we merge everything else and keep this as it is for now? Unless of course there are reasons to change this.

jaheba: Pytest prints *a lot* of warnings, and I wanted to get rid of them.

I think "his is not a problem in practise" is an interesting way to see things and I would prefer to think about the edge cases.

lostella: > Pytest prints _a lot_ of warnings, and I wanted to get rid of them.
> 
> I think "his is not a problem in practise" is an interesting way to see things and I would prefer to think about the edge cases.

Of course, but dividing by 1.0 here is an arbitrary choice: I'm just pointing this out since an alternative would be that of dividing by a very small number. But that's ok, again I think this problem should mostly arise in tests, so any positive number is ok.

PR #739, opened 2020-04-02
Mp data loader updates

AaronSpieler:
*Description of changes:*
* Adds multiprocessing support for the InferenceDataLoader (add additional test too)
* Minor refactoring regarding data handling of workers

### DIFF
--- a/test/dataset/test_multiprocessing_loader.py
+++ b/test/dataset/test_multiprocessing_loader.py
@@ -24,7 +24,11 @@ import pytest
 
 # First-party imports
 from gluonts.dataset.field_names import FieldName
-from gluonts.dataset.loader import TrainDataLoader, ValidationDataLoader
+from gluonts.dataset.loader import (
+    TrainDataLoader,
+    ValidationDataLoader,
+    InferenceDataLoader,
+)
 from gluonts.dataset.common import ListDataset
 from gluonts.transform import Chain, UniformSplitSampler, InstanceSplitter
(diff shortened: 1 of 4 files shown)


test/dataset/test_multiprocessing_loader.py
```diff
@@ -228,7 +291,7 @@ def test_training_loader_soft_constraint_01() -> None:
 
 # CASE 02: if we have say 5 workers, but let each only cover 1/10, then
 # it should be impossible to cover the whole underlying dataset
-@flaky(max_runs=3, min_passes=3)
+@flaky(max_runs=2, min_passes=2)
```
lostella, line 294: Minor: this decorator makes this equivalent to running this test twice, and expecting two successes. From flaky's README:

> By default, flaky will retry a failing test once, but that behavior can be overridden by passing values to the flaky decorator. It accepts two parameters: max_runs, and min_passes; flaky will run tests up to max_runs times, until it has succeeded min_passes times. Once a test passes min_passes times, it's considered a success; once it has been run max_runs times without passing min_passes times, it's considered a failure.

Is that intended?

AaronSpieler: Yes, because it's a soft constraint, I don't want it to just pass it by luck. It is imaginable that if this should fail, it still passes simply based on luck. Never seen it happen, however, I think it should at least pass twice. But since this has never happened I'm not to keen on keeping it.

PR #740, opened 2020-04-03
Multiprocessing support for metric calculation (inference)

AaronSpieler:
*Description of changes:*
* Added multiprocessing support for metric calculation (inference)

I have measured speedups up to 3x on lager datasets like `m4_yearly` and up to 10% on smaller dataset like `m4_hourly`. Whats essentially limiting is the read/write speeds of the hard disk now. 


AaronSpieler:
Ups, seem I have worked off the wrong branch.

Issue #683, opened 2020-02-28
"map-reduce" compatible interface


vafl:
For [generating predictions](https://github.com/awslabs/gluon-ts/blob/167772820d87f7481a2305691678741d67d9326e/src/gluonts/evaluation/backtest.py#L40) and for [running evaluations](https://github.com/awslabs/gluon-ts/blob/167772820d87f7481a2305691678741d67d9326e/src/gluonts/evaluation/backtest.py#L118) we currently have a streaming based api. This is nice, because it allows out of core evaluations for larger datasets.

However, we do not really support parallelization directly using multiple CPUs/GPUs or e.g. to run predictions on a cluster of machines.

I think we should change our APIs to make them more friendly to "map-reduce" type operations. This would make it very easy to parallelize predictions and evaluations and would allow the user to directly plug into e.g. [Dask](https://dask.org/) or Spark for larger scale predictions on a cluster.

Maybe this is already doable with the current APIs, in that case, we can make it more explicit (perhaps renaming some methods or function signatures) and have a tutorial for how to do this with e.g. Dask.

PR #741, opened 2020-04-03
Added multiprocessing support for evaluation.

AaronSpieler:
*Description of changes:*

* Added multiprocessing support for metric calculation (inference)

I have measured speedups up to 3x on lager datasets like m4_yearly and up to 10% on smaller dataset like m4_hourly. Whats essentially limiting is the read/write speeds of the hard disk now.


src/gluonts/evaluation/_base.py
```diff
@@ -65,6 +67,24 @@ def get_seasonality(freq: str) -> int:
     return seasonality // multiple
 
 
+# This is required for the multiprocessing to work.
+_worker_evaluator = None
+
+
+def _worker_init(evaluator):
```
lostella, line 74:
```suggestion
def _worker_init(evaluator: Evaluator):
```

src/gluonts/evaluation/_base.py
```diff
@@ -90,6 +110,14 @@ class Evaluator:
         which is computationally expensive to evaluate and thus slows
         down the evaluation process considerably.
         By default False.
+    num_worders
+        The number of multiprocessing workers that will be used to process
+        the data in parallel.
+        Default is `int(multiprocessing.cpu_count()/(2/3)))`.
```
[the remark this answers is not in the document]
AaronSpieler: I found it more performant to not use all CPUs. Essentially it just more work to initialise more, and some processes are busy with the main thread anyways. Maybe `multiprocessing.cpu_count()-1` could be an idea.

AaronSpieler:
> See also: #683

How is this meant?
So I would argue this doesn't fix this issue completely, but I guess Its the best performance/effort option. And it's probably close to the best performance one is going to get on a single machine. The cluster inference will have to be addressed separately.

src/gluonts/evaluation/_base.py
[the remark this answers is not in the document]
AaronSpieler: Ahh, but yeah, that was only for small datasets I guess, so thats why I left it finally at the cpu count.

src/gluonts/evaluation/_base.py
```diff
@@ -142,8 +161,20 @@ def __call__(
...
         ) as it, np.errstate(invalid="ignore"):
-            for ts, forecast in it:
-                rows.append(self.get_metrics_per_ts(ts, forecast))
+            if self.num_workers > 0:
+                mp_pool = multiprocessing.Pool(
+                    initializer=_worker_init(self), processes=self.num_workers
+                )
+                rows = mp_pool.map(
+                    func=_worker_fun,
+                    iterable=iter(it),
```
AaronSpieler, line 170: this iter(...) call is necessary for some reason, otherwise tests fail. I checked, and without it in some tests the first instance returned by next is just a string showing the progress bar?

src/gluonts/evaluation/_base.py
```diff
@@ -90,6 +92,14 @@ class Evaluator:
...
         down the evaluation process considerably.
         By default False.
+    num_workers
+        The number of multiprocessing workers that will be used to process
+        the data in parallel.
+        Default is multiprocessing.cpu_count().
+        Setting it to 0 means no multiprocessing.
+    chunk_size
+        Controls the approximate chunk size each workers handles at a time.
+        Default is 32.
```
AaronSpieler, line 102: Optimal `chunck_size` will depend on the operating system and the page sizes etc. and how much the cpu is prefetching.

AaronSpieler: 32 seems to be good, but basically 8 to 64 was about as good too.

src/gluonts/evaluation/_base.py
lostella: Since we are iterating a tqdm object, I'm not sure what the progress bar will look like when you enable multiprocessing. Maybe what you observing is related to tqdm interfering with multiprocessing?

AaronSpieler: Yes, it's due to tqdm, however, the progress bar does work in mp.

AaronSpieler:
> This looks neat: could you make sure that both cases `num_workers == 0` and `not num_workers == 0` are tested?

done.

test/evaluation/test_evaluator.py
```diff
@@ -417,7 +417,31 @@ def test_MASE_sMAPE_M4(timeseries, res):
...
+                "Scores for the metric {} do not match: \nexpected: {} "
+                "\nobtained: {}".format(metric, res[metric], score)
+            )
+
+
+@pytest.mark.parametrize(
+    "timeseries, res, has_nans, input_type",
+    zip(TIMESERIES, RES, HAS_NANS, INPUT_TYPE),
+)
+def test_metrics(timeseries, res, has_nans, input_type):
```
lostella, line 441: This test function and the above one have the same name: I'm not sure how that will work out, but for clarity you could rename them as `test_metrics_single_processing` and `test_metrics_multi_processing`, or something like that

AaronSpieler: or, renamed it

PR #742, opened 2020-04-03
MP DataLoader Improvements

AaronSpieler:
*Description of changes:*
* dataset iterations are now cache aligned (instead of modulo based, just assign subsets)
* switched to default Pool creation method (for linux that would be fork currently: this improves initialisation times significantly on linux systems)
* added caching support for file dataset
* added support for num_batches_for_shuffling, with default 8
* added much more typing, and documentation
* some refactoring


AaronSpieler:
@vafl maybe you could check wether you run into problems, or see improvements?

AaronSpieler:
So for WaveNetEstimator i get 4x performance if `num_batches_for_schuffling=1` otherwise `3x` (in which `num_batches_for_schuffling=8`).

The question is whether to leave the default 8 or 1. So far this arguments default was 8, on the other hand, there was a bug in the previous data loader code (before multiprocessing), namely:

```
@property
    def stream(self) -> Iterable:
        s = self.transform(self.dataset, is_train=self.is_train)
        if self.shuffle_for_training:
            return shuffler(s, self.num_batches_for_shuffling)
        return s
```

```
def shuffler(stream: Iterable[T], batch_size: int) -> Iterator[T]:
    """Modifies a stream by shuffling items in windows.

    It continously takes `batch_size`-elements from the stream and yields
    elements from each batch  in random order."""

    for batch in batcher(stream, batch_size):
        random.shuffle(batch)
        yield from batch
```

which means that only `num_batches_for_schuffling` samples were shuffled, and not `batch_size`*`num_batches_for_schuffling`, which is most likely even worse than having `num_batches_for_schuffling=1`.

And the model training seems to have worked that way fine too. However, I can imagine fixing this bug could yield better performance now in some cases.

Any preference @vafl ?

AaronSpieler:
Regarding the read speeds, if I use `m4_yearly` and run the following code (which is not an exact measurement, but stark differences should show):

```
num_iter=100
print(f"Runtime: {timeit.timeit(lambda: list(dataset.train), number=num_iter)/num_iter}")
```

I consistently get the following results:

```
Chunk: 0.04972371574986027
Modulo: 0.05123165169003187
```

However, while the difference in speed is measurable, it's also negligible.

AaronSpieler:
And setting `cache=True` in the `FileDataset` yields significant boost to performance (on Wavenet). I get essentially the same performance as if the dataset was converted beforehand to a `list`, except, there is not significant hit in performance in the beginning of training (due to dataset needing to be copied to all the workers), and overall there is less data cached in memory, as opposed to if the dataset was converted to a list beforehand (only 1/num_workers as much cashed in total)

For example, for this configuration on the electricity dataset:
```
estim = WaveNetEstimator(
    freq=meta.freq,
    prediction_length=24,
    seasonality=48,
    trainer=Trainer(
        batch_size=32,
        epochs=7,
        hybridize=True,
        learning_rate=0.01,
        num_batches_per_epoch=400
    )
)
estim.train(train_01, num_workers=8, num_batches_for_shuffling=1, shuffle_for_training=True)
```

I get ~17 vs 24 it/sec uncached vs cached.

AaronSpieler:
With the last commit I changed the functionality of `num_batches_for_schuffling` to probabilistically choosing every `num_batches_for_schuffling ` enty, which:

* massively simplifies the code
* improves the performance in the case of `num_batches_for_schuffling=8` in my testing that it's basically as performant as `num_batches_for_schuffling=1`, which is a 50% improvement over the previous implementation
* the caching mechanism is unaffected
* "drawback": we can't guarantee (in fact it's unlikely) that in `num_batches_for_schuffling` passes every data entry is processed exactly once. (but the InstanceSplitter is also probabilistic at the moment for training, so that would be the case anyways.)

In order to achieve this, I had to switch back to the cache aligned data iteration, because in the none cached FileDataset the difference between the techniques is significantly noticeable (since now `num_batches_for_schuffling` as much has to be read from the dataset, i.e. from the disk).

AaronSpieler:
With this I think every feature is added that was missing (that I am aware of), and I tested every design choice that I though needed validation.

src/gluonts/dataset/loader.py
```diff
@@ -67,13 +65,14 @@ def __init__(
         dataset: Dataset,
         *,
         transform: Transformation,
+        cyclic: bool,
         is_train: bool,
         batch_size: int,
         ctx: mx.Context,
-        dtype: DType = np.float32,
-        cyclic: bool = False,
+        dtype: Optional[DType] = np.float32,
```
lostella, line 72: Why is this optional? If I go up the hierarchy of constructions, I end up in the ParallelDataLoader that has this argument optional, and if this is None then sets it to np.float32. Maybe None should not be allowed, and np.float32 be the default in all these classes

src/gluonts/dataset/common.py
```diff
@@ -185,26 +186,39 @@ class FileDataset(Dataset):
...
+        Indicates whether the dataset should be cached or not.
     """
 
     def __init__(
-        self, path: Path, freq: str, one_dim_target: bool = True,
+        self,
+        path: Path,
+        freq: str,
+        one_dim_target: bool = True,
+        cache: Optional[bool] = False,
```
lostella, line 198: How is this optional? Shouldn’t this be either True or False?

I think Optional and a default None should be used in two cases:
1. Having None as value for an argument has some meaning, like “don’t use this feature”
2. The default value for the argument requires a complex expression and potentially depends on other arguments values, so one defaults it to None and sets it appropriately in the constructor body

lostella:
I have some minor comments, see inline

src/gluonts/dataset/common.py
AaronSpieler: You are right, I think I have over used it.

src/gluonts/dataset/common.py
```diff
@@ -185,26 +186,39 @@ class FileDataset(Dataset):
...
 
     def __init__(
-        self, path: Path, freq: str, one_dim_target: bool = True,
+        self,
+        path: Path,
+        freq: str,
+        one_dim_target: bool = True,
+        cache: bool = False,
     ) -> None:
+        self.cache = cache
```
lostella, line 200: I don't get it: where is this used within the class?

lostella: I see, it's just stored, and `cache` is used in the constructor

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -438,8 +478,9 @@ class ParallelDataLoader(object):
         Floating point type to use.
     shuffle
         Whether to shuffle the samples.
-    sampler
-        The sampler to use. Either specify sampler or shuffle, not both.
+    num_batches_for_shuffling
+        The number of batches among which samples are shuffled. So for example if num_batches_for_shuffling = 8
+        then the next num_batches_for_shuffling * 8 samples will be shuffled and then batched.
```
lostella, line 483: If `num_batches_for_shuffling = 8` then `num_batches_for_shuffling * 8 = 64`? Or did you mean `batch_size * 8`?

src/gluonts/dataset/loader.py
```diff
@@ -67,13 +65,14 @@ def __init__(
         dataset: Dataset,
         *,
         transform: Transformation,
+        cyclic: bool,
```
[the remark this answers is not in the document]
AaronSpieler: It fundamentally changes how the dataset is being handled, this shouldn't be just set to some default. Similar to `is_train`; one can set it to a default like False, but probably shouldn't.

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -167,28 +168,84 @@ def _as_in_context(batch: dict, ctx: mx.Context):
...
+    num_batches_for_shuffling: int,
+) -> Iterator[DataEntry]:
+    # Approximate shuffling `num_batches_for_shuffling` probabilistically
+    def skip_data_iter(data: Dataset):
+        for data_entry in data:
+            if (
+                random.randint(1, num_batches_for_shuffling)
+                == num_batches_for_shuffling
+            ):
+                yield data_entry
```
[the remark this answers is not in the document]
AaronSpieler: `p=1/num_batches_for_shuffling ` which leads to `num_batches_for_shuffling` as many time series being considered for a single batch, which is approximately like taking `num_batches_for_shuffling * batch_size` elements, shuffling them, and then returning `batch_size` chunks.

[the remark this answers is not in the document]
AaronSpieler: This method is much much faster than the alternative described above, essentially because one does not need to store temporarily batches, and because if one batch is requested only one is processed and not `num_batches_for_shuffling` times as many.
The difference is like a 50% increase in performance in mp.

src/gluonts/dataset/parallelized_loader.py
AaronSpieler: I should maybe say: if num_batches_for_shuffling = 8 and batch_size=8 then effectively 64 samples will be shuffled and then batched.

src/gluonts/dataset/jsonl.py
```diff
@@ -55,28 +56,43 @@ class JsonLinesFile:
...
+        if not self.cache or (self.cache and not self._data_cache):
+            with open(self.path) as jsonl_file:
+                for line_number, raw in enumerate(jsonl_file):
+                    lower_bound = MPWorkerInfo.worker_id * segment_size
+                    upper_bound = (
+                        (MPWorkerInfo.worker_id + 1) * segment_size
+                        if MPWorkerInfo.worker_id + 1
+                        != MPWorkerInfo.num_workers
+                        else np.inf
                     )
```
lostella, line 79: Maybe it's better to have the exact length of the last segment be apparent here. The way `segment_size` is computed above, you could end up with the last worker having to do almost twice as much work as the other ones. For example, suppose `len(self) == 19` and `num_workers = 4`, then `segment_size = int(19 / 4) = 4`, and you'll have the first three worker process [0, 4), [4, 8) and [8, 12) respectively, and the last one [12, 19) which is 7 elements instead of 4. If all workers synchronize at some point, this may keep some workers idle for too long.

Instead, you can round up the division, and have `segment_size = int(ceiling(19 / 4)) = 5`, and have workers split the load as follows: [0, 5), [5, 10), [10, 15), [15, 19), which seems more balanced. In this case the upper bound here would become:

```python
upper_bound = min(len(self), (MPWorkerInfo.worker_id + 1) * segment_size)
```

vafl: 👍

src/gluonts/dataset/common.py
```diff
@@ -185,26 +186,39 @@ class FileDataset(Dataset):
         Must be a valid Pandas frequency.
     one_dim_target
         Whether to accept only univariate target time series.
+    cache
```
vafl, line 189: Nice.

src/gluonts/dataset/jsonl.py
```diff
@@ -55,28 +56,43 @@ class JsonLinesFile:
...
+                        parsed_line = Line(json.loads(raw), span=span)
+                        if self.cache:
+                            self._data_cache.append(parsed_line)
+                        yield parsed_line
+                    except ValueError:
+                        raise GluonTSDataError(
+                            f"Could not read json line {line_number}, {raw}"
+                        )
+        else:
+            for i in range(len(self._data_cache)):
```
vafl, line 94: maybe just 
```python
yield from self._data_cache
```

src/gluonts/dataset/jsonl.py
vafl: 👍

AaronSpieler: I like this, however this brakes the code for example for `len(dataset)=5` and `num_workers=4`, because the first two workers will get 2 each, the other two nothing, which leads to workers running into an infinite fetch loop and ultimately to a timeout.

As far as I can see the worst overhead a single worker could have is `num_workers-1` sample , which in case for large datasets is negligible, which is the main usecase for mp data leader?

What do you think?

lostella: > As far as I can see the worst overhead a single worker could have is `num_workers-1` sample , which in case for large datasets is negligible, which is the main usecase for mp data leader?

I think the maximum overhead is actually `int(len(dataset) / num_workers) - 1` (see my example above).

Maybe to solve the case you mention, it is sufficient to change how `num_worker` is bounded by `len(dataset)`: currently we assert essentially that `num_worker <= len(dataset)` (if I'm not wrong), we could assert that `num_worker <= len(dataset)/2` or something like that

AaronSpieler: I commited your simplified logic with `min(len(self),...` though. Turns out there was a test with a one off error too.

AaronSpieler: `num_worker <= len(dataset)/2` should indeed work.

`int(len(dataset) / num_workers) - 1` is in fact  ` num_workers - 1` in your example, but its also the wrong formula. 

It should be impossible to get more than ` num_workers - 1` because if it was, then just distribute ` num_workers` time series among the workers, and give the rest to the last, in which case is again bounded by ` num_workers - 1`.

Or am I missing something?

lostella: Yes, the remainder must be in {0, ..., num_workers-1}, that's true. Then I would just replace `else np.inf` with `else len(self)`, right?

AaronSpieler: Yes!, Thx!

src/gluonts/dataset/jsonl.py
```diff
@@ -67,6 +67,8 @@ def __iter__(self):
         # with lower and upper bound, where each worker is assigned one segment
         segment_size = int(len(self) / MPWorkerInfo.num_workers)
 
+        print("dataset_len:, ", len(self))
```
vafl, line 70: remove print

test/dataset/test_multiprocessing_loader.py
```diff
@@ -48,9 +48,12 @@
 CONTEXT_LEN = 7
 SPLITTING_SAMPLE_PROBABILITY = 1  # crucial for the ValidationDataLoader test
 CD_NUM_STEPS = 14
-CD_NUM_TIME_SERIES = 50  # too small and batch test might fail
+CD_NUM_TIME_SERIES = 47  # too small and batch test might fail
```
lostella, line 51: A prime number, good choice ;-)

AaronSpieler: yep, also 47 looks cool

lostella:
I merged upstream changes to let the tests run again, there was a timeout in on test which I want to make sure is not occurring consistently. @AaronSpieler any idea where these time outs come from?

AaronSpieler:
The windows timeout error? IDK: It should be non mp: "You have set `num_workers` to a non zero value, however, currently multiprocessing is not supported on windows and therefore`num_workers will be set to 0."

AaronSpieler:
Should be fixed now. The error came from mp evaluation being enabled on windows. I think this was shadowed by the continual windows errors we were getting so far from other commits.

AaronSpieler:
So windows is still failing, however, it has nothing to do with multiprocessing. So I think we could at some point reenable multiprocessing evaluation for windows, however, how the error is now more clearly visible because its not happening in the forked processes anymore:

```
File "D:\a\gluon-ts\gluon-ts\test\model\conftest.py", line 100, in test_accuracy
    evaluator=Evaluator(calculate_owa=statsmodels is not None),
  File "d:\a\gluon-ts\gluon-ts\src\gluonts\evaluation\backtest.py", line 224, in backtest_metrics
    ts_it, forecast_it, num_series=maybe_len(test_dataset)
  File "d:\a\gluon-ts\gluon-ts\src\gluonts\evaluation\_base.py", line 178, in __call__
    for ts, forecast in it:
  File "c:\hostedtoolcache\windows\python\3.6.8\x64\lib\site-packages\tqdm\std.py", line 1127, in __iter__
    for obj in iterable:
  File "d:\a\gluon-ts\gluon-ts\src\gluonts\model\predictor.py", line 317, in predict
    num_samples=num_samples,
  File "d:\a\gluon-ts\gluon-ts\src\gluonts\model\forecast_generator.py", line 204, in __call__
    outputs = prediction_net(*inputs).asnumpy()
  File "c:\hostedtoolcache\windows\python\3.6.8\x64\lib\site-packages\mxnet\ndarray\ndarray.py", line 2535, in asnumpy
    ctypes.c_size_t(data.size)))
```

PR #745, opened 2020-04-07
change the default value to a scaler instance

yx1215:
*Issue #, if available:*
The default value for the scaler parameter in MLP2QRForecaster, RNN2QRForecaster and CNN2QRForecaster is incorrect. It should be an instance of NOPscaler.

*Description of changes:*
Parenthesis should be added to make it instances of NOPscaler.


lostella:
That's right, thanks!

PR #746, opened 2020-04-07
Categorical Distribution

steverab:
*Description of changes:*
This PR adds a categorical distribution which allows for more flexible binnings being learned. In contrast to the binned distribution, the categorical distribution only consists of a number of probabilities per category and users only need to provide the desired number of bins as part of `CategoricalOutput`.
This distribution is designed to work in conjunction with the upcoming input/output binning strategies (corresponding PR to follow soon), which can pre-process the target into a categorical variable and which can post-process samples from the categorical distribution into values from the original target domain/range.


test/distribution/test_distribution_inference.py
```diff
@@ -808,6 +810,46 @@ def test_binned_likelihood(
     ), f"bin_prob did not match: bin_prob = {bin_prob}, bin_prob_hat = {bin_prob_hat}"
 
 
+@pytest.mark.parametrize("num_cats", [6])
+@pytest.mark.parametrize(
+    "cat_probs", [np.array([0.3, 0.1, 0.05, 0.2, 0.1, 0.25])]
+)
+@pytest.mark.parametrize("hybridize", [False])
```
lostella, line 817: is there anything wrong with `hybridize=True`?

steverab: I copied this part over from the binned distribution. Just tested this with hybridize=True and it seems to work.

PR #755, opened 2020-04-15
Representation module

steverab:
*Description of changes:*

This PR adds a representation module to GluonTS, which enables users to define which kinds of input and output representations they desire. Input representations transform the target values which are fed into the model, while output representations define the labels during loss computation, the samples during prediction, and how to transform these values back to the original domain.

This model now encompasses the MeanScaler and the NOPScaler, but also adds various binning strategies, as well as a hybrid representation which enables users to combine multiple input representations (like binnings with different bin sizes, binning and scaling, etc.). 
In addition to the module itself, this PR also adds some relevant test.

The corresponding changes to make this module work with some of our popular models (DeepAR, WaveNet, Transformer, and SimpleFeedForward) will follow once this PR is approved.


kashif:
Perhaps here it might make sense to have a flag for when the tensors are not `NTC` but rather `NCT` e.g. when the instance splitter you can set `output_NTC=False`, i.e. instead of hard-coding the `dim=1` you might set it to the appropriate dim if this flag is `True` or `False`?

src/gluonts/representation/representation.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        pass
+
+    # noinspection PyMethodOverriding
+    def hybrid_forward(
+        self,
+        F,
+        data: Tensor,
+        observed_indicator: Tensor,
+        scale: Optional[Tensor],
```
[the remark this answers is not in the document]
steverab: In my experiments this did not prove to be an issue, not even with `hybridize=True`.

src/gluonts/representation/representation.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+
+    @validated()
+    def __init__(self, is_output: bool = False):
+        super().__init__()
+        self.is_output = is_output
+
+    def initialize_from_dataset(self, input_dataset: Dataset):
+        r"""
+        Perform some computation based on an entire dataset.
```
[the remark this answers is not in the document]
steverab: That sounds better indeed, fixed!

src/gluonts/representation/representation.py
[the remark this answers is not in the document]
lostella: Yes, I just checked: I think it's only an issue if you explicitly pass in or return a `None`. In this case if the input is not passed then the output is not `None` because the scale gets computed internally, so it's fine.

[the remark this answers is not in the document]
steverab: It also seems to work when explicitly passing `None`. Calling hybrid_forward like this for example works fine in `SimpleFeedForwardEstimator` with both `hybridize=False` and `hybridize=True`:
```
input_tar_repr, scale = self.input_repr(
     past_target, F.ones_like(past_target), None
)
```

src/gluonts/representation/hybrid_representation.py
```diff
@@ -0,0 +1,81 @@
...
+            if self.is_output:
+                representation_data = representation_data.expand_dims(
+                    -1
+                ).swapaxes(1, 2)
+
+            if representation_data_agg is None:
+                representation_data_agg = representation_data
+            else:
+                representation_data_agg = F.concat(
+                    representation_data_agg, representation_data, dim=1
```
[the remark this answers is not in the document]
steverab: That's cleaner indeed, I changed this to your suggestion!

src/gluonts/representation/binning.py
```diff
@@ -0,0 +1,122 @@
...
+        scale: Optional[Tensor],
+    ) -> Tuple[Tensor, Tensor]:
+        # Calculate local scale if scale is not already supplied.
+        if scale is None:
+            scale = F.expand_dims(
+                F.sum(data, axis=-1) / F.sum(observed_indicator, axis=-1), -1
+            )
+            # Clip scale on the bottom to prevent division by zero.
+            scale = F.clip(scale, 1e-20, np.inf)
+        self.scale = scale.asnumpy()
```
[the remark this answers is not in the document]
steverab: Absolutely true, storing the scale is not necessary for `Binning`! Fixed!

src/gluonts/representation/representation.py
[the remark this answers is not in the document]
lostella: Maybe it's a new thing: on MXNet 1.6 the following snippet

```python
import mxnet as mx

class MyNet(mx.gluon.HybridBlock):
    def hybrid_forward(self, F, x, y=None):
        return 2*x, y

net = MyNet()
net.initialize()
net.hybridize()

print(net(
    mx.nd.ones(shape=(10,)),
    None,
))
```

gives me

```
AssertionError: HybridBlock input must be (nested) list of Symbol or NDArray, but got None of type <class 'NoneType'>
```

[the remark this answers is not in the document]
steverab: Correct me if I'm wrong, but I guess this is closer to a minimal working example for what we actually use and this works fine on my machine. Maybe you need to ensure that you actually pass a symbol object when having called `net.hybridize()` before?
```python
import mxnet as mx
from typing import Optional, Union, Tuple

Tensor = Union[mx.nd.NDArray, mx.sym.Symbol]

class MyNet(mx.gluon.HybridBlock):
    def hybrid_forward(self, F, x: Tensor, y: Optional[Tensor]
    ) -> Tuple[Tensor, Tensor]:
        return 2*x, y

net = MyNet()
net.initialize()

hybridize = True
if hybridize:
    F = mx.sym
    net.hybridize()
else:
    F = mx.nd

print(net(F.full(10,1),None))
```

src/gluonts/representation/mean_scaling.py
```diff
@@ -0,0 +1,105 @@
...
+
+        scaled_data = F.broadcast_div(data, scale)
+
+        if self.clip_max != -1:
+            scaled_data = F.clip(scaled_data, -self.clip_max, self.clip_max)
+
+        if self.is_output:
+            return scaled_data, scale
+        else:
+            return F.expand_dims(scaled_data, axis=1), scale
```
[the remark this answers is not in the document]
steverab: I have generalized the models (`WaveNet`, `DeepAR`, `Transformer`, `SimpleFeedForward`) to deal with 3-dimensional tensors. Since binnings require a three dimensional tensor of shape `(batch_size, embedding_size, pred_length)`, I have also adapted the mean scaler to output `(batch_size, 1, pred_length)`-shaped tensors. This prevents us from having to handle different dimensions in the models and abstracts this into the representation. However, tensor expansion is only required in input mode, and not in output mode. See larger comment below for more on this.

src/gluonts/representation/mean_scaling.py
```diff
@@ -0,0 +1,105 @@
...
+        data: Tensor,
+        observed_indicator: Tensor,
+        scale: Optional[Tensor],
+    ) -> Tuple[Tensor, Tensor]:
+        data = F.cast(data, dtype="float32")
+
+        if scale is None:
+            scale = self.compute_scale(F, data, observed_indicator)
+            scale = scale.expand_dims(axis=1)
+        self.means = scale
```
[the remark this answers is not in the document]
steverab: See large comment below.

src/gluonts/representation/global_relative_binning.py
```diff
@@ -0,0 +1,235 @@
...
+            x = F.array(x_np)
+
+        bin_cent = F.array(self.bin_centers_hyb)
+        x_oh = F.one_hot(F.squeeze(x), self.num_bins)
+
+        # Pick corresponding bin centers for all samples
+        x = F.sum(bin_cent * x_oh, axis=1).expand_dims(-1)
+
+        # Transform bin centers back to the oiginal scale
+        x = F.broadcast_mul(F.array(self.scale), x)
```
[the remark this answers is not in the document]
steverab: See large comment below.

steverab:
Thanks for that very detailed feedback @lostella! Let me address your main points:

---

**Splitting up scaler, binner, and embedder**: 

As I'm not too happy with the `is_output` parameter either, I do generally think that separating the embedding into an additional `Emedding` class would be a good idea. This could definitely be done for all the binnings. One problem however, as you already noted in one of your inline comments, arises in the `MeanScaling` setting. Here, we don't embed the data in input mode, but we still need to treat inputs and outputs differently. Specifically, we need to expand the tensor to have the same number of dimensions as the binnings to ensure that models can deal with both representation types flexibly, which is why I decided to stick to the `is_output` parameter for now.

Separating scaler and binner in the way that you proposed might again be possible for some of the binnings (`Binning`, `GlobalRelativeBinning`) and would conceptually separate these two components. However, it will not work for `LocalAbsoluteBinning`. Here, we need to know which data points are observed (i.e. we need to know `observed_indicator`) as each series is binned individually.

---

**Scale storing/passing, statelessness**:

I do think that it is possible to not store the scale as part of the representation, i.e. to return it from `hybrid_forward` (we already do this) and to have it passed to `post_transform`. However, binnings will have members which we cannot pass around (bin centers or bin edges), i.e. completely making the representations stateless is probably not doable without major changes to how models interact with the representations (if you want you can already find the respective implementations for `WaveNet`, `DeepAR`, `Transformer`, `SimpleFeedForward` in [this branch](https://github.com/steverab/gluon-ts/tree/repr_models), PR to follow once we are done with this PR). If removing the scale as a member variable but keeping the bin centers/edges as such is fine, then I can do that.

steverab:
After discussion with @jgasthaus:

---

**Splitting up scaler, binner, and embedder**: 

We could add both an `Embedding` representation and a `DimExpansion` representation to solve the problem of input vs output. It's also not the cleanest of solutions, but probably better than the current boolean indicator. Examples:
- `GlobalRelativeBinning()` would turn into `Embedding(GlobalRelativeBinning())`
- `GlobalRelativeBinning(is_output=True)` would turn into `GlobalRelativeBinning()`
- `MeanScaling()` would turn into `DimExpansion(MeanScaling())`
- `MeanScaling(is_output=True)` would turn into `MeanScaling()`

---

**Scale storing/passing, statelessness**:

Jan agrees that we should try to make the representations stateless as far as possible. I will try to add a third return parameter from the `hybrid_forward`s containing variables that are required in `post_transform`, i.e.

```python
target_repr, scale = self.output_repr(
    target,
    past_observed_values,
    None
)

...

self.output_repr.post_transform(F, samples)
```
would change to
```python
target_repr, scale, pt_params = self.output_repr(
    target,
    past_observed_values,
    None
)

...

self.output_repr.post_transform(F, samples, pt_params)
```

---

Let me know what you think @lostella!

steverab:
@kashif I'll have a look at #790 and see how I can bring those contributions to this module, too!

src/gluonts/representation/dim_expansion.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    A class representing a dimension expansion operation of a given representation along a specified axis.
+
+    Parameters
+    ----------
+    representation
+        The underlying representation.
+    axis
+        Axis on which to expand the tensor.
+        (default: 1)
```
lostella, line 35: -1

src/gluonts/representation/embedding.py
```diff
@@ -0,0 +1,82 @@
...
+    ----------
+    binning
+        The underlying binning.
+    size
+        The desired embedding size.
+        (default: round(num_bins**(1/4)))
+    """
+
+    @validated()
+    def __init__(self, binning: Binning, size: int = -1, *args, **kwargs):
```
lostella, line 47: For optional arguments (size in this case) it’s better to use Optional (from typing import Optional) and default to None

src/gluonts/representation/mean_scaling.py
```diff
@@ -0,0 +1,103 @@
...
+        The maximum value to which re-scaled values will be clipped to. By default, max values will not be clipped.
+        (default: -1)
+    minimum_scale
+        The minimum value to which re-scaled values will be clipped to.
+        (default: 1e-10)
+    """
+
+    @validated()
+    def __init__(
+        self, clip_max: int = -1, scale_min: float = 1e-10, *args, **kwargs
```
lostella, line 44: Shouldn't `clip_max` be a `float`? If it can be left unspecified it's better to use `Optional[float]` and default it to `None`

src/gluonts/representation/dim_expansion.py
steverab: Good catch!

src/gluonts/representation/embedding.py
steverab: Oh yeah right, that's indeed way cleaner!

src/gluonts/representation/mean_scaling.py
steverab: Fixed!

src/gluonts/representation/representation_chain.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class RepresentationChain(Representation):
+    """
+    A class representing a hybrid approach of combining multiple representations into a single representation.
+    Representations will be combined by concatenating them on dim=-1.
+
+    Parameters
+    ----------
+    representations
```
lostella, line 34: This was probably changed in the code:

```suggestion
    chain
```

src/gluonts/representation/custom_binning.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class CustomBinning(Representation):
+    """
+    A class representing binned representations with custom centers.
+
+    Parameters
+    ----------
+    bin_centers
+        The bins to be used to discritize the data.
```
lostella, line 37:
```suggestion
        The bins to be used to discretize the data.
```

src/gluonts/representation/discrete_pit.py
```diff
@@ -0,0 +1,103 @@
...
+        **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+
+        self.num_bins = num_bins
+        self.mlp_transf = mlp_transf
+
+        if embedding_size is None:
+            # Embedding size heuristic that seems to work well in practice. For reference see:
+            # https://developers.googleblog.com/2017/11/introducing-tensorflow-feature-columns.html
```
lostella, line 66: Maybe this reference can go in the docstring directly, so that it's more exposed

src/gluonts/representation/global_relative_binning.py
```diff
@@ -0,0 +1,170 @@
...
+        rep_params: List[Tensor],
+        **kwargs,
+    ) -> Tuple[Tensor, Tensor, List[Tensor]]:
+        bin_edges = kwargs["bin_edges"]
+        bin_centers = kwargs["bin_centers"]
+
+        # Calculate local scale if scale is not already supplied.
+        if scale is None:
+            scale = F.expand_dims(
+                F.sum(data, axis=-1) / F.sum(observed_indicator, axis=-1), -1
```
lostella, line 137: Looks like the scale here is computed as the average observation: however the numerator assumes that unobserved data is simply zero, which may not necessarily be the case. Maybe the numerator could be
```
F.sum(data * observed_indicator, axis=-1)
```
The same holds anywhere else the scale is computed this way

src/gluonts/representation/mean_scaling.py
```diff
@@ -0,0 +1,109 @@
...
+        observed_indicator: Tensor,
+        scale: Optional[Tensor],
+        rep_params: List[Tensor],
+        **kwargs,
+    ) -> Tuple[Tensor, Tensor, List[Tensor]]:
+        data = F.cast(data, dtype="float32")
+
+        if scale is None:
+            scale = self.compute_scale(F, data, observed_indicator)
+            scale = scale.expand_dims(axis=1)
```
lostella, line 96: I'm not sure how this plays out with chaining. This example might seem like it's cherry-picked, but: what happens if you chain two scalers?

src/gluonts/representation/global_relative_binning.py
lostella: Related to this: is it necessary to handle the scaling in the binning components? If I understand correctly the design, binnings use `rep_params` to pass down the information to `post_transform` about how to reconstruct the data in the original domain: couldn't scalers do the same? This way `scale` would disappear from `hybrid_forward` and `post_transform`, and just be included in `rep_params` by the scaler, and the "scaling+binning" could be implemented as a chain of the two.

But that I guess would require some strategy on how to organize the `rep_params` in composed representations: for example, a chain could use some sort of stack.

Do I understand correctly that right now it is assumed that a chain contains at most one binning and that any other representation should not "store" anything in its output `rep_params`?

src/gluonts/representation/representation_chain.py
```diff
@@ -0,0 +1,80 @@
...
+        observed_indicator: Tensor,
+        scale: Optional[Tensor],
+        rep_params: List[Tensor],
+        **kwargs,
+    ) -> Tuple[Tensor, Tensor, List[Tensor]]:
+        for representation in self.chain:
+            data, scale, rep_params = representation(
+                data, observed_indicator, scale, rep_params,
+            )
+        return data, scale, rep_params
```
lostella, line 71: How are the representation parameters of each individual representation in the chain organized in `rep_params` here? I was expecting like `rep_params: List[List]` and the for loop doing something like
```python
for representation, params in zip(self.chain, rep_params):
    data, scale, rep_params = representation(
        data, observed_indicator, scale, params,
    )
```

src/gluonts/representation/custom_binning.py
```diff
@@ -0,0 +1,110 @@
...
+        self,
+        F,
+        data: Tensor,
+        observed_indicator: Tensor,
+        scale: Optional[Tensor],
+        rep_params: List[Tensor],
+        **kwargs,
+    ) -> Tuple[Tensor, Tensor, List[Tensor]]:
+        bin_edges = kwargs["bin_edges"]
+        bin_centers = kwargs["bin_centers"]
```
lostella, line 77: Why are these taken from `kwargs` and not from `self`?

lostella:
I checked in the code and the following seems to work out, but please @steverab check that I'm not missing something. In particular, I hope that the local absolute binning fits this.

The base class and interface could be as follows:

```python
class Representation(nn.HybridBlock):
    @validated()
    def __init__(self):
        super().__init__()

    def initialize_from_dataset(
        self, input_dataset: Dataset, ctx: mx.Context = get_mxnet_context()
    ):
        pass

    def initialize_from_array(
        self, input_array: np.ndarray, ctx: mx.Context = get_mxnet_context()
    ):
        pass

    def hybrid_forward(self, F, data: Tensor, observed_indicator: Tensor) -> Tuple[Tensor, List]:
        return data, []

    def post_transform(self, F, samples: Tensor, rep_params: List) -> Tensor:
        return samples
```

Here:

* `hybrid_forward` takes the data and returns the transformed data along with the list of representation parameters
* `post_transform` takes data in the transformed domain along with the list of representation parameters, and return the data in the original domain

Examples of returned representation parameters:

* Scalers will return a list containing the scale tensor.
* Binning representations will return a list containing bin centers and edges; they may also return the scale in the same list, but that's a matter of taste (I prefer components to do one thing only, if possible, and so use a chain to do scaling+binning).
* Embeddings will not have any exposed parameters, so they return an empty list.
* DiscretePIT also doesn't seem to expose any parameter, so empty list.
* A chain of representation will return the list of representation parameters of all the components (so, a list of lists). In `post_transform`, this can be zipped with the list of representation to anti-transform the samples.

Also note that the `Representation` class above is essentially the "identity" representation, and as a consequence

```python
NOPScaler = Representation
```

I feel like we're almost there

src/gluonts/representation/custom_binning.py
```diff
@@ -63,22 +86,14 @@ def hybrid_forward(
 
         # Discretize the data.
         # Note: Replace this once there is a clean way to do this in MXNet.
-        data_binned = np.digitize(
-            data.asnumpy(), bins=self.bin_edges, right=False
-        )
-
-        data = F.array(data_binned)
+        data = F.Custom(data, bin_edges, op_type="digitize")
```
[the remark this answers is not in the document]
steverab: I did not notice any performance issues in my tests.

src/gluonts/representation/custom_binning.py
steverab: This is where some dark magic happens. I'm not sure how this works exactly, but only the binning parameters in `kwargs` passed via `hybrid_forward` are ensured to be flexible. Accessing the same parameters via self does not have this guarantee.

src/gluonts/representation/representation_chain.py
lostella: Agreed offline to refine this design at a later stage

steverab:
Thanks again for your detailed feedback @lostella. I updated the code to reflect the changes that we agreed upon.

PR #773, opened 2020-04-21
Allow wavenet to output DistributionForecast directly

rshyamsundar:
*Description of changes:*
This PR allows wavenet model to output `DistributionForecast` directly apart from `SampleForecast`. Currently this part works only in NDarray mode; will update it to hybrid mode if this is really useful (added a TODO).

Updated test to include this prediction mode. Currently pickling fails for `DistributionForecast` so running this particular test with `num_workers=0`.

Also running benchmark to see the difference in results if we use `DistributionForecast` instead of samples.


src/gluonts/model/wavenet/_network.py
```diff
@@ -547,59 +586,204 @@ def blow_up(u):
...
+        )
+
+        # get the batch shape right
+        scale = F.broadcast_add(scale, F.zeros_like(feat_static_cat))
+        bin_centers = F.broadcast_add(
+            F.array(self.bin_values), F.zeros_like(feat_static_cat)
+        )
+
+        dist_args = F.log(bin_probs), bin_centers
+        return dist_args, scale
```
lostella, line 789: If `loc` is needed by `DistributionOutput.distribution`, you could return `loc = F.zeros_like(scale)` here

src/gluonts/model/forecast_generator.py
```diff
@@ -118,8 +118,8 @@ def __call__(
                 )
 
             distributions = [
-                self.distr_output.distribution(*u)
-                for u in _extract_instances(outputs)
+                self.distr_output.distribution(distr_args, scale=scale)
+                for distr_args, scale in _extract_instances(outputs)
```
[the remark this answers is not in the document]
rshyamsundar: I initially had `loc` as well (as you notice it in the comment below) but removed while fixing a test and forgot to add it back. I will add that.

`DistributionForecastGenerator` is not used currently anywhere. This change was necessary because if any model wants to directly output distributions, then their prediction network should return all `dist_args`, `scale`, `loc` so that distribution object can be constructed here by passing them to `distr_output.distribution` function.

src/gluonts/model/wavenet/_network.py
rshyamsundar: ya.. I had it like that but removed while fixing a test. Forgot to add it back.

src/gluonts/model/wavenet/_network.py
```diff
@@ -547,59 +586,204 @@ def blow_up(u):
...
+            scale of the time series: (batch_size, 1)
+
+        Returns
+        -------
+        Tuple containing `dist_args`, `loc` and `scale`.
+
+            `dist_args` is a pair of tensors containing logarithm of bin probabilities with shape
+            (batch_size, pred_length, num_bins) and bin values with shape (num_bins,).
+
+            `loc` is a Tensor with shape (batch_size, 1).
```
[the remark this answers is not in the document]
rshyamsundar: It is supposed to!

src/gluonts/model/wavenet/_network.py
```diff
@@ -547,59 +586,204 @@ def blow_up(u):
...
+
+            `scale` is a Tensor with shape (batch_size, 1).
+        """
+
+        bin_probs = self.predict(
```
vafl, line 770: Using a single sample for the autoregressive part here is probably not ideal. 
Two ideas to improve this:

- pass the temperature for drawing samples. Then we could run this with temperature 0 picking the bin with max probability in each step
- run the prediction with more than one sample and consolidate afterwards, by averaging the bin probabilities over the samples. 

The second option is the correct way, but we have to draw more samples. The first is not really correct but may work in practice.

vafl: Mmh, I think this is not what you want. We probably want to use the T=0 for drawing the samples, but T=1 for the distribution output, right?

rshyamsundar: Yes, that is why the [returned `bin_probs`](https://github.com/awslabs/gluon-ts/pull/773/files/3fa3f5c6929af57049ef037d5098fdcd79afebcd#diff-0ca3301b70787528429402b5e0015586R514) are not affected by temperature (T=1 while computing these probs). I also added a [note](https://github.com/awslabs/gluon-ts/pull/773/files/3fa3f5c6929af57049ef037d5098fdcd79afebcd#diff-0ca3301b70787528429402b5e0015586R501) clarifying this.

vafl: Ok, makes sense.

src/gluonts/model/wavenet/_network.py
```diff
@@ -735,7 +734,7 @@ def hybrid_forward(
         past_time_feat: Tensor,
         future_time_feat: Tensor,
         scale: Tensor,
-    ) -> Tuple[Tuple[Tensor, Tensor], Tensor]:
+    ) -> Tuple[Tuple[Tensor, Tensor], Tensor, Tensor]:
```
lostella, line 737: For tuples of homogenous type (but unknown number of elements) we can use ellipsis, see https://docs.python.org/3/library/typing.html#typing.Tuple

```suggestion
    ) -> Tuple[Tuple[Tensor, ...], Tensor, Tensor]:
```

src/gluonts/model/wavenet/_estimator.py
```diff
@@ -175,7 +175,7 @@ def __init__(
         seasonality: Optional[int] = None,
         embedding_dimension: int = 5,
         num_bins: int = 1024,
-        bin_centers: Optional[List[float]] = None,
+        bin_centers: Optional[np.ndarray] = None,
```
lostella: This could be more general: `Optional[Collection[float]]` and then you cast it to `np.array` further below

PR #775, opened 2020-04-21
New preliminary values for MQDNN.

AaronSpieler:
*Description of changes:*
Updated Evaluations according to the changes in #668 


AaronSpieler:
@lovvge These are the results of the tests I just ran.
There are some significant improvements for the MQRNN for some reason. Otherwise it's a mixed bag.

I think I will work tomorrow on creating a proper evaluations script with regression tests to get this right once and for all.

AaronSpieler:
> @lovvge These are the results of the tests I just ran.
> There are some significant improvements for the MQRNN for some reason. Otherwise it's a mixed bag.
> 
> I think I will work tomorrow on creating a proper evaluations script with regression tests to get this right once and for all.

I suspect that the relatively few epochs and the single run just have a too high variance in the results.

AaronSpieler:
> @AaronSpieler if these numbers were obtained with the final code in #668 (at least for what concerns the models network and transformation) I think this can be merged too

Im pretty sure its out of date.

Issue #384, opened 2019-10-11
Input Data-Format Definitions


jaheba:
## Goal

At the moment we have no way to define what data-format a model supports.

We also assume that all algorithm use similar fields and that their datatypes are identical. This has the advantage that a single dataset can be used across multiple algorithms.

Still, to achieve more flexibility, it is desirable to have an explicit way to define what data-format a given model has. In other words, we want to decouple the user's input data-format from the model's input data-format.
 
### Field names

In GluonTS we use generic names to describe the function of a given field. For example, we use `target` to describe the time-series which should be predicted into the future.

However, in concrete cases it makes sense to have more descriptive naming. E.g., when dealing with sales-data, one wants to predict the `sales` for a product and thus use such a name for the `target`-field. Similarly, the price could be encoded as such as a dynamic feature:

```json
{"sales": [10, 12, 11, "..."], "price": [5.99, 5.49, 5.99, "..."]}
```

### TC vs CT layout

Two-dimensional data can be layout in two ways. With respect to time-series, we can either use a `TC`-layout were for every time-point we have multiple values:

```
xs = [(1a, 1b), (2a, 2b), ...]
```

Or use a `CT`-layout where for every "channel" we have independent time-series:

```
xs = ([1a, 2a, ...], [1b, 2b, ...])
```

In some contexts one layout might feel more natural than the other. Still, they are exchangeable. 

## Proposal

Each algorithm defines a data-model it requires. We offer type-specifiers to describe what kind of data-layout the algorithm expects:

* `C`: a static (time independent) array
* `T`: a single time-series
* `TC`: a time series, containing multiple values at each time-point
* `CT`: a time-series for each channel

For example, DeepAR could define such a model like this:

```python
class DeepARModel:
    target: T[float]
    feat_static_cat: Optional[C]
    feat_dynamic_real: Optional[CT[float]]
```

Then, given a dataset, we use a second model to describe the layout of the dataset:

```python
class MyDataModel:
    sales: T[int]
    price: T[float]

    name_mapping = {
        "sales": "target",
        "price": "dynamic_feat"
    }
```

We define transformations for `T -> {TC, CT}`, `TC -> {CT}` and `CT -> {TC}`:

```python
def ct_to_tc(xs):
    return xs.transpose()

def tc_to_ct(xs):
    return xs.transpose()

def t_to_tc(xs):
    return xs.reshape((-1, 1))

def t_to_ct(xs):
    return xs.reshape((1, -1))


[...]

PR #776, opened 2020-04-21
Introduce schema for FileDataset

vafl:
This change makes the Datasets more generic, by supporting any fields and types.
It introduces a schema for Datasets that can be inferred from the data or specified.
Also introduces support for gzip json files and fallback from ujson to json.


jaheba:
I think this is similar to what I had worked on before.

https://github.com/awslabs/gluon-ts/pull/650 and https://github.com/awslabs/gluon-ts/issues/384

I would argue for discussing the designs first, before going with this or any other approach.

vafl:
> I think this is similar to what I had worked on before.
> 
> #650 and #384
> 
> I would argue for discussing the designs first, before going with this or any other approach.

There is some overlap, but I'm trying to solve a simpler problem: loading fields other than the standard fields defined in `process`. So it is not about any layout questions. Let me know if you have other ideas on this.

Issue #294, opened 2019-09-09
Bringing Point Processes into GluonTS


canerturkmen:
Hello everyone. 🎉 

[Here](https://github.com/awslabs/gluon-ts/compare/master...canerturkmen:rmtpp) you may find the branch I'm working on to include capability for working with temporal point processes in GluonTS. It has a few days of extra work left to complete the sampler, and potentially throw in a tutorial, but the design as I see it is in place.

So integrating ideas from TPP to GluonTS meant a bit of work, and extending some of the core facilities the library provides. 

The main difficulty is that TPP realizations are "ragged" tensors, i.e., observations occur in a fixed length interval in the real line and have "variable length." This is why the Tensor objects passed around are ragged, and common arguments such as `prediction_length` and `context_length` are void. I tried, to the best of my ability, to conform to the GluonTS API while allowing TPP-type data structures. This meant extending some of GluonTS's common interfaces. 

Specifically,
- A new `Forecast` object is there, which allows for forecasts to be available as ragged tensors.
- The `DataLoader` interfaces are extended to include logic on `stack`ing ragged tensors. Particularly, this means extending the `BatchBuffer` object, and we now have a `VariableLengthBatchBuffer`.
- A new transform, the continuous-time analogue of `InstanceSplitter`, had to go in there. `ContinuousTimeInstanceSplitter` (extending `FlatMapTransform`) samples sub-intervals of the real line and produces ragged tensors which include the point observations. This introduces a dependency on a `ContinuousTimePointSampler` object, which we also define an interface for.
- Extending `GluonPredictor`, The `PointProcessGluonPredictor` is again aware of point process-style data structures and yields `PointProcessSampleForecast`s.
- Finally, of course, we have estimator and network objects that diverge from the GluonTS API sparingly.

#### Some Other Issues

- Hybridization, unfortunately at this stage, is not an option. This is mainly driven by the fact that our implementation depends on the `LSTMCell` facility rather than `LSTM` (the latter cannot work with ragged sequences), and has to be unrolled *explicitly*---with the length of unrolling known ahead of time. 
- The univariate Hawkes process will wait a little since the [operator](https://github.com/apache/incubator-mxnet/pull/14683) was released as part of MXNet 1.5.0, which we don't seem to allow at the moment. The current model is the [RMTPP](https://dl.acm.org/citation.cfm?id=2939875).
- I did not try extra hard at this point for "representability," though it would be great if you could point out what's missing on that front.

[...]

PR #780, opened 2020-04-22
Add variable length functionality to DataLoaders

canerturkmen:
*Issue #, if available:* Related to #294 

*Description of changes:*

Currently data loaders assume that arrays (e.g., that come out of a splitter type transform) are easily stackable into data batches. This is not convenient when working with time series of variable length that a model may want to see, such as when working with temporal point processes.

Here, we add variable-length awareness to the `batchify` and `stack` functions in the parallelized loader. We also expose the batchify function as a constructor argument to `ParallelDataLoader`. A consumer of DataLoaders can now specify `DataLoader(..., variable_length=True)` in order to make the loader aware that some input fields may not be stackable. In this case, the loader pads the arrays on the right along the first dimension (this is a hardcoded assumption for now) before stacking them. By default, the loader falls back to business as usual. 

### DIFF
--- a/src/gluonts/dataset/parallelized_loader.py
+++ b/src/gluonts/dataset/parallelized_loader.py
@@ -14,6 +14,7 @@
 
 # Standard library imports
 import collections
+import functools
 import itertools
 import logging
 import pathlib
@@ -25,7 +26,7 @@ import time
 
 from collections.abc import Sized
 from multiprocessing.managers import SyncManager
-from typing import Callable, Iterable, Optional, List, Iterator
+from typing import Callable, Iterable, Optional, List, Iterator, Union, Any
 
 import multiprocessing
 import multiprocessing.queues
@@ -84,16 +85,73 @@ else:
 ForkingPickler.register(nd.NDArray, reduce_ndarray)
 
 
+def _is_stackable(
+    arrays: List[Union[np.ndarray, mx.nd.NDArray, Any]],
+) -> bool:
+    """
+    Check if elements are scalars, have too few dimensions, or their
+    time axes have equal length; i.e. they are directly `stack` able.
+    """
+    if isinstance(arrays[0], (mx.nd.NDArray, np.ndarray)):
+        s = set(arr.shape[0] for arr in arrays)
+        return len(s) <= 1
+    return True
+
+
+def _pad_arrays(
+    data: List[Union[np.ndarray, mx.nd.NDArray]], pad_axis: int = 0,
+) -> List[Union[np.ndarray, mx.nd.NDArray]]:
+    assert isinstance(data[0], (np.ndarray, mx.nd.NDArray))
+
+    max_len = functools.reduce(max, (x.shape[pad_axis] for x in data))
+    arr_lib = np if isinstance(data[0], np.ndarray) else mx.nd
+
+    padded_data = []
+
+    for x in data:
+        pad_size = max_len - x.shape[pad_axis]
+
+        pad_lengths = [(0, 0)] * x.ndim
+        pad_lengths[pad_axis] = (0, pad_size)
+
+        padded_data.append(
+            arr_lib.pad(x, mode="constant", pad_width=pad_lengths)
+        )
+
+    return padded_data
+
+
 def stack(
     data,
     multi_processing: bool,
     dtype: DType,
     single_process_ctx: Optional[mx.Context] = None,
+    variable_length: bool = False,
 ):
-    """Stack a list of data.
-        Used when creating a single batch from list of dicts
-        depending on whether multiprocessing is turned on, the batches will be
-        constructed using different memory allocation techniques"""
+    """
+    Stack a list of data. Used when creating a single batch from list of dicts
+    depending on whether multiprocessing is turned on, the batches will be
+    constructed using different memory allocation techniques. If `variable_length`
+    is specified, the data will be 'padded' with zeros along the first axis.
+
+    Parameters
+    ----------
+    data: List
+        Lists of array-like, stacked into data batches and loaded to appropriate
+        memory (according to whether `multi_processing` is specified).
+    multi_processing: bool
+        If True, data will be loaded to mxnet ndarrays on shared CPU memory.
+    dtype: DType
+    single_process_ctx: Optional[mx.Context]
+    variable_length: bool
+        If True, the function will check if the list of data are "stackable",
+        i.e., they have matching axes. If not, it will assume that the first
+        dimension of each array is heterogeneous (i.e., ragged) and will pad
+        this axis before stacking.
+    """
+    if variable_length and not _is_stackable(data):
+        data = _pad_arrays(data, pad_axis=0)
+
     if isinstance(data[0], mx.nd.NDArray):
         if multi_processing:
             out = nd.empty(
@@ -131,11 +189,12 @@ def stack(
     return data
 
 
-def _batchify_fn(
+def batchify(
     data: List[dict],
     dtype: DType,
     multi_processing: bool,
     single_process_ctx: Optional[mx.Context] = None,
+    variable_length: bool = False,
 ) -> DataBatch:
     """reduce the list of dictionaries to a single dictionary, where values
         referenced by identical key are reduced using the stack function"""
@@ -145,6 +204,7 @@ def _batchify_fn(
             multi_processing=multi_processing,
             dtype=dtype,
             single_process_ctx=single_process_ctx,
+            variable_length=variable_length,
         )
         for key in data[0].keys()
     }
@@ -504,6 +564,7 @@ class ParallelDataLoader(object):
         ctx: mx.Context,
         dtype: DType = np.float32,
         shuffle: bool = False,
+        batchify_fn: Callable = batchify,
         num_batches_for_shuffling: Optional[int] = None,
         num_prefetch: Optional[int] = None,
         num_workers: Optional[int] = None,
@@ -553,6 +614,7 @@ class ParallelDataLoader(object):
         self.is_train = is_train
         self.batch_size = batch_size
         self.ctx = ctx
+        self.batchify_fn = batchify_fn
 
         self.dtype = dtype
         self.shuffle = shuffle
@@ -634,7 +696,7 @@ class ParallelDataLoader(object):
                         return
 
                     # make them into a single batch
-                    batch = _batchify_fn(
+                    batch = self.batchify_fn(
                         data=batch_samples,
                         multi_processing=False,
                         dtype=self.dtype,
@@ -657,7 +719,7 @@ class ParallelDataLoader(object):
                     batch_size=self.batch_size,
                     shuffle=self.shuffle,
                     num_batches_for_shuffling=self.num_batches_for_shuffling,
-                    batchify_fn=_batchify_fn,
+                    batchify_fn=self.batchify_fn,
                     dtype=self.dtype,
                     ctx=self.ctx,
                     is_train=self.is_train,
--- a/src/gluonts/dataset/parallelized_loader.py
+++ b/src/gluonts/dataset/parallelized_loader.py
@@ -86,27 +86,27 @@ ForkingPickler.register(nd.NDArray, reduce_ndarray)
 
 
 def _is_stackable(
-    arrays: List[Union[np.ndarray, mx.nd.NDArray, Any]],
+    arrays: List[Union[np.ndarray, mx.nd.NDArray, Any]], axis: int = 0,
 ) -> bool:
     """
     Check if elements are scalars, have too few dimensions, or their
-    time axes have equal length; i.e. they are directly `stack` able.
+    target axes have equal length; i.e. they are directly `stack` able.
     """
     if isinstance(arrays[0], (mx.nd.NDArray, np.ndarray)):
-        s = set(arr.shape[0] for arr in arrays)
-        return len(s) <= 1 and arrays[0].shape[0] != 0
+        s = set(arr.shape[axis] for arr in arrays)
+        return len(s) <= 1 and arrays[0].shape[axis] != 0
     return True
 
 
 def _pad_arrays(
-    data: List[Union[np.ndarray, mx.nd.NDArray]], pad_axis: int = 0,
+    data: List[Union[np.ndarray, mx.nd.NDArray]], axis: int = 0,
 ) -> List[Union[np.ndarray, mx.nd.NDArray]]:
     assert isinstance(data[0], (np.ndarray, mx.nd.NDArray))
     is_mx = isinstance(data[0], mx.nd.NDArray)
 
     # MxNet causes a segfault when persisting 0-length arrays. As such,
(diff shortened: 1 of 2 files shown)


canerturkmen:
Test fails are likely related to the GPU. @AaronSpieler  can you maybe take a look when you have time? Tests run fine locally.

AaronSpieler:
> Test fails are likely related to the GPU. @AaronSpieler can you maybe take a look when you have time? Tests run fine locally.

So MXNet cant handle pickling empty arrays, could this be the problem?

AaronSpieler:
In general: I'm not sure whether this is the place to do this. As our codebase currently is, I would suspects its the `InstanceSplitter`, since it has also functionality like `pick_incomplete` in which case time series are also padded. Putting this code here leads to a preprocessing step happening outside the Transformation Chain.

I would suggest to consider for your use case using the `InstanceSplitter`: https://gluon-ts.mxnet.io/api/gluonts/gluonts.transform.html?highlight=instancesplitter#gluonts.transform.InstanceSplitter and 
1) extending the functionality you need there if it't not there. 
2) Then, defining a transformation chain with the splitter
The Transformation Chain is always executed in the DataLoader.

What do you think @canerturkmen ?

canerturkmen:
> > Test fails are likely related to the GPU. @AaronSpieler can you maybe take a look when you have time? Tests run fine locally.
> 
> So MXNet cant handle pickling empty arrays, could this be the problem?

Yes this appears to be the case, ends up in (I think) a segfault. Bad thing is, this is a valid edge case.

canerturkmen:
> In general: I'm not sure whether this is the place to do this. As our codebase currently is, I would suspects its the `InstanceSplitter`, since it has also functionality like `pick_incomplete` in which case time series are also padded. Putting this code here leads to a preprocessing step happening outside the Transformation Chain.
> 
> I would suggest to consider for your use case using the `InstanceSplitter`: https://gluon-ts.mxnet.io/api/gluonts/gluonts.transform.html?highlight=instancesplitter#gluonts.transform.InstanceSplitter and
> 
>     1. extending the functionality you need there if it't not there.
> 
>     2. Then, defining a transformation chain with the splitter
>        The Transformation Chain is always executed in the DataLoader.
> 
> 
> What do you think @canerturkmen ?

So this was tricky in the original contribution as well. Unfortunately, the padding is a function of the batch itself, which materializes only through the data loader. So cramming it into a transformer would, in my mind, be the same as introducing a "batch transformer," or imitate "batching" in a transformer. Our transformers are designed to operate on logical `DataEntry`s so I think we should have it in the loader.

This is a good discussion though, thanks for bringing this up @AaronSpieler . What do you think @lostella?

AaronSpieler:
> > > Test fails are likely related to the GPU. @AaronSpieler can you maybe take a look when you have time? Tests run fine locally.
> > 
> > 
> > So MXNet cant handle pickling empty arrays, could this be the problem?
> 
> Yes this appears to be the case, ends up in (I think) a segfault. Bad thing is, this is a valid edge case.

I ended up converting those matrices to `numpy`, but in your use-case this probably doesn't work.

AaronSpieler:
> "batch transformer," or imitate "batching"

Well, wouldn't then the current `InstanceSplitter` already be a  "batch transformer" since it's padding sequences to a common length, which is exactly your use case as I understand?

canerturkmen:
1. For the 0-length issue, I'll probably just have a dummy array and handle it downstream. Ugly, but necessary and should be easy to handle.
2. On the splitter issue. Yes, but two key things are different here:
   - the `InstanceSplitter` pads for the *entire data set,* here pad length is a function of the *batch* itself. 
   - the `InstanceSplitter` knows the lengths of arrays ahead of time (indeed, it's a parameter). Here the lengths are variable (from the `ContinuousTimeInstanceSplitter`).

A potential workaround is to have a PadEverythingSplitter which would be wasteful in two ways.
- It would need to fully scan the data set to know the lengths of everyone, which goes against the spirit of our Datasets. I understand from the current design that transformers should not be picky about which type of dataset they work on (so, not necessarily a `ListDataset`), and be able to handle iterators. A PadEverythingSplitter maybe would have to exhaust the data set to decide which length to pad to.
- It could be potentially very wasteful with memory, up to 2x, 3x more footprint. Given the GPU constraint this blocks other nice things downstream. Our TPPs will be multivariate, so very demanding with memory (and hence the initial thinking).

AaronSpieler:
> 1. For the 0-length issue, I'll probably just have a dummy array and handle it downstream. Ugly, but necessary and should be easy to handle.
> 2. On the splitter issue. Yes, but two key things are different here:
>    
>    * the `InstanceSplitter` pads for the _entire data set,_ here pad length is a function of the _batch_ itself.
>    * the `InstanceSplitter` knows the lengths of arrays ahead of time (indeed, it's a parameter). Here the lengths are variable (from the `ContinuousTimeInstanceSplitter`).
> 
> A potential workaround is to have a `PadEverythingTransformer` which would be wasteful in two ways.
> 
> * It would need to fully scan the data set to know the lengths of everyone, which goes against the spirit of our Datasets. I understand from the current design that transformers should not be picky about which type of dataset they work on (so, not necessarily a `ListDataset`), and be able to handle iterators. A `PadEverythingTransformer` maybe would have to exhaust the data set to decide which length to pad to.
> * It could be potentially very wasteful with memory, up to 2x, 3x more footprint. Given the GPU constraint this blocks other nice things downstream. Our TPPs will be multivariate, so very demanding with memory (and hence the initial thinking).
 
* Regarding 1: yeah, probably the easiest fix

* Regarding 2: 
Does this mean that individual batches can have different length of dimension 1:

"the InstanceSplitter pads for the entire data set, here pad length is a function of the batch itself."

How would setting a `max_length` affect the implementation, and would it be sensible? So to say to always pad to this length, even if the longest time series in the batch is not that long.

canerturkmen:
@AaronSpieler Yes. Individual batches can have different lengths. Unfortunately, `max_length` in this case is not really an option both on account of the memory constraint and more importantly it is near-impossible to set intuitively.

AaronSpieler:
> @AaronSpieler Yes. Individual batches can have different lengths. Unfortunately, `max_length` in this case is not really an option both on account of the memory constraint and more importantly it is near-impossible to set intuitively.

I see, Thx for the clarification!

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -84,16 +85,77 @@ def reduce_ndarray(data):
 ForkingPickler.register(nd.NDArray, reduce_ndarray)
 
 
+def _is_stackable(
+    arrays: List[Union[np.ndarray, mx.nd.NDArray, Any]],
```
AaronSpieler, line 89: Is there any point in having this Union, if it contains Any?

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -84,16 +85,77 @@ def reduce_ndarray(data):
...
+    assert isinstance(data[0], (np.ndarray, mx.nd.NDArray))
+    is_mx = isinstance(data[0], mx.nd.NDArray)
+
+    # MxNet causes a segfault when persisting 0-length arrays. As such,
+    # we add a dummy pad of length 1 to 0-length dims.
+    max_len = max(1, functools.reduce(max, (x.shape[pad_axis] for x in data)))
+    padded_data = []
+
+    for x in data:
+        # MxNet lacks the functionality to pad n-D arrays consistently.
```
AaronSpieler, line 113: So `mxnet.ndarray.pad`: https://beta.mxnet.io/api/ndarray/_autogen/mxnet.ndarray.pad.html is not an option here?

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -84,16 +85,77 @@ def reduce_ndarray(data):
...
+    return True
+
+
+def _pad_arrays(
+    data: List[Union[np.ndarray, mx.nd.NDArray]], pad_axis: int = 0,
+) -> List[Union[np.ndarray, mx.nd.NDArray]]:
+    assert isinstance(data[0], (np.ndarray, mx.nd.NDArray))
+    is_mx = isinstance(data[0], mx.nd.NDArray)
+
+    # MxNet causes a segfault when persisting 0-length arrays. As such,
```
AaronSpieler, line 107: So I'm wondering, why does it come to this, i.e. why would we want to persist 0-length arrays? Can we not just pad the whole length with 0s or whatever the padding value is? 
What about removing that entry from the batch?

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -84,16 +85,77 @@ def reduce_ndarray(data):
...
+    # we add a dummy pad of length 1 to 0-length dims.
+    max_len = max(1, functools.reduce(max, (x.shape[pad_axis] for x in data)))
+    padded_data = []
+
+    for x in data:
+        # MxNet lacks the functionality to pad n-D arrays consistently.
+        # We fall back to numpy if x is an mx.nd.NDArray.
```
AaronSpieler, line 114: So this could mean a lot of overhead, we should be extra sure that there is no way of padding MXNet arrays properly.

src/gluonts/dataset/parallelized_loader.py
canerturkmen: If I put `int`, `float` instead of `Any`, mypy complains despite the isinstance() check below. So, those are just meant to be "type hints" I guess.

src/gluonts/dataset/parallelized_loader.py
canerturkmen: From the docs:
```
Current implementation only supports 4D and 5D input arrays with padding applied only on axes 1, 2 and 3. Expects axes 4 and 5 in pad_width to be zero.
```
🤷‍♂️

src/gluonts/dataset/parallelized_loader.py
canerturkmen: Persisting is not really relevant here. Under normal circumstances we do pad 0-length arrays to the appropriate length. This line only kicks in when the *entire batch* has 0-length, in which case we pad to a dummy dimensionality of 1.

Unfortunately the entry is valid data. You want to know if you observed nothing--can't kick it out of the batch.

src/gluonts/dataset/parallelized_loader.py
AaronSpieler: Ok, super weird.

src/gluonts/dataset/parallelized_loader.py
AaronSpieler: I see.

vafl:
Perhaps we can expose the batch / stack function as a parameter of the Trainer and pass it along? Then you could plug in whatever you want. For inference if we use batch_size=1 it should work without changes, or we can also pass this to a predictor somehow.

src/gluonts/dataset/parallelized_loader.py
AaronSpieler: Can you please test whether the workaround described here: https://mxnet.apache.org/versions/1.4.1/tutorials/gluon/gotchas_numpy_in_mxnet.html
makes sense?

canerturkmen:
> Perhaps we can expose the batch / stack function as a parameter of the Trainer and pass it along? Then you could plug in whatever you want. For inference if we use batch_size=1 it should work without changes, or we can also pass this to a predictor somehow.

Thanks for popping in @vafl. I'm not sure I fully understand but I'll try to comment nevertheless :) Both `batchify` and `stack` make sense to me in the context of loaders, and best remain within the loader module boundaries.  With this change, the loader exposes the batching function as a constructor argument,
```python
TrainDataLoader(
                num_batches_per_epoch=22, 
                batchify_fn=partial(batchify, variable_length=True), 
                **kwargs
            )
```
I understand our Trainers do not know about TrainDataLoaders (don't keep references to them in their state), and only get them as call arguments. It's the Estimator that calls the loader's constructor, and passes it to the trainer. Do you think Estimators should expose batchify functions?

vafl:
> > Perhaps we can expose the batch / stack function as a parameter of the Trainer and pass it along? Then you could plug in whatever you want. For inference if we use batch_size=1 it should work without changes, or we can also pass this to a predictor somehow.
> 
> Thanks for popping in @vafl. I'm not sure I fully understand but I'll try to comment nevertheless :) Both `batchify` and `stack` make sense to me in the context of loaders, and best remain within the loader module boundaries. With this change, the loader exposes the batching function as a constructor argument,
> 
> ```python
> TrainDataLoader(
>                 num_batches_per_epoch=22, 
>                 batchify_fn=partial(batchify, variable_length=True), 
>                 **kwargs
>             )
> ```
> 
> I understand our Trainers do not know about TrainDataLoaders (don't keep references to them in their state), and only get them as call arguments. It's the Estimator that calls the loader's constructor, and passes it to the trainer. Do you think Estimators should expose batchify functions?

Sorry, yeah. The trainer is not the right place.

What I meant in general was whether we can make this more flexible and configurable from the outside. So for example, we could make `stack_fn` a parameter of `batchify` with the default being the standard `stack` and then you could tie that to another stack function in your estimator. This is just an idea. I think your change also makes sense as is.

src/gluonts/dataset/parallelized_loader.py
canerturkmen: So I implemented this, see latest commit `c10ded2`.

Tests pass locally, but still fail with a segfault (?). If you also can't spot the problem in the code, I suggest we go back to the old solution. This should only be used by TPPs for a long time, and we can try to work around it.

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -84,16 +85,77 @@ def reduce_ndarray(data):
 ForkingPickler.register(nd.NDArray, reduce_ndarray)
 
 
+def _is_stackable(
+    arrays: List[Union[np.ndarray, mx.nd.NDArray, Any]],
+) -> bool:
+    """
+    Check if elements are scalars, have too few dimensions, or their
+    time axes have equal length; i.e. they are directly `stack` able.
```
lostella, line 93: Just to align this with the padding function that follows: the axis being checked could be an (optional) argument

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -84,16 +85,77 @@ def reduce_ndarray(data):
...
+    """
+    if isinstance(arrays[0], (mx.nd.NDArray, np.ndarray)):
+        s = set(arr.shape[0] for arr in arrays)
+        return len(s) <= 1 and arrays[0].shape[0] != 0
+    return True
+
+
+def _pad_arrays(
+    data: List[Union[np.ndarray, mx.nd.NDArray]], pad_axis: int = 0,
```
lostella, line 102: Maybe just “axis” instead of “pad_axis”: more like mxnet B-) and less redundant

src/gluonts/dataset/parallelized_loader.py
AaronSpieler: I see, then we'll have to look into it some other time. Anyways, thx for trying!

Issue #379, opened 2019-10-10
Regression tests


timoschowski:
## Description
For the 0.4 release, I propose to add regression test coverage so that we are sure not to deteriorate accuracy on any of the datasets that are listed in publications.  For example, there recently seems to have occurred a regression in the M4 data sets and GluonTS should be protected against this.

jaheba:
I think we could use nightly builds and GitHub actions for that.

Also, I think we should store the results to have data over time.

vafl:
Yes, nightly builds is a good way.
@AaronSpieler 's work on sagemaker framework should make it much simpler to write these tests.

jaheba:
Since this is not blocking for 0.4, let's punt it to 0.5.

lostella:
I'm removing this one from milestones, since this is about testing

AaronSpieler:
> Yes, nightly builds is a good way.
> @AaronSpieler 's work on sagemaker framework should make it much simpler to write these tests.

So the SagemakerSDK including GluonTSFramework is now  done.
However, they require credentials to launch jobs, so we would need to think about how this can work with GitHub actions.

AaronSpieler:
I'm working on this now.

PR #782, opened 2020-04-22
Run unstable tests in separate environments

arangatang:
*Issue #, if available:*764

*Description of changes:*
This commit splits up the troublesome tests of models into two parts, one stable and one unstable part. 
The unstable part of the tests are moved to a new file and are run during a seperate test execution.

The unstable parts of the tests are moved to the folder:

`/tests/model/unstable_tests/test_name.py`

for example the troublesome parts of the file:

`tests/model/deepstate/test_model.py`

is moved to:

`/tests/model/unstable_tests/test_deepstate.py`

In order to run these tests after a pull request ive modified the file:

`.github/workflows/tests.yml`

More specifically the change made should allow for the tests to run in parallel. However since this file is run on GitHub itself I have not been able to try this locally. I did however try partial changes on a private github repo.

The change

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         python-version: [3.6, 3.7]
         platform: [ubuntu-latest, windows-latest]
-        pytest-parameters: [--ignore=test/model/unstable_tests, test/model/unstable_tests]
+        pytest-parameters: [test --ignore=test/model/unstable_tests, test/model/unstable_tests]
     runs-on: ${{ matrix.platform }}
 
     steps:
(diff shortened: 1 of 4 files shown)


lostella:
Hey @arangatang the style checks are failing: can you make sure to run

```
pip install -r requirements/requirements-setup.txt
./dev_setup.sh
```

This will install black and mypy, and set some pre-commit hooks which will make sure you don't commit type-unsafe or not-properly-styled code. To fix the style in this PR, you can just run

```
black src test
```

and commit & push the changes.

arangatang:
Will do, however I am going to redo the commit with another approach before that though as I need to add regression tests for the accuracy instead of quarantining the flaky values.

arangatang:
I believe we can close this pull request as the proposed solution is to be replaced by #379

PR #787, opened 2020-04-23
fix bug with iteration over datasets

vafl:
Fixes a bug, where we keep iterating over the same data with each epoch instead of continuing from the last point. Also some minor improvements to the `ListDataset`.

FYI, this bug was introduced a few month ago (before the parallel data loader), so training with longer datasets has since then only used the first `num_batches_per_epoch x batch_size` entries.

### DIFF
--- a/src/gluonts/dataset/parallelized_loader.py
+++ b/src/gluonts/dataset/parallelized_loader.py
@@ -175,15 +175,6 @@ def _sequential_sample_generator(
     cyclic: bool,
     num_batches_for_shuffling: int,
 ) -> Iterator[DataEntry]:
-    # Approximate shuffling `num_batches_for_shuffling` probabilistically
-    def skip_data_iter(data: Dataset):
-        for data_entry in data:
-            if (
-                random.randint(1, num_batches_for_shuffling)
-                == num_batches_for_shuffling
-            ):
-                yield data_entry
-
     # sanity check: since `num_batches_for_shuffling` works by skipping
     # entries, this should not be used for non cyclic datasets
     assert (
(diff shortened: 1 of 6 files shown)


test/dataset/test_multiprocessing_loader.py
```diff
@@ -191,6 +202,81 @@ def test_validation_loader_equivalence() -> None:
...
+        {
+            "start": "2012-01-01",
+            "target": np.random.uniform(size=40).astype(float).tolist(),
+            "item_id": i,
+        }
+        for i in range(batch_size * num_batches_per_epoch * X)
+    ]
+
+    if num_workers:
+        assert len(simple_data) % num_workers == 0
```
lostella, line 225: Why is this required? I guess it should work in any case, right?

vafl: removed it

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -239,29 +239,16 @@ def _sequential_sample_generator(
...
     ), "Setting num_batches_for_shuffling >= 1 only makes sense in the context of cyclic datasets currently."
 
     while True:
-        for sample in transformation(
-            data_it=skip_data_iter(dataset)
-            if num_batches_for_shuffling > 1
-            else dataset,
-            is_train=is_train,
-        ):
-            yield sample
```
lostella, line 264: So I think this was put here to have a way of shuffling an iterable: this change removes this completely I suppose? If so, `num_batches_for_shuffling` is maybe not required anymore

vafl: Yes, let's maybe remove this for now. There are some issues with performance when using this and I think our assumption has always been training datasets should be shuffled. I think we could reintroduce this in a simpler way, by keeping a longer batch of data and shuffling this.

AaronSpieler: Yes, we also saw some performance issues with this in MQCNN, the read speeds are just to slow. Greater batch size should work.

Issue #784, opened 2020-04-23
Seasonal Naive method is not available


alexhallam:
## Description
Seasonal Naive method is not available

## To Reproduce
```python
import pandas as pd 
import numpy as np 
import mxnet as mx
from mxnet import gluon
from gluonts.dataset import common
from gluonts.model.seasonal_naive import SeasonalNaiveEstimator
from gluonts.trainer import Trainer

N = 10  # number of time series
T = 100  # number of timesteps
prediction_length = 24
custom_dataset = np.random.normal(size=(N, T))
start = pd.Timestamp("01-01-2019", freq='1H')  # can be different for each time series
print(custom_dataset.shape)
print(custom_dataset)

# train dataset: cut the last window of length "prediction_length", add "target" and "start" fields
train_ds = [{'target': x, 'start': start} for x in custom_dataset[:, :-prediction_length]]
# test dataset: use the whole dataset, add "target" and "start" fields
test_ds = [{'target': x, 'start': start} for x in custom_dataset]


estimator = SeasonalNaiveEstimator(
    prediction_length=prediction_length,
    context_length=100,
    freq='1H',
    trainer=Trainer(ctx="cpu")
)
```

## Error message or code output

```
from gluonts.model.seasonal_naive import SeasonalNaiveEstimator
ModuleNotFoundError: No module named 'gluonts.model.seasonal_naive'
```

## Environment
- Operating system: Windows
- Python version: 3.7.6
- GluonTS version: 0.4.3.dev109+g6cc6ed4

lostella:
@alexhallam this was moved, you can do `from gluonts.model.baseline import SeasonalNaivePredictor`. Could you confirm that this import works?

Note that this is a predictor and not an estimator, since it doesn’t require offline training.

alexhallam:
Seems to be getting closer. I got an `ModuleNotFoundError: No module named 'statsmodels'` error. I am installing `statsmodels` now.

alexhallam:
Works like a charm! Thanks.

A couple of points on documentation:
1. `statsmodels` was required
2. `num_samples` was needed, but `num_eval_samples` was in docs

I am not sure if I will be able to add documentation, but if you point me in the right direction I may be able to fix those problems.

```python
import pandas as pd 
import numpy as np 
import mxnet as mx
from mxnet import gluon
from gluonts.dataset import common
from gluonts.model.baseline import SeasonalNaivePredictor
from gluonts.trainer import Trainer
[...]

PR #788, opened 2020-04-24
Move naive predictors to independent subpackages

### LINKED ISSUE #786: SeasonalNaivePredictor depends on statsmodels
Because of how the baseline models are grouped within the `gluonts.models.baseline` subpackage, one requires `statsmodels` in order to be able to import `SeasonalNaivePredictor`, see #784.

This is not ideal, we should move back both baseline models into individual subpackages:

```
gluonts.model.seasonal_naive
gluonts.model.naive2
```

lostella:
*Issue #, if available:* fixes #786 (related to #784)

*Description of changes:* The seasonal naive method was moved together with the naive2 method in a common subpackage. This has the drawback that all of the sudden `statsmodels` is required for using the seasonal naive predictor, which is not required. This PR puts back the seasonal naive predictor to `gluonts.model.seasonal_naive`, and the more recent naive 2 method to `gluonts.model.naive_2`.


AaronSpieler:
This is a good idea, thx!

Issue #765, opened 2020-04-16
LSTNet's scaling 


kashif:
## Description
The scaling in LSTNet is only valid for `NTC` tensors however we are feeding it `NCT` tensors since `output_NTC=False` and thus are not using the scaled values.

## To Reproduce
(Please provide minimal example of code snippet that reproduces the error. For existing examples, please provide link.)

```python
put code here
```

## Error message or code output
(Paste the complete error message, including stack trace, or the undesired output that the above snippet produces.)

```
put error or undesired output here
```

## Environment
- Operating system:
- Python version:
- GluonTS version:

(Add as much information about your environment as possible, e.g. dependencies versions.)

lostella:
Thanks @kashif, you’re right. I think this requires having an `axis` option in scaler blocks, what do you think?

kashif:
@lostella right I have that working, also the the LSTNet implementation is wrong, and so I can send both the fixes in 1 PR... is that ok?

lostella:
> @lostella right I have that working, also the the LSTNet implementation is wrong, and so I can send both the fixes in 1 PR... is that ok?

Hey thanks for that, it would be best to have two separate PRs I think

kashif:
ok PR with `axis` coming first

kashif:
fixed

PR #790, opened 2020-04-27
Scaler for (N,T,C) and (N,C,T) tensors

kashif:
fixed the scaler to work on time first and time last shaped tensors.

*Issue #, if available:*
#765 

*Description of changes:*

The scaler now takes axis argument to specify the axis over which to calculate the scaling.

### DIFF
--- a/src/gluonts/block/scaler.py
+++ b/src/gluonts/block/scaler.py
@@ -31,8 +31,8 @@ class Scaler(nn.HybridBlock):
     keepdims
         toggle to keep the dimension of the input tensor.
     input_NTC
-        toggle the shape of the input tensors from the default (N, T, C)
-        to (N, C, T) when this is False.
+        specify the shape of the input tensors either the default (N, T, C)
+        or (N, C, T) when this is False.
     """
 
     def __init__(self, keepdims: bool = False, input_NTC: bool = True):
@@ -52,7 +52,7 @@ class Scaler(nn.HybridBlock):
             API in MXNet.
 
         data
-            tensor of shape (N, T, C) containing the data to be scaled
+            tensor of shape (N, T, C) or (N, C, T) containing the data to be scaled.
 
         observed_indicator
             observed_indicator: binary tensor with the same shape as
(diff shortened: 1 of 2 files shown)


src/gluonts/block/scaler.py
```diff
@@ -30,12 +30,16 @@ class Scaler(nn.HybridBlock):
     ----------
     keepdims
         toggle to keep the dimension of the input tensor.
+    input_NTC
```
lostella, line 33: I think it would be best to configure these blocks directly with `axis`? And default that to 1 for backward compat

src/gluonts/block/scaler.py
```diff
@@ -48,7 +52,7 @@ def compute_scale(self, F, data: Tensor, observed_indicator: Tensor):
             API in MXNet.
 
         data
-            tensor of shape (N, T, C) containing the data to be scaled
+            tensor of shape (N, T, C) or (N, C, T) containing the data to be scaled.
```
lostella, line 55: With the suggested change above, this could become just “tensor containing the data to be scaled”

src/gluonts/block/scaler.py
```diff
@@ -79,7 +83,7 @@ def hybrid_forward(
         Returns
         -------
         Tensor
-            Tensor containing the "scaled" data, shape: (N, T, C).
+            Tensor containing the "scaled" data.
         Tensor
             Tensor containing the scale, of shape (N, C) if ``keepdims == False``, and shape
             (N, 1, C) if ``keepdims == True``.
```
lostella, line 89:
```suggestion
            Tensor containing the scale: this has the same shape as the data, except for the axis ``axis``
            along which the scale is computed, which is removed if ``keepdims == False``, and kept with
            length 1 otherwise. For example, if ``data`` has shape ``(N, T, C)`` and ``axis ==1 ``, then
            ``scale`` has shape ``(N, C)`` if ``keepdims == False``, and ``(N, 1, C)`` otherwise.
```

kashif:
@lostella if you can kindly merge, i'll send you the LSTNet fix next.

kashif:
thanks @lostella

PR #792, opened 2020-04-27
Add temperature scaling to categorical distribution

steverab:
*Description of changes:*

This PR adds an optional constructor parameter `temperature` to `CategoricalOutput` to enable temperature scaling in the categorical distribution.

### DIFF
--- a/src/gluonts/distribution/categorical.py
+++ b/src/gluonts/distribution/categorical.py
@@ -16,6 +16,7 @@ from typing import Tuple, List
 
 # Third-party imports
 import numpy as np
+import mxnet as mx
 
 # First-party imports
 from gluonts.core.component import validated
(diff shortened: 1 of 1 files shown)


steverab:
Right, this parameter needs to be tuned on a validation set (like a lot of other parameters used by our models should). But this parameter doesn't necessarily need to be treated as a constant. Since `temperature` is a member variable you can access and set it anywhere in the model via `distr_output.temperature` and modify it if you want. I don't see any harm in having temperature be settable via the constructor.

**Background**: The main reason for this PR is to enable the same kind of temperature scaling that is currently being done in the `WaveNetEstimator`. As part of my changes to GluonTS via the representation module, the `WaveNetEstimator` will loose its `temperature` parameter (since we want `WaveNetEstimator` to support flexible distributions and not just discretized representations) and the best way to enable temperature scaling then is to add it to `CategoricalOutput`. I'm not sure how `temperature` is currently tuned in `WaveNetEstimator` but I would claim that the same tuning could still be done here with very little to no overhead. Maybe @rshyamsundar @vafl can comment?

Issue #774, opened 2020-04-21
Windows Wavenet asnumpy Timeout


AaronSpieler:
## Description
Wavenet produces output that when trying to convert it to numpy fails.

I identified the pull that introduced this bug: https://github.com/awslabs/gluon-ts/pull/727 see https://github.com/awslabs/gluon-ts/pull/727/checks.
(The preceding pulls, including the most recent before this in the commit history https://github.com/awslabs/gluon-ts/pull/728 do not have this error: https://github.com/awslabs/gluon-ts/pull/728/checks.)

## Examples:

* https://github.com/awslabs/gluon-ts/pull/772/checks?check_run_id=606008914
```
  File "c:\hostedtoolcache\windows\python\3.6.8\x64\lib\site-packages\pluggy\callers.py", line 187, in _multicall
    res = hook_impl.function(*args)
  File "c:\hostedtoolcache\windows\python\3.6.8\x64\lib\site-packages\_pytest\python.py", line 166, in pytest_pyfunc_call
    testfunction(**testargs)
  File "D:\a\gluon-ts\gluon-ts\test\model\wavenet\test_model.py", line 40, in test_accuracy
    accuracy_test(WaveNetEstimator, hyperparameters, accuracy=0.7)
  File "D:\a\gluon-ts\gluon-ts\test\model\conftest.py", line 100, in test_accuracy
    evaluator=Evaluator(calculate_owa=statsmodels is not None),
  File "d:\a\gluon-ts\gluon-ts\src\gluonts\evaluation\backtest.py", line 224, in backtest_metrics
    ts_it, forecast_it, num_series=maybe_len(test_dataset)
  File "d:\a\gluon-ts\gluon-ts\src\gluonts\evaluation\_base.py", line 178, in __call__
    for ts, forecast in it:
  File "c:\hostedtoolcache\windows\python\3.6.8\x64\lib\site-packages\tqdm\std.py", line 1127, in __iter__
    for obj in iterable:
  File "d:\a\gluon-ts\gluon-ts\src\gluonts\model\predictor.py", line 317, in predict
    num_samples=num_samples,
  File "d:\a\gluon-ts\gluon-ts\src\gluonts\model\forecast_generator.py", line 204, in __call__
    outputs = prediction_net(*inputs).asnumpy()
  File "c:\hostedtoolcache\windows\python\3.6.8\x64\lib\site-packages\mxnet\ndarray\ndarray.py", line 2535, in asnumpy
[...]

PR #793, opened 2020-04-28
Fixing timeout issue for Wavenet 

arangatang:
*Issue #, if available: #774*

*Description of changes:*
Changed the number of batches run per epoch to lower both training time and evaluation time.
This causes a speedup of the test by about 35%
To handle the potential difference in accuracy I performed some benchmarks of the test and adjusted the allowed accuracy to be below 1 since it fluctuated between ~0.67 - ~0.9.


test/model/wavenet/test_model.py
```diff
@@ -27,21 +27,18 @@ def hyperparameters(dsinfo):
         num_cells=2,
         num_layers=1,
         context_length=2,
-        num_batches_per_epoch=1,
+        num_batches_per_epoch=5,
```
jaheba, line 30: Why not always run with `1`?

arangatang: @lostella asked for half of what is currently see the comment below
https://github.com/awslabs/gluon-ts/issues/774#issuecomment-620164912

lostella: If there are strong feelings about making this test much faster, it's fine for me to lower it further. Maybe just make sure that more than one batch is run, since the network is used differently in the first two forward steps, when hybridizing the network.

arangatang: Does `num_batches_per_epoch=3` sound good then? Like this we both initialise the algorithm and then run one pass after it being initialised so that all parts are properly exercised.

jaheba: `3` sounds alright. Does it pass consistently then?

lostella:
FYI the failing tests on jenkins are 1) some data loading throughput tests, which are probably affected by the machine the tests run on, and 2) some other flaky model tests :-)

test/model/wavenet/test_model.py
arangatang: Yes it does, it passes after running 50 iterations hybridised and 50 iterations not hybridised for each dataset. I.e. it passes 200 tests in a row.

PR #797, opened 2020-04-30
Run nightly test

### LINKED ISSUE #785: Nightly build on latest release
It sometimes happens that dependencies introduce breaking changes, despite not bumping the major version number. This appears to have happened a few days ago with pydantic (see #768).

Since most of the developers work on the master branch anyway, it would be nice to catch whenever a dependency breaks. One idea to do this would be to have nightly builds installing the latest released tag and running tests.

arangatang:
*Issue #, if available:* #785 

*Description of changes:*
Added two new jobs, one for Windows and one for Unix systems that builds gluon-ts from scratch every night. 
The jobs are modified from these files: [mxnet_nightly.yml](https://github.com/awslabs/gluon-ts/blob/master/.github/workflows/mxnet_nightly.yml) and [pythonpackage.yml](https://github.com/awslabs/gluon-ts/blob/v0.4.3/.github/workflows/pythonpackage.yml)

They function like this:

1. The latest released tag is fetched from the GitHub API
2. The repository is cloned and the branch with the corresponding name is checked out
3. Dependencies get installed

Both of the nightly builds run well when debugging on my own GitHub fork.

*The .yml files would need to be updated in case of new requirements files being needed to run the tests successfully* 

For example in the latest test suite we have added the following lines of code:
* [line 29](https://github.com/awslabs/gluon-ts/blob/8e6823a071dd5f4b6210b9ba47d22041806f8cde/.github/workflows/tests.yml#L29)
* [line 30](https://github.com/awslabs/gluon-ts/blob/8e6823a071dd5f4b6210b9ba47d22041806f8cde/.github/workflows/tests.yml#L30)

If we do not end up merging the contents of these files into `requirements-test.txt` we would need to modify the files in this commit so that the build succeeds for the next release. 

fixes #785 


.github/workflows/build_release_unix_nightly.yml
```diff
@@ -0,0 +1,36 @@
...
+  build:
+
+    strategy:
+      max-parallel: 4
+      fail-fast: false
+      matrix:
+        python-version: [3.6, 3.7]
+        platform: [ubuntu-latest]
+
+    runs-on: ${{ matrix.platform }}
```
jaheba, line 19: I guess this can be simplified.

arangatang: Fixed this in the last revision.

.github/workflows/test_release_unix_nightly.yml
```diff
@@ -0,0 +1,36 @@
...
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Clone and install dependencies
+      run: |
+        git clone https://github.com/awslabs/gluon-ts --branch $(curl https://api.github.com/repos/awslabs/gluon-ts/releases/latest | grep tag_name | cut -d : -f 2,3 | tr -d \"\ | tr -d \,\ )
+        cd gluon-ts
+        python -m pip install --upgrade pip
+        pip install -r requirements/requirements-mxnet.txt
+        pip install gluonts
```
lostella, line 30: I think this will install gluonts from pypi? And that's totally fine, but probably the logic for getting the latest tag is superfluous then

lostella: I see, that's needed to get the right dependencies

lostella: There might be an issue here in case we ever start doing release branches, and releasing bug fixes for previous releases: the latest tag might not be the same as the latest available release from PyPI. But until then, this is fine.

arangatang: yep, also if we add more requirement files that the tests depend on as is done [here](https://github.com/awslabs/gluon-ts/blob/53120d4af666e24fe92a2f5051f865399ad7a7da/.github/workflows/tests.yml#L29) we will run into issues for future releases

PR #798, opened 2020-04-30
fix in RForecastPredictor

kashif:
fix pandas Timestamp

*Issue #, if available:*

*Description of changes:*

the pandas API has changed and this is the preferred way as far as i am aware.


jaheba:
There is `forecast_start` in `gluonts.support.pandas`. Would that work?

PR #799, opened 2020-04-30
Fix how cardinality is handled in DeepState

### LINKED ISSUE #794: deepstate asks for cardinality while use_feat_static_cat is set to False
Using gluonts 0.4.2 got error message when trying 
`estimator = DeepStateEstimator(
    freq='D',
    past_length=context_length,
    use_feat_static_cat=False,
    use_feat_dynamic_real=True,
    trainer=Trainer(ctx="cpu", 
                    epochs=50,  
                    learning_rate=1e-3, 
                    num_batches_per_epoch=100
                   )
)
`
error message as follows. What makes sense to put for cardinality when use_feat_static_cat is set to False? Thank you!
````
--------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
<ipython-input-454-045457f005f5> in <module>
      9                     epochs=50,
     10                     learning_rate=1e-3,
---> 11                     num_batches_per_epoch=100
     12                    )
     13 )

/anaconda3/lib/python3.7/site-packages/gluonts/core/component.py in init_wrapper(*args, **kwargs)
    422                 if name != "self"
    423             }
--> 424             model = PydanticModel(**{**nmargs, **kwargs})
    425 
    426             # merge nmargs, kwargs, and the model fields into a single dict

/anaconda3/lib/python3.7/site-packages/pydantic/main.py in __init__(__pydantic_self__, **data)
    281         values, fields_set, validation_error = validate_model(__pydantic_self__.__class__, data)
    282         if validation_error:
--> 283             raise validation_error
    284         object.__setattr__(__pydantic_self__, '__dict__', values)
    285         object.__setattr__(__pydantic_self__, '__fields_set__', fields_set)

ValidationError: 1 validation error for DeepStateEstimatorModel
cardinality
  field required (type=value_error.missing)
```

lostella:
*Issue #, if available:*  fixes #794 

### DIFF
--- a/src/gluonts/model/deepstate/_estimator.py
+++ b/src/gluonts/model/deepstate/_estimator.py
@@ -153,7 +153,7 @@ class DeepStateEstimator(GluonEstimator):
         self,
         freq: str,
         prediction_length: int,
-        cardinality: List[int],
+        cardinality: Optional[List[int]] = None,
         add_trend: bool = False,
         past_length: Optional[int] = None,
         num_periods_to_train: int = 4,
@@ -189,7 +189,8 @@ class DeepStateEstimator(GluonEstimator):
             num_parallel_samples > 0
         ), "The value of `num_parallel_samples` should be > 0"
         assert dropout_rate >= 0, "The value of `dropout_rate` should be >= 0"
-        assert not use_feat_static_cat or any(c > 1 for c in cardinality), (
+        assert use_feat_static_cat == (cardinality is not None)
+        assert cardinality is None or any(c > 1 for c in cardinality), (
             f"Cardinality of at least one static categorical feature must be larger than 1 "
             f"if `use_feat_static_cat=True`. But cardinality provided is: {cardinality}"
         )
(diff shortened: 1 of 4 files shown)


src/gluonts/model/deepstate/_estimator.py
```diff
@@ -153,7 +153,7 @@ def __init__(
         self,
         freq: str,
         prediction_length: int,
-        cardinality: List[int],
+        cardinality: Optional[List[int]] = None,
```
rshyamsundar, line 156: Now you cannot construct the estimator with default values since `cardinality` and `use_feat_static_cat` (which is `True` by default) are in conflict. This was why we made this argument non-optional. 

I think we agreed that when this estimator is used with `use_feat_static_cat=False`, then the user explicitly sets `cardinality = [1]` (which is the right thing to do) and is an additional check so that it is not misused (this case should not happen most of the times). We should update the docstring for `cardinality` argument to reflect this.

lostella: Agreed, let's update the docstring instead then

lostella: @rshyamsundar on a second thought: I think the way I fixed it in this PR just makes things run smoothly in case you set `use_feat_static_cat=False`. With default values (`use_feat_static_cat=True`) you would get a validation error prior to this PR, and an assertion error after this PR.

rshyamsundar: > if no cardinality is provided then you would get a validation error prior to this PR,

`cardinality` was non-optional argument prior to this PR, so it has to be given by default like `freq` or `pred_length` argument. Prior to this PR, this estimator can be instantiated by providing only the non-optional arguments but now if one instantiates that way they get an assertion error.

PR #802, opened 2020-05-01
Fix lstnet scaling by circumventing defer init bug

ehsanmok:
*Description of changes:*

The known deferred init [bug](https://github.com/apache/incubator-mxnet/issues/14373) prevents using `scaling=False` in hybridized mode. This is a work around at the expense of some computation until it's fix.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #804, opened 2020-05-03
Fix LSTNet observed indicator usage

lostella:
*Issue #, if available:* Addresses the same issue as #802 

*Description of changes:* This PR addresses the same issue as #802 (issues with scaling=False and hybridize=True) by
* using the observed indicator as input to the model blocks (CNN and FF)

Furthermore
* future observed indicator is added to the training network and used to mask the loss
* tests updates are copied from #802, except that the number of test cases is here reduced

cc @ehsanmok 

### DIFF
--- a/src/gluonts/model/lstnet/_network.py
+++ b/src/gluonts/model/lstnet/_network.py
@@ -97,7 +97,7 @@ class LSTNetBase(nn.HybridBlock):
                 (num_series, kernel_size),
                 activation="relu",
                 layout="NCHW",
-                in_channels=1,
+                in_channels=2,
             )  # NC1T
             self.cnn.cast(dtype)
             self.dropout = nn.Dropout(dropout_rate)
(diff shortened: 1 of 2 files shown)


src/gluonts/model/lstnet/_network.py
```diff
@@ -183,9 +183,13 @@ def _skip_rnn_layer(self, F, x: Tensor) -> Tensor:
...
         ar_x = F.slice_axis(x, axis=2, begin=-self.ar_window, end=None)  # NCT
-        ar = self.ar_fc(ar_x)  # NxCx(1 or prediction_length)
+        ar_observed = F.slice_axis(
+            observed, axis=2, begin=-self.ar_window, end=None
+        )  # NCT
+        ar_fc_inputs = F.concat(ar_x, ar_observed, dim=-1)
+        ar = self.ar_fc(ar_fc_inputs)  # NxCx(1 or prediction_length)
         return ar
 
     # noinspection PyMethodOverriding,PyPep8Naming
```
ehsanmok, line 195: Are we still keeping these things?

ehsanmok:
Thanks Lorenzo! this is a better solution. LGTM!

src/gluonts/model/lstnet/_network.py
lostella: they don't hurt :-)

PR #811, opened 2020-05-07
Add logit normal distribution

steverab:
*Description of changes:*

This PR adds the [logit-normal distribution](https://en.wikipedia.org/wiki/Logit-normal_distribution). This distribution is useful for probability-integral-transformed inputs. While the marginal distribution over probability-integral-transformed data is uniform, the conditional distribution is not and that's where this distribution (and especially a mixture) gives us more flexibility to model data lying in (0,1). Since there are no analytic solutions for the mean and the variance of this distribution, the corresponding tests requiring these properties were omitted.

### DIFF
--- a/src/gluonts/distribution/__init__.py
+++ b/src/gluonts/distribution/__init__.py
@@ -48,7 +48,7 @@ from .student_t import StudentT, StudentTOutput
 from .transformed_distribution import TransformedDistribution
 from .transformed_distribution_output import TransformedDistributionOutput
 from .uniform import Uniform, UniformOutput
-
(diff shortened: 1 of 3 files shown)


steverab:
@lostella There is a method which does that (see [here](https://github.com/awslabs/gluon-ts/blob/master/test/distribution/test_distribution_sampling.py#L157)). Unfortunately, that same method also tests for the means (which the logit-normal does not have an analytic solution for). Maybe that functionality should be refactored into a separate testing method? But yeah, this probably deserves its own issue/PR.

PR #812, opened 2020-05-07
fix scaling in negative binomial

lostella:
*Issue #, if available:* #718

*Description of changes:* From the relationship with mean and standard deviation, the scaled parameter `alpha` should be `alpha' = alpha + (nu-1)/(nu*mu)`. The latest fix just missed `nu` in the denominator.

### DIFF
--- a/src/gluonts/distribution/neg_binomial.py
+++ b/src/gluonts/distribution/neg_binomial.py
@@ -130,7 +130,9 @@ class NegativeBinomialOutput(DistributionOutput):
         else:
             F = getF(mu)
             mu = F.broadcast_mul(mu, scale)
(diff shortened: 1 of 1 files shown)


src/gluonts/distribution/neg_binomial.py
```diff
@@ -130,7 +130,9 @@ def distribution(
         else:
             F = getF(mu)
             mu = F.broadcast_mul(mu, scale)
-            alpha = F.broadcast_add(alpha, F.broadcast_div(scale - 1, mu))
+            alpha = F.broadcast_add(
```
canerturkmen, line 133: OK, so isn't the scale is already multiplied into `mu` on line 132?

lostella: Of course, that was me being tired

PR #823, opened 2020-05-13
Model averaging

benidis:
Model averaging

Averaging of best models (based on best epochs) during training. The weight of each model is either `1/N` or proportional to the metric value (better metric value more weight).

### DIFF
--- a/src/gluonts/trainer/_base.py
+++ b/src/gluonts/trainer/_base.py
@@ -360,7 +360,11 @@ class Trainer:
                 else:
                     logging.info("Computing averaged parameters.")
                     averaged_params_path = ma.average_parameters(
-                        gluonts_temp, num_models=self.num_averaged_models
+                        gluonts_temp,
+                        num_models=self.num_averaged_models,
+                        metric="score",
+                        maximize=False,
+                        weight="exp-metric",
                     )
 
                     logging.info("Loading averaged parameters.")
--- a/src/gluonts/trainer/_base.py
+++ b/src/gluonts/trainer/_base.py
@@ -110,6 +110,7 @@ class Trainer:
         init: Union[str, mx.initializer.Initializer] = "xavier",
         hybridize: bool = True,
         num_averaged_models: int = None,
+        averaged_models_weight: str = "average",
     ) -> None:
 
         assert (
@@ -150,6 +151,7 @@ class Trainer:
         self.init = init
         self.hybridize = hybridize
         self.num_averaged_models = num_averaged_models
+        self.averaged_models_weight = averaged_models_weight
         self.ctx = ctx if ctx is not None else get_mxnet_context()
         self.halt = False
 
@@ -364,7 +366,7 @@ class Trainer:
                         num_models=self.num_averaged_models,
                         metric="score",
                         maximize=False,
-                        weight="exp-metric",
+                        weight=self.averaged_models_weight,
                     )
 
                     logging.info("Loading averaged parameters.")
--- a/src/gluonts/trainer/_base.py
+++ b/src/gluonts/trainer/_base.py
@@ -195,6 +195,12 @@ class Trainer:
             ):
                 batch_size = train_iter.batch_size
 
+                best_epoch_info = {
+                    "params_path": "%s-%s.params" % (base_path(), "init"),
+                    "epoch_no": -1,
+                    "score": np.Inf,
+                }
+
                 lr_scheduler = lrs.MetricAttentiveScheduler(
                     objective="min",
                     patience=self.patience,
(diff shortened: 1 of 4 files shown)


benidis:
> Nice! It would be great to have a test with dummy parameters to check whether the parameters are averaged correctly.

Yes I agree. I will write a test for that.

src/gluonts/trainer/model_averaging.py
```diff
@@ -0,0 +1,164 @@
...
+import json
+
+# Third-party imports
+import mxnet as mx
+import numpy as np
+
+EPOCH_INFO_STRING = "epoch-info"
+
+
+def save_epoch_info(tmp_path, epoch_info):
```
vafl, line 25: Could you add type annotations to all these functions?

src/gluonts/trainer/_base.py
```diff
@@ -311,43 +316,60 @@ def loop(
...
+                        bp = base_path()
+                        best_epoch_info = {
+                            "params_path": f"{bp}-0000.params",
+                            "epoch_no": epoch_no,
+                            "score": loss_value(epoch_loss),
+                        }
+
                         net.save_parameters(
-                            best_epoch_info.params_path
+                            best_epoch_info["params_path"]
```
vafl, line 328: Perhaps we can simplify the trainer logic a bit:

1. Checkpoint after every epoch no matter if we do model averaging or not (should not be a big overhead)
2. Use an `AveragingStrategy` similar to what you already introduced below. If we don't want model averaging this just picks the best epoch, or the last epoch etc

I'm also fine with merging as is for now. We can do these changes later.

src/gluonts/trainer/model_averaging.py
```diff
@@ -0,0 +1,164 @@
...
+
+def average_parameters(
+    model_path, num_models=5, metric="score", maximize=False, weight="average"
+):
+    """
+    Averages model parameters of serialized models based on the selected model strategy and metric.
+    IMPORTANT: Depending on the metric the user might want to minimize or maximize. The maximize flag has to be
+    chosen appropriately to reflect this.
+
+    :param model_path: str, path to the models directory.
```
lostella, line 43: Could you convert all these docstrings to numpy-style? You can configure PyCharm to automatically style do strings like that when you open triple-quotes

src/gluonts/trainer/_base.py
```diff
@@ -57,12 +58,6 @@ def loss_value(loss: mx.metric.Loss) -> float:
     return loss.get_name_value()[0][1]
 
 
-class BestEpochInfo(NamedTuple):
```
lostella, line 60: Why removing this class? If I understand correctly, it’s being replace by a dictionary. I thought we rename this into “EpochInfo”, but maybe there are reasons to remove it instead?

benidis: NamedTuple is not json serializable. The structure is destroyed and when you deserialize you get back just a list. Dictionaries on the other hand do not have this issue.

src/gluonts/trainer/model_averaging.py
```diff
@@ -0,0 +1,250 @@
...
+        Parameters
+        ----------
+        model_path
+            Path to the models directory.
+
+        Returns
+        -------
+        Path to file with the averaged model.
+        """
+        checkpoints = self.load_top_checkpoints(model_path)
```
vafl, line 86: I think this can be more flexible here, by not pre-selecting the top n. For example how would we pick the last n epochs even if they didn't have the best scores?

At this point we should get a list with all available checkpoints and their metdata (score, epoch etc) and then call a method `select_checkpoints` with this data. The method can then return the weights and checkpoint paths that should be averaged.

The `select_checkpoints` method could then differ for different strategies. So we could then have `SelectNBest`, `AverageLastN` etc.

Does that make sense?

benidis: The assumption here was that you always want to select the best and not some random epochs. I kept this part (selecting top checkpoints) common is all classes since it was a common assumption, regardless of how each class does the averaging later. I am not sure if it is useful to average the last N and not the best N, but still we can include this option. 

This increases a bit the complexity of defining things. I see two options here:

1) The class should take both some info on which epochs to select and how to average them, e.g. 

`MyStrategy(num_models=N, select='last', average='mean')`

which basically removes the need for subclasses but creates an `if ... elif` block for checking all the options of `select` and `average` (unless we actually create classes for selection and averaging and pass instances which makes things more complex in defining things e.g. `MyStrategy(num_models=N, select=SelectLast() average=MeanAvg())`). 

2) Create specific classes for each option but then the number of classes grows too much, e.g.

`MeanAvgLast(num_models=N)  -> implicit select='last', average='mean'`
`MeanAvgBest(num_models=N )-> implicit  select='best', average='mean'`
`MetricValueAvgBest(num_models=N)-> implicit  select='best', average='metric'`

Any suggestions on what is better?

vafl: Yeah, I think keeping it flexible is best. For instance it may be good to pick the best epoch and average the n surrounding epochs. Specific classes and the `select_checkpoints` method is the most flexible approach I think.

PR #840, opened 2020-05-25
Representation models

steverab:
*Description of changes:*
This PR updates four of our core models (SimpleFeedForward, DeepAR, Transformer, WaveNet) to the representations module (#755). This enables these models to deal with flexible input and output representations.

### DIFF
--- a/src/gluonts/model/deepar/_estimator.py
+++ b/src/gluonts/model/deepar/_estimator.py
@@ -20,6 +20,7 @@ from mxnet.gluon import HybridBlock
 
 # First-party imports
 from gluonts.core.component import DType, validated
+from gluonts.dataset.common import Dataset
 from gluonts.dataset.field_names import FieldName
 from gluonts.dataset.stat import calculate_dataset_statistics
 from gluonts.distribution import DistributionOutput, StudentTOutput
@@ -44,6 +45,12 @@ from gluonts.transform import (
     SetField,
     Transformation,
     VstackFeatures,
+)
+from gluonts.representation import (
+    Representation,
+    MeanScaling,
+    DimExpansion,
+    RepresentationChain,
 )
 
 # Relative imports
@@ -96,11 +103,15 @@ class DeepAREstimator(GluonEstimator):
     embedding_dimension
         Dimension of the embeddings for categorical features
         (default: [min(50, (cat+1)//2) for cat in cardinality])
+    input_repr
+        Representation for the model inputs.
+        (default: DimExpansion(MeanScaling()))
+    output_repr
+        Representation for the model outputs.
+        (default: Representation())
     distr_output
         Distribution to use to evaluate observations and sample predictions
         (default: StudentTOutput())
-    scaling
-        Whether to automatically scale the target values (default: true)
     lags_seq
         Indices of the lagged target values to use as inputs of the RNN
         (default: None, in which case these are automatically determined
@@ -129,8 +140,11 @@ class DeepAREstimator(GluonEstimator):
         use_feat_static_real: bool = False,
         cardinality: Optional[List[int]] = None,
         embedding_dimension: Optional[List[int]] = None,
+        input_repr: Representation = RepresentationChain(
+            chain=[MeanScaling(), DimExpansion()]
+        ),
+        output_repr: Representation = Representation(),
         distr_output: DistributionOutput = StudentTOutput(),
-        scaling: bool = True,
         lags_seq: Optional[List[int]] = None,
         time_features: Optional[List[TimeFeature]] = None,
         num_parallel_samples: int = 100,
@@ -165,6 +179,8 @@ class DeepAREstimator(GluonEstimator):
             context_length if context_length is not None else prediction_length
         )
         self.prediction_length = prediction_length
+        self.input_repr = input_repr
+        self.output_repr = output_repr
         self.distr_output = distr_output
         self.distr_output.dtype = dtype
         self.num_layers = num_layers
@@ -182,7 +198,6 @@ class DeepAREstimator(GluonEstimator):
             if embedding_dimension is not None
             else [min(50, (cat + 1) // 2) for cat in self.cardinality]
         )
-        self.scaling = scaling
         self.lags_seq = (
             lags_seq
             if lags_seq is not None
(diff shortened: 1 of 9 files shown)


src/gluonts/model/wavenet/_network.py
```diff
@@ -568,40 +568,51 @@ def blow_up(u):
...
-                # (batch_size, 1)
-                y = F.sample_multinomial(probs)
-            else:
-                # (batch_size, 1)
-                y = F.argmax(unnormalized_outputs, axis=-1)
-            y = y.astype("int32")
+            distr_args = self.proj_distr_args(unnormalized_outputs)
+            distr = self.distr_output.distribution(distr_args, scale=scale_bu)
+
+            y = distr.sample()
```
rshyamsundar, line 610: Earlier we had a way to sample only the mode if we set `temperature` to zero or negative. We don't have this now. Also it looks like `distr_output` does not take into account `temperature`.

src/gluonts/model/wavenet/_network.py
```diff
@@ -292,8 +282,7 @@ def target_feature_embedding(
 
         """
         # (batch_size, embed_dim, sequence_length)
-        o = self.target_embed(target).swapaxes(1, 2)
-        o = F.concat(o, features, dim=1)
+        o = F.concat(target, features, dim=1)
```
rshyamsundar, line 285: So there is no target embedding now?

src/gluonts/model/wavenet/_network.py
steverab: `CategoricalOutput()` includes a `temperature` parameter which you can set to your liking to regain that same functionality: https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/distribution/categorical.py#L106

src/gluonts/model/wavenet/_network.py
steverab: This is all taken care by the respective representation that you choose. For example, a mean scaled representation (`MeanScaling()`) will not embed target values while a binned representation (e.g. `GlobalRelativeBinning()`) will do that. This takes place here: https://github.com/steverab/gluon-ts/blob/representation_models/src/gluonts/model/wavenet/_network.py#L383-L389

src/gluonts/model/wavenet/_network.py
rshyamsundar: Yes, I saw that but you are not setting the `temperature` that is passed to `WavenetSampler`. At the moment passing a different value to `temperature` and using the default value for `distr_output` in `WavenetEstimator` cannot reproduce earlier behavior (`temperature` is silently ignored). 

I see that `temperature` is set inside `CategoricalOutput` and it becomes difficult to reset it given that both `temperature` and `distr_output` are constructor arguments of `WavenetEstimator`.  

This brings the question: should `temperature` be part of `CategoricalOutput`? Currently it is used to modify sampling alone. If there are no other cases, why not pass it as an explicit argument to `sample` method?

src/gluonts/model/deepar/_estimator.py
```diff
@@ -222,6 +237,18 @@ def derive_auto_fields(cls, train_iter):
...
+        self,
+        training_data: Dataset,
+        validation_data: Optional[Dataset] = None,
+        num_workers: Optional[int] = None,
+        num_prefetch: Optional[int] = None,
+        **kwargs,
+    ) -> Predictor:
+        self.input_repr.initialize_from_dataset(training_data)
+        self.output_repr.initialize_from_dataset(training_data)
+        return super().train(training_data)
```
lostella, line 250: I think other arguments should be passed on to `super().train`, right? You can probably capture all but `training_data` in **kwargs

src/gluonts/model/deepar/_network.py
```diff
@@ -251,20 +248,27 @@ def unroll_encoder(
...
         # from (batch_size, sub_seq_len, *target_shape, num_lags)
         # to (batch_size, sub_seq_len, prod(target_shape) * num_lags)
         input_lags = F.reshape(
-            data=lags_scaled,
+            data=lags,
             shape=(
                 -1,
                 subsequences_length,
-                len(self.lags_seq) * prod(self.target_shape),
+                # len(self.lags_seq) * prod(self.target_shape),
```
lostella, line 258: Remove?

src/gluonts/model/deepar/_network.py
```diff
@@ -251,20 +248,27 @@ def unroll_encoder(
...
         )
 
+        # input_lags = F.reshape(
+        #     data=lags,
+        #     shape=(
+        #         lags.shape[0],
+        #         lags.shape[1],
+        #         lags.shape[2] * lags.shape[3],
+        #     ),
+        # )
```
lostella, line 270: Remove the above lines?

src/gluonts/model/simple_feedforward/_estimator.py
```diff
@@ -136,11 +149,24 @@ def __init__(
...
+        self,
+        training_data: Dataset,
+        validation_data: Optional[Dataset] = None,
+        num_workers: Optional[int] = None,
+        num_prefetch: Optional[int] = None,
+        **kwargs,
+    ) -> Predictor:
+        self.input_repr.initialize_from_dataset(training_data)
+        self.output_repr.initialize_from_dataset(training_data)
+        return super().train(training_data)
```
lostella, line 168: Same as above

src/gluonts/model/transformer/_estimator.py
```diff
@@ -205,6 +220,11 @@ def __init__(
             self.prediction_length, self.config, prefix="dec_"
         )
 
+    def train(self, training_data: Dataset) -> Predictor:
+        self.input_repr.initialize_from_dataset(training_data)
+        self.output_repr.initialize_from_dataset(training_data)
+        return super().train(training_data)
```
lostella, line 226: Same as above

src/gluonts/model/deepar/_network.py
steverab: Must have missed deleting this. Fixed!

src/gluonts/model/deepar/_network.py
steverab: Must have missed deleting this. Fixed!

src/gluonts/model/simple_feedforward/_estimator.py
steverab: Fixed!

src/gluonts/model/deepar/_estimator.py
steverab: Ah yeah, right. Fixed!

src/gluonts/model/simple_feedforward/_estimator.py
steverab: Fixed!

src/gluonts/model/transformer/_estimator.py
steverab: Seems like the Transformer does not have the data loader and validation stuff as of this moment. I added those and fixed the call to `super`.

PR #841, opened 2020-05-25
Dockerfile for R container added

Schmedu:
*Issue #, if available:*

*Description of changes:*
Dockerfile for R container added


jaheba:
I'd like to argue for creating a new folder where we put our dockerfiles. Maybe `/examples/dockerfiles/` is a good place to start.

That way it's more of a suggestion on how one can create such a container, rather than instructions how that is generally done.

Dockerfile.r
```diff
@@ -0,0 +1,13 @@
+FROM rpy2/rpy2:2.9.x
+
+ADD . /gluonts
+
+RUN pip3 install /gluonts[shell]
```
jaheba, line 5: We could also just suggest to pip install gluonts from PyPi.

Schmedu: You're right. Sounds better

examples/dockerfiles/Dockerfile.r
```diff
@@ -0,0 +1,10 @@
+FROM rpy2/rpy2:2.9.x
+
+RUN pip3 install --upgrade mxnet==1.6 "gluonts[shell]"
+
+RUN R -e 'install.packages(c("forecast", "nnfor"), repos="https://cloud.r-project.org")'
+
+ENV LC_ALL="C.UTF-8"
+ENV LANG="C.UTF-8"
+
+ENTRYPOINT ["python3", "-m", "gluonts.shell"]
```
jaheba, line 10:
```suggestion
ENTRYPOINT ["python3", "-m", "gluonts.shell"]

```

PR #843, opened 2020-05-26
New methods for missing value imputation

Francois-Aubet:
Hi,

I added a couple of options in the AddObservedValuesIndicator in order to replace the missing values with other methods. There is now this new parameter:
imputation_method
        Select the method to replace the missing values.
        - "standard" to just replace them with the dummy_value
        - "mean" to replace them with the mean of the array
        - "median" to replace them with the median of the array
        _ "last_val" to replace them with the last non missing value

I added this parameter in the constructor of all the models using this transformation.

I did a quick test to show cast the improvements that it brings.
Training the MQCNN model with the 4 different imputation methods. We obtain:
image

Results were obtained with this code.
```
import mxnet as mx
from mxnet import gluon
import numpy as np
import pandas as pd
import json
import os
from itertools import islice
from pathlib import Path

from copy import deepcopy

mx.random.seed(0)
np.random.seed(0)


from gluonts.dataset.artificial import ComplexSeasonalTimeSeries

from gluonts.model.seq2seq import MQCNNEstimator

# creating the datasets:
min_dataset_val = -1
max_dataset_val = 20

number_series = 10

prediction_length=400

freq_string="min"

length_low=1900
length_high=2000

artificial_dataset_miss = ComplexSeasonalTimeSeries(
    num_series=number_series,
    freq_str=freq_string,
    length_low=length_low,
    length_high=length_high,
    min_val=min_dataset_val,
    max_val=max_dataset_val,
    is_integer=False,
    proportion_missing_values=0.2,
    is_noise=True,
    is_scale=True,
    percentage_unique_timestamps=1,
    is_out_of_bounds_date=True,
)

# spliting in train and test sets

train_ds_miss = ListDataset(artificial_dataset_miss.train, freq=artificial_dataset_miss.metadata.freq)
test_ds_miss = ListDataset(artificial_dataset_miss.test, freq=artificial_dataset_miss.metadata.freq)

train_entry = next(iter(train_ds_miss))
train_series = to_pandas(train_entry)

test_entry_miss = next(iter(test_ds_miss))
test_series_miss = to_pandas(test_entry_miss)
test_entry = next(iter(test_ds_miss))
test_series = to_pandas(test_entry_miss)


# creating the estimator

freq="1min"
lead_time = 50
hidden_size = 40
num_layers = 3
epochs = 10

l_imputation_methods = ["standard", "mean", "median", "last_val"]

d_rnn = {}
d_predictors = {}

for imputation_method in l_imputation_methods:
    rnn = MQCNNEstimator(
        freq=freq,
        context_length=48,
        trainer=Trainer(
            ctx="cpu",
            epochs=epochs,
            batch_size=32,
            num_batches_per_epoch=100,
            learning_rate=0.002,
            hybridize=False
        ),
        imputation_method=imputation_method
    )

    d_rnn[imputation_method] = rnn




def train_and_wrap(model, dataset):
    predictor = model.train(dataset)

    return predictor


for imputation_method in l_imputation_methods:
    d_predictors[imputation_method] = train_and_wrap(d_rnn[imputation_method], train_ds_miss)


# Evaluation


l_eval_metric = ["MSE", "MASE", "sMAPE", "MSIS", "RMSE"]

l_predictors = d_predictors.values()

l_pred_names = list(d_predictors.keys())



def get_metric(predictor, dataset, l_metrics):
    # not efficient if one is interested in more than one metric

        dataset=dataset,  # test dataset
        num_samples=100,  # number of sample paths we want for evaluation
    )
    forecasts = list(forecast_it)
    tss = list(ts_it)

    evaluator = Evaluator(quantiles=[0.1, 0.5, 0.9])
    agg_metrics, item_metrics = evaluator(iter(tss), iter(forecasts), num_series=len(test_ds_miss))

    print(json.dumps(agg_metrics, indent=4))

    return [agg_metrics[metric] for metric in l_metrics]


def gather_metrics(l_predictors, dataset, l_metrics):
    num_pred = len(l_predictors)
    num_metrics = len(l_metrics)

    # gather the metrics
    metric_result = np.zeros((num_pred, num_metrics))

    for i, pred in enumerate(l_predictors):
        metric_result[i][:] = get_metric(pred, dataset, l_metrics)

        print(metric_result)

    return metric_result


def create_table_from_metrics(metric_result, l_pred_names, l_eval_metric):
    metric_result = np.round(metric_result, 4)

    # create the table
    fig, ax = plt.subplots()

    # hide axes
    fig.patch.set_visible(False)
    ax.axis('off')
    # ax.axis('tight')

    colors = [['w' for i in range(len(l_eval_metric))] for i in range(len(l_pred_names))]
    for i in range(len(l_eval_metric)):
        j = np.argmin(metric_result[:, i])
        colors[j][i] = 'g'
    for i in range(len(l_eval_metric)):
        j = np.argmax(metric_result[:, i])
        colors[j][i] = 'r'

    ax.table(cellText=metric_result, cellColours=colors, colLabels=l_eval_metric, rowLabels=l_pred_names, loc='center')

    fig.tight_layout()

    plt.show()


metric_result = gather_metrics(l_predictors, test_ds_miss, l_eval_metric)


create_table_from_metrics(metric_result, l_pred_names, l_eval_metric)
```


conftest.py
```diff
@@ -125,7 +125,7 @@ def pytest_runtest_makereport(item, call):
     setattr(item, "rep_" + rep.when, rep)
 
 
-@pytest.fixture(scope="function", autouse=True)
+# @pytest.fixture(scope="function", autouse=True)
```
lostella, line 128: I guess this change is unintended?

src/gluonts/transform/feature.py
```diff
@@ -45,6 +45,12 @@ class AddObservedValuesIndicator(SimpleTransformation):
         Field name to use for the indicator
     dummy_value
         Value to use for replacing missing values.
+    imputation_method
+        Select the method to replace the missing values.
+        - "standard" to just replace them with the dummy_value
```
lostella, line 50: Maybe “dummy_value” is a more appropriate name for this option?

src/gluonts/transform/feature.py
```diff
@@ -45,6 +45,12 @@ class AddObservedValuesIndicator(SimpleTransformation):
...
+        - "mean" to replace them with the mean of the array
+        - "median" to replace them with the median of the array
+        _ "last_val" to replace them with the last non missing value
```
lostella, line 53: “last_value”

src/gluonts/model/deepar/_estimator.py
```diff
@@ -111,6 +111,13 @@ class DeepAREstimator(GluonEstimator):
     num_parallel_samples
         Number of evaluation samples per time series to increase parallelism during inference.
         This is a model optimization that does not affect the accuracy (default: 100)
+    imputation_method
+        Select the method to replace the missing values.
+        - "standard" to just replace them with the dummy_value
+        - "mean" to replace them with the mean
+        - "median" to replace them with the median
+        - "last_val" to replace them with the last non missing value
+        (default: "standard")
```
lostella, line 120: I think repeating this docstring for each estimator may not scale well (eg in case other options are added). Maybe we could mention the role of this option and the default value, and point to the docs for AddObservedIndicator to know more about the available options. What do you think?

src/gluonts/transform/feature.py
```diff
@@ -57,21 +63,58 @@ def __init__(
...
+        nan_indices = np.where(np.isnan(value))
+        value[nan_indices] = np.nanmean(value)
+        return value
+
+    def _median_impute_missing(self, value: np.ndarray) -> np.ndarray:
+        nan_indices = np.where(np.isnan(value))
+        value[nan_indices] = np.nanmedian(value)
+        return value
+
+    def _last_missing_values(self, value: np.ndarray) -> np.ndarray:
```
benidis, line 92: Maybe `_last_value_impute_missing` or something would be more consistent and accurate naming?

src/gluonts/transform/feature.py
```diff
@@ -57,21 +63,58 @@ def __init__(
...
+        return value
+
+    def _mean_impute_missing(self, value: np.ndarray) -> np.ndarray:
+        nan_indices = np.where(np.isnan(value))
+        value[nan_indices] = np.nanmean(value)
+        return value
+
+    def _median_impute_missing(self, value: np.ndarray) -> np.ndarray:
+        nan_indices = np.where(np.isnan(value))
+        value[nan_indices] = np.nanmedian(value)
```
lostella, line 89: I think that both mean and median, implemented this way, will “leak” future observations into past missing values. The running mean and median should be used in order to avoid that, but I’m not sure how efficiently these can be implemented?

conftest.py
Francois-Aubet: Actually it was, I just forgot to change it back.

My pytests would not run if this was not commented out.

src/gluonts/transform/feature.py
Francois-Aubet: Fair enough, I can change this to "dummy_value"

src/gluonts/model/deepar/_estimator.py
Francois-Aubet: Yeah sure, I was not sure how to do that best.

Sounds good, will change this.

src/gluonts/transform/feature.py
Francois-Aubet: Fair enough. Will edit.

jaheba:
I'm not sure adding a new argument to Estimators is the best approach.

Value imputation is something which is independent from the actual model used, isn't it?

lostella:
> I'm not sure adding a new argument to Estimators is the best approach.
> 
> Value imputation is something which is independent from the actual model used, isn't it?

The other option would be to wrap the dataset before feeding it into the estimator?

benidis:
> I'm not sure adding a new argument to Estimators is the best approach.
> 
> Value imputation is something which is independent from the actual model used, isn't it?

Not exactly, a model can be such that it imputes missing values through its defined transformations. I agree that not ALL estimators should have this. But the models that "support" the missing value imputation could have it. 

I don't have a strong opinion though on what is the best approach here.

lostella:
> Not exactly, a model can be such that it imputes missing values through its defined transformations. I agree that not ALL estimators should have this. But the models that "support" the missing value imputation could have it.
> 
> I don't have a strong opinion though on what is the best approach here.

I would add that many steps in the transformations could be ideally "taken out" of the estimator so that one has complete control on what happens. But I guess that's a design choice about estimators: the way we implemented them so far, they are meant to encapsulate everything that happens to the data, from its raw state to the loss/prediction.

I don't know, I think ultimately this is a bigger question about how all the steps of the model are exposed.

jaheba:
The problem I'm seeing is that the proposed changes fix the `imputation_method` values to the Estimator class. If we decide to add a new method, we would need to touch all classes again.

This goes against the [Open-closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle).

On the other hand, from an end-user's perspective, the current design makes a lot of sense. When I use an estimator, I don't care how things are implemented, I want to have a unified interface.

We have a somewhat similar situation with the `Trainer`-class. It is an argument of estimators, but we separate internal responsibilities. When used with SageMaker, we pass all hyper-parameters to `Estimator` and `Trainer` during initialisation.

---

I think what we can do, is to still add a new argument to Estimators to enable imputation, but do so by delegating handling of the values to a new class:

```python
class ImputationStrategy(str, Enum):
    none = "none"
    standard = "standard"
    mean = "mean"
    median = "median"
    last_val = "last_val"

    def impute(self, value):
        ...
```

and then use it like this:

```python
def __init__(self,
    ...
    imputation_method=ImputationStrategy.none,
    ...
)
```

This has the advantage that `pydantic` automatically does the handling of strings to `ImputationStrategy`, and the outside usage does not change.

jaheba:
> I don't know, I think ultimately this is a bigger question about how all the steps of the model are exposed.

Yes! I think the boundary of what belongs to the model and what should be considered pre-processing is not always clear. I think something similar can be said about the `Trainer`.

Francois-Aubet:
Ok, I think that this is a good idea, and we could even add this to the GluonEstimator class, what do you think?
With the addition of the "none" option with would work.


> The problem I'm seeing is that the proposed changes fix the `imputation_method` values to the Estimator class. If we decide to add a new method, we would need to touch all classes again.
> 
> This goes against the [Open-closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle).
> 
> On the other hand, from an end-user's perspective, the current design makes a lot of sense. When I use an estimator, I don't care how things are implemented, I want to have a unified interface.
> 
> We have a somewhat similar situation with the `Trainer`-class. It is an argument of estimators, but we separate internal responsibilities. When used with SageMaker, we pass all hyper-parameters to `Estimator` and `Trainer` during initialisation.
> 
> I think what we can do, is to still add a new argument to Estimators to enable imputation, but do so by delegating handling of the values to a new class:
> 
> ```python
> class ImputationStrategy(str, Enum):
>     none = "none"
>     standard = "standard"
>     mean = "mean"
>     median = "median"
>     last_val = "last_val"
> 
>     def impute(self, value):
>         ...
> ```
> 
> and then use it like this:
> 
> ```python
> def __init__(self,
>     ...
>     imputation_method=ImputationStrategy.none,
>     ...
> )
> ```
> 
> This has the advantage that `pydantic` automatically does the handling of strings to `ImputationStrategy`, and the outside usage does not change.

vafl:
Nice, thanks!
A couple of comments:

I also think a class would be better. Ideally, we can make it more flexible, by having a generic base class e.g.

```python
class MissingValueImputation:
    def impute(self, values: np.ndarray) -> np.ndarray:
        raise NotImplementedError()
```
or just `__call__` and implement the specific logic in sub classes. Then the user can add whatever imputation they want, e.g. exponential smoothing etc.

I would prefer not to include this in the `Estimator` api, since this doesn't really gain or simplify  anything. Each subclass still has to call the `AddObservedValuesIndicator`, pass it along etc.

src/gluonts/transform/feature.py
```diff
@@ -30,6 +30,66 @@ def target_transformation_length(
...
 
 
+class ImputationStrategy:
+    none = "none"
+    dummy_value = "dummy_value"
+    mean = "mean"
+    median = "median"
+    last_value = "last_value"
+
+    @staticmethod
```
jaheba, line 40: We do not want to define static methods.

src/gluonts/transform/feature.py
```diff
@@ -30,6 +30,66 @@ def target_transformation_length(
...
+    def impute(values: np.ndarray, method: str, dummy_value: float = 0.0) -> np.ndarray:
```
jaheba, line 41: Change this to

```python
def impute(self, values: np.ndarray,, dummy_value: float = 0.0) -> np.ndarray:
        ...
        return dic_imput_methods[self](values, dummy_value)
```

Francois-Aubet:
I put the methods and different options in an outer class.

I put it here to have your opinion.

I am not sure I would have framed it with a parent class for all imputation methods, because then we would have still needed another class for the enum.

I have not implemented yet the "causal" mean and median, I am still looking into how to best code these.
Also, I will apply the same changes to the _estimator.py of the other models if you agree with the change of the one of DeepAR.

src/gluonts/transform/feature.py
```diff
@@ -30,6 +30,66 @@ def target_transformation_length(
     return target.shape[-1] + (0 if is_train else pred_length)
 
 
+class ImputationStrategy:
+    none = "none"
+    dummy_value = "dummy_value"
+    mean = "mean"
+    median = "median"
+    last_value = "last_value"
```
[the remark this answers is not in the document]
Francois-Aubet: Thanks! Addressed.

src/gluonts/transform/feature.py
Francois-Aubet: Thanks! Fixed.

src/gluonts/transform/feature.py
Francois-Aubet: Thanks! Fixed.

vafl:
> I am not sure I would have framed it with a parent class for all imputation methods, because then we would have still needed another class for the enum.

Why? We can just pass an instance of the class. See e.g. how the `Trainer` is configured as a parameter of the estimator. This would make things more flexible.

Francois-Aubet:
> Why? We can just pass an instance of the class. See e.g. how the `Trainer` is configured as a parameter of the estimator. This would make things more flexible.

Is see this, but so you mean that we would just have the class instance as parameter and no string? So, just like the Trainer?

I am not sure if it is worth having something like that as it may make it more complicated for users that would just want to pick an existing method, but it is actually fine. If we all agree on this, then sure.

Francois-Aubet:
I made the change suggested by Valentin.

Please let me know what you think of this organisation, if we are all happy with it, then I will change the other estimators and improve the tests.

src/gluonts/model/deepar/_estimator.py
```diff
@@ -134,6 +141,7 @@ def __init__(
         lags_seq: Optional[List[int]] = None,
         time_features: Optional[List[TimeFeature]] = None,
         num_parallel_samples: int = 100,
+        imputation_method: MissingValueImputation = None,
```
jaheba, line 144: This should `Optional[MissingValueImputation]` then.

Francois-Aubet: True, thanks!

src/gluonts/model/deepar/_estimator.py
```diff
@@ -34,6 +34,7 @@
 from gluonts.trainer import Trainer
 from gluonts.transform import (
     AddAgeFeature,
+    ImputationStrategy,
```
lostella, line 37: Is this class defined?

src/gluonts/model/deepstate/_estimator.py
```diff
@@ -174,6 +181,7 @@ def __init__(
         noise_std_bounds: ParameterBounds = ParameterBounds(1e-6, 1.0),
         prior_cov_bounds: ParameterBounds = ParameterBounds(1e-6, 1.0),
         innovation_bounds: ParameterBounds = ParameterBounds(1e-6, 0.01),
+        imputation_method: str = "standard",
```
lostella, line 184: So maybe we should remove this string-based selection. I would say we only modify the DeepAREstimator for now, using the MissingValueImputation constructor argument, and then we go about updating the rest

src/gluonts/model/deepvar/_estimator.py
```diff
@@ -226,6 +232,7 @@ def __init__(
         time_features: Optional[List[TimeFeature]] = None,
         conditioning_length: int = 200,
         use_marginal_transformation=False,
+        imputation_method: str = "standard",
```
lostella, line 235: Same as for DeepState

src/gluonts/model/deepvar/_estimator.py
```diff
@@ -296,6 +295,7 @@ def __init__(
         else:
             self.output_transform = None
 
+
```
lostella, line 298:
```suggestion
```

test/test_transform.py
```diff
@@ -879,6 +879,67 @@ def test_ctsplitter_train_short_intervals(point_process_dataset):
...
+        assert np.array_equal(d_expected_result[method], res["target"])
+        assert np.array_equal(
+            expected_missindicator, res[FieldName.OBSERVED_VALUES]
+        )
+
+
+
+
+
+
```
lostella, line 942: These blank lines should probably go

src/gluonts/transform/feature.py
```diff
@@ -30,6 +30,126 @@ def target_transformation_length(
     return target.shape[-1] + (0 if is_train else pred_length)
 
 
+class MissingValueImputation:
+    """
+    The parent class for all the missing value imputation classes.
+    You can just implement your own inheriting this class.
+    """
+
+    def __init__(self) -> None:
```
lostella, line 39: You probably need to decorate this constructor with the `@validated()` decorator to make this type compatible with the serialization/deserialization mechanism that is used. See how this is done e.g. here: https://github.com/awslabs/gluon-ts/blob/60a0d5ca1ba6d1d062aab3065b6e9c68bf45a31f/src/gluonts/transform/feature.py#L54

src/gluonts/transform/feature.py
```diff
@@ -30,6 +30,126 @@ def target_transformation_length(
...
+        raise NotImplementedError()
+
+
+class LeavesMissingValues(MissingValueImputation):
+    def __call__(self, values: np.ndarray) -> np.ndarray:
+        return values
+
+
+class DummyValueImputation(MissingValueImputation):
+    def __init__(self, dummy_value: float = 0.0) -> None:
```
lostella, line 52: `@validated()`

src/gluonts/transform/feature.py
```diff
@@ -30,6 +30,126 @@ def target_transformation_length(
...
+        values[mask] = ar_res[mask]
+
+        # make sure that we do not leave the potential nan in the first position:
+        values[0] = value_no_nans[0]
+
+        return values
+
+
+class RollingMeanValueImputation(MissingValueImputation):
+    def __init__(self, window_size: int = 10) -> None:
```
lostella, line 121: `@validated()`

lostella:
@Francois-Aubet are you sure you have run the `./dev_setup.sh` script? The style checks are failing, and you should get an error message about those when you try to commit. Also make sure you run

```
pip install -r requirements/requirements-setup.txt
```

to have the right versions of mypy and black installed.

test/test_transform.py
```diff
@@ -879,6 +879,61 @@ def test_ctsplitter_train_short_intervals(point_process_dataset):
...
+            target_field=FieldName.TARGET,
+            output_field=FieldName.OBSERVED_VALUES,
+            imputation_method=d_method_instances[method],
+        )
+
+        d = {"target": array_value.copy()}
+
+        res = transfo.transform(d)
+
+        print(res)
```
lostella, line 929: This should be removed I guess

test/test_transform.py
```diff
@@ -879,6 +879,61 @@ def test_ctsplitter_train_short_intervals(point_process_dataset):
...
+    """
+    from gluonts.transform import (
+        MissingValueImputation,
+        LeavesMissingValues,
+        DummyValueImputation,
+        MeanValueImputation,
+        LastValueImputation,
+        CausalMeanValueImputation,
+        RollingMeanValueImputation,
+    )
```
lostella, line 894: It's maybe better to import this on top of the script

test/test_transform.py
```diff
@@ -879,6 +879,61 @@ def test_ctsplitter_train_short_intervals(point_process_dataset):
...
+
+    l_methods = ["dummy_value", "mean", "causal_mean", "last_value"]
+
+    d_method_instances = {
+        "dummy_value": DummyValueImputation(),
+        "mean": MeanValueImputation(),
+        "causal_mean": CausalMeanValueImputation(),
+        "last_value": LastValueImputation(),
+    }
```
lostella, line 907: Is RollingMeanValueImputation not tested?

src/gluonts/transform/feature.py
```diff
@@ -30,6 +30,129 @@ def target_transformation_length(
...
+        nan_indices = np.where(np.isnan(values))
+        values[nan_indices] = self.dummy_value
+        return values
+
+
+class MeanValueImputation(MissingValueImputation):
+    """
+    Careful this is not a 'causal' method in the sense that it leaks information about the furture in the imputation.
+    You may prefer to use CausalMeanValueImputation instead.
+    """
```
lostella, line 67: It would be nice to have a description of what these classes do: maybe you could have a docstring here, for each classes, clarifying what the __call__ method does

lostella:
Looks like you need to run black against the code once more:

```
black src test
```

And commit the changes. Make sure you are using the versions of black and mypy listed in: https://github.com/awslabs/gluon-ts/blob/master/requirements/requirements-setup.txt

```
pip install -r requirements/requirements-setup.txt
./dev_setup.sh
```

After running the above, you shouldn’t be able to commit code if it doesn’t make black happy

Issue #831, opened 2020-05-18
Implementing rolling forecasts


arangatang:
## Description

Implementing error evaluation using rolling forecasts (also known as time series cross validation) would bring more accurate error metrics to Gluon-TS. 

There are a couple of different ways to implement this:
1. Implement this to run it in memory in `backtest.py`
2. Edit predictors such that they reuse target values in datasets
3. Rerun backtest_metrics several times with different datasets
4. Create a new dataset class

Below I present these ideas in further detail.
## Method 1 edit backtest.py to run rolling forecasts in memory with multiple datasets
It would work something like this:

1. Model is trained on training data as usual
2. If flag is set for using rolling_forecasting, generate new dataset for testing
   1. for each timeseries in the test dataset 
      1. if timeseries fullfill some quality criterion (length, frequency, etc)
         1. generate rolling forecast dataset with X different "prediction windows" 
3. Generate forecasts using the predictor on the new dataset which is X times larger than the original one
4. Some metrics should be extracted based on both individual (original) timeseries and across all timeseries

Where would this be implemented in GluonTS:
Most of the changes would be made in [backtest.py](https://github.com/awslabs/gluon-ts/blob/5ecf0ea772f4109cd937c5a316304e36d38012a8/src/gluonts/evaluation/backtest.py)

1. is already handled and is a non-issue. 
2. can be implemented [here](https://github.com/awslabs/gluon-ts/blob/5ecf0ea772f4109cd937c5a316304e36d38012a8/src/gluonts/evaluation/backtest.py#L101) by truncating the same dataset multiple times and merging the results. See below for some pseudo code:

    ```python
    # define the timeseries used for forecasts
    dataset_rolling_trunc = dataset_trunc = TransformedDataset(dataset, transformations=[transform.AdhocTransform(truncate_target)])

    # X is the number of forecasts that should make up the rolling forecast window
    for step in range(X):
        
        # generate test set
        if(step == X-1):
            dataset = dataset + dataset_rolling_trunc
        
        dataset_trunc = TransformedDataset(
            dataset_trunc, transformations=[transform.AdhocTransform(truncate_target)]
        )
        dataset_rolling_trunc = dataset_rolling_trunc + dataset_trunc # This may be too slow
    ```

3. The predictor can then predict as normally
4. The evaluator class need changes
   1. It could be useful to generate error metrics on a per timeseries basis and not only for all timeseries. Thus it would need to be able to identify which series are copies. 

### cons

* High memory overhead

[...]

PR #844, opened 2020-05-26
Implemented generate_rolling_datasets

arangatang:
*Issue #, if available:* #831

*Description of changes:*
### Exposed a method for generating datasets for rolling forecasting evaluations. 

the method `generate_rolling_datasets` takes a dataset, two dates and a window size. Thereafter it does the following:

1. It truncates any target values in the dataset which appears after the last of the provided dates. Since these will not be used neither when evaluating nor when forecasting they are superfluous. 
2. Generates a dataset used when evaluating the rolling forecasts. This dataset currently contains copies of the original timeseries. This will possibly be changed to instead contain only the values that is to be compared with the forecasts and not extra values. 
3. The method generates the dataset to use for generating forecasts on a rolling origin based on the window size and the provided time range. Any remainders is handled as a new window of a smaller size
4. the datasets are returned as generators contained in a dictionary

TODO
1. Debug current implementation
2. Write tests 
3. Possibly rewrite how the `to_evaluate` dataset is generated (depends on how the `evaluator` handles superfluous values.)
4. Implement `evaluator` functionality to handle rolling forecasts
5. write tests for `evaluator`
6. integrate into `make_evaluation_predictions`
7. integrate into `backtest_metrics`


src/gluonts/evaluation/backtest.py
```diff
@@ -38,9 +38,148 @@
...
+
+    Returns
+    -------
+    dict 
+        Contains the dataset to provide to the predictor and the dataset to provide to the evaluator
+    """
+
+    assert(dataset, start_time, end_time, window_size)
+    
+    _freq = start_time.freq
```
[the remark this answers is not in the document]
arangatang: To make it easier to see which scope it is accessible from.

src/gluonts/evaluation/backtest.py
```diff
@@ -38,9 +38,148 @@
...
+        #calc number datapoints needed from start of timeseries until end of rolling test range
+        timerange = pd.date_range(
+            start=data['start'], 
+            end=end_time, 
+            freq=_freq
+        )
+        max_items = len(timerange)
+
+        # keep values until end of test range
+        if max_items < len(data['target']):
```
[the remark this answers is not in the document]
arangatang: Ok, good to know will change this

src/gluonts/evaluation/backtest.py
```diff
@@ -38,9 +38,148 @@
...
+            data['target'] = data['target'][:max_items]
+        
+        return data
+    
+    # generator to create rolling datasets
+    def perform_roll(dataset):
+        for timeseries in dataset:
+            windows_todo = _num_window_iterations
+            window_start_date = end_time
+            while windows_todo:
```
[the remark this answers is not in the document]
arangatang: Cause we use time series a few lines further down

src/gluonts/evaluation/backtest.py
```diff
@@ -38,9 +38,148 @@
...
+                _ts = timeseries.copy()
+                
+                # 1. calc length of target 
+                target_length = len(_ts['target'])
```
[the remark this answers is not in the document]
arangatang: it is used 6 times

src/gluonts/evaluation/backtest.py
```diff
@@ -38,9 +38,148 @@
...
+                # 4. yield the truncated/rolled timeseries even if it is "not touched" 
+                windows_todo = windows_todo - 1
+                yield _ts
+
+    # creates a generator for a evaluation dataset of the same 
+    # number timeseries as the rolling set
+    def generate_test(dataset):
+        for timeseries in dataset:
+            ts = timeseries.copy()
+            for _ in range(_num_window_iterations):
```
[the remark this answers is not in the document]
arangatang: if we perform in total N rolls then this will output N copies of each timeseries. This is to ensure the same amount of timeseries of the dataset used for prediction and evaluation.

src/gluonts/evaluation/backtest.py
```diff
@@ -38,9 +38,148 @@
...
+                
+                # 4. yield the truncated/rolled timeseries even if it is "not touched" 
+                windows_todo = windows_todo - 1
+                yield _ts
+
+    # creates a generator for a evaluation dataset of the same 
+    # number timeseries as the rolling set
+    def generate_test(dataset):
+        for timeseries in dataset:
+            ts = timeseries.copy()
```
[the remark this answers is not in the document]
arangatang: I thought we would need copies of it for evaluation. Do you recon that we can simply reuse the same timeseries over and over?
 I considered reusing them but assumed it would be safer to provide copies as to ensure that values are not overwritten at a later stage.

src/gluonts/evaluation/backtest.py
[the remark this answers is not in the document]
arangatang: Not needed, thought it would make the code as a whole more readable

src/gluonts/evaluation/backtest.py
```diff
@@ -38,9 +38,148 @@
...
+    end_time
+        The end of the time period which rolling forecasts should be applied
+
+    Returns
+    -------
+    dict 
+        Contains the dataset to provide to the predictor and the dataset to provide to the evaluator
+    """
+
+    assert(dataset, start_time, end_time, window_size)
```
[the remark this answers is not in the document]
arangatang: Ok, good to know.
This was a temporary check in any case, going to expand it to use proper error messages for each parameter in any case.

src/gluonts/evaluation/backtest.py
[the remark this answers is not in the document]
jaheba: > This is to ensure the same amount of timeseries of the dataset used for prediction and evaluation.

Why should we enforce the same amount of time-series?

src/gluonts/evaluation/backtest.py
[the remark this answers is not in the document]
jaheba: Strong disagree. This makes it more noisy in my view.

src/gluonts/evaluation/backtest.py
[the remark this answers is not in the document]
jaheba: If that is an issue, then that means we should re-think the design and not fix it by introducing some new convention.

examples/run_rolling_forecast_backtest.py
```diff
@@ -0,0 +1,156 @@
...
+    forecast_it, ts_it = make_evaluation_predictions(
+        dataset.test, predictor=predictor, num_samples=100, rolling_time_range=(t0,t1)
+    )
+
+    agg_metrics, item_metrics = Evaluator()(
+        ts_it, forecast_it, num_series=len(dataset.test)
+    )
+
+    pprint.pprint(agg_metrics)
+    '''
```
jaheba, line 156: Always have a new line at the end of the file!

src/gluonts/evaluation/backtest.py
```diff
@@ -38,9 +38,148 @@
...
 from mxnet.ndarray import NDArray
 
+def generate_rolling_datasets(
+    dataset: Dataset,
+    window_size: int,
+    start_time: pd.Timestamp, 
+    end_time: pd.Timestamp
+    ) -> dict:
+    """
+    Returns a dictionary containing two dataset generators for using when performing rolling origin evaluations.
```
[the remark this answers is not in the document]
arangatang: Fixing this

src/gluonts/evaluation/backtest.py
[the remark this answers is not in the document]
jaheba: You got me more confused.

src/gluonts/evaluation/backtest.py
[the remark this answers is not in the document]
arangatang: I will fix this, so ill resolve this comment

src/gluonts/evaluation/backtest.py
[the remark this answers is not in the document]
arangatang: okay, ill change this then

src/gluonts/evaluation/backtest.py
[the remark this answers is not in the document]
arangatang: oki, then ill fix this

src/gluonts/evaluation/backtest.py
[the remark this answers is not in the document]
arangatang: sorry my bad, thought you meant we should change line 121 while you were speaking of line 124. 

Applying your change should work yes.

src/gluonts/evaluation/backtest.py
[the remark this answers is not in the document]
arangatang: This method is a little bit work in progress as it depends on how the evaluator works/should be changed, I only quickly skimmed the evaluator as of now.  
By providing this restrictions we (should) simplify the logic of the evaluator as it can compare timeseries on a 1-by-1 basis, i.e. 1 forecast to 1 real-timeseries without needing to calculate which real-timeseries belongs to which forecast.

examples/run_rolling_forecast_backtest.py
arangatang: yep, missed that

src/gluonts/evaluation/backtest.py
```diff
@@ -38,9 +38,148 @@
...
+                transform.AdhocTransform(remove_remainder)
+            ]
+        )
+    dataset_rolled = perform_roll(dataset)
+
+    # we now have the shortened test dataset and we need to generate the rolls
+    #dataset_rolled = TransformedDataset(
+    #        dataset_cut_remainder, transformations=[transform.AdhocTransform(perform_roll)]
+    #    )
+    return {'to_predict':dataset_rolled, 'to_evaluate':dataset_eval}
```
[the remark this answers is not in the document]
arangatang: Agree, fixed it in the last commit

src/gluonts/evaluation/backtest.py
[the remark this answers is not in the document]
arangatang: Implemented in the latest commit

src/gluonts/evaluation/backtest.py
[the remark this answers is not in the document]
arangatang: discussed offline, this will be remade for next commit

jaheba:
I think using `pd.Series` objects would make this code easier.

What do you think about this approach:

```python
from functools import partial
from typing import Optional

import numpy as np
import pandas as pd

from gluonts.dataset.common import Dataset


def generate_rolling_datasets(
    dataset: Dataset,
    strategy,
    start_time: pd.Timestamp,
    end_time: Optional[pd.Timestamp] = None,
) -> Dataset:

    for item in dataset:
        series = to_pandas(item, start_time.freq)
        base = series[:start_time][:-1].to_numpy()
        prediction_window = series[start_time:end_time]

        for window in strategy(prediction_window):
            new_item = item.copy()
            new_item["target"] = np.concatenate([base, window.to_numpy()])
            yield new_item


def xstrat(window, prediction_length):
    yield window


print(
    list(
        generate_rolling_datasets(
            [{"start": "2020", "target": list(range(1, 101))}],
            start_time=pd.Timestamp("2020-02", freq="D"),
            strategy=partial(xstrat, prediction_length=10),
        )
    )
)
```

arangatang:
I like the suggestion, however I have some concerns:

1. since `yield` only generates once we would need to generate the dataset at least twice, once for making the evaluations and once for generating the forecasts. If one would like to run `make_evaluation_predictions` on the complete generated dataset we need to report the amount of time series in the dataset. This is also difficult when using a generator as it is not known until after it is consumed.
2. I like the strategy approach that you propose as it allows for new variations in the future, however using a partial strategy as it is now is not as user friendly as a simple function call imo, how about this:
```python3
def xstrat(prediction_length):
    def part_function(window, prediction_length):
        # whathever logic to create the window here
        yield window
    return partial(part_function, prediction_length=prediction_length)
```
3. the `to_pandas` function does indeed simplify it

src/gluonts/evaluation/backtest.py
```diff
@@ -39,8 +41,147 @@
...
+            if number_rolls < 0:
+                number_rolls = 0
+
+        # print('number_rolls:', number_rolls, '\nitems_to_roll:', items_to_roll, '\nmodifier:', modifier)
+
+        return number_rolls, modifier
+
+    # generator to create rolling datasets
+    def perform_roll(dataset):
+        ds = []
```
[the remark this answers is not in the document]
arangatang: In order to reuse the values.

We may need to reuse the dataset up to two times
1. when forecasting using the dataset
2. when fetching the length of the dataset one would need to count all the values see `num_samples` parameter to `make_evaluation_predictions`

If we can precalculate the number of entries in the dataset and return the length as a separate variable we can use yield.

src/gluonts/evaluation/backtest.py
```diff
@@ -39,8 +41,211 @@
...
+        The augmented dataset
+    """
+
+    assert dataset, "a dataset to perform rolling evaluation on is needed"
+    assert prediction_length, "prediction_length is needed"
+    assert prediction_length > 0, "prediction length needs to be > 0"
+    assert start_time, "a pandas Timestamp object is needed for the start time"
+
+    freq = start_time.freq
+    if ignore_end:
```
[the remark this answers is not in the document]
arangatang: This would require the user to explicitly ignore the end time, thus the user would get a warning in case they simply forgot to add a end_time.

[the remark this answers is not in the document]
arangatang: Thought it would be more intuitive for a user, and less error prone

src/gluonts/evaluation/backtest.py
```diff
@@ -39,8 +41,211 @@
...
+
+        # ensure that the window size is small enough to allow rolling
+        assert (
+            num_items_in_rolling > prediction_length
+        ), "timestamps to close for the prediction_length"
+
+    # removes target values appearing after end_time
+    def truncate_end(data):
+        data = data.copy()
+        data["rolling"] = (start_time, end_time)
```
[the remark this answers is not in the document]
arangatang: Nope, should be removed, thanks

src/gluonts/evaluation/backtest.py
```diff
@@ -39,8 +41,211 @@
...
+    start_time: pd.Timestamp,
+    end_time: Optional[pd.Timestamp] = None,
+    use_unique_rolls: Optional[bool] = False,
+    ignore_end: Optional[bool] = False,
+) -> Dataset:
+    """
+    Returns an augmented version of the input dataset where each timeseries has 
+    been rolled upon based on the parameters supplied. See below for examples of
+    the behaviour of the function and what effect the parameters has on its output.
+
```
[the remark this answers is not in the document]
arangatang: Makes sense

src/gluonts/evaluation/backtest.py
```diff
@@ -39,8 +41,147 @@
...
+            # perform the rolling
+            for _ in range(number_rolls):
+                ts_copy = ts.copy()
+                ts_copy["target"] = ts_copy["target"][:new_length]
+                new_length = new_length - modifier
+                ds.append(ts_copy)
+
+        return ds
+
+    dataset_to_roll = dataset
```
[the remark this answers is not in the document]
arangatang: yep, agree

jaheba:
I think it's generally fine to iterate multiple times over the dataset. Then again, I don't like our dataset abstraction too much and would like to change how `make_evaluation_predictions` works in the future.

The strategy approach does not require the use of partial functions. You could also define some classes for this. Thinking about it, we could even shift the responsibility of ensuring that the prediction_length is considered to it.

src/gluonts/evaluation/backtest.py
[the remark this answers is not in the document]
arangatang: Fixed in latest commit

src/gluonts/evaluation/backtest.py
```diff
@@ -36,11 +35,10 @@
 from gluonts.model.predictor import GluonPredictor, Predictor
 from gluonts.support.util import maybe_len
 from gluonts.transform import TransformedDataset
-from mxnet.ndarray import NDArray
 
 
 def make_evaluation_predictions(
-    dataset: Dataset, predictor: Predictor, num_samples: int
+    dataset: Dataset, predictor: Predictor, num_samples: int,
```
jaheba, line 41: I think these changes shouldn't be commited.

src/gluonts/dataset/rolling_dataset.py
```diff
@@ -0,0 +1,197 @@
...
+    strategy
+        The strategy that is to be used when rolling
+    start_time
+        The start of the window where rolling forecasts should be applied
+    end_time
+        The end time of the window where rolling should be applied
+    Returns
+    -------
+    Dataset 
+        The augmented dataset
```
jaheba, line 106: This should be at the top of the docstring.

src/gluonts/evaluation/backtest.py
arangatang: sure, since I moved all changes to other files than `backtest.py` I will simply revert all changes made to backtest to when I started the implementation instead.

src/gluonts/dataset/rolling_dataset.py
arangatang: Oki, thought it needed to be in this order, will change it

test/dataset/test_rolling.py
```diff
@@ -0,0 +1,306 @@
...
+    else:
+        raise ValueError
+    return dataset
+
+
+@pytest.mark.parametrize(
+    "prediction_length, unique",
+    [(p, u) for p in [-1, 0] for u in [True, False]],
+)
+def test_fails(prediction_length, unique):
```
jaheba, line 245: I think that's a bad test-name. When you want to select tests with `-k` you generally want to use more descriptive names.

test/dataset/test_rolling.py
```diff
@@ -0,0 +1,306 @@
...
+            dataset=generate_dataset("constant"),
+            start_time=pd.Timestamp("2000-01-01-20", freq="1H"),
+            end_time=pd.Timestamp("2000-01-02-00", freq="1H"),
+            strategy=StepStrategy(
+                step_size=prediction_length if unique else 1,
+                prediction_length=prediction_length,
+            ),
+        )
+        # program should have failed at this point
+        raise RuntimeWarning
```
jaheba, line 257: This looks wrong. There is `pytest.raises` to ensure that a test-case fails.

test/dataset/test_rolling.py
```diff
@@ -0,0 +1,306 @@
...
+            [25, 23],  # test 2
+            [23],  # test 3
+            [],  # test 4
+            [],  # test 5
+            [5, 3],  # test 6
+            [3],  # test 7
+            [],  # test 8
+            [],  # test 9
+            [3],  # test 10
+        ]
```
jaheba, line 133: I don't like this. Data for test-cases should be defined together, not at different places.

test/dataset/test_rolling.py
```diff
@@ -0,0 +1,306 @@
...
+from math import floor
+import pandas as pd
+from gluonts.dataset.artificial import constant_dataset
+from gluonts.dataset.common import ListDataset
+from gluonts.dataset.rolling_dataset import (
+    StepStrategy,
+    generate_rolling_datasets,
+)
+import pytest
+import itertools
```
jaheba, line 27: There is an ordering for imports. See https://www.python.org/dev/peps/pep-0008/#imports

test/dataset/test_rolling.py
```diff
@@ -0,0 +1,306 @@
...
+            {  # test 6: starts on start ends after end, ts == t0, te > t1
+                "target": [0.0 for i in range(10)],
+                "start": pd.Timestamp(2000, 1, 1, 20, 0),
+            },
+            {  # test 7: starts in between ts and te, ts < t0 < te < t1
+                "target": [0.0 for i in range(10)],
+                "start": pd.Timestamp(2000, 1, 1, 22, 0),
+            },
+            {  # test 8: starts on end time, te == t0
+                "target": [0.0 for i in range(10)],
```
jaheba, line 223: You can abbreviate these with

```suggestion
                "target": [0.0] *  10],
```

src/gluonts/dataset/rolling_dataset.py
```diff
@@ -0,0 +1,177 @@
...
+    [1, 2, 3, 4, 5, 6, 7]
+    """
+    assert dataset, "a dataset to perform rolling evaluation on is needed"
+    assert start_time, "a pandas Timestamp object is needed for the start time"
+    assert strategy, """a strategy to use when rolling is needed, for example
+        gluonts.dataset.rolling_dataset.StepStrategy"""
+    if end_time:
+        assert end_time > start_time, "end time has to be after the start time"
+
+    ds = []
```
lostella, line 166: Sorry to jump in late, not sure this was already discussed: does this need to be a list? Couldn't we use a generator here?

jaheba: I think we at least discussed this offline.

The issue is that we need to run this twice, when using `backtest.make_evaluation_predictions`. But maybe that's an issue of `make_evaluation_predictions`

src/gluonts/dataset/rolling_dataset.py
```diff
@@ -0,0 +1,173 @@
...
+    strategy
+        The strategy that is to be used when rolling
+    start_time
+        The start of the window where rolling forecasts should be applied
+    end_time
+        The end time of the window where rolling should be applied
+    Returns
+    -------
+    Dataset
+        The augmented dataset
```
lostella, line 87: I think the convention is that these go to the bottom of the docstring

src/gluonts/dataset/rolling_dataset.py
```diff
@@ -0,0 +1,173 @@
...
+    def get_windows(self, window):
+        assert num_splits > 1, """num_splits should be > 1"""
+        for slice_idx in np.linspace(
+            start=self.prediction_length, stop=len(window), num=self.num_splits
+        ):
+
+            yield window[: int(round(slice_idx))]
+
+
+def generate_rolling_datasets(
```
lostella, line 67:
```suggestion
def generate_rolling_dataset(
```

src/gluonts/dataset/rolling_dataset.py
```diff
@@ -0,0 +1,173 @@
...
+    The 'rolling' will happen on the data available in the provided window between the
+    start_time and the end_time for each timeseries. If end_time is omitted, rolling
+    happens on all datapoints from start_time until the end of the timeseries.
+    The way the data is rolled is governed by the strategy used.
+
+    Below examples will be based on this one timeseries long dataset
+    [{
+    target=[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15]
+    start='2000-1-1-01'
+    }]
```
lostella, line 103:
```suggestion
    [{
    "target": np.array([1,2,3,4,5,6,7,8,9,10,11,12,13,14,15]),
    "start": pd.Timestamp('2000-1-1-01', freq='1H')
    }]
```

src/gluonts/dataset/rolling_dataset.py
```diff
@@ -0,0 +1,173 @@
...
+
+            yield window[: int(round(slice_idx))]
+
+
+def generate_rolling_datasets(
+    dataset: Dataset,
+    strategy,
+    start_time: pd.Timestamp,
+    end_time: Optional[pd.Timestamp] = None,
+) -> Dataset:
```
lostella, line 72: Maybe minor, but we can keep a #TODO comment here about having the options to specify which fields should be rolled: for now it's just `"target"`, but one may need to roll other arrays that have a time axis, like time dependent features

src/gluonts/dataset/rolling_dataset.py
arangatang: Sounds good

src/gluonts/dataset/rolling_dataset.py
arangatang: This contradicts what @jaheba reviewed on the 8th of June, the examples I found online also has them on the bottom but im not sure. @jaheba what do you think?

src/gluonts/dataset/rolling_dataset.py
arangatang: Ill apply this

src/gluonts/dataset/rolling_dataset.py
arangatang: Ill add one

src/gluonts/dataset/rolling_dataset.py
```diff
@@ -0,0 +1,174 @@
...
+class StepStrategy(BaseModel):
+    """
+    Removes datapoints equivalent to step_size for each iteration until
+    amount of data left is less than prediction_length
+
+    Parameters
+    ----------
+    prediction_length
+        The prediction length of the Predictor that the dataset will be
+        used with
```
lostella, line 32: description of `step_size` is missing here

src/gluonts/dataset/rolling_dataset.py
```diff
@@ -0,0 +1,174 @@
...
+        assert self.step_size > 0, """step_size should be > 0"""
+
+        while len(window) >= self.prediction_length:
+            yield window
+            window = window[: -self.step_size]
+
+
+class NumSplitsStrategy(BaseModel):
+    prediction_length: int
+    num_splits: int
```
lostella, line 55: I think it would be good to have a docstring similar to the one for the other strategy

src/gluonts/dataset/rolling_dataset.py
```diff
@@ -0,0 +1,174 @@
...
+
+# TODO Add parameter allowing for rolling of other arrays
+# with a time axis such as time dependent features
+def generate_rolling_dataset(
+    dataset: Dataset,
+    strategy,
+    start_time: pd.Timestamp,
+    end_time: Optional[pd.Timestamp] = None,
+) -> Dataset:
+    """
```
lostella, line 74: Also having at least a high-level description here would be nice

src/gluonts/dataset/rolling_dataset.py
lostella: That's fine, as long as it displays nicely when API docs are generated, I can take a look how sphinx reads it

lostella: it looks like this is displayed weirdly:

image

src/gluonts/dataset/rolling_dataset.py
arangatang: Done

src/gluonts/dataset/rolling_dataset.py
arangatang: Added it

src/gluonts/dataset/rolling_dataset.py
arangatang: Updated the formatting in the latest commit, it looks nice on my machine currently :)

src/gluonts/dataset/rolling_dataset.py
arangatang: I recon this got solved by moving things around, let me know if you want me to expand upon it.

arangatang:
Im not sure why this fails, it seems unrelated to my changes

PascalIversen:
@arangatang Yes, the failing tests are unrelated, addressing this here #1147

lostella (approved):

PR #849, opened 2020-05-28
Fix the quantiles used to compute MSIS

### LINKED ISSUE #848: MSIS computation is too implicit in Evaluator
## Description
The computation of MSIS in the Evaluator class depends on the Evaluator settings in a rather implicit way. This could be improved by taking as constructor arguments exactly the quantiles to be used as lower and upper bound for the MSIS computation, and using exactly those.

## References

#845

https://github.com/awslabs/gluon-ts/issues/845#issuecomment-635161515

lostella:
*Issue #, if available:* Fixes #848 

*Description of changes:* The quantiles used for computing MSIS were not really controlled by the alpha parameter of the Evaluator class, despite what the documentations says. This PR tries to fix that and make them entirely dependent on alpha.

### DIFF
--- a/src/gluonts/evaluation/_base.py
+++ b/src/gluonts/evaluation/_base.py
@@ -84,11 +84,11 @@ class Evaluator:
         uses the default seasonality
         for the given series frequency as returned by `get_seasonality`
     alpha
-        parameter of the MSIS metric from M4 competition that
-        defines the confidence interval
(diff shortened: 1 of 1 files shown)


lostella:
The failing test is unrelated to the PR, and concerns data loading throughout (it appears to be flaky on Jenkins)

PR #855, opened 2020-06-03
Added __encode__ __decode__ to serde.

jaheba:
*Issue #, if available:*

*Description of changes:*


jaheba:
Example usage:

```python
import random

class MyClass:
    def __init__(self):
        self.number = random.randrange(1, 100)

    def __encode__(self):
        return {"number": self.number}

    @classmethod
    def __decode__(cls, state):
        m = MyClass()
        m.number = state["number"]
        return m
```

PR #858, opened 2020-06-04
Add data loader shuffling function

yx1215:
*Description of changes:*
Added ShuffleIter Class to enable persudo shuffling in data loader. The following are the plots.
single_process_1
single_process_10
two_process_1
two_process_10
compare

### DIFF
--- a/src/gluonts/model/seq2seq/_forking_estimator.py
+++ b/src/gluonts/model/seq2seq/_forking_estimator.py
@@ -108,7 +108,7 @@ class ForkingSeq2SeqEstimator(GluonEstimator):
         The age feature starts with a small value at the start of the time series and grows over time.
     enable_decoder_dynamic_feature
         Whether the decoder should also be provided with the dynamic features (``age``, ``time``
-        and ``feat_dynamic_real`` if enabled respectively). (default: True)
+        and ``feat_dynamic_real`` if enabled respectively). (default: False)
         It makes sense to disable this, if you dont have ``feat_dynamic_real`` for the prediction range.
     trainer
         trainer (default: Trainer())
@@ -131,9 +131,9 @@ class ForkingSeq2SeqEstimator(GluonEstimator):
         use_feat_static_cat: bool = False,
         cardinality: List[int] = None,
         embedding_dimension: List[int] = None,
-        add_time_feature: bool = True,
-        add_age_feature: bool = True,
-        enable_decoder_dynamic_feature: bool = True,
+        add_time_feature: bool = False,
+        add_age_feature: bool = False,
+        enable_decoder_dynamic_feature: bool = False,
         trainer: Trainer = Trainer(),
         scaling: bool = False,
         dtype: DType = np.float32,
--- a/src/gluonts/model/seq2seq/_transform.py
+++ b/src/gluonts/model/seq2seq/_transform.py
@@ -171,7 +171,12 @@ class ForkingSequenceSplitter(FlatMapTransformation):
                         ):
                             dec_field[:] = ts[:, idx : idx + self.dec_len].T
 
-                    out[self._future(ts_field)] = np.squeeze(forking_dec_field)
+                    if forking_dec_field.shape[-1] == 1:
+                        out[self._future(ts_field)] = np.squeeze(
+                            forking_dec_field, axis=-1
+                        )
+                    else:
+                        out[self._future(ts_field)] = forking_dec_field
 
             # So far pad indicator not in use
             pad_indicator = np.zeros(self.enc_len)


src/gluonts/dataset/artificial/_base.py
```diff
@@ -848,3 +848,26 @@ def constant_dataset() -> Tuple[DatasetInfo, Dataset, Dataset]:
     )
 
     return info, train_ds, test_ds
+
+
+def shuffle_testing_dataset(n):
```
lostella, line 853: Can this be moved directly in the test script?

src/gluonts/dataset/artificial/__init__.py
```diff
@@ -19,6 +19,7 @@
     RecipeDataset,
     constant_dataset,
     default_synthetic,
+    shuffle_testing_dataset,
```
lostella, line 22: We don't need to have this in gluonts.dataset.artificial, I think

examples/draw_shuffling_graph.py
```diff
@@ -0,0 +1,88 @@
...
+    plt.legend()
+    plt.show()
+
+
+def length_vs_randomness(n, n_worker=None):
+    for i in range(1, n):
+        plot_heap_map(i, n_worker, subplot=True, shuffle=True)
+
+
+if __name__ == '__main__':
```
lostella, line 86: This script is really nice, but doesn't need to be merged; could you instead edit the original post of the PR and add there plots showing results (heatmaps, running times) obtained with this?

src/gluonts/dataset/artificial/_base.py
yx1215: just removed this

src/gluonts/dataset/artificial/__init__.py
yx1215: just removed this

examples/draw_shuffling_graph.py
yx1215: Where should I add these plots?

lostella: on the top post in this issue, you can edit that and add code snippets or images

yx1215: I've uploaded the graphs.

examples/draw_shuffling_graph.py
```diff
@@ -0,0 +1,212 @@
...
+        [
+            {FieldName.TARGET: a, FieldName.START: start, FieldName.ITEM_ID: b}
+            for a, b in zip(targets, number)
+        ],
+        freq="D",
+    )
+    return train_ds
+
+
+def plot_heap_map(c, num_worker, subplot, shuffle):
```
lostella, line 120: I think we can remove this script from the PR. But the plots are really cool!

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -631,9 +664,30 @@ def __iter__(self) -> Iterator[DataBatch]:
...
+                            itertools.islice(generator, self.batch_size)
+                        )
+                        batch_samples = []
+                        # O(batch_size) since every operation in the loop is O(1)
+                        for i in range(self.batch_size):
+                            idx = random.randint(
+                                0, self.shuffle_buffer_length - 1
+                            )
+                            batch_samples.append(self.shuffle_buffer[idx])
+                            self.shuffle_buffer[idx] = next_samples[i]
```
lostella, line 690: This code segment is repeated in a very similar way in the multi process iterator class above. I’m wondering if we could avoid this and exploit the following pattern:

What is being done is transforming an iterator into a new one which consumes slightly more items, from the original iterator, and manages the buffer sampling and replenishing logic.

This could be a generic class (extending Iterator) which *wraps* a given iterator: you would just need to wrap two different iterators, depending on whether we’re in single or multi processing, but the shuffling logic would be managed by the same code in both cases.

What do you think? Could this work?

yx1215:
> This looks nice! I have some comments inline on how to potentially improve the code, let me know what you think.
> 
> Also, did we conclude whether the increase in running time with increasing buffer size is expected? I don’t remember

No we didn't conclude for that, but I have some thoughts. Since the time is used to generate 50 batches, without shuffling, it takes around 0.55s, 0.011s per batch. When shuffle buffer equals 30 * batch size, filling the buffer in the first round is equivalent to generate 30 extra batches, so in total we actually generate 80 batches, which should take around 0.88s in theory, and in the graph the time given is close to this(or even shorter). So I think this increase in with batch size makes sense. What do you think?

lostella:
> So I think this increase in with batch size makes sense. What do you think?


Yes, makes sense. Maybe you can then experiment with iterating over more than 50 batches, to get a more accurate average time per batch.

Also, it’s sufficient to keep the buffer length between 2 and 10 times the batch size, not more. So if you iterate over, say, 200 batches, the initial overhead should become much less relevant already

src/gluonts/core/component.py
```diff
@@ -263,17 +263,21 @@ def equals_parameter_dict(
...
+            name = prefix.rstrip("0123456789") + args
+
+        return name
+
+    this_param_names_stripped = [
+        strip_prefix_enumeration(key, this.prefix) for key in this.keys()
+    ]
+    that_param_names_stripped = [
+        strip_prefix_enumeration(key, that.prefix) for key in that.keys()
+    ]
```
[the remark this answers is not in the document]
lostella: @yx1215 could you try to get rid of these changes? They are not supposed to be in this PR

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -346,6 +355,45 @@ def _worker_fn(
     return buf.getvalue()
 
 
+class _ShuffleBatchIter(Iterator[DataEntry]):
```
lostella, line 358: You can call this `ShuffleIterator` (without underscore, and without “batch” since this object knows nothing about batches)

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -346,6 +355,45 @@ def _worker_fn(
...
+    def __init__(
+        self,
+        base_iterator: Iterator[DataEntry],
+        shuffle_buffer_length: Optional[int],
+        multiprocess: bool,
+    ):
+        if multiprocess:
+            self.shuffle_buffer = _WorkerData.shuffle_buffer
```
lostella, line 366: Why is this needed? Couldn’t you just set `self.shuffle_buffer = []` in any case?

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -346,6 +355,45 @@ def _worker_fn(
     return buf.getvalue()
 
 
+class _ShuffleBatchIter(Iterator[DataEntry]):
+    def __init__(
+        self,
+        base_iterator: Iterator[DataEntry],
+        shuffle_buffer_length: Optional[int],
```
lostella, line 362: I would not make this optional: if one use this class, it’s because they want to shuffle.

The decision on whether to wrap (hence shuffle) the base_iterator or not (ie using it as it is) should be taken by the caller.

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -346,6 +355,45 @@ def _worker_fn(
...
+                    ):
+                        self.shuffle_buffer.append(element)
+                idx = random.randint(0, self.shuffle_buffer_length - 1)
+                next_sample = self.shuffle_buffer[idx]
+                self.shuffle_buffer[idx] = next(self.base_iterator)
+                return next_sample
+            else:
+                return next(self.base_iterator)
+        except StopIteration:
+            return {}
```
lostella, line 387: Why is this needed? Without this you could remove the try...except

src/gluonts/core/component.py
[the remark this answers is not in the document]
yx1215: Yes, I've reverted these changes.

src/gluonts/dataset/parallelized_loader.py
yx1215: Previously I am using next(self.base_iterator) so StopIteration Exception may occur. Now I've  edited the way I've written and removed the try/except.

src/gluonts/dataset/parallelized_loader.py
yx1215: I've changed this.

PR #860, opened 2020-06-05
fix hanging data loader bug

### LINKED ISSUE #837: Process hangs when num_workers >= 1
## Description
Data loading seems to hang when `num_workers` is set to `1` or more. The same doesn't seem to happen if `num_workers=None` or `num_workers=0` (I'm not sure what the latter implies).

## To Reproduce

Copy-paste the following in a script and run it 

```python
import mxnet as mx

from gluonts.dataset.loader import TrainDataLoader

freq = "H"
prediction_length = 24
data = ListDataset(
    data_iter=[
        {
            "item_id": k,
            "start": "2020-01-01 00:00:00",
            "target": [1.0 for _ in range(1000)]
        }
        for k in range(100)
    ],
    freq=freq,
)

epochs = 5
batch_size = 32
num_batches_per_epoch = 10

total_batches = num_batches_per_epoch * epochs
total_samples = total_batches * batch_size

estimator = DeepAREstimator(
    freq=freq,
)

transform = estimator.create_transformation()

loader = TrainDataLoader(
    data,
    transform=transform,
    batch_size=batch_size,
    ctx=mx.cpu(),
    num_batches_per_epoch=num_batches_per_epoch,
    num_workers=1,
)

all_batches = []
for epoch_no in range(epochs):
    print(f"epoch #{epoch_no}")
    batches = list(loader)
    assert len(batches) == num_batches_per_epoch
    all_batches.extend(batches)

assert len(all_batches) == total_batches
```

The process just hangs and needs to be terminated with SIGINT

## Environment
- Operating system: macOS Mojave 10.14.6 (18G4032) on dual-core i7
- Python version: 3.7
- GluonTS version: master
- MXNet version: 1.6.0

lostella:
Co-authored with @AaronSpieler 

*Issue #, if available:* Fixes #837 

### DIFF
--- a/src/gluonts/dataset/parallelized_loader.py
+++ b/src/gluonts/dataset/parallelized_loader.py
@@ -13,20 +13,17 @@
 
 
 # Standard library imports
-import collections
 import functools
 import itertools
 import logging
-import pathlib
 import pickle
 import io
-import random
 import sys
 import time
 
 from collections.abc import Sized
 from multiprocessing.managers import SyncManager
-from typing import Callable, Iterable, Optional, List, Iterator, Union, Any
+from typing import Callable, Optional, List, Iterator, Union, Any
 
 import multiprocessing
 import multiprocessing.queues
@@ -49,6 +46,9 @@ from gluonts.core.component import DType
 from gluonts.dataset.common import Dataset, DataEntry, DataBatch, FileDataset
 from gluonts.transform import Transformation
 from gluonts.dataset.util import MPWorkerInfo
+
+# this is needed for pickle to be able to use numpy
+sys.modules["numpy"] = np
 
 # ForkingPickler related functions:
 if sys.platform == "darwin" or sys.platform == "win32":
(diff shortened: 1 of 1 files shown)


lostella:
> So no time.sleep(...) is needed, and there are no memory leaks?

Sleeping is not needed. Memory leaks I'm not sure how they could occur: if the process terminates then Python will have garbage collected everything I suppose. How can one check memory leaks?

lostella:
However: https://github.com/awslabs/gluon-ts/pull/860/checks?check_run_id=742422660

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -474,17 +474,24 @@ def __iter__(self) -> Iterator[DataBatch]:
...
+        # efficient implicit garbage collection
         try:
             for k in list(self._data_buffer.keys()):
-                res = pickle.loads(self._data_buffer.pop(k).get(self._timeout))
-                del res
+                handle = self._data_buffer.pop(k)
+                # In some cases the workers might have been terminated
+                # before they finished their respective batch,
+                # in which case the results will never be available
+                if handle.ready():
```
yx1215, line 485: It seems that this is not the correct way to fix this problem. The ```handle.ready()``` here should always be False since if we need garbage collection to retrieve, these results are always not ready since we never used them. The real problem of the hang bug here should be that the module needed for ```pickle.loads```, for example, numpy and pandas, are deleted in the system before they are used. See the explanation [here](url: https://stackoverflow.com/questions/14986568/python-how-to-ensure-that-del-method-of-an-object-is-called-before-the-mod/14986683#14986683)

yx1215, line 485: It seems that this is not the correct way to fix this problem. The ```handle.ready()``` here should always be False since if we need garbage collection to retrieve, these results are always not ready since we never used them. The real problem of the hang bug here should be that the module needed for ```pickle.loads```, for example, numpy and pandas, are deleted in the system before they are used. See the explanation [here](https://stackoverflow.com/questions/14986568/python-how-to-ensure-that-del-method-of-an-object-is-called-before-the-mod/14986683#14986683)

lostella: I'm not sure I follow, do you want to take a shot at fixing this? You can open a separate PR

AaronSpieler:
Already fixed in https://github.com/awslabs/gluon-ts/pull/944 .

PR #861, opened 2020-06-05
Using pytest xdist.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,4 +1,5 @@
 [pytest]
+addopts = -n8
 markers =
     serial: mark a test that requires more resources to run that are thus only suitable for serial run.
     remote_required: mark a test that requires internet access.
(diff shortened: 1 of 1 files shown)


pytest.ini
```diff
@@ -1,9 +1,10 @@
...
+addopts = -n8
 markers =
     serial: mark a test that requires more resources to run that are thus only suitable for serial run.
     remote_required: mark a test that requires internet access.
     gpu: mark a test that requires GPU.
     integration: mark an integration test
     skip_master: mark a test that is temporarily skipped for mxnet master validation.
 
-timeout = 30
+timeout = 300
```
lostella, line 10: Why this?

jaheba: Some tests timed out before. Now, with using more workers, individual tests might even be slower.

jaheba:
We are using `xdist` in our workflows now. Users are free to use it themselves.

PR #867, opened 2020-06-11
MQCNN fixes

AaronSpieler:
*Description of changes:*
* fixing edge case of prediction length 1
* and some defaults

### DIFF
--- a/src/gluonts/model/seq2seq/_forking_estimator.py
+++ b/src/gluonts/model/seq2seq/_forking_estimator.py
@@ -108,7 +108,7 @@ class ForkingSeq2SeqEstimator(GluonEstimator):
         The age feature starts with a small value at the start of the time series and grows over time.
     enable_decoder_dynamic_feature
         Whether the decoder should also be provided with the dynamic features (``age``, ``time``
-        and ``feat_dynamic_real`` if enabled respectively). (default: True)
+        and ``feat_dynamic_real`` if enabled respectively). (default: False)
         It makes sense to disable this, if you dont have ``feat_dynamic_real`` for the prediction range.
     trainer
         trainer (default: Trainer())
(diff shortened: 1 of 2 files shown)


src/gluonts/model/seq2seq/_forking_estimator.py
```diff
@@ -131,9 +131,9 @@ def __init__(
         use_feat_static_cat: bool = False,
         cardinality: List[int] = None,
         embedding_dimension: List[int] = None,
-        add_time_feature: bool = True,
-        add_age_feature: bool = True,
-        enable_decoder_dynamic_feature: bool = True,
+        add_time_feature: bool = False,
+        add_age_feature: bool = False,
+        enable_decoder_dynamic_feature: bool = False,
```
lostella, line 136: Why disabling these by default?

src/gluonts/model/seq2seq/_transform.py
```diff
@@ -171,7 +171,12 @@ def flatmap_transform(
                         ):
                             dec_field[:] = ts[:, idx : idx + self.dec_len].T
 
-                    out[self._future(ts_field)] = np.squeeze(forking_dec_field)
+                    if forking_dec_field.shape[-1] == 1:
```
lostella, line 174: Shape won’t really work with hybridize=True I guess, or I’m missing something?

lostella: Sorry, I misunderstood: we're in numpyland

lostella: Wouldn't it make sense to add a test for this?

src/gluonts/model/seq2seq/_forking_estimator.py
AaronSpieler: So they are by default disabled in MQCNN. This is just to make this consistent.

src/gluonts/model/seq2seq/_transform.py
AaronSpieler: Yeah we probably should. We should add much more edge case tests in general.

PR #873, opened 2020-06-16
Add shuffling function

yx1215:
*Issue #, if available:*

*Description of changes:*

*Note: replaces #858*

Added ShuffleIter class to enable shuffling  in data loader.

In the first time comparison plot, the x-axis represents the shuffle buffer length, that is, if the x-axis is c, then the shuffle buffer is of size c * batch size, y-axis is the time used to generate 200 batches of size 32 from a dataset of length 300.

comparison

In the following heatmaps, x-axis is the batch index, y-axis is the time series index, and color intensity indicates how many elements in the x-batch are from the y-time-series

single_process_1
single_process_10
two_process_1
two_process_10

### DIFF
--- a/src/gluonts/dataset/parallelized_loader.py
+++ b/src/gluonts/dataset/parallelized_loader.py
@@ -368,12 +368,16 @@ class ShuffleIter(Iterator[DataEntry]):
         self.base_iterator = base_iterator
 
     def __next__(self) -> DataEntry:
+        # if the buffer is empty, fill the buffer first.
+        # (should only executed in the first round)
         if len(self.shuffle_buffer) == 0:
             self.shuffle_buffer = list(
                 itertools.islice(
                     self.base_iterator, self.shuffle_buffer_length
                 )
             )
+        # choose an element at a random index and yield it
+        # and fill it with the next element in the sequential generator
         idx = random.randint(0, self.shuffle_buffer_length - 1)
         next_sample = self.shuffle_buffer[idx]
         self.shuffle_buffer[idx] = next(self.base_iterator)
--- a/src/gluonts/dataset/parallelized_loader.py
+++ b/src/gluonts/dataset/parallelized_loader.py
@@ -257,8 +257,6 @@ class _WorkerData:
     transformation: Transformation
     # replicate transformation
     dataset_iterator: Iterator[DataEntry]
-    # shuffle buffer
-    shuffle_buffer: list = []
     # indicates which cycle the iterator has been reset last
     iterator_latest_reset_cycle: int = 0
     # indicates whether the iterator was previously depleted
@@ -366,21 +364,34 @@ class ShuffleIter(Iterator[DataEntry]):
         self.shuffle_buffer: list = []
         self.shuffle_buffer_length = shuffle_buffer_length
         self.base_iterator = base_iterator
+        self.base_iter_finished = False
 
     def __next__(self) -> DataEntry:
         # if the buffer is empty, fill the buffer first.
         # (should only executed in the first round)
-        if len(self.shuffle_buffer) == 0:
+        if not self.shuffle_buffer:
             self.shuffle_buffer = list(
                 itertools.islice(
                     self.base_iterator, self.shuffle_buffer_length
                 )
             )
+        # if buffer still empty, means all elements used,
+        # return a signal of end of iterator
+        if not self.shuffle_buffer:
+            return {}
         # choose an element at a random index and yield it
         # and fill it with the next element in the sequential generator
-        idx = random.randint(0, self.shuffle_buffer_length - 1)
+        idx = random.randint(0, len(self.shuffle_buffer) - 1)
         next_sample = self.shuffle_buffer[idx]
-        self.shuffle_buffer[idx] = next(self.base_iterator)
+
+        # replace the index with the next element in the iterator if the iterator has not finished.
+        # delete the index otherwise.
+        to_replace: DataEntry = next(self.base_iterator, {})
+        if to_replace:
+            self.shuffle_buffer[idx] = to_replace
+        else:
+            del self.shuffle_buffer[idx]
+
         return next_sample
 
     def __iter__(self) -> Iterator[DataEntry]:
@@ -652,9 +663,7 @@ class ParallelDataLoader(object):
         self.worker_id_queue: Optional[Queue] = None
         # In order to recycle unused but pre-calculated batches from last epoch for training:
         self.multi_worker_cache: Optional[Iterator[DataBatch]] = None
-        # Do shuffle or not
         self.shuffle_buffer_length: Optional[int] = shuffle_buffer_length
-        self.shuffle_buffer: list = []
 
         if self.num_workers > 0:
             # generate unique ids for processes
--- a/src/gluonts/dataset/parallelized_loader.py
+++ b/src/gluonts/dataset/parallelized_loader.py
@@ -257,6 +257,8 @@ class _WorkerData:
     transformation: Transformation
     # replicate transformation
     dataset_iterator: Iterator[DataEntry]
+    # shuffle buffer
+    shuffle_buffer: list = []
     # indicates which cycle the iterator has been reset last
     iterator_latest_reset_cycle: int = 0
     # indicates whether the iterator was previously depleted
@@ -265,16 +267,25 @@ class _WorkerData:
 
 # needed because some iterators are not cyclic
 def _worker_reset_iterator(
-    is_train: bool, cyclic: bool, cycle_num: int,
+    is_train: bool,
+    cyclic: bool,
+    cycle_num: int,
+    shuffle_buffer_length: Optional[int],
 ) -> None:
     """Initialize or reset iterators of workers."""
 
-    _WorkerData.dataset_iterator = _sequential_sample_generator(
+    generator = _sequential_sample_generator(
         dataset=_WorkerData.dataset,
         transformation=_WorkerData.transformation,
         is_train=is_train,
         cyclic=cyclic,
     )
+    if shuffle_buffer_length is not None:
+        generator = ShuffleIter(
+            base_iterator=generator,
+            shuffle_buffer_length=shuffle_buffer_length,
+        )
+    _WorkerData.dataset_iterator = generator
 
     _WorkerData.iterator_latest_reset_cycle = cycle_num
     _WorkerData.iterator_exhausted_indicator = False
@@ -308,6 +319,7 @@ def _worker_fn(
     is_train: bool,
     cyclic: bool,
     cycle_num: int,
+    shuffle_buffer_length: int,
 ):
     """Function for processing data in worker process."""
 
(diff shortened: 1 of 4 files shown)


AaronSpieler:
If I read you code correctly, you are shuffling the batches, but not the samples from which then different batches are generated?

yx1215:
> If I read you code correctly, you are shuffling the batches, but not the samples from which then different batches are generated?

I am shuffling the samples which are put into each batch. Originally samples are generated sequentially from the function _sequential_sample_generator. Now I take these samples and put them in a shuffle buffer(a list) which has length larger than batch size, randomly generate a index, take the element at that index, put it in the batch and replace it with the next element in the original sequential generator. Also you can refer to the heat map in the PR. The x-axis means the batch number, and the y-axis means the time series id. So if the number at (x, y) is z, that means in batch number x, it takes z element from time series number y. It clearly shows that with a larger buffer size, the elements will become more shuffled.

AaronSpieler:
> > If I read you code correctly, you are shuffling the batches, but not the samples from which then different batches are generated?
> 
> I am shuffling the samples which are put into each batch. Originally samples are generated sequentially from the function _sequential_sample_generator. Now I take these samples and put them in a shuffle buffer(a list) which has length larger than batch size, randomly generate a index, take the element at that index, put it in the batch and replace it with the next element in the original sequential generator. Also you can refer to the heat map in the PR. The x-axis means the batch number, and the y-axis means the time series id. So if the number at (x, y) is z, that means in batch number x, it takes z element from time series number y. It clearly shows that with a larger buffer size, the elements will become more shuffled.

Ok, got it, thx!

AaronSpieler:
Ok, looks nice. Could you add some comments to the `ShuffleIter` class maybe?
Also, how did you test the speed of the respective shuffle setting?

richardk53:
It looks like it will have a semi-random stopping time, I think between dataset_length - shuffle_buffer_length and dataset_length iterations (where dataset_length is the number of iterations until the base_iterator is exhausted).

E.g. if shuffle_buffer_length has the same length as the dataset, then the base_iterator will be exhausted in the first iteration. It will start putting Nones in the list until ShuffleIter randomly encounters one. This happens latest after all elements have been replaced with None.

yx1215:
> Ok, looks nice. Could you add some comments to the `ShuffleIter` class maybe?
> Also, how did you test the speed of the respective shuffle setting?

I've updated the description in the PR about the plots. And comments are added to the shuffle iter.

lostella:
> E.g. if shuffle_buffer_length has the same length as the dataset, then the base_iterator will be exhausted in the first iteration. It will start putting Nones in the list until ShuffleIter randomly encounters one. This happens latest after all elements have been replaced with None.

I think this may not be a problem, since at training time the data loader is initialized as "cyclic", see here: https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/dataset/loader.py#L166. In fact, I think the same mechanism is used to ensure that if in an epoch you want to iterate more data than the time series you have (by setting num_batches_per_epoch) then the data loader starts back from the beginning once the dataset is exhausted.

The logic is a bit convoluted probably, and we should improve it.

lostella:
Maybe related to what @richardk53 mentioned: the following snippet gives weird results

```python
data = range(20)
shuffled_data = ShuffleIter(base_iterator=iter(data), shuffle_buffer_length=10)
print(len(list(shuffled_data)))  # I would expect 20 here
```

I would expect to the shuffled data to always have the same size as the original data, in case that's finite (so 20 in the above example). But I'm getting all sorts of numbers instead, and sometimes errors.

@yx1215 I think `ShuffleIter` as a standalone component should should have that property right? Maybe you can fix it to make sure it does, and add a test for it?

yx1215:
> Maybe related to what @richardk53 mentioned: the following snippet gives weird results
> 
> ```python
> data = range(20)
> shuffled_data = ShuffleIter(base_iterator=iter(data), shuffle_buffer_length=10)
> print(len(list(shuffled_data)))  # I would expect 20 here
> ```
> 
> I would expect to the shuffled data to always have the same size as the original data, in case that's finite (so 20 in the above example). But I'm getting all sorts of numbers instead, and sometimes errors.
> 
> @yx1215 I think `ShuffleIter` as a standalone component should should have that property right? Maybe you can fix it to make sure it does, and add a test for it?

yx1215:
> Maybe related to what @richardk53 mentioned: the following snippet gives weird results
> 
> ```python
> data = range(20)
> shuffled_data = ShuffleIter(base_iterator=iter(data), shuffle_buffer_length=10)
> print(len(list(shuffled_data)))  # I would expect 20 here
> ```
> 
> I would expect to the shuffled data to always have the same size as the original data, in case that's finite (so 20 in the above example). But I'm getting all sorts of numbers instead, and sometimes errors.
> 
> @yx1215 I think `ShuffleIter` as a standalone component should should have that property right? Maybe you can fix it to make sure it does, and add a test for it?

I've fixed the problem in the following way: once the base iterator reaches its end, each time we select an element from the shuffle buffer, we will delete that element instead of replacing it. This, I think, will not have much influence on the performance because this is only used to deal with edge case. And I've updated the time comparison plot.

richardk53:
> > E.g. if shuffle_buffer_length has the same length as the dataset, then the base_iterator will be exhausted in the first iteration. It will start putting Nones in the list until ShuffleIter randomly encounters one. This happens latest after all elements have been replaced with None.
> 
> I think this may not be a problem, since at training time the data loader is initialized as "cyclic", see here: https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/dataset/loader.py#L166. In fact, I think the same mechanism is used to ensure that if in an epoch you want to iterate more data than the time series you have (by setting num_batches_per_epoch) then the data loader starts back from the beginning once the dataset is exhausted.
> 
> The logic is a bit convoluted probably, and we should improve it.


I think this component should not need to make these assumptions about other components using it. 
With the fix it should be possible to implement e.g. sampling without replacement (which often works better) by choosing the buffer size as the dataset length and iterating only once over the dataset.

lostella:
> I think this component should not need to make these assumptions about other components using it.

Yes, totally agree. @yx1215 has fixed some of these edge cases already, I’ll check if that covers them all (we should test it well)

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -346,6 +357,51 @@ def _worker_fn(
...
+        if not self.shuffle_buffer:
+            self.shuffle_buffer = list(
+                itertools.islice(
+                    self.base_iterator, self.shuffle_buffer_length
+                )
+            )
+        # if buffer still empty, means all elements used,
+        # return a signal of end of iterator
+        if not self.shuffle_buffer:
+            return {}
```
lostella, line 381: I think the right signal to send when the iterator is exhausted is raising a `StopIteration` exception, see here: https://docs.python.org/3/library/stdtypes.html#iterator-types

src/gluonts/dataset/parallelized_loader.py
```diff
@@ -346,6 +357,51 @@ def _worker_fn(
...
+            del self.shuffle_buffer[idx]
+
+        return next_sample
+
+    def __iter__(self) -> Iterator[DataEntry]:
+        while True:
+            next_sample = next(self)
+            if not next_sample:
+                return
+            yield next_sample
```
lostella, line 402: I think this can simply return `self`, see here: https://docs.python.org/3/library/stdtypes.html#iterator-types

src/gluonts/dataset/parallelized_loader.py
yx1215: yes, already changed.

src/gluonts/dataset/parallelized_loader.py
yx1215: Actually I found out that we don't need to implement ```__iter__``` method. Changing ```__next__``` is everything needed.

AaronSpieler:
So it seems there is in general some documentation missing regarding `shuffle_buffer_length` in `loader.py` and  in `parallelized_loader.py`.

Additionally, I guess in most cases, the argument shuffle_buffer_length is implicitly passed as an argument. It should, however , be explicit just like `num_workers` or `num_prefetch`. Example below:

```
    def train(
        self,
        training_data: Dataset,
        num_workers: Optional[int] = None,
        num_prefetch: Optional[int] = None,
        **kwargs,
    ) -> Predictor:
        has_negative_data = any(np.any(d["target"] < 0) for d in training_data)
        low = -10.0 if has_negative_data else 0
        high = 10.0
        bin_centers = np.linspace(low, high, self.num_bins)
        bin_edges = np.concatenate(
            [[-1e20], (bin_centers[1:] + bin_centers[:-1]) / 2.0, [1e20]]
        )

        logging.info(
            f"using training windows of length = {self.train_window_length}"
        )

        transformation = self.create_transformation(
            bin_edges, pred_length=self.train_window_length
        )

        training_data_loader = TrainDataLoader(
            dataset=training_data,
            transform=transformation,
            batch_size=self.trainer.batch_size,
            num_batches_per_epoch=self.trainer.num_batches_per_epoch,
            ctx=self.trainer.ctx,
            num_workers=num_workers,
            num_prefetch=num_prefetch,
            **kwargs,
        )

        validation_data_loader = None
        if validation_data is not None:
            validation_data_loader = ValidationDataLoader(
                dataset=validation_data,
                transform=transformation,
                batch_size=self.trainer.batch_size,
                ctx=self.trainer.ctx,
                dtype=self.dtype,
                num_workers=num_workers,
                num_prefetch=num_prefetch,
                **kwargs,
            )
```

@yx1215 Do you think you could look into this, maybe fix it using a PR?

yx1215:
> So it seems there is in general some documentation missing regarding `shuffle_buffer_length` in `loader.py` and in `parallelized_loader.py`.
> 
> Additionally, I guess in most cases, the argument shuffle_buffer_length is implicitly passed as an argument. It should, however , be explicit just like `num_workers` or `num_prefetch`. Example below:
> 
> ```
>     def train(
>         self,
>         training_data: Dataset,
>         validation_data: Optional[Dataset] = None,
>         num_workers: Optional[int] = None,
>         num_prefetch: Optional[int] = None,
>         **kwargs,
>     ) -> Predictor:
>         has_negative_data = any(np.any(d["target"] < 0) for d in training_data)
>         low = -10.0 if has_negative_data else 0
>         high = 10.0
>         bin_centers = np.linspace(low, high, self.num_bins)
>         bin_edges = np.concatenate(
>             [[-1e20], (bin_centers[1:] + bin_centers[:-1]) / 2.0, [1e20]]
>         )
> 
>         logging.info(
>             f"using training windows of length = {self.train_window_length}"
>         )
> 
>         transformation = self.create_transformation(
>             bin_edges, pred_length=self.train_window_length
>         )
> 
>         training_data_loader = TrainDataLoader(
>             dataset=training_data,
>             transform=transformation,
>             batch_size=self.trainer.batch_size,
>             num_batches_per_epoch=self.trainer.num_batches_per_epoch,
>             ctx=self.trainer.ctx,
>             num_workers=num_workers,
>             num_prefetch=num_prefetch,
>             **kwargs,
>         )
> 
>         validation_data_loader = None
>         if validation_data is not None:
>             validation_data_loader = ValidationDataLoader(
>                 dataset=validation_data,
>                 transform=transformation,
>                 batch_size=self.trainer.batch_size,
>                 ctx=self.trainer.ctx,
>                 dtype=self.dtype,
>                 num_workers=num_workers,
>                 num_prefetch=num_prefetch,
>                 **kwargs,
>             )
> ```
> 
> @yx1215 Do you think you could look into this, maybe fix it using a PR?

Yes, I've created a new PR to add documentation as well as fix the problem of implicit input. Actually I've already add ```shuffle_buffer_length``` to the general ```GluonEstimator``` class in ```model/estimator.py```, so the problem only occurs when sub estimators reimplement the train method. Current I only find one model that does this which is the wavenet. Feel free to tell me if there is more. Here is the new PR #910 .

PR #874, opened 2020-06-16
Fix to_time_series_item in splitter.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/test/dataset/split/test_split.py
+++ b/test/dataset/split/test_split.py
@@ -1,0 +1,35 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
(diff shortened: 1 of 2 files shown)


lostella:
Maybe we can add a small test for these components?

dcmaddix:
> Maybe we can add a small test for these components?

Yes, let's add a smoke test for this.  I can add a smoke test for the failing case we saw.

dcmaddix:
Great, this should fix the current error:
 ```
ValidationError: 2 validation errors for TimeSeriesItem
feat_dynamic_real -> 0
  value is not a valid list (type=type_error.list)
feat_dynamic_real -> 1
  value is not a valid list (type=type_error.list)
```
on line `--> 109             feat_dynamic_real=[real.values for real in self.feat_dynamic_real],`

where  feat_dynamic_real is of type `List[np.ndarray]` but the `TimeSeriesItem` requires type `List[List[float]]`.

**Comment by jaheba** [comment]
I've added a quick test. Without the changes it fails.

File: test/dataset/split/test_split.py [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+    sl = TimeSeriesSlice(
+        target=make_series(range(100)),
+        item="",
+        feat_static_cat=[1, 2, 3],
+        feat_static_real=[0.1, 0.2, 0.3],
+        feat_dynamic_cat=[make_series(range(100))],
+        feat_dynamic_real=[make_series(range(100))],
+    )
+
+    sl.to_time_series_item()
```
dcmaddix, line 35: Looks good, otherwise this line would fail.  Thanks!

PR #875, opened 2020-06-16
updated Pandas deprecation

pmohamma:
*Description of changes:*
Addition of integers and Timestamps is deprecated in Pandas 0.23+. Changed the addition in DateSplitter to be compatible with newer Pandas versions by casting the addition to a Timedelta object.


lostella:
Hey @pmohamma thanks for this! We use black for style checking, and it’s apparently detecting some badly styled code in the PR, hence one of the build steps is failing.

You can set up a git pre-commit hook locally by running 

```
pip install requirements/requirements-setup.txt
./dev_setup.sh
```

so that every time you commit, style checks are run and an error message is displayed, with instructions on how to fix the code.

In this case you can fix the error by running

```
black src
```

and committing & pushing the changes it makes.

pmohamma:
Thanks @lostella ! Just ran the comments and looks like it is now hopefully in the right format.

src/gluonts/dataset/split/splitter.py
```diff
@@ -268,4 +268,9 @@ def _train_slice(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
     def _test_slice(
         self, item: TimeSeriesSlice, offset: int = 0
     ) -> TimeSeriesSlice:
-        return item[: self.split_date + self.prediction_length + offset]
+        freq = item.start.freq
+        return item[
+            : self.split_date
+            + pd.Timedelta(self.prediction_length, unit=freq)
+            + pd.Timedelta(offset, unit=freq)
```
jaheba, line 275:
```suggestion
            + pd.Timedelta(self.prediction_length + offset, unit=freq)
```

I think we can simplify this a bit.

PR #876, opened 2020-06-17
Added mx module.

jaheba:
Moved mxnet dependent modules into dedicated mx-module.

Added stub files which import module-contents from new place.

*Issue #, if available:*

*Description of changes:*


lostella:
This creates lots of files, which may create confunsion. So I would have two comments:
* Isn't there a more compact way to go about it, which doesn't keep all these subpackages and modules around?
* If the above is not possible, I guess we will want to eventually remove all these files: maybe we want to deprecate importing from these old modules? We could put a warning for when someone tries to do that maybe

PR #886, opened 2020-06-19
Aggregate lag transformation

benidis:
Creates a transformation that add lags of a selected frequency as a feature. It ignores lags that need values from the `prediction_length` so it applies in both training and inference.

The aggregation of the lags is done by specifying the following:
- `base_freq`: the initially frequency of the data, e.g., `1min`.
- `agg_freq`: the frequency at which the data will be aggregated, e.g., `1H`. It can be the same as the `base_freq` in which case the transformation adds lags in the original frequency.
- `agg_fun`: the function that will be used for aggregation. This depends on the data type, e.g., `mean` could be used for latency  and `sum` for request rate.
- `rolling_agg`: boolean indicating if the aggregation should be done with a rolling window or by calendar. Consider the following example: hourly data, `1D` aggregation and we are interested in the first aggregate lag (previous day).
  - If `rolling_agg=False` any hour today will have the same value for the first `1D` aggregate lag since the aggregation of the lags is based on calendar and the previous day is common for every hour today.
  - if `rolling_agg=True` every hour today will have potentially different values for the first `1D` aggregate lag since the aggregation is done based on a rolling window centered 24H prior to the current hour.


src/gluonts/transform/feature.py
```diff
@@ -433,3 +433,198 @@ def map_transform(self, data: DataEntry, is_train: bool) -> DataEntry:
...
+                ),
+                index=pd.date_range(
+                    data["start"],
+                    periods=len(data[self.target_field]) + self.pred_length,
+                    freq=self.base_freq,
+                ),
+            )
+
+        if not self.rolling_agg:
+            # compute how many time stamps are in the last (potentially not full) aggregation window
```
vafl, line 551: I don't fully understand the details for the non-rolling case, but I wonder if it could be simplified.
Perhaps we can use some pandas logic such as

```python
pd_ts.resample('1H').mean().repeat(60).values
```

to create an array of lagged values that behave in this non-rolling manner. Maybe this is missing some edge cases though, so feel free to ignore if this is missing something.

src/gluonts/transform/feature.py
```diff
@@ -433,3 +433,198 @@ def map_transform(self, data: DataEntry, is_train: bool) -> DataEntry:
...
+            )
+
+    def map_transform(self, data: DataEntry, is_train: bool) -> DataEntry:
+        assert self.base_freq == data["start"].freq
+
+        # convert to pandas Series for easier indexing and aggregation
+        if is_train:
+            pd_ts = pd.Series(
+                data[self.target_field],
+                index=pd.date_range(
```
vafl, line 528: If you find the transformation to be slow the pandas `date_range` may be a bottleneck. We could avoid this by computing a long date range once and then just indexing into this long array. 
Let me know if you have performance issues and want to look into this.

src/gluonts/transform/feature.py
benidis: About `resample` I do not directly apply it because it aggregates windows that are not fully observed. For example if a time series starts at 12.50, with `1H` resample we will get a value for 12 that is the aggregate of the last 10min and this can be way off. Therefore, I cut the first and last windows IF they are not full (last window does not matter eitherway value-wise because you cannot use lag 0). Then I apply the `resample` method to the fully observed windows (see line 595).

About `repeat`, this was an option but in the end I decided not to do it. One thing that we cannot avoid is the indexes that point to the correct lags of each timestep which should be K x T (number of lags, length of time series). These indexes point directly to the correct place of the aggregated time series without repeating the timesteps. This is just to reduce the computational and memory burdens of the transformation. Having said that, by repeating the two cases (rolling and not rolling) could potentially be a bit more unified and the code a bit more clean (with the tradeoff of repeating the aggregate time series). I am happy to do this change if you think it is better.

src/gluonts/transform/feature.py
benidis: I have not done large scale tests to observe this but I will keep it in mind if it is too slow and then we can discuss it.

PR #890, opened 2020-06-23
splitting features and uses DataEntry instead of TimeSeriesItem

pmohamma:
*Description of changes:* 

Enabled splitters to split features as well as the target for datasets. Also converted Splitters to intake and output type DataEntry instead of TimeSeriesItem, so that it is standard with the rest of the gluon-ts library.

### DIFF
--- a/src/gluonts/dataset/split/splitter.py
+++ b/src/gluonts/dataset/split/splitter.py
@@ -14,26 +14,19 @@
 """
 Train/test splitter
 ~~~~~~~~~~~~~~~~~~~
-
 This module defines strategies to split a whole dataset into train and test
 subsets.
-
 For uniform datasets, where all time-series start and end at the same point in
 time `OffsetSplitter` can be used::
-
     splitter = OffsetSplitter(prediction_length=24, split_offset=24)
     train, test = splitter.split(whole_dataset)
-
 For all other datasets, the more flexible `DateSplitter` can be used::
-
     splitter = DateSplitter(
         prediction_length=24,
         split_date=pd.Timestamp('2018-01-31', freq='D')
     )
     train, test = splitter.split(whole_dataset)
-
 The module also supports rolling splits::
-
     splitter = DateSplitter(
         prediction_length=24,
         split_date=pd.Timestamp('2018-01-31', freq='D')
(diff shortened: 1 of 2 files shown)


src/gluonts/dataset/split/splitter.py
```diff
@@ -71,47 +71,70 @@ class Config:
...
+            item["feat_dynamic_real"] if "feat_dynamic_real" in item else []
+        )
         feat_dynamic_real = [
-            pd.Series(real, index=index) for real in item.feat_dynamic_real
+            pd.Series(real, index=index) for real in item_dynamic_real
         ]
 
+        feat_static_cat = (
+            list(item["feat_static_cat"]) if "feat_static_cat" in item else []
+        )
```
jaheba, line 100: I think this is a bit more concise:

```suggestion
        feat_static_cat = list(item.get("feat_static_cat", []))
```

src/gluonts/dataset/split/splitter.py
```diff
@@ -71,47 +71,70 @@ class Config:
...
-            feat_static_real=self.feat_static_real,
-            feat_dynamic_cat=[
+    def to_data_entry(self) -> DataEntry:
+        ret = {
+            "start": self.start,
+            "item": self.item,
+            "target": self.target.values,
+        }
+
+        if len(self.feat_static_cat) > 0:
```
jaheba, line 124: In Python, you almost never need to use `len` to check whether an element has entries:

```suggestion
        if self.feat_static_cat:
```

src/gluonts/dataset/split/splitter.py
```diff
@@ -129,14 +152,21 @@ def __getitem__(self, slice_: slice) -> "TimeSeriesSlice":
...
+                feat[slice_] for feat in self.feat_dynamic_real
+            ]
 
         if self.feat_dynamic_cat is not None:
-            feat_dynamic_cat = [feat for feat in self.feat_dynamic_cat]
+            feat_dynamic_cat = [feat[slice_] for feat in self.feat_dynamic_cat]
+
+        target = self.target[slice_]
+
+        assert all([len(target) == len(i) for i in feat_dynamic_real])
```
jaheba, line 164: Don't use `i` if it is not an integer.

```suggestion
        assert all([len(target) == len(feat) for feat in feat_dynamic_real])
```

Issue #891, opened 2020-06-24
CloudWatch logging on SageMaker with version 0.5.0


MaximilianProll:
I upgraded `gluon-ts` from `0.4.2` to `0.5.0` and now I am experiencing some inconsistent logging behaviour when **training on SageMaker** (when running locally without SageMaker, I am not experiencing this issue).

On version `0.4.2` the very last few logging lines looked something like this:
```
INFO:root:Epoch[98] Learning rate is 5e-05
#015  0%|          | 0/100 [00:00<?, ?it/s]#015 24%|██▍       | 24/100 [00:10<00:32,  2.34it/s, avg_epoch_loss=1.74]#015 50%|█████     | 50/100 [00:20<00:20,  2.40it/s, avg_epoch_loss=1.74]#015 76%|███████▌  | 76/100 [00:31<00:09,  2.40it/s, avg_epoch_loss=1.73]#015100%|██████████| 100/100 [00:40<00:00,  2.46it/s, avg_epoch_loss=1.72]
INFO:root:Epoch[98] Elapsed time 40.639 seconds
INFO:root:Epoch[98] Evaluation metric 'epoch_loss'=1.724552
INFO:root:Epoch[99] Learning rate is 5e-05
#015  0%|          | 0/100 [00:00<?, ?it/s]#015 24%|██▍       | 24/100 [00:10<00:31,  2.40it/s, avg_epoch_loss=1.69]#015 50%|█████     | 50/100 [00:20<00:20,  2.45it/s, avg_epoch_loss=1.7] #015 76%|███████▌  | 76/100 [00:30<00:09,  2.44it/s, avg_epoch_loss=1.7]#015100%|██████████| 100/100 [00:40<00:00,  2.49it/s, avg_epoch_loss=1.7]
INFO:root:Epoch[99] Elapsed time 40.239 seconds
INFO:root:Epoch[99] Evaluation metric 'epoch_loss'=1.697624
INFO:root:Loading parameters from best epoch (99)
INFO:root:Final loss: 1.6976238334178924 (occurred at epoch 99)
INFO:root:End model training
```

On version `0.5.0` the very last logging lines changed to the following:
```
[...]

PR #895, opened 2020-06-25
Readded final loss info.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/mx/trainer/_base.py
+++ b/src/gluonts/mx/trainer/_base.py
@@ -344,6 +344,11 @@ class Trainer:
                             best_epoch_info["params_path"], self.ctx
                         )
 
+                logger.info(
+                    f"Final loss: {best_epoch_info['metric_value']} "
+                    f"(occurred at epoch {best_epoch_info['epoch_no']})"
+                )
+
                 logging.info("Computing averaged parameters.")
                 averaged_params_path = self.avg_strategy.apply(gluonts_temp)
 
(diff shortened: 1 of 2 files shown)


jaheba:
This got me thinking: Should we add a centralised metric-API?

jaheba:
See: #891

src/gluonts/mx/trainer/_base.py
```diff
@@ -344,6 +344,11 @@ def loop(
                             best_epoch_info["params_path"], self.ctx
                         )
 
+                logger.info(
```
benidis, line 347: I had removed this part after we introduced model averaging. I am not sure how informative it to add the final loss info if it is does not represent the loss of the final model. It may be misleading.

Also not sure if going all the way and print the losses of all the checkpoints that we average would be informative.

jaheba: Oh, I see. But what if model_averaging is not enabled?

benidis: There is always model averaging (at least defined). The equivalent of "not enabled" is to average only one model which is effectively selecting the best (unless the averaging strategy does something weird and does not select the best for a particular reason). In this case the final loss would be meaningful. 

I guess a way to include this is to check the num_models in the averaging strategy (see constructor of trainer) and if it is 1 you can log the loss and otherwise a message of the form "FInal loss: average over K model using X averaging strategy". Just a suggestion...

MaximilianProll: @benidis's suggestion sounds like a transparent compromise!

PR #897, opened 2020-06-26
Update bibliographic references in README

lostella:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,8 @@ your own models and quickly experiment with different solutions.
 
 * [Documentation (stable version)][stable docs url]
 * [Documentation (latest)][latest docs url]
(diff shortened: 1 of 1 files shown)


kashif:
@lostella may I kindly suggest to create a reference which also acknowledges the work of the open source contributors as well? 

For example, of my many open source contributions to various frameworks etc. Lasagne was the only project that officially acknowledged this e.g https://zenodo.org/record/27878#.XvWlqJMzZpI In other projects we have been involved in we have tried to add the open source contributors to the official paper as well, e.g. FlairNLP.

Perhaps one can adapt a similar reference setup here?

lostella:
@kashif I like the idea of something like zenodo to be able to cite the software directly. I've never used it, but I'll check what we can do about it.

Issue #854, opened 2020-06-02
Enable different tensor types in batches 


lostella:
The [very last step of the data loader](https://github.com/awslabs/gluon-ts/blob/60a0d5ca1ba6d1d062aab3065b6e9c68bf45a31f/src/gluonts/dataset/parallelized_loader.py#L643) is to stack a sequence of dictionaries item-wise to get a single dictionary (a batch).

The way things are stacked depends on their type, and currently np.ndarray objects are converted into mx.nd.NDArray ones and stacked along a new first axis, see [here](https://github.com/awslabs/gluon-ts/blob/60a0d5ca1ba6d1d062aab3065b6e9c68bf45a31f/src/gluonts/dataset/parallelized_loader.py#L181).

However, the batchify function could somehow be parametrized by the type of array that it should produce, and yield batches containing PyTorch or TensorFlow arrays instead, if desired. This rather minimal change would enable using the datasets, data transformations, and data loader from gluonts to feed models written using frameworks other than mxnet.

One way to do that would be passing a parameter down the data loading stack (some string?); a less invasive option could be that of using a context manager.

canerturkmen:
We could have both a parameter and a context manager, as in mxnet context.

AaronSpieler:
We also need to extend the pickling methods to support pytorch, essentially this:
```
# retrieve the batch from shared memory along with metadata
success, worker_id, batch = pickle.loads(got)
```
needs to work for pytorch (tensors) too.

jaheba:
> We also need to extend the pickling methods to support pytorch, essentially this:
> 
> ```
> # retrieve the batch from shared memory along with metadata
> success, worker_id, batch = pickle.loads(got)
> ```
> 
> needs to work for pytorch (tensors) too.

We need to be fully `serde` compatible, not just pickle.

jaheba:
Couldn't we just use type-specific strategies for this? Basically define a `Batcher` type, which then is implemented for different types?

lostella:
> Couldn't we just use type-specific strategies for this? Basically define a `Batcher` type, which then is implemented for different types?

Yeah, something like that as well

kashif:
thanks for looking into this... on the pytorch side though it would be ideal to continue to use the `DataLoader` which then return the batches from some map or Iterable style `Dataset`. The `DataLoader` also supports workers etc. and other libraries like pytorch-Lightning then work with this abstraction. On the tensorflow side I would need to check what `tf-data` expects or returns...

[...]

PR #898, opened 2020-06-26
Refactoring data loading utilities

lostella:
Joint work with @AaronSpieler and @PascalIversen 

Among other things, this aims at solving #837 and #854.

This sorts out the data loading utilities a bit better, I think. Examples of how the current "training" and "inference" data loader behaviors can be recovered is included in `examples/loader_example.py`.

Main ideas:
* Extracting many of the responsibilities currently on the data loader into standalone components. Examples:
    - Having a "stateful" iteration, which resumes from where it left, can simply use Python's `iter`.
    - Having a "cyclic" iteration for when a single epoch needs to go through the time series more than once, can be done in a separate component (non-allocating, unlike Python's `cycle`).
    - Pseudo-shuffling an iterator can be independent (this was already so on master) and applied where one desires (e.g. before or after the transformation).
    - If we want to use indexable dataset types, we can instead use proper shuffling that samples uniformly among the time series (not there yet).
* The data loader then just assembles the right iterator using these components:
* The multi-processing batch iterator was also rewritten
* The batching function can be customized with an "array constructor", that stacks arrays to the appropriate type (numpy, mxnet, pytorch, tensorflow, ...)

Also there:
* Minor changes to `TransformedDataset`

Missing:
- [x] Add multiprocessing
- [x] Make sure processes don't hang (deadlocks)
- [x] Remove dependency on MXNet for the loader.py module
- [x] Benchmark against existing code
- [ ] Make sure all processes are correctly terminated in case of interrupts or error in a subprocess
- [x] Make tests pass
- [x] Fix mypy complaints
- [x] Check/solve #941 
- [ ] Add docstrings


lostella:
>       * Having a "stateful" iteration (i.e. which resumes from where it left)

This could probably be done by just using `iter`

src/gluonts/dataset/loader_v2.py
```diff
@@ -0,0 +1,189 @@
...
+        end_event,
+    ):
+        MPWorkerInfo.worker_process = True
+        MPWorkerInfo.worker_id = worker_id
+        MPWorkerInfo.num_workers = num_workers
+
+        for entry in iterable:
+            if end_event.is_set():
+                break
+            data_queue.put((worker_id, entry))
```
AaronSpieler, line 128: I'm concerned that putting the entries directly into the queue is too slow, essentially this is the kind of thing we would like to prevent in multiprocessing since its a bottleneck that requires synchronisation between processes and is capped by slow access speeds of the RAM.
- as @lostella mentioned, this could be partially circumvented by having a unique Queue per process
- sill, pickling the time series data into the Queue is not a good idea in general imo
- a more elegant way to do it is pickling the data into shared memory and communicating the address through the queue: like done here: https://github.com/awslabs/gluon-ts/blob/3774b1df6d2b9fb75fbabc181fadb188c34fc6db/src/gluonts/dataset/parallelized_loader.py#L353

src/gluonts/dataset/loader_v2.py
```diff
@@ -0,0 +1,189 @@
...
+        assert max_queue_size is None or max_queue_size >= num_workers
+
+        self.base_iterable = base_iterable
+        self.num_workers = num_workers
+        self.num_entries = num_entries
+        self.max_queue_size = (
+            max_queue_size if max_queue_size is not None else 5 * num_workers
+        )
+
+        self.data_queue: Queue = Queue(maxsize=self.max_queue_size)
```
AaronSpieler, line 93: I think this is the wrong Queue type, if you are sharing the Queue with multiple processes. The correct one is afaik this one: https://github.com/awslabs/gluon-ts/blob/3774b1df6d2b9fb75fbabc181fadb188c34fc6db/src/gluonts/dataset/parallelized_loader.py#L660 i.e. a multiprocessing.manager().Queue()

src/gluonts/dataset/loader_v2.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def __del__(self):
+        # Send termination message to workers
+        self.done_event.set()
+        # Empty queue to make sure workers get the message
+        self._empty_queue()
+        for p in self.processes:
+            p.join()
+        self.data_queue.cancel_join_thread()
+        self.data_queue.close()
```
AaronSpieler, line 164: So this method is not guaranteed to be called, just want to make sure you have taken this into consideration.

src/gluonts/dataset/loader_v2.py
```diff
@@ -0,0 +1,189 @@
...
+
+class PseudoShuffledIterable(Iterable):
+    def __init__(self, base_iterable: Iterable, buffer_length: int):
+        self.base_iterable = base_iterable
+        self.buffer_length = buffer_length
+        self.shuffle_buffer: list = []
+
+    def sample_from_buffer(self):
+        idx = random.randint(0, len(self.shuffle_buffer) - 1)
+        return self.shuffle_buffer.pop(idx)
```
yx1215, line 62: I think we should not always use pop when we select an element from the buffer. pop() is much more time consuming than just replacing an index.

src/gluonts/dataset/loader_v2.py
lostella: This doesn't appear to be an issue, but I've only quickly benchmarked this: will do more thorough testing

src/gluonts/dataset/loader_v2.py
lostella: Yes, this seems to have solved the deadlock issues I was facing. Thanks!

src/gluonts/dataset/loader_v2.py
lostella: Yes, good point. Things should now "gracefully" terminate without the need for this.

src/gluonts/dataset/loader_v2.py
lostella: Good point! Will fix this

lostella: @yx1215 I actually kept the class we have on master that you implemented: initially I thought that having it in the form of an iterable (as opposed to iterator) would be better, but I think the way you have implemented it is better.

AaronSpieler:
I refactored the code so that the new data-loader is supported.
However, there seems to be 3 errors due to the MultiProcessIterator not really supporting inference or validation datasets; will have to investigate.

lostella:
> I refactored the code so that the new data-loader is syntactically supported.
> However, there seems to be 3 errors due to the MultiProcessIterator not really supporting inference or validation datasets; will have to investigate.

At inference time you probably don't want to use multi processing, since you want to keep things in the original order.

lostella:
@AaronSpieler let's avoid merge commits please, the diff is now a mess. Just use rebase, the PR should only show relevant commits and diffs against master

AaronSpieler:
> > I refactored the code so that the new data-loader is syntactically supported.
> > However, there seems to be 3 errors due to the MultiProcessIterator not really supporting inference or validation datasets; will have to investigate.
> 
> At inference time you probably don't want to use multi processing, since you want to keep things in the original order.

Well, sometimes. Most of the time when evaluating a model we don't care about the ordering. I think this should be something the user should be able to set. In any case, we could run InferenceDataLoader using multiprocessing and then reorder the output if desired.

PascalIversen:
@AaronSpieler and I did a quick proof of concept of using this data loader and the Predictor/Evaluation framework with a simple PyTorch model.
[PyTorch_GluonTSPOC.zip](https://github.com/awslabs/gluon-ts/files/4994728/PyTorch_GluonTSPOC.zip)

We conclude that:
 1. The new TrainDataLoader is compatible with a PyTorch training setup.
 
 2. Except for the serialization, the PyTorch Predictor needs only minor changes (basically just concerning the device parameter).
 
 3. However, the SampleForecastGenerator is not framework agnostic, since the frameworks have different methods to transform tensors to numpy (.data.numpy() in PyTorch vs. asnumpy()). Currently, the PythonForecastGenerator needs a lot of code copying, which we need to address.

AaronSpieler:
@lostella I have a new branch based on this one with 3 commits that consolidate the new data loader with the rest of the framework. All tests pass locally.

- https://github.com/AaronSpieler/gluon-ts/tree/lostella_based_mp_dl
- I hard coded `num_workers=None` for the InferenceDataLoader and ValidationDataLoader.

AaronSpieler:
A few thoughts on the new data-loader concept:

- It is completely based on `Iterable`s an thus:
   - will not support splitting the dataset across workers efficiently
   - will not support multiprocessing for inference and validation loader
   - will not allow for generic global sampling strategies 
      - the "sampling through splitting the time series at different points" will have to remain a transformation
      - we restrict ourselves again to sequentially processing time series per worker, essentially imposing a sort of ordering on the dataset (relaxed through the buffer based shuffling)
- we could take the framework dependent batchify function concept to the old multiprocessing data loader (maybe together with the queue for the batches and the more elegant stopping logic) and recover the framework independent property whilst retraining the option to use multiprocessing on validation and inference data 
- The major benefit of the new data loader I see is the simplified code and the cleaner stopping logic

I would strongly advise against basing a new data loader on `Iterable`s since such a data loader will come with many restriction (some mentioned above), and rather base it on `Collection`s.

lostella:
> I would strongly advise against basing a new data loader on Iterables since such a data loader will come with many restriction (some mentioned above), and rather base it on Collections.
 
@AaronSpieler I'm not sure: data loading results in a sequence of batches, so iterators are perfectly fine to encode that.

What could be (optionally) represented as collections are datasets: that will allow more flexibility for example in shuffling.

Most importantly, this PR is not based on iterators neither more nor less than the current data loader from master: it's just as much I would say. Do you agree @AaronSpieler that all drawbacks you see are shared with the data loader in master? (Just as a reminder: the first goal here is to make the code simpler and better decoupled without sacrificing anything compared to master; it's not supposed to be final wrt to every aspect)

AaronSpieler:
> > I would strongly advise against basing a new data loader on Iterables since such a data loader will come with many restriction (some mentioned above), and rather base it on Collections.
> 
> @AaronSpieler I'm not sure: data loading results in a sequence of batches, so iterators are perfectly fine to encode that.
> 
> What could be (optionally) represented as collections are datasets: that will allow more flexibility for example in shuffling.
> 
> Most importantly, this PR is not based on iterators neither more nor less than the current data loader from master: it's just as much I would say. Do you agree @AaronSpieler that all drawbacks you see are shared with the data loader in master? (Just as a reminder: the first goal here is to make the code simpler and better decoupled without sacrificing anything compared to master; it's not supposed to be final wrt to every aspect)

Yes, to have the overall data-loader encoded as an Iterator is perfectly fine.
Yes, I meant to say that we should rather consider datasets to be collections instead of Iterables or Iterators.
Yes, the drawbacks I mentioned also apply to the current data-loader in the master branch. 

Ahh, I understand, I wan't aware that the only goal for now was the simplification of the code.
However, we would loose multiprocessing capability of the validation and test dataset with this data-loader design. A sacrifice  that is potentially worth it.

src/gluonts/dataset/loader.py
```diff
@@ -217,7 +232,7 @@ def __init__(
         self.batch_size = batch_size
         self.batchify_fn = batchify_fn
         self.num_batches_per_epoch = num_batches_per_epoch
-        self.num_workers = num_workers
+        self.num_workers = win32_guard(num_workers)
```
PascalIversen, line 235: You are using the unguarded num_workers afterwards (line 256). It should be
```
self.batch_iterator = (
            map(batchify_fn, batcher(shuffled_iterator, batch_size))
            if self.num_workers is None
            else MultiProcessBatcher(
                shuffled_iterator,
                batch_size=batch_size,
                batchify_fn=batchify_fn,
                num_workers=num_workers,
                max_queue_size=num_prefetch,
            )
        )
```

PascalIversen:
The GPU utilization during training is similar to the current data loader's performance:

Model training performance using the **data loader of this branch** (DeepAR, M4_hourly, ml.p3.2xlarge, 1000 epochs):
image

Model training performance using the **current master data loader** (DeepAR, M4_hourly, ml.p3.2xlarge, 2000 epochs):
image

vafl (dismissed):
This is great, guys! Very clean and neat.

We previously had multiple times regressions in the performance and in particular memory usage when this code was touched so the comparisons are very useful @PascalIversen. I wonder if we could include this as automated tests (or nightly runs) to catch such regressions. Perhaps you are already looking into this?

src/gluonts/mx/batchify.py
```diff
@@ -0,0 +1,112 @@
...
+        data = _pad_arrays(data, axis=0)
+    if isinstance(data[0], mx.nd.NDArray):
+        # TODO: think about using shared context NDArrays
+        #  https://github.com/awslabs/gluon-ts/blob/42bee73409f801e7bca73245ca21cd877891437c/src/gluonts/dataset/parallelized_loader.py#L157
+        return mx.nd.stack(*data)
+    if isinstance(data[0], np.ndarray):
+        data = mx.nd.array(data, dtype=dtype, ctx=ctx)
+    elif isinstance(data[0], (list, tuple)):
+        return list(stack(t, ctx=ctx) for t in zip(*data))
+    return data
```
jaheba, line 80: I'm not sure we actually do need all these cases. When I adapted `stack` for glide, this version worked:

```python
def stack(data, ctx=None, dtype=np.float32):
    peek = data[0]
    if isinstance(peek, (mx.nd.NDArray, np.ndarray)) and not 0 in peek.shape:
        return mx.nd.array(data, dtype=dtype, ctx=ctx)

    return data
```

src/gluonts/mx/batchify.py
```diff
@@ -0,0 +1,112 @@
...
+        k: v.as_in_context(ctx) if isinstance(v, mx.nd.NDArray)
+        # Workaround due to MXNet not being able to handle NDArrays with 0 in shape properly:
+        else (
+            stack(v, ctx=ctx, dtype=v.dtype, variable_length=False)
+            if isinstance(v[0], np.ndarray) and 0 in v[0].shape
+            else v
+        )
+        for k, v in batch.items()
+    }
+    return batch
```
jaheba, line 112: Similarly, I think this can be radically simplified.

src/gluonts/shell/train.py
```diff
@@ -133,11 +133,9 @@ def run_test(
...
     test_dataset = TransformedDataset(
         base_dataset=test_dataset,
-        transformations=[
-            FilterTransformation(
-                lambda x: x["target"].shape[-1] > predictor.prediction_length
-            )
-        ],
+        transformation=FilterTransformation(
+            lambda x: x["target"].shape[-1] > predictor.prediction_length
+        ),
```
jaheba, line 138: I'm not a fan of passing the keywords here.

src/gluonts/dataset/loader.py
```diff
@@ -11,186 +11,270 @@
...
+from gluonts.dataset.util import MPWorkerInfo, batcher
 from gluonts.transform import Transformation
+from gluonts.transform.dataset import TransformedDataset
+
+
+logger = logging.getLogger(__name__)
 
 
-class DataLoader(Iterable[DataEntry]):
+class CyclicIterable(Iterable):
```
jaheba, line 34:
```suggestion
class Cycle(Iterable):
```

I think it makes sense to abbreviate this. We are not Java after all.
`itertools.cycle`

src/gluonts/dataset/loader.py
```diff
@@ -11,186 +11,270 @@
...
+    def __init__(self, base_iterable: Iterable) -> None:
```
jaheba, line 35: Not a fan of `base_iterable`. Why not just `iterable` like here: https://docs.python.org/2/library/itertools.html#itertools.cycle

src/gluonts/dataset/loader.py
```diff
@@ -11,186 +11,270 @@
...
-        The length of the buffer used to do pseudo shuffle.
-        If not None, the loader will perform pseudo shuffle when generating batches.
-        Note that using a larger buffer will provide more randomized batches, but will make the job require a bit
-        more time to be done.
+    A wrapper class which takes a serialized iterator as an input and generates a
+    pseudo randomized iterator using the same elements from the input iterator.
     """
 
+    def __init__(self, base_iterator: Iterator, shuffle_buffer_length: int):
+        self.shuffle_buffer: list = []
```
jaheba, line 50:
```suggestion
        self.shuffle_buffer = []
```

src/gluonts/dataset/loader.py
```diff
@@ -11,186 +11,270 @@
...
+        self.shuffle_buffer_length = shuffle_buffer_length
+        self.base_iterator = base_iterator
+        self.base_iter_finished = False
+
+    def __next__(self):
+        # if the buffer is empty, fill the buffer first.
+        # (Should only executed in the first round)
```
jaheba, line 57: "Should"? When would that be also the case?

src/gluonts/dataset/loader.py
```diff
@@ -11,186 +11,270 @@
...
+            )
+
+        # If buffer still empty, means all elements used, return a signal of
+        # end of iterator
+        if not self.shuffle_buffer:
+            raise StopIteration
+
+        # Choose an element at a random index and yield it and fill it with
+        # the next element in the sequential generator
+        idx = random.randint(0, len(self.shuffle_buffer) - 1)
```
jaheba, line 72:
```suggestion
        idx = random.randrange(len(self.shuffle_buffer))
```

src/gluonts/dataset/loader.py
```diff
@@ -11,186 +11,270 @@
...
-        yield from self.parallel_data_loader
+        self.manager = Manager()
+        self.data_queue = self.manager.Queue(maxsize=self.max_queue_size)
+        self.terminate_event = self.manager.Event()
+        self.exhausted_events = [
+            self.manager.Event() for _ in range(self.num_workers)
+        ]
+        self.processes = []
 
+        for wid, event in enumerate(self.exhausted_events):
```
jaheba, line 115: What is `wid`?

src/gluonts/dataset/loader.py
```diff
@@ -11,186 +11,270 @@
...
+            self.terminate_event.set()
+        except FileNotFoundError:
+            pass
+        # Empty queue to make sure workers get the message
+        self._empty_queue()
+        for p in self.processes:
+            p.join()
+
+    def __del__(self):
+        self._halt_processes()
```
jaheba, line 205:
```suggestion
    def __del__(self):
        self._halt_processes()
```

src/gluonts/dataset/loader.py
```diff
@@ -11,186 +11,270 @@
...
-            **kwargs,
+        self.max_queue_size = (
+            max_queue_size if max_queue_size is not None else 5 * num_workers
         )
 
-    def __iter__(self) -> Iterator[DataBatch]:
-        # Will take all batches, so that all data is sampled exactly once
-        yield from self.parallel_data_loader
+        self.manager = Manager()
+        self.data_queue = self.manager.Queue(maxsize=self.max_queue_size)
```
jaheba, line 108: `data_queue` is not really a helpful name. Is this the input or the output queue?

src/gluonts/dataset/loader.py
```diff
@@ -11,186 +11,270 @@
...
+        self.terminate_event = self.manager.Event()
+        self.exhausted_events = [
+            self.manager.Event() for _ in range(self.num_workers)
+        ]
```
jaheba, line 112: Not sure I like this design. Personally, I prefer using sentinel values.

src/gluonts/dataset/loader.py
```diff
@@ -11,186 +11,270 @@
...
+        data_queue: Queue,
+        terminate_event,
+        exhausted_event,
+    ):
+        MPWorkerInfo.worker_process = True
+        MPWorkerInfo.worker_id = worker_id
+        MPWorkerInfo.num_workers = num_workers
+
+        for batch_list in batcher(iterable, batch_size):
+            batch = batchify_fn(batch_list)
```
jaheba, line 150: Don't like `batch_list`. Can we name `batch_list` -> `batch` and then `batch` -> `stacked_batch`?

src/gluonts/dataset/loader.py
```diff
@@ -11,186 +11,270 @@
...
-    """
+        self.count = 0
+
+    @staticmethod
+    def worker_fn(
+        worker_id: int,
+        num_workers: int,
+        iterable: Iterable,
+        batch_size: int,
+        batchify_fn: Callable,
```
jaheba, line 140: I don't understand this. We `batchify` a batch? This doesn't make sense to me.

Can we find better descriptions to what we actually do here?

src/gluonts/dataset/loader.py
```diff
@@ -11,186 +11,270 @@
...
+    def __iter__(self):
+        return self
 
+    def __next__(self):
+        if (
+            all(event.is_set() for event in self.exhausted_events)
+            and self.data_queue.empty()
+        ):
+            self._halt_processes()
+            raise StopIteration()
```
jaheba, line 173:
```suggestion
            raise StopIteration
```

src/gluonts/dataset/loader.py
```diff
@@ -11,186 +11,270 @@
...
-        dtype: DType = np.float32,
-        num_workers: Optional[int] = None,
-        num_prefetch: Optional[int] = None,
-        shuffle_buffer_length: Optional[int] = None,
-        **kwargs,
-    ) -> None:
+        batchify_fn: Callable,
+        num_workers: int,
+        max_queue_size: Optional[int] = None,
+        todevice_fn: Callable = lambda x: x,
```
jaheba, line 93: I think `todevice_fn` is not a good name.

In my experiments I had a pair of `encode() / decode()` functions, which were called just before putting things into the queue, and just after when popping a value from it.

src/gluonts/dataset/loader.py
```diff
@@ -11,186 +11,270 @@
...
+                self.data_queue.get(block=False)
+        except (Empty, FileNotFoundError):
+            pass
+
+    def _halt_processes(self):
+        try:
+            # Send termination message to workers
+            self.terminate_event.set()
+        except FileNotFoundError:
+            pass
```
jaheba, line 198: Just ignore it? This doesn't look right to me.

jaheba (changes requested):

src/gluonts/dataset/loader.py
lostella: `stack_fn` would be better?

jaheba: Sounds better to me.

src/gluonts/dataset/loader.py
lostella: This was to make mypy happy

src/gluonts/dataset/loader.py
lostella: depends who you ask: is the output queue for the workers, it's the input queue for the main process

src/gluonts/dataset/loader.py
lostella: not sure what the suggestion here is

src/gluonts/dataset/loader.py
lostella: if it's just a matter of taste, then let's leave it as it is

src/gluonts/dataset/loader.py
jaheba: Not sure I can follow that logic. Can we name it `result_queue` or something similar?

src/gluonts/dataset/loader.py
jaheba: Delete the method?

src/gluonts/dataset/loader.py
jaheba: I think mypy might have complained because it didn't know what the containing type is. With `list` probably indicate `List[Any]`.

jaheba (approved):

PR #900, opened 2020-06-29
Corrected the formula for the stddev of the MixtureDistribution.

PascalIversen:
*Description of changes:*
Corrected the formula of the stddev of the MixtureDistribution.

### DIFF
--- a/src/gluonts/mx/distribution/mixture.py
+++ b/src/gluonts/mx/distribution/mixture.py
@@ -122,14 +122,22 @@ class MixtureDistribution(Distribution):
     @property
     def stddev(self) -> Tensor:
         F = self.F
-        sq_mean_values = F.square(F.stack(*[c.mean for c in self.components], axis=-1))
-        sq_std_values = F.square(F.stack(*[c.stddev for c in self.components], axis=-1))
+        sq_mean_values = F.square(
+            F.stack(*[c.mean for c in self.components], axis=-1)
+        )
+        sq_std_values = F.square(
+            F.stack(*[c.stddev for c in self.components], axis=-1)
+        )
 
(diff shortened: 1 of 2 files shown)


lostella:
Nice, thanks! I think a test should have caught this: in fact, there's a test script for mixture, but no assertion on the stddev/variance (except for a mention in a comment, see https://github.com/awslabs/gluon-ts/blob/master/test/distribution/test_mixture.py#L128-L132). Could you modify that test to also do assertions on the variance?

src/gluonts/mx/distribution/mixture.py
```diff
@@ -122,10 +122,21 @@ def cdf(self, x: Tensor) -> Tensor:
...
+        )
+
+        return F.sqrt(
+            F.sum(
+                F.broadcast_mul(
+                    sq_mean_values + sq_std_values, self.mixture_probs, axis=-1
+                ),
+                axis=-1,
+            )
+            - F.square(self.mean)
```
lostella, line 139: I think this will work for the univariate case only, but it's fine like this for now and we can open a separate issue about further improving it

test/distribution/test_mixture.py
```diff
@@ -127,9 +127,12 @@ def test_mixture(
 
     # check mean and stddev
     calc_mean = mixture.mean.asnumpy()
+    calc_std = mixture.std.asnumpy()
     sample_mean = samples_mix.asnumpy().mean(axis=0)
+    sample_std = np.sqrt(np.mean(np.square(samples_mix - sample_mean)))
```
lostella, line 132: Probably `sample_std = samples_mix.asnumpy().std(axis=0)` will give you the same result

PR #904, opened 2020-07-01
Removed TimeSeriesItem.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/dataset/common.py
+++ b/src/gluonts/dataset/common.py
@@ -100,6 +100,44 @@ class TrainDatasets(NamedTuple):
     train: Dataset
     test: Optional[Dataset] = None
 
+    def save(self, path_str: str, overwrite=True) -> None:
+        """
+        Saves an TrainDatasets object to a JSON Lines file.
+
+        Parameters
+        ----------
+        path_str
+            Where to save the dataset.
+        overwrite
+            Whether to delete previous version in this folder.
+        """
+        import shutil
+        import ujson as json
+
+        path = Path(path_str)
+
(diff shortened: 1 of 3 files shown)


test/dataset/test_loader.py
```diff
@@ -79,81 +77,14 @@ def load_file_dataset_numpy(path: Path, freq: str) -> Iterator[Any]:
...
-        )
+    yield from FileDataset(path, freq)
 
 
 def load_list_dataset(path: Path, freq: str) -> Iterator[Any]:
     lines = (line.content for line in load_json_lines_file(path, freq))
     return iter(ListDataset(lines, freq))
 
 
-# @pytest.mark.skip()
```
lostella, line 93: Why not just skip this? I'm assuming you're removing this because it raises false positives, right?

src/gluonts/dataset/common.py
```diff
@@ -500,46 +433,6 @@ def load_datasets(
     return TrainDatasets(metadata=meta, train=train_ds, test=test_ds)
 
 
-def save_datasets(
-    dataset: TrainDatasets, path_str: str, overwrite=True
-) -> None:
```
lostella, line 505: Why removing this?

jaheba: It's never used.

test/dataset/test_loader.py
jaheba: I'm not sure about the value this test provides to us. Also, it uses `save_dataset` which I want to remove :)

PR #905, opened 2020-07-02
Generalized pareto distribution && Unit test

BingLiHanShuang:
*Issue #, if available:*

*Description of changes: Add new distribution: Generalized pareto distribution which used in Extreme value theroy. And add unit test for Generalized pareto distribution.


BingLiHanShuang:
Update and pull another pull  request about Generalized pareto distribution && Unit test.

Issue #906, opened 2020-07-02
DeepAR + NegativeBinomialOutput performance degradation after upgrading from v0.4.2 to 0.5.0


jsirol:
Continuing the discussion started as a part of [#872](https://github.com/awslabs/gluon-ts/issues/872) regarding the experienced performance degradation of NegativeBinomialOutput (NB) with DeepAR after upgrading from `v0.4.2` to `v0.5.0`,  I thought to open a dedicated issue for this as I've conducted some extra steps to try to understand this without major success except narrowing the problem down to the convergence of the algorithm after the scaling change done in #719 . 

With the data I am modelling with NB in `0.5.0`, I tried various hyperparameter settings including changing the learning rate, but all of the trained models had the same issue of too high forecasts and poor convergence. This seems to be a problem especially for the time series in the data for which the scale of the series is high.

Today I tried to replicate the issue with some synthetic data, and managed to get similar behaviour rather easily. The code to replicate is attached below. Basically the algorithm with NegativeBinomialOutput doesn't seem to converge very well and I end up with forecasts that are way above the actuals. I compared this with e.g. the StudentTOutput and PoissonOutput which both work fine out of the box (the same is true for the real data I am working on).

The created data (first 16 time series) look like this:
![image](https://user-images.githubusercontent.com/5380581/86370492-7ef85700-bc88-11ea-9dbc-75d62229addd.png)

After training the models and comparing the forecasts on the test sets the results (first 64 series) look like this, the NB forecasts are way off even though the other forecasts are fine, and the only change in the model definition is the `distr_output`:
![image](https://user-images.githubusercontent.com/5380581/86370623-a818e780-bc88-11ea-9f47-fb1cf34b41a5.png)

To try to isolate the problem, I tried changing the line that was changed in #719 for the NB back to what it was in `0.4.2` i.e. `alpha = F.broadcast_mul(alpha, F.sqrt(scale + 1.0))`, and with this I see better convergence and similar results to Student-t and Poisson outputs. So there seem to be some convergence/other issues with the new scaling that I have not managed to find a way to fix. 

This really makes me wonder whether I am missing something important in the optimiser settings or if there is an issue somewhere else. Any help in understanding this is appreciated. In the meanwhile, I am using the Poisson/Student-t Outputs to give me comparable results in `v0.5.0` to what I used to get with `v0.4.2` and the NB.

The code to produce the results below:

```python
import matplotlib.pyplot as plt
import numpy as np
import pandas as pd
[...]

PR #909, opened 2020-07-03
Scale the negative binomial's gamma()

kashif:
For issue #906 

The `theta` param. of Gamma gets scaled by `scale *theta = alpha * mu * scale`. So only need to scale the `mu` . See: https://en.wikipedia.org/wiki/Gamma_distribution#Scaling

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/mx/distribution/neg_binomial.py
+++ b/src/gluonts/mx/distribution/neg_binomial.py
@@ -133,9 +133,7 @@ class NegativeBinomialOutput(DistributionOutput):
             return NegativeBinomial(mu, alpha)
         else:
             F = getF(mu)
-            scale = 1.0 + softplus(F, scale - 1.0)
             mu = F.broadcast_mul(mu, scale)
-            alpha = F.broadcast_add(alpha, F.broadcast_div(scale - 1, mu))
             return NegativeBinomial(mu, alpha, F)
 
     @property


lostella:
Thanks @kashif, this looks good to me. It would be nice to validate this against some artificial examples, both with small and large scale (eg the snippet in #814). But the example you included in #906 looks already very good

kashif:
```python
data = np.random.negative_binomial(n=10, p=0.9, size=(200,))

data_series = pd.Series(
    data=data,
    index=pd.date_range(
        start='2014-01-05 00:00:00',
        periods=len(data),
        freq='w'
    )
)

dataset = ListDataset(
    data_iter=[
        {
            "start": '2014-01-05 00:00:00',
            "target": list(data)
        }
    ],
    freq="w"
)

    freq="w",
    prediction_length=20,
    distr_output=NegativeBinomialOutput(),
    trainer=Trainer(epochs=20, device='cuda'),
)

predictor = estimator.train(dataset)

forecast = next(iter(predictor.predict(dataset)))
data_series.plot()
forecast.plot()
plt.show()
```

download

canerturkmen:
Hi Kashif, Thanks for your contribution and not letting go of this issue. Just to help me understand, I understand #814 led to a degradation. I can't immediately make the connection between how Gamma scaling would apply directly here, although I have some intuition about how it would apply approximately. Perhaps, for very small scales this could still fail, for reasons analogous to the discussion in #814?

kashif:
@canerturkmen ok so the degradation was not due to #814, but my attempt to fix the original scaling issue in the NB output i.e. #718. There i was trying to "fix" the scaling as proposed by the deepAR paper. This I have now come realise is wrong... after looking at the issue again and questioning my assumptions I realized that the proper way of scaling the NB is to sample from a `scale*rate` gamma distribution which then means due to the scaling property of the gamma distribution in the parametrization of gluton-ts we just need to scale the `mu`. I hope that clears up the issue?

canerturkmen:
Yes, this makes sense. Though I'm surprised this didn't drop out of the algebra for #814. Thanks!

kashif:
@canerturkmen I too was / am surprised at that... I'm gonna go through the algebra again but perhaps not around midnight haha :sleeping:

PR #911, opened 2020-07-07
Shape squeeze edge case bug fix

benidis:
Fixes an edge case where we pass a single value and `np.squeeze` suppresses all dimensions.

### DIFF
--- a/src/gluonts/transform/feature.py
+++ b/src/gluonts/transform/feature.py
@@ -107,7 +107,8 @@ class LastValueImputation(MissingValueImputation):
         out = values[np.arange(idx.shape[0])[:, None], idx]
 
         values = np.squeeze(out)
-
+        if values.shape == ():  # edge case when we pass a single value
+            values = values.reshape(1,)
         # in case we need to replace nan at the start of the array
         mask = np.isnan(values)
         values[mask] = np.interp(
--- a/src/gluonts/transform/feature.py
+++ b/src/gluonts/transform/feature.py
@@ -87,6 +87,8 @@ class MeanValueImputation(MissingValueImputation):
     """
 
     def __call__(self, values: np.ndarray) -> np.ndarray:
+        if len(values) == 1:
+            return DummyValueImputation()(values)
         nan_indices = np.where(np.isnan(values))
         values[nan_indices] = np.nanmean(values)
         return values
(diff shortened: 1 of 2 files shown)


src/gluonts/transform/feature.py
```diff
@@ -99,6 +99,8 @@ class LastValueImputation(MissingValueImputation):
     """
 
     def __call__(self, values: np.ndarray) -> np.ndarray:
+        if len(values) == 1:
+            return DummyValueImputation()(values)
```
lostella, line 103: This would need to be converted to a float I guess

```suggestion
            return DummyValueImputation()(values.item())
```

benidis: No it should be numpy array: `def __call__(self, values: np.ndarray) -> np.ndarray:`. 

A simple example (copying what the dummy value imputation does) with numpy array:
```
values = np.array([1])
nan_indices = np.where(np.isnan(values))
values[nan_indices] = 0
print(values)
[1]
```

and with float:
```
values = np.array([1]).item()
nan_indices = np.where(np.isnan(values))
values[nan_indices] = 0
print(values)

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-97-8448c7edc37d> in <module>
      1 values = np.array([1]).item()
      2 nan_indices = np.where(np.isnan(values))
----> 3 values[nan_indices] = 0
      4 print(values)

TypeError: 'int' object does not support item assignment

```

benidis:
> Is this edge case covered by tests?

No, I am adding a test.

src/gluonts/transform/feature.py
lostella: yes, sorry, I got confused between __init__ and __call__

src/gluonts/transform/feature.py
```diff
@@ -87,6 +87,8 @@ class MeanValueImputation(MissingValueImputation):
...
```
lostella, line 91: Would it make sense to pass a user specified dummy_value when constructing the DummyValueImputation object here? Essentially, all these classes would allow specifying the dummy_value, to be used in extreme cases.

But maybe not urgent and can be addressed later.

benidis: Yeah, so the `DummyValueImputation` accepts a value that uses to impute. With the current implementation it uses the default. We can always add one more argument to all the imputation methods to override the default but maybe this is too much. If needed we can always include it easily.

PR #913, opened 2020-07-08
NanMixture: Distribution to model missing values

PascalIversen:
*Description of changes:*
Added a distribution to model data which contains missing values.

### DIFF
--- a/src/gluonts/model/simple_feedforward/_network.py
+++ b/src/gluonts/model/simple_feedforward/_network.py
@@ -159,6 +159,8 @@ class SimpleFeedForwardTrainingNetwork(SimpleFeedForwardNetworkBase):
         )
 
         if isinstance(distr, NanMixture):
+            # demasking the missing values for the future_targets
+
             loss = distr.loss(
                 F.where(
                     future_observed_values,
--- a/src/gluonts/model/simple_feedforward/_network.py
+++ b/src/gluonts/model/simple_feedforward/_network.py
@@ -23,7 +23,7 @@ from typing import Tuple
 
 # First-party imports
 from gluonts.mx.block.scaler import MeanScaler, NOPScaler
-from gluonts.mx.distribution import DistributionOutput
+from gluonts.mx.distribution import DistributionOutput, NanMixture
 from gluonts.support.util import weighted_average
 
 
@@ -158,15 +158,22 @@ class SimpleFeedForwardTrainingNetwork(SimpleFeedForwardNetworkBase):
             distr_args, loc=loc, scale=scale
         )
 
-        # (batch_size, prediction_length, target_dim)
-        loss = distr.loss(future_target)
-
-        weighted_loss = weighted_average(
-            F=F, x=loss, weights=future_observed_values, axis=1
-        )
-
-        # (batch_size, )
-        return weighted_loss
+        if isinstance(distr, NanMixture):
+            loss = distr.loss(
+                F.where(
+                    future_observed_values,
+                    future_target,
+                    0.0 / future_target.zeros_like(),
+                )
+            )
+            return loss
+        else:
+            loss = distr.loss(future_target)
+            weighted_loss = weighted_average(
+                F=F, x=loss, weights=future_observed_values, axis=1
+            )
+            # (batch_size, )
+            return weighted_loss
 
 
 class SimpleFeedForwardSamplingNetwork(SimpleFeedForwardNetworkBase):
--- a/src/gluonts/model/simple_feedforward/_network.py
+++ b/src/gluonts/model/simple_feedforward/_network.py
@@ -23,7 +23,7 @@ from typing import Tuple
 
 # First-party imports
 from gluonts.mx.block.scaler import MeanScaler, NOPScaler
-from gluonts.mx.distribution import DistributionOutput, NanMixture
+from gluonts.mx.distribution import DistributionOutput
 from gluonts.support.util import weighted_average
 
 
@@ -108,11 +108,11 @@ class SimpleFeedForwardNetworkBase(mx.gluon.HybridBlock):
         Returns
         -------
         Tuple[Tensor, Tensor Tensor]
-            A tuple of three tensors ``distr_args``, ``loc``, and ``scale``: the 
-            first contains the output distribution parameters, and the others are 
+            A tuple of three tensors ``distr_args``, ``loc``, and ``scale``: the
+            first contains the output distribution parameters, and the others are
             location and scale of the distribution.
- 
-        
+
+
         """
         scaled_target, target_scale = self.scaler(
             past_target, F.ones_like(past_target),
@@ -158,24 +158,15 @@ class SimpleFeedForwardTrainingNetwork(SimpleFeedForwardNetworkBase):
             distr_args, loc=loc, scale=scale
         )
 
-        if isinstance(distr, NanMixture):
-            # demasking the missing values for the future_targets
+        # (batch_size, prediction_length, target_dim)
+        loss = distr.loss(future_target)
 
-            loss = distr.loss(
-                F.where(
-                    future_observed_values,
-                    future_target,
-                    0.0 / future_target.zeros_like(),
-                )
-            )
-            return loss
-        else:
-            loss = distr.loss(future_target)
-            weighted_loss = weighted_average(
-                F=F, x=loss, weights=future_observed_values, axis=1
-            )
-            # (batch_size, )
-            return weighted_loss
+        weighted_loss = weighted_average(
+            F=F, x=loss, weights=future_observed_values, axis=1
+        )
+
+        # (batch_size, )
+        return weighted_loss
 
 
 class SimpleFeedForwardSamplingNetwork(SimpleFeedForwardNetworkBase):
@@ -238,7 +229,7 @@ class SimpleFeedForwardDistributionNetwork(SimpleFeedForwardNetworkBase):
         -------
         distr_args: the parameters of distribution
         loc: an array of zeros with the same shape of scale
-        scale: 
+        scale:
         """
         distr_args, loc, scale = self.get_distr_args(F, past_target)
         return distr_args, loc, scale
--- a/src/gluonts/mx/distribution/categorical.py
+++ b/src/gluonts/mx/distribution/categorical.py
@@ -91,7 +91,7 @@ class Categorical(Distribution):
             return indices
 
         return _sample_multiple(s, self.probs, num_samples=num_samples).astype(
-            "int32"
+            dtype
         )
 
     @property
--- a/src/gluonts/mx/distribution/categorical.py
+++ b/src/gluonts/mx/distribution/categorical.py
@@ -121,7 +121,7 @@ class CategoricalOutput(DistributionOutput):
     def distribution(
         self, distr_args, loc=None, scale=None, **kwargs
     ) -> Distribution:
-        distr = Categorical(distr_args)
+        distr = Categorical(*distr_args)
         return distr
 
     @property
--- a/src/gluonts/mx/distribution/categorical.py
+++ b/src/gluonts/mx/distribution/categorical.py
@@ -121,7 +121,7 @@ class CategoricalOutput(DistributionOutput):
     def distribution(
         self, distr_args, loc=None, scale=None, **kwargs
     ) -> Distribution:
-        distr = Categorical(*distr_args)
+        distr = Categorical(distr_args)
         return distr
 
     @property
--- a/src/gluonts/mx/distribution/deterministic.py
+++ b/src/gluonts/mx/distribution/deterministic.py
@@ -1,0 +1,161 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+# Standard library imports
+import math
+from functools import partial
+from typing import Dict, List, Optional, Tuple
+
+# Third-party imports
+import numpy as np
+
+from gluonts.core.component import validated
+
+# First-party imports
+from gluonts.model.common import Tensor
+from gluonts.support.util import erf, erfinv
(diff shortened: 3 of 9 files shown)


src/gluonts/mx/distribution/nan_mixture.py
```diff
@@ -0,0 +1,173 @@
...
+        # stack log probabilities of components
+        component_log_likelihood = F.stack(
+            *[non_nan_dist_log_likelihood, self.components[1].log_prob(x)],
+            axis=-1,
+        )
+        # compute mixture log probability by log-sum-exp
+        summands = log_mix_weights + component_log_likelihood
+        max_val = F.max_axis(summands, axis=-1, keepdims=True)
+
+        # catch edge case max_val = -inf
```
benidis, line 110: When is this edge case happening? In theory you should never have this unless the values of both distributions are `-inf`, right? But if the value is not nan then the proper distribution should not give `-inf`.

src/gluonts/mx/distribution/categorical.py
```diff
@@ -121,7 +121,7 @@ def domain_map(self, F, probs):
     def distribution(
         self, distr_args, loc=None, scale=None, **kwargs
     ) -> Distribution:
-        distr = Categorical(distr_args)
+        distr = Categorical(*distr_args)
```
lostella, line 124: This could go in a separate PR, along with some tests

src/gluonts/mx/distribution/deterministic.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def sample(
+        self, num_samples: Optional[int] = None, dtype=np.int32
+    ) -> Tensor:
+        # TODO implement dtype?
+        return _sample_multiple(
+            lambda value: value, value=self.value, num_samples=num_samples
+        )
+
+    def quantile(self, level: Tensor) -> Tensor:
```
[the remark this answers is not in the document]
PascalIversen: Mostly, but there are some edge cases: The quantile of p=0 is for example -inf. Furthermore, if the deterministic distribution is NaN-valued the quantile is always NaN.

src/gluonts/mx/distribution/nan_mixture.py
PascalIversen: I forgot to remove that, thanks! This is needed if we allow nan_prob=1, but we realized  that nan_prob has to be in (0,1) for multiple reasons. I will remove it.

PascalIversen:
> Thanks Pascal, looks great overall. I see there are some TODOs here and there... are these things you are working on including now?

Thank you for reviewing, Konstantinos. 
I forgot to remove some of these TODO's. The only thing left is concerning the testing of the NanMixture. Since we haven't decided if the mean/stddev should return NaN or the values of the Non-NaN component, I would leave it for now.

src/gluonts/mx/distribution/categorical.py
PascalIversen: As discussed offline, this was a mistake and I removed it again.

test/distribution/test_nan_mixture.py
```diff
@@ -0,0 +1,344 @@
...
+
+    # TODO check mean and stddev
+
+    x = mx.nd.array([[[np.nan, 3.5, -0.5]], [[np.nan, 3.5, np.nan]]])
+
+    # check log_prob
+    log_prob = nan_mixture.log_prob(x)
+    log_prob_true = mx.nd.log(mx.nd.where(x != x, p, (1 - p) * distr.prob(x)))
+
+    assert np.allclose(log_prob.asnumpy(), log_prob_true.asnumpy(), atol=1e-1)
```
lostella, line 106: I’m wondering if these allclose could use lower tolerance (or just the default one). Same for other calls below in this same test. Are expecting these two values to be “somewhat close” or essentially the same?

test/distribution/test_nan_mixture.py
```diff
@@ -0,0 +1,344 @@
...
+            Categorical(log_probs=mx.nd.log(mx.nd.array(p_cat))),
+            mx.nd.array(p),
+            mx.nd.array(p_cat),
+        ),
+    ],
+)
+@pytest.mark.parametrize("serialize_fn", serialize_fn_list)
+def test_nan_mixture(
+    distr: Distribution,
+    cat_distr: Distribution,
```
lostella, line 75: Why do we take two distributions as input? Couldn’t these two just be two test cases? Same for p and p_cat.

Unless the test is significantly different between the categorical case vs other distributions: but in that case maybe we could split this test function into two test functions

test/distribution/test_nan_mixture.py
```diff
@@ -0,0 +1,344 @@
...
+    for _ in t:
+        with mx.autograd.record():
+            distr_args = args_proj(input)
+            d = nmdo.distribution(distr_args)
+            loss = d.loss(mixture_samples)
+        loss.backward()
+
+        loss_value = loss.mean().asnumpy()
+        t.set_postfix({"loss": loss_value})
+        trainer.step(BATCH_SIZE)
```
lostella, line 285: This is just an idea (and we could do this in a separate PR): we know that EM works better than SGD for the job, would it make sense to use EM in this type of tests? As long as this would involve invoking the loss (and its backward) then it should be fine. In that case we could probably hope for “stricter” assertions on the recovered parameters below

test/distribution/test_nan_mixture.py
```diff
@@ -0,0 +1,344 @@
...
+
+    distr_args = args_proj(input)
+    d = nmdo.distribution(distr_args)
+
+    cat_probs_hat = d.distribution.probs.asnumpy()
+    nan_prob_hat = d.nan_prob.asnumpy()
+
+    print(loss_value)
+    print(cat_probs_hat)
+    print(nan_prob_hat)
```
lostella, line 338: These prints we can remove I think

test/distribution/test_nan_mixture.py
PascalIversen: We could do that for the Mixture distribution. However, for the NanMixture the maximisation step is not defined. Also, EM does not use the gradient. Maybe in a separate PR I could implement something like  ["A Gradient Algorithm Locally Equivalent to the EM Algorithm"](https://doi.org/10.1111/j.2517-6161.1995.tb02037.x) which uses the gradient and would, I think, work for the NanMixture.

test/distribution/test_nan_mixture.py
PascalIversen: Yes, thanks. Default 1e-8 makes sense and works.

PascalIversen:
I created some artificial time series data with missing values and fitted a Gaussian NanMixture using the SimpleFeedForward estimator.
I had to make one more adjustment to the SimpleFeedForward's hybrid_forward() and I am not sure if it should be part of this PR. Namely, I have to demask the NaN values before the loss calculation:

        if isinstance(distr, NanMixture):
            # demasking the missing values for the future_targets

            loss = distr.loss(
                F.where(
                    future_observed_values,
                    future_target,
                    0.0 / future_target.zeros_like(),
                )
            )
            return loss
        else:
            loss = distr.loss(future_target)
            weighted_loss = weighted_average(
                F=F, x=loss, weights=future_observed_values, axis=1
            )
            # (batch_size, )
            return weighted_loss

In my experiments, the parameters are recovered:


This was done using this code:
```

# Third-party imports
get_ipython().run_line_magic('matplotlib', 'inline')
import mxnet as mx
from mxnet import gluon
import numpy as np
import pandas as pd


from gluonts.dataset.repository.datasets import get_dataset, dataset_recipes



def nan_prob_gen():
    while(True):
        for i in np.linspace(0.01, 0.3, 12):
            yield i
        for i in np.linspace(0.3, 0.01, 12):
            yield i




def create_dataset(num_series, num_steps, period=24, mu=1, sigma=0.3, true_mean=False):
    # create target: noise + pattern    
    # noise
    noise = np.random.normal(mu, sigma, size=(num_series, num_steps))
    
    # pattern - sinusoid with different phase
    sin_minumPi_Pi = np.sin(np.tile(np.linspace(-np.pi, np.pi, period), int(num_steps / period)))
    sin_Zero_2Pi = np.sin(np.tile(np.linspace(0, 2 * np.pi, period), int(num_steps / period)))
    
    target = np.tile(sin_minumPi_Pi.reshape(1, -1), 
                                      (int(np.ceil(num_series)),1))

    if(not true_mean):
        nan_gen = nan_prob_gen()
        nan_prob = []
        for _ in range(num_steps):
            nan_prob.append(next(nan_gen))

        is_nan = np.random.uniform(size = target.shape)<nan_prob
        target = np.where(is_nan, np.nan, target)
        
        target = noise + target
    
        
    else:
        target += mu
    return target


# define the parameters of the dataset
n_series = 1200
custom_ds_metadata = {'num_series': n_series,
                      'num_steps': 24 * 7,
                      'prediction_length': 24 ,
                      'freq': '1H',
                      'start': [pd.Timestamp("01-01-2019", freq='1H') 
                                for _ in range(n_series)]
                     }



sigma = 0.3
target = create_dataset(custom_ds_metadata['num_series'], 
                          custom_ds_metadata['num_steps'],                                                      
                          custom_ds_metadata['prediction_length'],
                        sigma=sigma
                         )
train_ds = ListDataset([{FieldName.TARGET: target, 
                         FieldName.START: start} 
                        for (target, start) in zip(target[:, :-custom_ds_metadata['prediction_length']], 
                                                             custom_ds_metadata['start'])],
                      freq=custom_ds_metadata['freq'])
test_ds = ListDataset([{FieldName.TARGET: target, 
                        FieldName.START: start} 
                       for (target, start) in zip(target, 
                                                            custom_ds_metadata['start'])],
                     freq=custom_ds_metadata['freq'])


train_entry = next(iter(train_ds))
test_entry = next(iter(test_ds))



test_series = to_pandas(test_entry)
train_series = to_pandas(train_entry)

fig, ax = plt.subplots(2, 1, sharex=True, sharey=True, figsize=(10, 7))

train_series.plot(ax=ax[0])
ax[0].grid(which="both")
ax[0].legend(["train series"], loc="upper left")

test_series.plot(ax=ax[1])
ax[1].axvline(train_series.index[-1], color='r') # end of train dataset
ax[1].grid(which="both")
ax[1].legend(["test series", "end of train series"], loc="upper left")

plt.show()
```


```


from gluonts.model.simple_feedforward import SimpleFeedForwardEstimator
from gluonts.mx.distribution import NanMixtureOutput, GaussianOutput




    num_hidden_dimensions=[24, 24],
    prediction_length=custom_ds_metadata['prediction_length'],
    context_length=2*custom_ds_metadata['prediction_length'],
    freq=custom_ds_metadata['freq'],
    trainer=Trainer(ctx="cpu", 
                    epochs=30, 
                    learning_rate=1e-3, 
                    hybridize=True, 
                    num_batches_per_epoch=n_series//16,
                    batch_size = 16
                   ),
    distr_output=NanMixtureOutput(GaussianOutput()),
    sampling=False
)



predictor = estimator.train(train_ds)





    dataset=test_ds,  # test dataset
    num_samples=4000,  # number of sample paths we want for evaluation
)
forecasts = list(forecast_it)
tss = list(ts_it)




# get the true means
target_mean = create_dataset(custom_ds_metadata['num_series'], 
                          custom_ds_metadata['num_steps'],                                                      
                          custom_ds_metadata['prediction_length'],
                        true_mean=True
                         )

test_ds_mean = ListDataset([{FieldName.TARGET: target, 
                        FieldName.START: start} 
                       for (target, start) in zip(target_mean, 
                                                            custom_ds_metadata['start'])],
                     freq=custom_ds_metadata['freq'])
true_mean = next(iter(test_ds_mean))["target"]



# get predicted distribution parameter
forecast_entry = forecasts[0]
mu_hat = forecast_entry.distribution.distribution.base_distribution.mu.asnumpy()
sigma_hat = forecast_entry.distribution.distribution.base_distribution.sigma.asnumpy()
nan_prob_hat = forecast_entry.distribution.nan_prob.asnumpy()




# plot the predicted distribution paremeters and the true parameters

ts_entry = tss[0]
plot_length = 80 
legend = ["observations", "median prediction"] 

fig, ax = plt.subplots(1, 1, figsize=(10, 7))
ax.set_ylabel('target', color="g") 
ax.set_xlabel('time') 

ax.tick_params(axis='y', labelcolor="g")

ax.plot(ts_entry.index[-plot_length:], true_mean[-plot_length:], color = "g", label="true target mean")

ax.plot(ts_entry.index[-plot_length:], true_mean[-plot_length:]+sigma, color = "g", linestyle =":")
ax.plot(ts_entry.index[-plot_length:], true_mean[-plot_length:]-sigma, color = "g", linestyle =":")

ax.plot(ts_entry.index[-custom_ds_metadata['prediction_length']:], mu_hat , color = "red", linestyle ="-",
        label="predicted target mean")
ax.plot(ts_entry.index[-custom_ds_metadata['prediction_length']:], mu_hat-sigma_hat, color = "red", linestyle =":")
ax.plot(ts_entry.index[-custom_ds_metadata['prediction_length']:], mu_hat+sigma_hat, color = "red", linestyle =":")

ax.legend()
ax2 = ax.twinx()

ax2.set_ylabel('probablity of missing values', color="b")  

start = pd.Timestamp("01-01-2019", freq='1H')
end = (pd.Timestamp("01-01-2019", freq='1H') + pd.to_timedelta(custom_ds_metadata["num_steps"]-1, unit='h'))

nan_prob = []
nan_gen = nan_prob_gen()
for i in range(custom_ds_metadata["num_steps"]):
    nan_prob.append(next(nan_gen))

ax2.plot(pd.date_range(start,end, freq="h"), nan_prob, color = "blue", linestyle ="-", label="true missing value probability")
ax2.plot(ts_entry.index[-custom_ds_metadata['prediction_length']:], nan_prob_hat, color = "darkred", linestyle ="-",
         label="predicted missing value probability")

ax2.tick_params(axis='y', labelcolor="blue")
ax2.set_ylim([0,1])
ax2.set_xlim([ts_entry.index[-plot_length], end])
fig.tight_layout() 
plt.title("Predicting a NanMixture: sin pattern with gaussian noise and changing missing value probability")
plt.legend()

plt.show()
```

src/gluonts/model/simple_feedforward/_network.py
```diff
@@ -158,15 +158,24 @@ def hybrid_forward(
             distr_args, loc=loc, scale=scale
         )
 
-        # (batch_size, prediction_length, target_dim)
-        loss = distr.loss(future_target)
+        if isinstance(distr, NanMixture):
+            # demasking the missing values for the future_targets
```
lostella, line 162: I agree with you that maybe these changes don’t belong in this PR. The results look great and everything else looks fine, so maybe you can remove the changes here and then we can merge!

PascalIversen: Thanks! I removed the changes.

PR #916, opened 2020-07-09
MQCNN: Support for past dynamic features and scaling

dcmaddix:
*Description of changes:*
- Added support for past dynamic features with `use_past_feat_dynamic_real` boolean.
- Added scaling for dynamic features for the decoder using `scaling_decoder_dynamic_feature` and set default target `scaling = False`
- Enabled dynamic features to be used in only the encoder, only the decoder and both using the booleans `enable_encoder_dynamic_feature` and `enable_decoder_dynamic_feature`
- Updated default parameter values: set `add_time_feature` default to `True`
- Updated default `dilation_seq = [1, 3, 9]`, since that minimized the variance in the experiments

### DIFF
--- a/src/gluonts/dataset/field_names.py
+++ b/src/gluonts/dataset/field_names.py
@@ -27,6 +27,11 @@ class FieldName:
     FEAT_STATIC_REAL = "feat_static_real"
     FEAT_DYNAMIC_CAT = "feat_dynamic_cat"
     FEAT_DYNAMIC_REAL = "feat_dynamic_real"
+    PAST_FEAT_DYNAMIC_REAL = "past_feat_dynamic_real"
+    FEAT_DYNAMIC_REAL_LEGACY = "dynamic_feat"
+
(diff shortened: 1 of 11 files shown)


khuyentran1401:
I wonder if we could use the same code logic here to scale the dynamic features in DeepAR?

dcmaddix:
Yes, we could also use the `MeanScaler` to scale the dynamic features in DeepAR.

khuyentran1401:
Thank you for your suggestion. I tried to add support for the dynamic feature scaling to the source code of DeepAR, but it doesn't seem to be straightforward. I wonder if there is any way I can scale the dynamic features without fixing the source code? I'm considering using [sklearn's `StandardScaler`](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html) to scale the dynamic features before feeding them to the model. What do you think about this approach?

PR #917, opened 2020-07-10
Addded trimming to encoded sagemaker parameters in shell package.

timforby:
SageMaker API docs note a hard limit on sagemaker parameter value lengths:
https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_HyperParameterTrainingJobSummary.html

We can encode the sagemaker parameters and trim them to a specific length.
Then on decode we reconstruct the parameters.

This just works by simply cutting every parameter value by the max length (256) and adding it as another value and prefixing the keys with `_idx_`, where idx is the index of the cut segment.

Example with max_len = 4:
```
    >>> trim_encoded_sagemaker_parameters({
    ...     'foo': '[1, 2, 3]',
    ...     'bar': 'hello'
    ... }, max_len = 4)
    {
        '_0_foo': '[1, ', '_1_foo': '2, 3', '_2_foo': ']',
        '_0_bar': 'hell', '_1_bar': 'o'
    }
```

*Description of changes:* 
- Just added an additional 4 functions that trim and un-trim hyperparameters to allow for parameters that exceed the sagemaker API limit.
- There is **another** significant change, line 50. I previously had to do the following `encode_sagemaker_parameter(serde.encode(parameters))`. Not sure why not use `serde.dump_json`?

This was a quick addition that I do on my end because I use shell.params to encode my hyperparameters for my gluonts container on sagemaker. I'm not entirely sure how you guys use the shell module so if this is the wrong place to do it then sorry for the bother!


src/gluonts/evaluation/_base.py
```diff
@@ -550,7 +550,7 @@ def abs_target_mean(target):
 
 class MultivariateEvaluator(Evaluator):
     """
-    
+
```
lostella, line 553: This was unintended I guess: 

```suggestion
```

src/gluonts/evaluation/_base.py
```diff
@@ -156,7 +156,7 @@ def __call__(
         ts_iterator = iter(ts_iterator)
         fcst_iterator = iter(fcst_iterator)
 
-        rows = []
+        rows: List = []
```
jaheba, line 159:
```suggestion
        rows = []
```

jaheba: Let's remove this as well.

jaheba:
Thanks for the PR!

I'm wondering whether this could become an issue:

> You can specify a maximum of 100 hyperparameters. Each hyperparameter is a key-value pair. Each key and value is limited to 256 characters, as specified by the Length Constraint.

We have a max of 100 parameters, which in some cases could not be enough?

src/gluonts/shell/sagemaker/params.py
```diff
@@ -79,3 +78,89 @@ def encode_sagemaker_parameters(decoded_params: dict) -> dict:
...
+            while f"_{idx}_{key[3:]}" in trimmed_params:
+                detrimmed_params[key[3:]] += trimmed_params[
+                    f"_{idx}_{key[3:]}"
+                ]
+                idx += 1
+        elif key[0] == "_":
+            continue
+        else:
+            detrimmed_params[key] = value
+    return detrimmed_params
```
jaheba, line 166:
```suggestion
def detrim_sagemaker_parameters(trimmed_params: dict) -> dict:
    """DE-trim parameters that have already been trimmed.
    Example:
    >>> detrim_sagemaker_parameters({
    ...     '_0_foo': '[1, ', '_1_foo': '2, 3', '_2_foo': ']',
    ...      '_0_bar': 'hell', '_1_bar': 'o'
    ... })
    {'foo': '[1, 2, 3]', 'bar': 'hello'}
    """
    detrimmed_params = trimmed_params.copy()

    trimmed_param_names = [
        param[3:] for param in detrimmed_params if param.startswith("_0_")
    ]

    for name in trimmed_param_names:
        value = ""
        for idx in count():
            part = detrimmed_params.pop(f"_{idx}_{name}", None)
            if part is None:
                break
            value += part

        detrimmed_params[name] = value

    return detrimmed_params
```

timforby:
> We have a max of 100 parameters, which in some cases could not be enough?

I hope that there will never be a model that requires anything close to that many parameters 🤣 . 
In that case I think some people [here](https://github.com/aws/sagemaker-python-sdk/issues/314) were suggesting using a manifest file.

src/gluonts/shell/sagemaker/params.py
timforby: Much cleaner thanks!

src/gluonts/evaluation/_base.py
timforby: Sorry, not sure what I did there. Friday nights 😕

src/gluonts/shell/sagemaker/params.py
```diff
@@ -79,3 +78,95 @@ def encode_sagemaker_parameters(decoded_params: dict) -> dict:
...
+
+
+def detrim_and_decode_sagemaker_parameters(trimmed_params: dict) -> dict:
+    """Decode a SageMaker parameters dictionary where all values are strings.
+
+    Example:
+
+    >>> detrim_sagemaker_parameters({
+    ...     '_0_foo': '[1, ', '_1_foo': '2, 3', '_2_foo': ']',
+    ...      '_0_bar': 'hell', '_1_bar': 'o'
```
jaheba, line 90: Can you change the doc-strings to be `key: value` for each line?

This ordering is confusing to me.

src/gluonts/evaluation/_base.py
```diff
@@ -550,7 +550,6 @@ def abs_target_mean(target):
 
 class MultivariateEvaluator(Evaluator):
     """
-
```
jaheba, line 553: Can you remove this change?

src/gluonts/shell/sagemaker/params.py
```diff
@@ -79,3 +78,95 @@ def encode_sagemaker_parameters(decoded_params: dict) -> dict:
...
+
+
+def encode_and_trim_sagemaker_parameters(
+    decoded_params: dict, max_len: int = 256
+) -> dict:
+    """Encode a SageMaker parameters dictionary where all values are strings then
+    trim them to account for Sagemaker character size limit.
+
+    Example:
+
```
jaheba, line 105:
```suggestion
```

src/gluonts/shell/sagemaker/params.py
```diff
@@ -79,3 +78,95 @@ def encode_sagemaker_parameters(decoded_params: dict) -> dict:
...
+    >>> encode_and_trim_sagemaker_parameters({
+    ...     "foo": [1, 2, 3],
+    ...     "bar": "hello"
+    ... }, max_len = 4)
+    {
+        '_0_foo': '[1, ', '_1_foo': '2, 3', '_2_foo': ']',
+        '_0_bar': 'hell', '_1_bar': 'o'
```
jaheba, line 112: Same as above. One line for each pair.

src/gluonts/shell/sagemaker/params.py
```diff
@@ -79,3 +78,95 @@ def encode_sagemaker_parameters(decoded_params: dict) -> dict:
...
+    }
+    """
+    trimmed_params = {}
+    for key, value in encoded_params.items():
+        if len(value) > max_len:
+            for idx, sub_len in enumerate(range(0, len(value), max_len)):
```
jaheba, line 137: There is `gluonts.dataset.util.batcher`, which would make this a bit more readable in my opinion:

```python
for idx, substr in enumerate(batcher(value, max_len)):
    trimmed_params[f"_{idx}_{key}"] = "".join(substr)
```

timforby: added! ideally it would have been nice to remove the  `if len(value) > max_len` using the batcher too, but then all values with appropriate length would have a key prefixed with `_0_`.

src/gluonts/shell/sagemaker/params.py
timforby: Agreed!

src/gluonts/shell/sagemaker/params.py
```diff
@@ -79,3 +79,105 @@ def encode_sagemaker_parameters(decoded_params: dict) -> dict:
...
+    >>> encode_and_trim_sagemaker_parameters({
+    ...     "foo": [1, 2, 3],
+    ...     "bar": "hello"
+    ... }, max_len = 4)
+    {
+        '_0_foo': '[1, ',
+        '_1_foo': '2, 3',
+        '_2_foo': ']',
+        '_0_bar': 'hell',
+        '_1_bar': 'o'
```
timforby, line 117: This is breaking doctest. Should we remove readability or I think we could add a doctest directive `NORMALIZE_WHITESPACE` ? Not familiar with doctest at all.

src/gluonts/shell/sagemaker/params.py
```diff
@@ -79,3 +79,105 @@ def encode_sagemaker_parameters(decoded_params: dict) -> dict:
...
+    >>> trim_encoded_sagemaker_parameters({
+    ...     'foo': '[1, 2, 3]',
+    ...     'bar': 'hello'
+    ... }, max_len = 4)
+    {
+        '_0_foo': '[1, ',
+        '_1_foo': '2, 3',
+        '_2_foo': ']',
+        '_0_bar': 'hell',
+        '_1_bar': 'o'
```
timforby, line 140: Same issue here as above.

src/gluonts/shell/sagemaker/params.py
```diff
@@ -79,3 +79,105 @@ def encode_sagemaker_parameters(decoded_params: dict) -> dict:
...
+    """
+    detrimmed_params = trimmed_params.copy()
+
+    trimmed_param_names = [
+        param[3:] for param in detrimmed_params if param.startswith("_0_")
+    ]
+
+    for name in trimmed_param_names:
+        value = ""
+        for idx in count():
```
jaheba, line 175: `count` needs to be imported from itertools.

src/gluonts/shell/sagemaker/params.py
jaheba: I see, didn't know of that.

I think readability is more important here. If `NORMALIZE_WHITESPACE` works, I'm all for it.

jaheba:
Thanks for the changes.

Would it work if we make `NORMALIZE_WHITESPACE` a global option for doctest? It might be that we just need to add it to `conftest.py`.

timforby:
> Thanks for the changes.
> 
> Would it work if we make `NORMALIZE_WHITESPACE` a global option for doctest? It might be that we just need to add it to `conftest.py`.

Should we do it in this PR?

jaheba:
Sure!

setup.cfg
```diff
@@ -12,4 +12,4 @@ description-file = public/README.md
 
 [tool:pytest]
 testpaths = test
-addopts = -v --doctest-modules --color=yes
+addopts = -v --doctest-modules --color=yes
```
jaheba, line 15: Can we revert this?

```suggestion
addopts = -v --doctest-modules --color=yes

```

PR #924, opened 2020-07-14
Simplify backtest_metrics

lostella:
*Description of changes:* The `backtest_metrics` utility function is currently trying to do too much stuff: it can take optional training data, and both a predictor or an estimator, and behave differently based on what it's provided with. As a result, `backtest_metrics` needs to take all sorts of parameters as input (for example regarding multi-processing options in case training is run), and depends on a bunch of model definitions (like GluonEstimator and so on) and data loading utilities (like InferenceDataLoader).

I think this is bad design first of all, but in addition to it this makes it hard (if possible at all) to remove its dependency on MXNet: this prevents us from potentially making it nicely usable with other frameworks.

This PR is simplifying it by letting it take a `predictor` and `test_data` only. Note that this only implies that `estimator.train` is invoked outside, which seems nicer to me.

**PROs**

* Much less code in `backtest_metrics`, at the (rather low) cost of having to invoke `estimator.train` outside

**CONs**

* The mechanism to reconstruct an experiment from the logs of `backtest_metrics` is lost. I think this is fine, and is just a symptom of the fact that this function is trying to do too much. In a sense, logging a whole experiment and being able to reconstruct it would really be the responsibility of some (potentially more general) "experiment runner" component.

### DIFF
--- a/src/gluonts/evaluation/backtest.py
+++ b/src/gluonts/evaluation/backtest.py
@@ -30,9 +30,9 @@ from gluonts.dataset.stat import (
     calculate_dataset_statistics,
 )
 from gluonts.evaluation import Evaluator
-from gluonts.model.estimator import Estimator, GluonEstimator
+from gluonts.model.estimator import Estimator
 from gluonts.model.forecast import Forecast
-from gluonts.model.predictor import GluonPredictor, Predictor
+from gluonts.model.predictor import Predictor
 from gluonts.support.util import maybe_len
 from gluonts.transform import TransformedDataset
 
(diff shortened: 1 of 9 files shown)


src/gluonts/evaluation/backtest.py
```diff
@@ -179,49 +159,9 @@ def backtest_metrics(
...
-            if use_symbol_block_predictor:
-                predictor = predictor.as_symbol_block_predictor(
-                    batch=next(iter(inference_data_loader))
-                )
-    else:
-        predictor = forecaster
-
     forecast_it, ts_it = make_evaluation_predictions(
         test_dataset, predictor=predictor, num_samples=num_samples
     )
```
lostella, line 167: Note that now the role of `backtest_metrics` is *essentially* just to call this here, plus run the evaluator through the results (two lines below). This makes you think: do we need `backtest_metrics` at all?

lostella:
Another relevant fact: `backtest_metric` is not even used in any of the tutorial notebooks, currently

PR #926, opened 2020-07-14
Implemented moving average

melopeo:
*Issue #, if available:*

*Description of changes:*

- Added moving average.
-- This is a deterministic task (no distribution of observations is assumed)
-- For the case where _prediction_length_ = 1, the output is the standard moving average of the previous _context_length_ observations
-- For the case where _prediction_length_ > 1, the input target is appended with previous moving averages, further computing moving averages until _prediction_length_ of them are computed.

### DIFF
--- a/src/gluonts/model/trivial/mean.py
+++ b/src/gluonts/model/trivial/mean.py
@@ -70,10 +70,9 @@ class MeanPredictor(RepresentablePredictor, FallbackPredictor):
         std = np.nanstd(target)
         normal = np.random.standard_normal(self.shape)
 
-        start_date = forecast_start(item)
         return SampleForecast(
             samples=std * normal + mean,
-            start_date=start_date,
+            start_date=forecast_start(item),
             freq=self.freq,
             item_id=item.get(FieldName.ITEM_ID),
         )
@@ -112,9 +111,10 @@ class MovingAveragePredictor(RepresentablePredictor):
     ) -> None:
         super().__init__(freq=freq, prediction_length=prediction_length)
 
-        # assert (
-        #     context_length >= 1
-        # ), "The value of `context_length` should be >= 1"
+        if context_length is not None:
+            assert (
+                context_length >= 1
+            ), "The value of `context_length` should be >= 1 or None"
 
         self.context_length = context_length
 
@@ -129,10 +129,9 @@ class MovingAveragePredictor(RepresentablePredictor):
 
             target.append(np.nanmean(window))
 
-        start_date = forecast_start(item)
         return SampleForecast(
             samples=np.array([target[-self.prediction_length :]]),
-            start_date=start_date,
+            start_date=forecast_start(item),
             freq=self.freq,
             item_id=item.get(FieldName.ITEM_ID),
         )
--- a/src/gluonts/model/trivial/mean.py
+++ b/src/gluonts/model/trivial/mean.py
@@ -78,6 +78,55 @@ class MeanPredictor(RepresentablePredictor, FallbackPredictor):
             item_id=item.get(FieldName.ITEM_ID),
         )
 
+class MovingAveragePredictor(RepresentablePredictor):
+    """
+    A :class:`Predictor` that predicts the moving average based on the
+    last `context_length` elements of the input target.
+
+    If `prediction_length` = 1, the output is the moving average
+    based on the last `context_length` elements of the input target.
+
+    If `prediction_length` > 1, the output is the moving average based on the
+    last `context_length` elements of the input target, where
+    previously calculated moving averages are appended at the end of the input target.
+    Hence, for `prediction_length` larger than `context_length`, there will be
+    cases where the moving average is calculated on top of previous moving averages.
+
+    Parameters
+    ----------
+    context_length
+        Length of the target context used to condition the predictions.
+    prediction_length
+        Length of the prediction horizon.
+    freq
+        Frequency of the predicted data.
+    """
+
+    @validated()
+    def __init__(
+            self,
+            prediction_length: int,
+            freq: str,
+            context_length: int,
+    ) -> None:
+        super().__init__(freq=freq, prediction_length=prediction_length)
+        self.context_length = context_length
+
+    def predict_item(self, item: DataEntry) -> SampleForecast:
+
+        target = item['target'].tolist()
+
+        for _ in range(self.prediction_length):
+            window = target[-self.context_length:]
+            target.append(np.nanmean(window))
+
+        start_date = frequency_add(item["start"], len(item["target"]))
+        return SampleForecast(
+            samples=np.array([target[-self.prediction_length:]]),
+            start_date=start_date,
+            freq=self.freq,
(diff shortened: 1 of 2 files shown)


benidis:
Could we add some basic tests for this? Something like sanity checks, shape checks, trivial input/output and maybe edge cases (NaNs, empty sequence, length 1 sequence, etc.)? Maybe under `test/model/trivial/test_moving_avg.py`?

src/gluonts/model/trivial/mean.py
```diff
@@ -79,6 +79,54 @@ def predict_item(self, item: DataEntry) -> SampleForecast:
...
+
+    def predict_item(self, item: DataEntry) -> SampleForecast:
+
+        target = item["target"].tolist()
+
+        for _ in range(self.prediction_length):
+            window = target[-self.context_length :]
+            target.append(np.nanmean(window))
+
+        start_date = frequency_add(item["start"], len(item["target"]))
```
jaheba, line 121: I think we have a `forecast_start` helper-method for this.

test/model/trivial/test_moving_average.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def set_data(target, freq):
+    """
+    Sets test data in the right format
+    """
+
+    start = "2020"
+    ds = ListDataset([{"target": target, "start": start}], freq=freq)
+    data = list(ds)
```
jaheba, line 29: There is no need to call `list` on `ds`, I think

test/model/trivial/test_moving_average.py
```diff
@@ -0,0 +1,125 @@
...
+            )
+            assert predictions.shape == (prediction_length,)
+
+
+def test_constant_sequence():
+    constant_value = 1
+    target_length = 3
+    target = [constant_value] * target_length  # [1, 1, 1]
+    freq = "D"
+    data = set_data(target, freq)
```
jaheba, line 93: I think all these lines make it actually harder to read the tests.

test/model/trivial/test_moving_average.py
```diff
@@ -0,0 +1,125 @@
...
+
+# First-party imports
+from gluonts.dataset.common import ListDataset
+from gluonts.model.trivial.mean import MovingAveragePredictor
+
+# Third-party imports
+import numpy as np
+
+
+def set_data(target, freq):
```
jaheba, line 22: I think it's better to just remove this function, since it doesn't do much. Let's just inline the creation of `ListDataset`s.

src/gluonts/model/trivial/mean.py
```diff
@@ -79,6 +79,58 @@ def predict_item(self, item: DataEntry) -> SampleForecast:
...
+        Length of the target context used to condition the predictions.
+    prediction_length
+        Length of the prediction horizon.
+    freq
+        Frequency of the predicted data.
+    """
+
+    @validated()
+    def __init__(
+        self, prediction_length: int, freq: str, context_length: int,
```
jaheba, line 108: I think we should make context_length optional.

test/model/trivial/test_moving_average.py
jaheba: What we probably want to do is something like this:

```python
def test_constant_sequence():
    assert get_predictions([5, 5, 5], prediction_length=2) == [5, 5]

def test_empty_sequence():
    np.testing.assert_equal(
        get_predictions([], prediction_length=2), [np.nan, np.nan]
    )

```


We could even just define test-cases and use `parametrize` to feed them into the test-function.

src/gluonts/model/trivial/mean.py
```diff
@@ -26,7 +26,7 @@
 from gluonts.model.forecast import Forecast, SampleForecast
 from gluonts.model.predictor import FallbackPredictor, RepresentablePredictor
 from gluonts.model.trivial.constant import ConstantPredictor
-from gluonts.support.pandas import frequency_add
+from gluonts.support.pandas import frequency_add, forecast_start
```
jaheba, line 29: Now we don't need `frequency_add` anymore :)

test/model/trivial/test_moving_average.py
```diff
@@ -0,0 +1,79 @@
...
+
+    predictions = get_predictions(
+        data,
+        prediction_length=prediction_length,
+        context_length=context_length,
+    )
+
+    np.testing.assert_equal(predictions, expected_output)
+
+    np.testing.assert_equal(predictions.shape, (prediction_length,))
```
jaheba, line 79: This test looks redundant, since assert_equal should already test for shape.

test/model/trivial/test_moving_average.py
```diff
@@ -0,0 +1,79 @@
...
+):
+    mp = MovingAveragePredictor(
+        prediction_length=prediction_length,
+        context_length=context_length,
+        freq=freq,
+    )
+
+    ds = ListDataset([{"target": target, "start": start}], freq=freq)
+    item = next(iter(ds))
+    predictions = mp.predict_item(item).samples[0]
```
jaheba, line 34:
```suggestion
    predictions = mp.predict_item(item).mean
```

I like this a bit better.

src/gluonts/model/trivial/mean.py
```diff
@@ -79,6 +79,58 @@ def predict_item(self, item: DataEntry) -> SampleForecast:
...
+        Length of the target context used to condition the predictions.
+    prediction_length
+        Length of the prediction horizon.
+    freq
+        Frequency of the predicted data.
+    """
+
+    @validated()
+    def __init__(
+        self, prediction_length: int, freq: str, context_length: int = 1,
```
jaheba, line 108: Can we make context_length optional?

src/gluonts/model/trivial/mean.py
```diff
@@ -79,6 +79,65 @@ def predict_item(self, item: DataEntry) -> SampleForecast:
...
+        self,
+        prediction_length: int,
+        freq: str,
+        context_length: Optional[int] = None,
+    ) -> None:
+        super().__init__(freq=freq, prediction_length=prediction_length)
+
+        # assert (
+        #     context_length >= 1
+        # ), "The value of `context_length` should be >= 1"
```
jaheba, line 117: We should still have a check when `context_length` is an integer.

src/gluonts/model/trivial/mean.py
```diff
@@ -79,6 +79,65 @@ def predict_item(self, item: DataEntry) -> SampleForecast:
...
+
+            target.append(np.nanmean(window))
+
+        start_date = forecast_start(item)
+        return SampleForecast(
+            samples=np.array([target[-self.prediction_length :]]),
+            start_date=start_date,
+            freq=self.freq,
+            item_id=item.get(FieldName.ITEM_ID),
+        )
```
jaheba, line 138:
```suggestion
        return SampleForecast(
            samples=np.array([target[-self.prediction_length :]]),
            start_date=forecast_start(item),
            freq=self.freq,
            item_id=item.get(FieldName.ITEM_ID),
        )
```

PR #927, opened 2020-07-14
Fix batch-transform case, where request is empty.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/shell/serve/app.py
+++ b/src/gluonts/shell/serve/app.py
@@ -128,7 +128,15 @@ def batch_inference_invocations(
 
     def invocations() -> Response:
         request_data = request.data.decode("utf8").strip()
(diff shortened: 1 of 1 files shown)


src/gluonts/shell/serve/app.py
```diff
@@ -128,7 +128,15 @@ def batch_inference_invocations(
 
     def invocations() -> Response:
         request_data = request.data.decode("utf8").strip()
-        instances = list(map(json.loads, request_data.split("\n")))
+
+        # request_data can be empty, but .split() will produce a non-empty
+        # list, which then means we try to decode an emoty string, which
```
lostella, line 133:
```suggestion
        # list, which then means we try to decode an empty string, which
```

PR #928, opened 2020-07-16
Moved mp-parameter handling to Trainer.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/model/estimator.py
+++ b/src/gluonts/model/estimator.py
@@ -25,7 +25,6 @@ from gluonts.core import fqname_for
 from gluonts.core.component import DType, from_hyperparameters, validated
 from gluonts.core.exception import GluonTSHyperparametersError
 from gluonts.dataset.common import Dataset
-from gluonts.dataset.loader import TrainDataLoader, ValidationDataLoader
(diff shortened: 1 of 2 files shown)


AaronSpieler:
So I really like the idea of moving `num_workers`, `num_prefetch` and `shuffle_buffer_length` into the Trainer parameters, because it makes sense. Additionally these arguments get serialised with the estimator. Saving the train and validation loader there is also really good.

Some things to think about though:
- This change makes the way these 3 parameters are set inconsistent between .`.train(...)` and `.predict(...)`. All of these parameters can currently be passed to either method in the same way, but this change would break that.

Some things that are missing:
- for example the wavenet estimator also has its own `train` method which would need to implement this change
```
 def train(
        self,
        training_data: Dataset,
        num_workers: Optional[int] = None,
        num_prefetch: Optional[int] = None,
        shuffle_buffer_length: Optional[int] = None,
        **kwargs,
    ) -> Predictor:
```

jaheba:
Thanks for pointing these out!

Let me mark this PR as a draft then :)

PR #934, opened 2020-07-17
Pass listify_dataset as a hyperparameter through the shell

dcmaddix:
- Pass `listify_dataset` as a boolean field in `Hyperparameters` dictionary through the shell to determine whether or not to cache the dataset
- If `listify_dataset = True`, we call `ListDataset` otherwise the dataset type is `FileDataset`
- Added corresponding test: `test_listify_dataset`

### DIFF
--- a/src/gluonts/shell/sagemaker/__init__.py
+++ b/src/gluonts/shell/sagemaker/__init__.py
@@ -17,7 +17,7 @@ import json
 import os
 from pathlib import Path
 from pydantic import BaseModel
-from typing import Any, Dict, Optional
+from typing import Dict, Optional
 
 # First party imports
 from gluonts.dataset.common import FileDataset, ListDataset, MetaData
@@ -45,15 +45,8 @@ class TrainEnv:
         self.hyperparameters = _load_hyperparameters(
             self.path.hyperparameters, self.channels
         )
-        self.train_auxillary_parameters = _load_train_auxillary_parameters(
-            self.path.train_auxillary_parameters
-        )
         self.current_host = _get_current_host(self.path.resourceconfig)
-        self.datasets = _load_datasets(
-            self.hyperparameters,
-            self.train_auxillary_parameters,
-            self.channels,
-        )
+        self.datasets = _load_datasets(self.hyperparameters, self.channels)
 
 
 class ServeEnv:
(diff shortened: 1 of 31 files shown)


src/gluonts/shell/sagemaker/__init__.py
```diff
@@ -131,8 +134,19 @@ def _load_datasets(
...
+    )
+    dataset_dict = {}
+    for name in DATASET_NAMES:
+        if name in channels:
+            file_dataset = FileDataset(channels[name], freq)
+            dataset_dict[name] = (
+                ListDataset(file_dataset, freq)
+                if listify_dataset
+                else file_dataset
+            )
```
AaronSpieler, line 150: I would add the try catch logic for catching OOM due to the conversion of FileDataset to ListDataset here.

src/gluonts/shell/sagemaker/__init__.py
```diff
@@ -131,8 +134,19 @@ def _load_datasets(
...
-    return {
-        name: FileDataset(channels[name], freq)
-        for name in DATASET_NAMES
-        if name in channels
-    }
+    listify_dataset = (
+        strtobool(hyperparameters["listify_dataset"])
+        if "listify_dataset" in hyperparameters.keys()
+        else False
+    )
```
jaheba, line 141:
```suggestion
    listify_dataset = strtobool(hyperparameters.get("listify_dataset", "no"))
```

test/shell/test_shell.py
```diff
@@ -112,6 +125,7 @@ def test_train_shell(train_env: TrainEnv, caplog) -> None:
             assert line.endswith("270.0")
 
 
+@pytest.mark.parametrize("listify_dataset", [True])
```
jaheba, line 128: I guess we would want to test `True` and `False`.

dcmaddix: I can add the `False` cases to all the tests as well, but I added a separate test `test_listify_dataset` that tests both `True` and `False`.

src/gluonts/shell/sagemaker/__init__.py
dcmaddix: Great, will update thanks!

PR #936, opened 2020-07-20
Distribution prediction for Seq2Seq

YuntianYeAWS:
*Issue #, if available:*

*Description of changes:*
add distribution forecast to seq2seq model


AaronSpieler:
@YuntianYeAWS could you please fix the merge conflicts?

YuntianYeAWS:
> @YuntianYeAWS could you please fix the merge conflicts?

Hi Aaron. So I added some new code to the original files. Could you tell me how can I resolve these conflicts?

YuntianYeAWS:
> @YuntianYeAWS could you please fix the merge conflicts?
there is no conflict shown in this file and I could not resolve the conflict.
src/gluonts/model/seq2seq/_mq_dnn_estimator.py
Could you have a look?

src/gluonts/model/seq2seq/_mq_dnn_estimator.py
```diff
@@ -111,6 +115,8 @@ def __init__(
         self,
         freq: str,
         prediction_length: int,
+        sampling: bool = True,
+        distr_output: DistributionOutput = GaussianOutput(),
```
AaronSpieler, line 119: We usually use `StudentTOutput` as default. See DeepAR.

src/gluonts/model/seq2seq/_forking_estimator.py
```diff
@@ -363,51 +380,92 @@ def create_transformation(self) -> Transformation:
...
-            scaling=self.scaling,
-            scaling_decoder_dynamic_feature=self.scaling_decoder_dynamic_feature,
-            dtype=self.dtype,
-        )
+        if self.sampling:
+            return ForkingSeq2SeqTrainingNetwork(
+                encoder=self.encoder,
+                enc2dec=FutureFeatIntegratorEnc2Dec(),
+                decoder=self.decoder,
+                distr_output=self.distr_output,
```
AaronSpieler, line 388: The sampling should not yield quantile output by default. The sampling option should yield samples from the specified distribution. We possible need another if else to cover the quantile output option.

src/gluonts/model/seq2seq/_forking_network.py
```diff
@@ -231,6 +242,62 @@ def hybrid_forward(
...
+            future_feat_dynamic,
+            feat_static_cat,
+            past_observed_values,
+        )
+        distr_args = self.distr_args_proj(dec_output)
+        distr = self.distr_output.distribution(
+            distr_args, scale=target_scale.expand_dims(axis=1)
+        )
+        loss = distr.loss(future_target)
+        return loss.mean(axis = 1)
```
AaronSpieler, line 300: The loss masking with the `future_observed_values` might be missing (see `ForkingSeq2SeqTrainingNetwork `):
```
  # mask the loss based on observed indicator
        weighted_loss = weighted_average(
            F=F, x=loss, weights=future_observed_values, axis=1
        )
```

AaronSpieler:
Some questions/observations:

- [ ] You are missing something like a `ForkingSeq2SeqSamplingNetwork`, however, the output of which should have the same shape as `ForkingSeq2SeqPredictionNetwork`. Btw, there might be a mistake in the documentation regarding the output shape, I believe it should be (batch_size, num_quantiles, prediction_length)). `ForkingSeq2SeqSamplingNetwork` would have output shape (batch_size, num_samples, prediction_length). And since `num_quantiles` and `num_samples` is just a matter of interpretation their respective output shapes would be equivalent.
- [ ] The `ForkingSeq2SeqDistributionNetwork` might be missing something like we have in the `ForkingSeq2SeqPredictionNetwork`:
```
      # We only care about the output of the decoder for the last time step
     fcst_output = F.slice_axis(dec_output, axis=1, begin=-1, end=None)
     fcst_output = F.squeeze(fcst_output, axis=1)
```
- [ ] you could have ` self.get_decoder_network_output` also return the target scale so that it can be utilised by self.distr_output.distribution` instead of scaling twice
- [ ] there seem to be a lot of redundant imports in `_forking_network.py`
- [ ] `ForkingSeq2SeqSamplingNetwork` and `ForkingSeq2SeqPredictionNetwork` can probably be merged
- [ ] `ForkingSeq2SeqTrainingNetwork` and `ForkingSeq2SeqDistributionTrainingNetwork `can probably be merged
- [ ] In general it might be useful to think of Quantile output just another type of configurable distribution output, i.e. it might not be warranted to have separate Networks for the support of other distributions. I believe we could fix this by just passing additional arguments to the `ForkingSeq2SeqNetworkBase` and checking in the Network whether a QuantileOutput or other Distribution is set and then executing the argument projection and loss calculation conditionally on that. 
    -  [ ] A seperate `ForkingSeq2SeqDistributionNetwork ` might be warranted though, since it has a different return type all together (although I'm not aware of anyone using the network equivalent for other models, since we are always interested in samples usually, so we could think about leaving that out for now, which would simplify things because we wouldn't have the `sampling` option anymore since it would always be True)

What do you think about these concerns? 
Have I misunderstood/overlooked something?

AaronSpieler:
@YuntianYeAWS Any updates?

YuntianYeAWS:
> @YuntianYeAWS Any updates?

Hi, Aaron. Sorry for the late reply. Last week, I have worked on another project. 

I have made few changes according to your previous comments. So I think we can directly draw samples from the distribution output, so maybe it is not necessary to write another SamplingPrediction. The sampling parameter can be changed to quantile?

Also could you explain how to merge two training network together? The choice of network seems only available in the estimator. I am not sure how to do within the network.

Thank you for your reviews and suggestions.

src/gluonts/model/seq2seq/_forking_estimator.py
```diff
@@ -133,6 +146,8 @@ def __init__(
         quantile_output: QuantileOutput,
         freq: str,
         prediction_length: int,
+        distr_output: DistributionOutput = GaussianOutput(),
+        sampling: bool = True,
```
lostella, line 150: We probably don't need the "sampling" option here: until now, this model produces forecasts in terms of quantiles, as described by the `quantile_output` argument; what you are doing here is to add the option for a distribution output instead, as specified by `distr_output`. No sampling is involved here in any of the two cases.

The simplest change to the signature here is as follow, I think:

```suggestion
        freq: str,
        prediction_length: int,
        quantile_output: Optional[QuantileOutput] = None,
        distr_output: Optional[DistributionOutput] = None,
```

followed by an assertion that makes sure that one and only one between `quantile_output` and `distr_output` is set (i.e., not None)

src/gluonts/model/seq2seq/_forking_estimator.py
```diff
@@ -363,51 +380,92 @@ def create_transformation(self) -> Transformation:
...
-            decoder=self.decoder,
-            quantile_output=self.quantile_output,
-            context_length=self.context_length,
-            cardinality=self.cardinality,
-            embedding_dimension=self.embedding_dimension,
-            scaling=self.scaling,
-            scaling_decoder_dynamic_feature=self.scaling_decoder_dynamic_feature,
-            dtype=self.dtype,
-        )
+        if self.sampling:
```
lostella, line 383: Following up from my other comment, this check here would be

```suggestion
        if self.distr_output is not None:
```

src/gluonts/model/seq2seq/_forking_estimator.py
```diff
@@ -363,51 +380,92 @@ def create_transformation(self) -> Transformation:
...
         self,
         transformation: Transformation,
         trained_network: ForkingSeq2SeqNetworkBase,
     ) -> Predictor:
         # this is specific to quantile output
-        quantile_strs = [
-            Quantile.from_float(quantile).name
-            for quantile in self.quantile_output.quantiles
-        ]
+        if self.sampling:
```
lostella, line 417: Similarly to above:

```suggestion
        if self.quantile_output is not None:
```

src/gluonts/model/seq2seq/_mq_dnn_estimator.py
```diff
@@ -111,6 +115,8 @@ def __init__(
         self,
         freq: str,
         prediction_length: int,
+        sampling: bool = True,
+        distr_output: DistributionOutput = StudentTOutput(),
```
lostella, line 119: Same as for the other class

```suggestion
        distr_output: Optional[DistributionOutput] = None,
```

And internally assert that `distr_output is None != quantiles is None` (that's the logical XOR)

src/gluonts/model/simple_feedforward/_estimator.py
```diff
@@ -32,12 +32,16 @@
     AddObservedValuesIndicator,
 )
 from gluonts.model.forecast_generator import DistributionForecastGenerator
+
```
lostella, line 35:
```suggestion
```

src/gluonts/model/simple_feedforward/_estimator.py
```diff
@@ -32,12 +32,16 @@
...
+from gluonts.support.util import get_hybrid_forward_input_names
+
```
lostella, line 37:
```suggestion
```

src/gluonts/model/simple_feedforward/_estimator.py
```diff
@@ -32,12 +32,16 @@
...
 from gluonts.transform.feature import (
     DummyValueImputation,
     MissingValueImputation,
 )
 
 
+
```
lostella, line 44: This blank line is not needed

src/gluonts/model/simple_feedforward/_estimator.py
```diff
@@ -151,12 +155,14 @@ def __init__(
         self.mean_scaling = mean_scaling
         self.num_parallel_samples = num_parallel_samples
         self.sampling = sampling
+
```
lostella, line 158: Could you get rid of all these introduced blank lines?

test/model/seq2seq/test_model.py
```diff
@@ -43,11 +44,12 @@ def Estimator(request):
 
 @pytest.mark.parametrize("quantiles", [[0.1, 0.5, 0.9], [0.5]])
 @pytest.mark.parametrize("hybridize", [True, False])
+@pytest.mark.parametrize("sampling",[True,False])
 def test_accuracy(
-    Estimator, accuracy_test, hyperparameters, hybridize, quantiles
+    Estimator, accuracy_test, hyperparameters, hybridize, quantiles, sampling
```
lostella, line 49: Following from the other comments, here we would need to test when `quantiles is not None` and `distr_output is None`, and the other way around. We can remove the `sampling` flag, since there's never really any sampling happening (see other comments)

src/gluonts/model/estimator.py
```diff
@@ -137,7 +137,7 @@ def __init__(
         super().__init__(lead_time=lead_time)
         self.trainer = trainer
         self.dtype = dtype
-
+
```
lostella, line 140:
```suggestion

```

src/gluonts/model/seq2seq/_forking_estimator.py
```diff
@@ -85,6 +94,10 @@ class ForkingSeq2SeqEstimator(GluonEstimator):
         seq2seq decoder
     quantile_output
         quantile output
+    distr_output
+        distribution output
+    sampling
+        True for sample prediction and False for distribution prediction
```
lostella, line 100: We probably don't need the `sampling` option (at least for now): it's sufficient to have `quantile_output` specified or `distr_output` specified (but not both).

src/gluonts/model/seq2seq/_forking_network.py
```diff
@@ -281,3 +351,53 @@ def hybrid_forward(
         predictions = self.quantile_proj(fcst_output).swapaxes(2, 1)
 
         return predictions
+
+    
+class ForkingSeq2SeqDistributionNetwork(ForkingSeq2SeqNetworkBase):
```
lostella, line 356: Maybe a better name for this (although longer) is

```suggestion
class ForkingSeq2SeqDistributionPredictionNetwork(ForkingSeq2SeqNetworkBase):
```

src/gluonts/model/seq2seq/_forking_estimator.py
YuntianYeAWS: Hi, Lorenzo. Thank you for your suggestions.
Could you explain the part of line 148 and 149?
Do you require the input specify what quantile/distribution output?
Would it be possible that both quantile/distribution are not None?

lostella: @YuntianYeAWS no, one and only one should be not None. You can test this with the following snippet, which is essentially the logical XOR

```python
assert distr_output is None != quantiles is None
```

YuntianYeAWS:
> @YuntianYeAWS Any updates?

Hi Aaron. As suggested by Lorenzo, the sampling should be substituted by quantile. If it is possible, could you have a look if there are other issues?

src/gluonts/model/simple_feedforward/_estimator.py
```diff
@@ -157,6 +157,7 @@ def __init__(
             else DummyValueImputation(self.distr_output.value_in_support)
         )
 
+
```
lostella, line 160:
```suggestion
```

src/gluonts/model/simple_feedforward/_estimator.py
```diff
@@ -200,6 +201,7 @@ def create_training_network(self) -> HybridBlock:
 
     # we now define how the prediction happens given that we are provided a
     # training network.
+
```
lostella, line 204:
```suggestion
```

src/gluonts/model/simple_feedforward/_network.py
```diff
@@ -93,9 +93,11 @@ def __init__(
             )
             self.scaler = MeanScaler() if mean_scaling else NOPScaler()
 
+
```
lostella, line 96:
```suggestion
```

src/gluonts/model/simple_feedforward/_network.py
```diff
@@ -93,9 +93,11 @@ def __init__(
...
     def get_distr_args(
         self, F, past_target: Tensor
     ) -> Tuple[Tensor, Tensor, Tensor]:
+
```
lostella, line 100:
```suggestion
```

src/gluonts/model/simple_feedforward/_network.py
```diff
@@ -107,11 +109,13 @@ def get_distr_args(
             Shape: (batch_size, context_length, target_dim).
         Returns
         -------
+
```
lostella, line 112:
```suggestion
```

src/gluonts/model/simple_feedforward/_network.py
```diff
@@ -107,11 +109,13 @@ def get_distr_args(
...
         Tuple[Tensor, Tensor Tensor]
             A tuple of three tensors ``distr_args``, ``loc``, and ``scale``: the 
             first contains the output distribution parameters, and the others are 
             location and scale of the distribution.
  
+
```
lostella, line 118:
```suggestion
```

src/gluonts/model/simple_feedforward/_network.py
```diff
@@ -145,9 +149,11 @@ def hybrid_forward(
         future_target
             Tensor with future observations.
             Shape: (batch_size, prediction_length, target_dim).
+
```
lostella, line 152:
```suggestion
```

src/gluonts/model/simple_feedforward/_network.py
```diff
@@ -145,9 +149,11 @@ def hybrid_forward(
...
         future_observed_values
             Tensor indicating which values in the target are observed, and
             which ones are imputed instead.
+
```
lostella, line 156:
```suggestion
```

src/gluonts/model/simple_feedforward/_network.py
```diff
@@ -207,7 +213,9 @@ def hybrid_forward(self, F, past_target: Tensor) -> Tensor:
 
 
 class SimpleFeedForwardDistributionNetwork(SimpleFeedForwardNetworkBase):
+
```
lostella, line 216:
```suggestion
```

src/gluonts/model/simple_feedforward/_network.py
```diff
@@ -207,7 +213,9 @@ def hybrid_forward(self, F, past_target: Tensor) -> Tensor:
...
     @validated()
+
```
lostella, line 218:
```suggestion
```

test/model/seq2seq/test_model.py
lostella: They cannot be both `None` at the same time, so I would do (note that `MQCNNEstimator` takes `quantiles` or `distr_output` as inputs):

```python
@pytest.mark.parametrize("quantiles, distr_output", [
    ([0.5, 0.95], None),
    (None, StudentTOutput())
])
```

src/gluonts/model/seq2seq/_forking_estimator.py
```diff
@@ -85,6 +94,10 @@ class ForkingSeq2SeqEstimator(GluonEstimator):
         seq2seq decoder
     quantile_output
         quantile output
+    distr_output
+        distribution output
+    quantile
```
lostella, line 99:
```suggestion
    quantile_output
```

src/gluonts/model/estimator.py
```diff
@@ -137,7 +137,6 @@ def __init__(
         super().__init__(lead_time=lead_time)
         self.trainer = trainer
         self.dtype = dtype
```
lostella, line 139:
```suggestion
        self.dtype = dtype

```

PR #942, opened 2020-07-21
Shell: Fix writing of `output/failure` file in case of error in provided hyper-parameters.

Schmedu:
Missing or incorrect mandatory Hyperparameters are leading to errors that are not captured inside the `failure` file inside the `output` folder. This PR fixes this problem by including the `env = TrainEnv(Path(data_path))` inside the `try` block and instantiate the `path` beforehand.

### DIFF
--- a/src/gluonts/shell/__main__.py
+++ b/src/gluonts/shell/__main__.py
@@ -155,8 +155,7 @@ def train_command(data_path: str, forecaster: Optional[str]) -> None:
     from gluonts.shell import train
 
     logger.info("Run 'train' command")
-    path = Path(data_path)
-    train_paths = TrainPaths(path)
+    train_paths = TrainPaths(Path(data_path))
 
     try:
         env = TrainEnv(train_paths)
--- a/src/gluonts/shell/__main__.py
+++ b/src/gluonts/shell/__main__.py
@@ -28,6 +28,7 @@ from gluonts.model.estimator import Estimator
 from gluonts.model.predictor import Predictor
 
 # Relative imports
+from gluonts.shell.sagemaker import TrainPaths
 from gluonts.shell.serve import Settings
 
 from .sagemaker import ServeEnv, TrainEnv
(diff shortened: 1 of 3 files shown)


src/gluonts/shell/__main__.py
```diff
@@ -154,9 +154,11 @@ def train_command(data_path: str, forecaster: Optional[str]) -> None:
     from gluonts.shell import train
 
     logger.info("Run 'train' command")
-    env = TrainEnv(Path(data_path))
+    data_path = Path(data_path)
+    failure_path = data_path / "output" / "failure"
```
jaheba, line 158: I think it's better to instantiate a `TrainPaths`-object here and pass it to the `TrainEnv`.

src/gluonts/shell/sagemaker/__init__.py
```diff
@@ -35,8 +35,8 @@ class DataConfig(BaseModel):
 
 
 class TrainEnv:
-    def __init__(self, path: Path = Path("/opt/ml")) -> None:
-        self.path = TrainPaths(path)
+    def __init__(self, path: TrainPaths = TrainPaths(Path("/opt/ml"))) -> None:
+        self.path = path
```
Schmedu, line 39: the default value might be redundant because we already have a default in the click.option
```
@click.option(
    "--data-path",
    type=click.Path(exists=True),
    envvar="SAGEMAKER_DATA_PATH",
    default="/opt/ml",
    help="The root path of all folders mounted by the SageMaker runtime.",
)
```

src/gluonts/shell/__main__.py
Schmedu: changed it, please review it

src/gluonts/shell/__main__.py
```diff
@@ -154,9 +155,11 @@ def train_command(data_path: str, forecaster: Optional[str]) -> None:
     from gluonts.shell import train
 
     logger.info("Run 'train' command")
-    env = TrainEnv(Path(data_path))
+    path = Path(data_path)
+    train_paths = TrainPaths(path)
```
jaheba, line 159:
```suggestion
    train_paths = TrainPaths(Path(data_path))
```

PR #944, opened 2020-07-23
Removing mp worker del method.

AaronSpieler:
*Issue #, if available:*
Fixes issue https://github.com/awslabs/gluon-ts/issues/837 

*Description of changes:*
Removes legacy `__del__` method from `_MultiWorkerIter` that was used to prevent memory leaks.

*TODO*
- [x]  make sure we don't get memory leaks

### DIFF
--- a/src/gluonts/dataset/parallelized_loader.py
+++ b/src/gluonts/dataset/parallelized_loader.py
@@ -528,17 +528,6 @@ class _MultiWorkerIter(object):
                 return
             yield next_batch
 
-    def __del__(self):
-        # Explicitly load the content from shared memory to delete it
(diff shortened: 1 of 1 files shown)


AaronSpieler:
There is no continuous memory leak with this change, as expected.
I checked this using the ` df -h` command on a linux instance looking at the ` /dev/shm` partition.

However, the memory that was used by the remaining prefetched batches is not freed once the training of a model is finished, which is a small price to pay compared to possible deadlocks, and will most likely correspond to few dozen of KB up to 1MB of data. I suspect the native MXNet mp data loader is also affected by this.

Also, this will be permanently fixed by the new version of the mp data loader.

Issue #945, opened 2020-07-23
Dockerfiles crashing when run 


arangatang:
## Description
Mxnet is not installed so the image crashes during training.

```
ModuleNotFoundError: No module named 'mxnet'
```

This happens for both `Dockerfile` and `Dockerfile.gpu` 

Installing Mxnet dependencies into the dockerfiles should solve this.

lostella:
@arangatang solved?

PR #946, opened 2020-07-23
Dockerfile fixes

arangatang:
*Issue #, if available:*
solves #945

*Description of changes:*
the mxnet dependencies are now installed.
also fixed indentation of the Dockerfile.gpu. 


Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM python:3.7
 
 ADD . /gluonts
 
+RUN pip install mxnet==1.6
```
jaheba, line 5:
```suggestion
RUN pip install -r /gluonts/requirements/requirements-mxnet.txt
```

This should ensure that we install the currently recommended mxnet version.

Dockerfile.gpu
```diff
@@ -14,6 +14,7 @@ RUN echo '#!/bin/bash\necho Building container with GPU support' > /usr/bin/nvid
 
 ADD . /gluonts
 
+RUN pip3.7 install mxnet-cu92
```
AaronSpieler, line 17: This also be `mxnet-cu92==1.6` or whatever version is specified in the requirements.

arangatang: in the requirements the version is defined as >=1.6.0 so ill use this instead

AaronSpieler: ok, but then why not `mxnet-cu92>=1.6`

AaronSpieler:
- I added support for R in the cpu container.
    - Tested this using the `ets` method and the new experimentation framework.
- The resulting image is about 30% larger and takes much longer to build.

I like this because the cpu image now intuitively supports all our Predictors.
We could think about having a separate cpu Dockerfile with R support though.

What do you think @jaheba, @arangatang

jaheba:
> We could think about having a separate cpu Dockerfile with R support though.

I think I prefer that approach.

Maybe we can generally extend the use of https://github.com/awslabs/gluon-ts/tree/master/examples/dockerfiles

AaronSpieler:
> 
> 
> > We could think about having a separate cpu Dockerfile with R support though.
> 
> I think I prefer that approach.
> 
> Maybe we can generally extend the use of https://github.com/awslabs/gluon-ts/tree/master/examples/dockerfiles

Great idea. We should really move it to that folder.
Did that Dockerfile.r work? @Schmedu I tried installing the packages that way and it when I tested the shell it didn't work.

Schmedu:
@AaronSpieler yes, it works. I just checked it.

jaheba:
How about we move the Dockerfiles to `examples/dockerfiles` but we link to it in the readme?

arangatang:
Grouping them together would make sense, i agree.

examples/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:3.7
+
+RUN apt-get update && apt-get install -y r-base
+RUN R -e "install.packages('forecast',dependencies=TRUE, repos='http://cran.rstudio.com/')"
+RUN R -e "install.packages('nnfor',dependencies=TRUE, repos='http://cran.rstudio.com/')"
```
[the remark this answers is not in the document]
arangatang: Yeah, just saw that now too. I think ill remove this and add a readme

examples/dockerfiles/README.md
```diff
@@ -0,0 +1,28 @@
+# Instructions for using the Dockerfiles
+
+To build these dockerfiles go to the Gluon-TS root directory and run:
```
jaheba, line 3: I think what also works is doing

    docker build ../.. -f <dockerfile>

in the current folder. Maybe we should list both ways.

examples/dockerfiles/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+```bash 
+docker build . -f examples/dockerfiles/<the dockerfile>
+```
+
+To run the built image run:
```
jaheba, line 9: "run ... run" is a bit confusing.

examples/dockerfiles/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+```bash
+docker run <image_id> <shell_params>
```
jaheba, line 12: Let's maybe leave this out, and just say that these images can be used on SageMaker?

examples/dockerfiles/README.md
```diff
@@ -0,0 +1,28 @@
...
+For more information about the shell and the available params, see the [shell documentation](https://github.com/awslabs/gluon-ts/tree/master/src/gluonts/shell) or run:
+
+```bash
+docker run <image_id> --help
+```
+
+## How to choose between the dockerfiles
+
+* Dockerfile
+    - This is for running Gluon-TS algorithms on a machine using only the CPU. This image is the most basic one and allows the use of most of the gluonts forecasters.
```
jaheba, line 24: I think we can abbreviate this by quite a bit, e.g.

```suggestion
    - Gluon-TS models using CPU.
```

arangatang: Oki

examples/dockerfiles/README.md
arangatang: Removed this

examples/dockerfiles/README.md
arangatang: Removed this

examples/dockerfiles/README.md
arangatang: Added this

examples/dockerfiles/README.md
```diff
@@ -0,0 +1,29 @@
+# Instructions for using the Dockerfiles
+
+To build these dockerfiles go to the Gluon-TS root directory and enter this command:
+
+```bash 
+docker build . -f examples/dockerfiles/<the dockerfile>
```
jaheba, line 6:
```suggestion
docker build . -f examples/dockerfiles/<dockerfile>
```

examples/dockerfiles/README.md
```diff
@@ -0,0 +1,29 @@
...
+```bash
+docker build ../.. -f <dockerfile>
+```
+
+The built images are compatible with sagemaker.
+For more information about the shell and the available params, see the [shell documentation](https://github.com/awslabs/gluon-ts/tree/master/src/gluonts/shell) or run:
+
+```bash
+docker run <image_id> --help
+```
```
jaheba, line 20: Let's maybe just remove this as well.

```suggestion
```

examples/dockerfiles/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## How to choose between the dockerfiles
+
+* Dockerfile
+    - Gluon-TS models using CPU.
+* Dockerfile.gpu
+    -  Gluon-TS models on a GPU accelerated machine. 
+* Dockerfile.r
+    - This is used when one wants to run the models in the [gluonts.model.r_forecast](https://gluon-ts.mxnet.io/api/gluonts/gluonts.model.r_forecast.html) package.
```
jaheba, line 29:
```suggestion
    - This provides dependencies for models defined in [gluonts.model.r_forecast](https://gluon-ts.mxnet.io/api/gluonts/gluonts.model.r_forecast.html).
```

PR #952, opened 2020-07-26
Pass multiprocessing parameters through the shell

dcmaddix:
### Summary
- Exposed the multiprocessing parameters `num_workers`, `num_prefetch`, `shuffle_buffer_length` as `hyperparameter` keys through the shell
- Updated `run_train` method to read these keys in from the hyper-parameters and set their values
- For `num_workers` we set to be the minimum of inputed value if set and if not input default 4 and the sort of the number of cpus
- Added logging to log these parameter values in`parallelized_loader.py`
- Added logging to whether the dataset was cached or not and its type, i.e. `ListDataset` and `FileDataset`, respectively

### Testing
- Add `MQCNNEstimator` as `forecaster_type` to `test_train_shell` to extend the test for trainable algorithms unlike the current `MeanPredictor`
- Passed the three multiprocessing parameters as hyper-parameters and used the smoke test with the `MQCNNEstimator`

### DIFF
--- a/src/gluonts/dataset/parallelized_loader.py
+++ b/src/gluonts/dataset/parallelized_loader.py
@@ -559,7 +559,7 @@ class ParallelDataLoader(object):
         but will consume more shared_memory. Using smaller number may forfeit the purpose of using
         multiple worker processes, try reduce `num_workers` in this case.
         By default it defaults to `num_workers * 2`.
-     shuffle_buffer_length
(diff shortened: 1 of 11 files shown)


dcmaddix:
There is a timeout test error of 30s for the additional test I added with the `MQCNNEstimator`.  It takes longer than that to run.  Is there a way to increase the test timeout>

PR #957, opened 2020-07-29
Using freq.freqstr, since freq is not hashable anymore.

jaheba:
*Issue #, if available:*

*Description of changes:*


jaheba:
We are using `pandas==1.0.x` for now, which doesn't require this change.

PR #961, opened 2020-07-31
ConvTrans Implementation

Gandor26:
*Description of changes:*
This PR adds implementation of ConvTrans model [1]. 

| Datasets    | Context Length | Prediction Length | epochs / batches per epoch | Methods     | wP10QL | wP50QL | wP90QL | Running Time (s) |
|-------------|----------------|-------------------|----------------------------|-------------|--------|--------|--------|------------------|
| Electricity | 168            | 24                | 10/1000                    | DeepAR      | 0.0292 | 0.0687 | 0.0369 | 900              |
|             |                |                   |                            | Transformer | 0.0317 | 0.0761 | 0.0474 | 500              |
|             |                |                   |                            | ConvTrans   | 0.0432 | 0.1507 | 0.0776 | 600              |
| Traffic     | 168            | 24                | 10/1000                    | DeepAR      | 0.0618 | 0.1585 | 0.1361 | 1140             |
|             |                |                   |                            | Transformer | 0.0803 | 0.1656 | 0.1062 | 600              |
|             |                |                   |                            | ConvTrans   | 0.0757 | 0.1637 | 0.1347 | 480              |
| Parts       | 24             | 12                | 5/2000                     | DeepAR      | 0.2221 | 1.0792 | 1.6332 | 360              |
|             |                |                   |                            | Transformer | 0.2068 | 1.002  | 1.7712 | 480              |
|             |                |                   |                            | ConvTrans   | 0.2572 | 1.0246 | 1.1014 | 450              |
| Wiki-10k    | 28             | 7                 | 5/2000                     | DeepAR      | 0.0744 | 0.2185 | 0.2147 | 1300             |
|             |                |                   |                            | Transformer | NaN    | NaN    | NaN    | 720              |
|             |                |                   |                            | ConvTrans   | 0.0759 | 0.2044 | 0.2061 | 600              |
| M4-Daily    | 28             | 7                 | 10/1000                    | DeepAR      | 0.0174 | 0.0274 | 0.0151 | 400              |
|             |                |                   |                            | Transformer | 0.0455 | 0.0649 | 0.039  | 240              |
|             |                |                   |                            | ConvTrans   | 0.0403 | 0.0178 | 0.0348 | 180              |

[1] [Enhancing the Locality and Breaking the Memory Bottleneck of Transformer on Time Series Forecasting
](https://arxiv.org/abs/1907.00235)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #962, opened 2020-07-31
Add TFT model

Gandor26:
*Issue #, if available:*

*Description of changes:*
Add TFT model [1], benchmark 

| Datasets    | Context Length | Prediction Length | epochs / batches per epoch | Methods     | wP10QL | wP50QL | wP90QL | Running Time (s) |
|-------------|----------------|-------------------|----------------------------|-------------|--------|--------|--------|------------------|
| Electricity | 168            | 24                | 10/1000                    | DeepAR      | 0.0292 | 0.0687 | 0.0369 | 900              |
|             |                |                   |                            | Transformer | 0.0317 | 0.0761 | 0.0474 | 500              |
|             |                |                   |                            | TFT         | 0.0677 | 0.1509 | 0.0974 | 1200             |
| Traffic     | 168            | 24                | 10/1000                    | DeepAR      | 0.0618 | 0.1585 | 0.1361 | 1140             |
|             |                |                   |                            | Transformer | 0.0803 | 0.1656 | 0.1062 | 600              |
|             |                |                   |                            | TFT         | 0.0646 | 0.1531 | 0.1077 | 1500             |
| Parts       | 24             | 12                | 5/2000                     | DeepAR      | 0.2221 | 1.0792 | 1.6332 | 360              |
|             |                |                   |                            | Transformer | 0.2068 | 1.002  | 1.7712 | 480              |
|             |                |                   |                            | TFT         | 0.2135 | 1.0312 | 0.9918 | 1060             |
| Wiki-10k    | 28             | 7                 | 5/2000                     | DeepAR      | 0.0744 | 0.2185 | 0.2147 | 1300             |
|             |                |                   |                            | Transformer | NaN    | NaN    | NaN    | 720              |
|             |                |                   |                            | TFT         | 0.0646 | 0.2042 | 0.1539 | 1150             |
| M4-Daily    | 28             | 7                 | 10/1000                    | DeepAR      | 0.0174 | 0.0274 | 0.0151 | 400              |
|             |                |                   |                            | Transformer | 0.0455 | 0.0649 | 0.039  | 240              |
|             |                |                   |                            | TFT         | 0.0105 | 0.0189 | 0.009  | 1200             |

[1] [Temporal Fusion Transformers for Interpretable Multi-horizon Time Series Forecasting
](https://arxiv.org/abs/1912.09363)


src/gluonts/model/tft/_estimator.py
```diff
@@ -0,0 +1,294 @@
...
+        self.num_instance_per_series = num_instance_per_series
+
+        self.time_features = time_features or time_features_from_frequency_str(
+            self.freq
+        )
+        self.static_cardinalities = static_cardinalities or {}
+        self.static_feature_dims = static_feature_dims or {}
+        self.dynamic_cardinalities = dynamic_cardinalities or {}
+        self.dynamic_feature_dims = dynamic_feature_dims or {}
+        self.past_dynamic_features = past_dynamic_features or []
```
jaheba, line 98: I think pydantic is able to handle this.

E.g.:

```python
time_features:List[TimeFeature] = [],
```

Normally you wouldn't do this in Python, but pydantic will return you a copy of the empty list, so this should work.

src/gluonts/model/tft/_network.py
```diff
@@ -0,0 +1,488 @@
...
+        super().__init__(**kwargs)
+
+        assert (
+            len(feature_dims) > 0
+        ), "Length of `cardinalities` list must be greater than zero"
+        assert len(feature_dims) == len(
+            embedding_dims
+        ), "Length of `embedding_dims` and `embedding_dims` should match"
+        assert all(
+            [c > 0 for c in feature_dims]
```
jaheba, line 82: The `[]` are not needed here.

src/gluonts/model/tft/_engine.py
```diff
@@ -0,0 +1,315 @@
...
+from gluonts.support.util import HybridContext
+
+logger = logging.getLogger("gluonts").getChild("trainer")
+MODEL_ARTIFACT_FILE_NAME = "model"
+STATE_ARTIFACT_FILE_NAME = "state"
+LOG_CACHE = set([])
+OutputTransform = Callable[[DataEntry, np.ndarray], np.ndarray]
+
+
+class Trainer(BaseTrainer):
```
jaheba, line 55: Why can't use use the `Trainer` defined in `gluonts.mx.trainer`? Or ask differently, what would you need to be able to use it?

src/gluonts/model/tft/_network.py
```diff
@@ -0,0 +1,488 @@
...
+        ), "Elements of `feature_dims` should be > 0"
+        assert all(
+            [d > 0 for d in embedding_dims]
+        ), "Elements of `embedding_dims` should be > 0"
+
+        self.feature_dims = feature_dims
+        self.dtype = dtype
+        self.__num_features = len(feature_dims)
```
jaheba, line 90: It's uncommon to use two leading `_` in Python, is this needed here?

src/gluonts/model/tft/_network.py
```diff
@@ -0,0 +1,488 @@
...
+        feature_dims: List[int],
+        embedding_dims: List[int],
+        dtype: DType = np.float32,
+        **kwargs,
+    ) -> None:
+        super().__init__(**kwargs)
+
+        assert (
+            len(feature_dims) > 0
+        ), "Length of `cardinalities` list must be greater than zero"
```
jaheba, line 77:
```suggestion
        assert (feature_dims), "Length of `cardinalities` list must be greater than zero"
```

src/gluonts/model/tft/_network.py
Gandor26: Well, I just followed the implementation in gluonts.mx.block.feature.FeatureEmbedder, in which it also has such a property with name mangling
https://github.com/Gandor26/gluon-ts/blob/san/src/gluonts/mx/block/feature.py#L65

src/gluonts/model/tft/_engine.py
dcmaddix: +1.  In the ConvModel PR, he had also overrode it and said that the base trainer do not accept default argument values in hybrid_forward and overloaded the trainer to give default None to unused feature types.  But I think we can use the `Trainer` directly using boolean or whether to use the features or not in the estimator or directly extend the `Trainer` class.

src/gluonts/model/tft/_estimator.py
Gandor26: fixed in abdc7114efd290505ea64ef943baebb656e7b5d0

src/gluonts/model/tft/_network.py
Gandor26: fixed in abdc7114efd290505ea64ef943baebb656e7b5d0

src/gluonts/model/tft/_network.py
Gandor26: fixed in abdc7114efd290505ea64ef943baebb656e7b5d0

src/gluonts/model/tft/_engine.py
```diff
@@ -0,0 +1,315 @@
...
+                                        )
+                                if name in data_entry:
+                                    inputs.append(data_entry[name])
+                                elif not (arg.default is inspect._empty):
+                                    inputs.append(arg.default)
+                                else:
+                                    raise RuntimeError(
+                                        f"The value of argument `{name}` of HybridBlock is not provided, "
+                                        f"and no default value is set."
+                                    )
```
lostella, line 149: Some observations:
* The code here seems like a refined version of [this](https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/mx/trainer/_base.py#L250)-plus-[this](https://github.com/awslabs/gluon-ts/blob/df1e50d462f97ac2bf8ee5865f165cce2c066d83/src/gluonts/support/util.py#L169-L176) allowing for defaults;
* This is duplicated in the forecast generator defined below;
* Both in the trainer and forecast generator now have an obsolete `input_names` constructor argument.

If you really want to push for this mechanism for TFT, my suggestion here is to propose this new mechanism in a separate PR where you showcase it with a minimal example model. This way you can focus on getting this into the default `Trainer` class and `ForecastGenerator` types **in a backward compatible way** (so that other models keep working fine with the default `Trainer` and `ForecastGenerator` classes).

Then you should then be able to proceed with the current PR, but avoiding a lot of code duplication.

lostella: This of course applies to #961 as well, so the code duplication savings double :-)

dcmaddix (approved):
Looks great, Xiaoyong! I think we can merge it in.  Thanks for updating to using the default Trainer.

Gandor26:
@dcmaddix  Thanks Danielle. Just let you know I ran one of the benchmark tests and got similar metrics, so it's safe to merge it.

lostella:
Thanks @Gandor26 for pushing this through! If I remember correctly you had a mechanism to allow for optional arguments to `hybrid_forward` to work with the training loop, which I thought was pretty cool: we may want to revisit that at some point

PR #971, opened 2020-08-04
Moved get_seasonality from evaluation to time_feature.

jaheba:
*Issue #, if available:*

*Description of changes:*


src/gluonts/time_feature/seasonality.py
```diff
@@ -0,0 +1,43 @@
...
+
+import logging
+
+import pandas as pd
+
+
+logger = logging.getLogger(__name__)
+
+
+DEFAULT_SEASONALITIES = {"H": 24, "D": 1, "W": 1, "M": 12, "B": 5}
```
AaronSpieler, line 22: I believe we should change "D": 7, however, that is for another PR.

jaheba: It's strange, agreed. Also `W` could be `52`.

Are these seasonalities defined by M4, or something?

Issue #973, opened 2020-08-04
Type check fails on the master branch


shchur:
## Description
There seems to currently be a bug on the master branch. The classes `Forecaster` and `GluonTSForecasterNotFoundError` in [`src/gluonts/shell/__main__.py`](https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/shell/__main__.py) are undefined. Also, I suspect that the argument `forecaster` in [`src/gluonts/shell/train.py`](https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/shell/train.py#L102)  should be of type `GluonEstimator`, not `Estimator`.

The above two problems break the type check (`python setup.py type_check`), which results in failed tests for submitted PRs.

## To Reproduce
```bash
git pull origin master
python setup.py type_check
```

## Error message or code output
```
src/gluonts/shell/train.py:122: error: Unexpected keyword argument "num_workers" for "train" of "Estimator"
src/gluonts/shell/train.py:122: error: Unexpected keyword argument "num_prefetch" for "train" of "Estimator"
src/gluonts/shell/train.py:122: error: Unexpected keyword argument "shuffle_buffer_length" for "train" of "Estimator"
src/gluonts/shell/__main__.py:77: error: Name 'Forecaster' is not defined
src/gluonts/shell/__main__.py:128: error: Name 'GluonTSForecasterNotFoundError' is not defined
Found 5 errors in 2 files (checked 133 source files)
```

## Environment
- Operating system: MacOS
- Python version: 3.7.7
- GluonTS version: 0.5.1.dev69+gd859164
- MXNet version: 1.6.0

PR #974, opened 2020-08-04
Fix type check. (#973)

shchur:
*Issue #, if available:* #973 

*Description of changes:* 
- Add missing import statements.
- Change the type of `forecaster` to `GluonEstimator`.

### DIFF
--- a/src/gluonts/shell/train.py
+++ b/src/gluonts/shell/train.py
@@ -26,7 +26,7 @@ from gluonts.core.component import check_gpu_support
 from gluonts.core.serde import dump_code
 from gluonts.dataset.common import Dataset
 from gluonts.evaluation import Evaluator, backtest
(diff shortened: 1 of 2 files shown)


src/gluonts/shell/train.py
```diff
@@ -99,7 +99,7 @@ def run_train_and_test(
 
 
 def run_train(
-    forecaster: Estimator,
+    forecaster: GluonEstimator,
```
jaheba, line 102: There are different `Estimator`s, not just `GluonEstimator`s.

The mypy error is actually helpful. We only want to pass the num_workers argument when it's a `GluonEstimator`.

shchur:
@jaheba sorry, I thought that I needed to rebase my branch again before merging, and this undid your approval.

PR #976, opened 2020-08-04
DeepTPP: RNN-based temporal point processes model

shchur:
*Issue #, if available:* #294 

*Description of changes:* 
### DeepTPP
- `DeepTPP`, our RNN-based TPP model, works as follows. After each event `i`, we feed the inter-arrival time `tau_i` (i.e. time since the previous event) and the mark `m_i` (one of `K` possible classes) into the RNN. The RNN output `h_i` corresponds to the "history embedding". Then, the history embeddings `h_i` is used to parametrize the distribution `p(tau_{i+1} | h_i)` over the next inter-arrival time, and the distribution `p(m_{i+1} | h_i)` over the next mark.
- The structure of the `DeepTPP` model is rather similar to `DeepAR`. The difference lies in the fact that `DeepTPP` works on (padded) variable-length sequences and only predicts the time & the class of the next event. It should be easy to extend `DeepTPP` to handle arbitrary marks (continuous, ordinal).

### TPP distributions
- Two new classes `TPPDistributionOutput` & `TPPDistribution` are used to model the output distribution over the inter-arrival times. These classes are similar to `DistributionOutput` and `Distribution` but have two important differences:
  - Methods `log_intensity` and `log_survival` make computing the log-likelihood for variable-length sequences possible in vectorized form.
  - The method `sample` takes an optional argument ` lower_bound`, which allows to sample from the conditional distribution `p(x | x > lower_bound`). This is necessary when generating forecasts conditioned on the previous interval.
- While the new classes are similar to the old distributions (`gluonts.mx.distribution`), we decided to implement the TPP distributions in a separate module (`gluonts.model.tpp.distribution`), since (1) the new distributions require multiple new methods, and we didn't want to inadvertently break the old code, (2) new distributions are not used anywhere outside of the TPP module, (3) the new 'Weibull` and `Loglogistic` distributions were anyway absent from the old module.
- We could potentially integrate the `TPPDistribution` into `Distribution` in a future pull request.


test/model/tpp/test_tpp_distribution_sampling.py
```diff
@@ -0,0 +1,65 @@
...
+    distr = serialize_fn(distr)
+    samples = distr.sample()
+    assert samples.shape == (2,)
+    num_samples = 1_000_000
+    samples = distr.sample(num_samples)
+    assert samples.shape == (num_samples, 2)
+
+    np_samples = samples.asnumpy()
+    # avoid accuracy issues with float32 when calculating std
+    # see https://github.com/numpy/numpy/issues/8869
```
canerturkmen, line 54: 😱

test/model/tpp/test_tpp_distribution_inference.py
```diff
@@ -0,0 +1,187 @@
...
+TOL = 0.3
+START_TOL_MULTIPLE = 1
+
+np.random.seed(1)
+mx.random.seed(1)
+
+
+def inv_softplus(y: NPArrayLike) -> np.ndarray:
+    # y = log(1 + exp(x))  ==>  x = log(exp(y) - 1)
+    return np.log(np.exp(y) - 1)
```
canerturkmen, line 48: not sure if it's relevant here or elsewhere but mxnet (and numpy) support `expm1` and `log1p`

test/model/tpp/test_tpp_distribution_inference.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def maximum_likelihood_estimate_sgd(
```
canerturkmen, line 51: does this have any major difference from its original? if not can we get away with importing / slightly changing the original?

src/gluonts/model/tpp/deeptpp/_network.py
```diff
@@ -0,0 +1,397 @@
...
+
+        first_step = True
+        while F.sum(arrival_times < self.prediction_interval_length) > 0:
+            # Sample the next inter-arrival time
+            time_distr_args = self.time_distr_args_proj(history_emb)
+            time_distr = self.time_distr_output.distribution(
+                time_distr_args, scale=self.output_scale,
+            )
+            if first_step:
+                # Time from teh last event until the next event
```
shchur, line 328: typo

src/gluonts/model/tpp/distribution/base.py
```diff
@@ -0,0 +1,219 @@
...
+            for t in self.transforms:
+                x = t.f(x)
+        return x.astype(dtype)
+
+
+class TPPDistributionOutput(DistributionOutput):
+    r"""
+    Class to construct a distribution given the output of a network.
+
+    Two differences compared to the base class DistributionOutput:
```
lostella, line 185: I like these docstrings a lot 👍

src/gluonts/model/tpp/predictor.py
```diff
@@ -56,10 +56,12 @@ def __call__(
             outputs, valid_length = (
                 x.asnumpy() for x in prediction_net(*inputs)
             )
+            # outputs (num_parallel_samples, batch_size, max_seq_len, 2)
+            # valid_length (num_parallel_samples, batch_size)
```
lostella, line 60: General comment (not strictly concerning the changes in this PR): so the model network is assumed to only produce samples, i.e., there's currently no support for cases were you would want to produce parametric distributions only, is that right?

test/model/tpp/test_deeptpp.py
```diff
@@ -0,0 +1,135 @@
...
+        num_marks=num_marks,
+        interval_length=2,
+        time_distr_output=WeibullOutput(),
+    )
+    model.initialize()
+
+    loglike_pred = model(nd.stack(ia_times, marks, axis=-1), valid_length)
+
+    assert loglike_pred.shape == (ia_times.shape[0],)
+    assert _allclose(loglike, loglike_pred)
```
lostella, line 80: I see this assertion potentially creating problems in the future (say, if the rng method changes in mxnet): is this testing anything useful?

test/model/tpp/test_tpp_distribution_shape.py
```diff
@@ -0,0 +1,83 @@
...
+            WeibullOutput(),
+            nd.random.gamma(shape=(3, 4, 5, 6)),
+            # [None, nd.array([5.0])],
+            [None, nd.ones(shape=(1,)), nd.ones(shape=(3, 4, 5))],
+            (3, 4, 5),
+            (),
+        ),
+    ],
+)
+def test_distribution_output_shapes(
```
lostella, line 49: I understand that the organization of test modules tries to mimic the one of source files, but if there is so much overlap with test/distribution/test_distribution_shapes.py, then maybe we can add these test instances to that script?

src/gluonts/model/tpp/distribution/base.py
shchur: Thanks!

test/model/tpp/test_tpp_distribution_shape.py
shchur: I have moved all the tests related to `TPPDistribution` into the existing tests in `test/distribution`

src/gluonts/model/tpp/predictor.py
shchur: Yes, that is currently the case. As far as I understand, it would only be possible to return the distribution over the inter-arrival time & mark over the next time step (i.e. `p(tau_i | history)` and `p(m_i | history)`). The distribution over `tau_{i+1}` and `m_{i+1}` will depend on the values of `tau_i` and `m_i` that we sampled at the time step `i`.

test/model/tpp/test_deeptpp.py
shchur: I can imagine that some changes may happen inside `DeepTPPTrainingNetwork` or `Weibull`, and this test should hopefully catch if any mistakes are introduced in the log-likelihood computation. Still, I agree with you that this is not a very likely scenario, and the cons may outweigh the pros. @canerturkmen do you have any ideas about what other tests we could include to check the correctness of the `DeepTPPTrainingNetwork`? Or is it fine if we only check the shape of the loss?

lostella: that's ok, we can leave this in now, and reconsider it whenever (if) it starts giving issues

vafl:
Looks good!

One question, if I read the code correctly, the mark and the time interval are drawn independently given the hidden state. Is this unnecessarily limiting? I imagine it could be that certain marks arrive (on average) later or earlier than others.
You could first sample the mark and then include that in the computation of the arrival time. 
Sorry if this is a naive question or if I'm missing something.

canerturkmen:
> Looks good!
> 
> One question, if I read the code correctly, the mark and the time interval are drawn independently given the hidden state. Is this unnecessarily limiting? I imagine it could be that certain marks arrive (on average) later or earlier than others.
> You could first sample the mark and then include that in the computation of the arrival time.
> Sorry if this is a naive question or if I'm missing something.

Hi @vafl, yes this is the model's design in all of the referenced papers, the next mark and time are conditionally independent given the RNN output. I guess one hopes that the correlation is represented by the network, which in many cases is a reasonable enough assumption.

vafl:
> > Looks good!
> > One question, if I read the code correctly, the mark and the time interval are drawn independently given the hidden state. Is this unnecessarily limiting? I imagine it could be that certain marks arrive (on average) later or earlier than others.
> > You could first sample the mark and then include that in the computation of the arrival time.
> > Sorry if this is a naive question or if I'm missing something.
> 
> Hi @vafl, yes this is the model's design in all of the referenced papers, the next mark and time are conditionally independent given the RNN output. I guess one hopes that the correlation is represented by the network, which in many cases is a reasonable enough assumption.

Ok, got it. Thanks. So this is the reference implementation from the literature.

I do think it is somewhat limiting. The hidden state can only capture the correlation if the model assigns most mass to one of the marks and basically already takes the decision before sampling. If multiple marks are likely, the arrival time cannot depend on the mark. But of course it depends on the data, whether it is important to model this or not...

shchur:
> > > Looks good!
> > > One question, if I read the code correctly, the mark and the time interval are drawn independently given the hidden state. Is this unnecessarily limiting? I imagine it could be that certain marks arrive (on average) later or earlier than others.
> > > You could first sample the mark and then include that in the computation of the arrival time.
> > > Sorry if this is a naive question or if I'm missing something.
> > 
> > 
> > Hi @vafl, yes this is the model's design in all of the referenced papers, the next mark and time are conditionally independent given the RNN output. I guess one hopes that the correlation is represented by the network, which in many cases is a reasonable enough assumption.
> 
> Ok, got it. Thanks. So this is the reference implementation from the literature.
> 
> I do think it is somewhat limiting. The hidden state can only capture the correlation if the model assigns most mass to one of the marks and basically already takes the decision before sampling. If multiple marks are likely, the arrival time cannot depend on the mark. But of course it depends on the data, whether it is important to model this or not...

This model is definitely not the most expressive one, but since the hidden state gets updated after each event (i.e. time & mark), the model should still have a decent degree of flexibility.

vafl:
Yeah, would be interesting to check this. Maybe histograms of arrival times per mark would give an indication of whether its important to model it or not.

PR #980, opened 2020-08-09
Bug Fixes and Warnings for Rotbaum

pmohamma:
*Issue #, if available:*

*Description of changes:*

This PR first fixes an issue in Rotbaum where the context_length is an optional parameter when creating a TreePredictor or TreeEstimator but will break if you do not explicitly pass in the context_length. 

It also prints a warning when using Rotbaum that if you are to use the Evaluator class to evaluate the TreePredictor, that you should not use multiprocessing in the Evaluator class.

Finally, this PR should fix a typo in an assertion when preprocessing features for Rotbaum

### DIFF
--- a/src/gluonts/model/rotbaum/_predictor.py
+++ b/src/gluonts/model/rotbaum/_predictor.py
@@ -16,6 +16,7 @@ from enum import Enum
 from typing import Iterator, List, Optional
 from pathlib import Path
 import json
+import warnings
 
 
 # Third-party imports
@@ -134,8 +135,11 @@ class TreePredictor(GluonPredictor):
         ], "method has to be either 'QRX', 'QuantileRegression', or 'QRF'"
         self.method = method
         self.lead_time = lead_time
+        self.context_length = (
+            context_length if context_length is not None else prediction_length
+        )
         self.preprocess_object = PreprocessOnlyLagFeatures(
-            context_length,
+            self.context_length,
             forecast_horizon=prediction_length,
             stratify_targets=False,
             n_ignore_last=n_ignore_last,
@@ -155,11 +159,9 @@ class TreePredictor(GluonPredictor):
             or use_feat_dynamic_real
             or use_feat_static_cat
             or use_feat_static_real
-        ), "The value of `prediction_length` should be > 0 or there should be features for model training and prediction"
+        ), 'The value of `prediction_length` should be > 0 or there should be features for model training and ' \
+           'prediction '
 
-        self.context_length = (
-            context_length if context_length is not None else prediction_length
-        )
         self.model_params = model_params if model_params else {}
         self.prediction_length = prediction_length
         self.freq = freq
@@ -167,6 +169,11 @@ class TreePredictor(GluonPredictor):
         self.clump_size = clump_size
         self.quantiles = quantiles
         self.model_list = None
+
+        logging.info(
+            'If using the Evaluator class with a TreePredictor, set num_workers=0. The TreePredictor and Evaluator ' \
+            'classes both use multiprocessing which is slow in tandem.'
+        )
 
     def __call__(self, training_data):
         assert training_data
--- a/src/gluonts/model/rotbaum/_preprocess.py
+++ b/src/gluonts/model/rotbaum/_preprocess.py
@@ -356,8 +356,8 @@ class PreprocessOnlyLagFeatures(PreprocessGeneric):
             [(np.floor(elem) == elem) for elem in feat_static_cat]
         )  # asserts that the categorical features are encoded
 
-        assert (not feat_static_cat) or all(
-            [(np.floor(elem) == elem) for elem in feat_dynamic_cat]
+        assert (not feat_dynamic_cat) or all(
+            [(np.floor(elem) == elem) for ent in feat_dynamic_cat for elem in ent]
         )  # asserts that the categorical features are encoded
 
         feat_dynamics = feat_dynamic_real + feat_dynamic_cat
--- a/src/gluonts/model/rotbaum/_preprocess.py
+++ b/src/gluonts/model/rotbaum/_preprocess.py
@@ -357,7 +357,7 @@ class PreprocessOnlyLagFeatures(PreprocessGeneric):
         )  # asserts that the categorical features are encoded
 
         assert (not feat_dynamic_cat) or all(
-            [(np.floor(elem) == elem) for ent in feat_dynamic_cat for elem in ent]
+            [(np.floor(elem) == elem) for elem in feat_dynamic_cat]
         )  # asserts that the categorical features are encoded
 
         feat_dynamics = feat_dynamic_real + feat_dynamic_cat


src/gluonts/model/rotbaum/_predictor.py
```diff
@@ -16,6 +16,7 @@
 from typing import Iterator, List, Optional
 from pathlib import Path
 import json
+import warnings
```
jaheba, line 19: Is this used?

src/gluonts/model/rotbaum/_predictor.py
```diff
@@ -168,6 +172,11 @@ def __init__(
         self.quantiles = quantiles
         self.model_list = None
 
+        logging.info(
```
jaheba, line 175: We started to create new loggers for each module:

```python
logger = logging.getLogger(__main__)
```

You can search the code to see how it is generally used.

The advantage is that this prints some more context, i.e. the module the message came from.

dcmaddix: +1

src/gluonts/model/rotbaum/_predictor.py
```diff
@@ -134,8 +135,11 @@ def __init__(
         ], "method has to be either 'QRX', 'QuantileRegression', or 'QRF'"
         self.method = method
         self.lead_time = lead_time
+        self.context_length = (
+            context_length if context_length is not None else prediction_length
```
dcmaddix, line 139: Thanks for fixing this!

src/gluonts/model/rotbaum/_predictor.py
```diff
@@ -41,6 +41,8 @@
 from ._preprocess import PreprocessOnlyLagFeatures
 from ._model import QRX, QuantileReg, QRF
 
+logger = logging.getLogger("TreePredictor")
```
jaheba, line 44:
```suggestion
logger = logging.getLogger(__name__)
```

src/gluonts/model/rotbaum/_predictor.py
```diff
@@ -168,6 +173,11 @@ def __init__(
         self.quantiles = quantiles
         self.model_list = None
 
+        logger.info(
+            "If using the Evaluator class with a TreePredictor, set num_workers=0. The TreePredictor and Evaluator "
```
jaheba, line 177: Can you shorten the lines, so that they ideally don't exceed 79 characters each?

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -260,6 +260,8 @@ def __init__(
         use_feat_static_cat=False,
         use_feat_dynamic_real=False,
         use_feat_dynamic_cat=False,
+        static_cardinality: Optional[List] = None,
+        one_hot_encode: bool = True,  # should improve performance but will slow down model
```
jaheba, line 264: Can we use "accuracy" instead of "performance"? I think that term is less ambiguous.

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -270,10 +272,27 @@ def __init__(
...
+        # assert (
+        #         use_feat_dynamic_cat or not dynamic_cardinality
+        # ), "You should set `static_cardinality` if and only if `use_feat_static_cat=True`"
+        assert static_cardinality is None or all(
+            c > 0 for c in static_cardinality
+        ), "Elements of `static_cardinality` should be > 0"
+        # assert dynamic_cardinality is None or all(
+        #     c > 0 for c in dynamic_cardinality
+        # ), "Elements of `dynamic_cardinality` should be > 0"
+
```
jaheba, line 288: What about these?

src/gluonts/transform/field.py
```diff
@@ -63,7 +63,8 @@ def __init__(self, field_names: List[str]) -> None:
 
     def transform(self, data: DataEntry) -> DataEntry:
         for k in self.field_names:
```
jaheba, line 65: Should this be part of the PR?

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -351,13 +390,21 @@ def make_features(self, time_series: Dict, starting_index: int) -> List:
...
             else []
         )
+        # if self.use_feat_dynamic_cat:
+        #     feat_dynamic_cat = (
+        #         self.create_cat_feats(time_series["feat_dynamic_cat"], is_static=False)
+        #         if self.one_hot_encode
+        #         else [elem for ent in time_series["feat_dynamic_cat"] for elem in ent]
+        #     )
+        # else:
+        #     feat_dynamic_cat = []
```
jaheba, line 400: Let's not commit commented code.

src/gluonts/model/rotbaum/_predictor.py
```diff
@@ -122,6 +124,8 @@ def __init__(
         use_feat_static_cat: bool = False,
         use_feat_dynamic_real: bool = False,
         use_feat_dynamic_cat: bool = False,
+        static_cardinality: Optional[List] = None,
```
jaheba, line 127: I think we can use an empty list the default. This should simplify the code a bit.

Also, shouldn't this be `static_cardinalities`, since each element is a cardinality?

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -306,6 +320,15 @@ def _pre_transform(cls, time_series_window) -> Tuple:
             },
         )
 
+    def create_cat_feats(self, feat_list: List) -> List:
```
jaheba, line 323: Can you explain what this does? Looking at the code, it's not intuitive to me.

src/gluonts/model/rotbaum/_predictor.py
pmohamma: Yup, absolutely right about the cardinalities. I'll make that change. However, for other models, we use a boolean for the use_feat_* hyperparameters. I do agree that it might simplify the code a bit but did not want to do something that might confuse a user if they are using multiple Gluon models.

src/gluonts/model/rotbaum/_preprocess.py
pmohamma: It is just a helper function to make one hot encodings. I can change it's name to that and make it a private function now that you mention it since it should not be called from outside the class.

src/gluonts/model/rotbaum/_predictor.py
pmohamma: However, if you still think it would be alright to use a list here, I would be happy to make that change

jaheba: > Yup, absolutely right about the cardinalities. I'll make that change. However, for other models, we use a boolean for the use_feat_* hyperparameters. I do agree that it might simplify the code a bit but did not want to do something that might confuse a user if they are using multiple Gluon models.

I think it's fine to do things differently than other algorithms.

The reason why I prefer an empty list is that both `None` and `[]` encode the same thing: the absence of features. However, if we use `None`, we need to explicitly check for that, whilst the empty list just works in some cases.

src/gluonts/model/rotbaum/_preprocess.py
jaheba: Looking at the signature, it's not entirely clear that it does one hot-encoding. I would either state this in a doc-string or the method name.

Personally, I don't care too much about making it "private" or not.

jaheba: I think the code can be improved if you use `zip`:

```python
def one_hot(feat_list, static_cardinality):
    xs = [0] * sum(static_cardinality)

    current = 0

    for feat, cardinality in zip(feat_list, static_cardinality):
        idx = cardinality - feat - 1
        xs[current + idx] = 1
        current += cardinality

    return xs


def one_hot2(feat_list, static_cardinality):
    res = []

    for feat, cardinality in zip(feat_list, static_cardinality):
        idx = cardinality - feat - 1
        one_hot = [0] * cardinality
        one_hot[idx] = 1
        res.extend(one_hot)

    return res
```

I think it would make sense to add comments to how this method works.

Also, you version doesn't assign things in order.

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -306,6 +333,39 @@ def _pre_transform(cls, time_series_window) -> Tuple:
...
+
+        for feat, cardinality in zip(feat_list, self.cardinality):
+            assert (
+                np.floor(feat) == feat
+            )  # asserts that the categorical features are encoded
+            idx = int(cardinality - feat - 1)
+            xs[current + idx] = 1
+            current += cardinality
+
+        return xs
```
jaheba, line 349: Maybe something like this is even more understandable:

```python

from itertools import chain, starmap


def encode_one_hot(feat: int, cardinality: int) -> List[int]:
    result = [0] * cardinality
    result[feat] = 1
    return result


def encode_one_hot_all(feat_list):
    encoded = starmap(encode_one_hot, zip(feat_list, self.cardinality))
    return list(chain.from_iterable(encoded))
````

File: src/gluonts/model/rotbaum/_predictor.py [hunk 13]
```diff
@@ -119,9 +121,10 @@ def __init__(
         clump_size: int = 100,  # Used only for "QRX" method.
         context_length: Optional[int] = None,
         use_feat_static_real: bool = False,
-        use_feat_static_cat: bool = False,
         use_feat_dynamic_real: bool = False,
         use_feat_dynamic_cat: bool = False,
+        cardinality="auto",
```
jaheba, line 126: I think we want to still use type annotations here.

You can use `Union[List[int], ...]` to allow more than one type.

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -270,10 +287,21 @@ def __init__(
             num_samples=num_samples,
             **kwargs
         )
+
+        if one_hot_encode:
+            assert (isinstance(cardinality, List) and cardinality) or (
+                isinstance(cardinality, str) and cardinality != "ignore"
+            ), "You should set `one_hot_encode=True` if and only if cardinality is not empty or ignored"
+
+        assert cardinality in set(["ignore", "auto"]) or all(
```
jaheba, line 296: This looks like a check we want to do earlier.

I think you could just create an Enum for this. You can look at the pydantic documentation for an example.

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -306,6 +334,38 @@ def _pre_transform(cls, time_series_window) -> Tuple:
             },
         )
 
+    def encode_one_hot(self, feat: int, cardinality: int) -> List[int]:
+        assert (
+            np.floor(feat) == feat
+        )  # asserts that the categorical features are label encoded
```
jaheba, line 340: Can't we check for this earlier?

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -306,6 +334,38 @@ def _pre_transform(cls, time_series_window) -> Tuple:
...
+        result[feat] = 1
+        return result
+
+    def encode_one_hot_all(self, feat_list):
+        encoded = starmap(
+            self.encode_one_hot, zip(feat_list, self.cardinality)
+        )
+        return list(chain.from_iterable(encoded))
+
+    def create_cardinalities(self, time_series):
```
jaheba, line 351: This looks noisy, can we clean this up?

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -25,6 +25,11 @@
 from gluonts.core.component import validated
 
 
+class CardinalityEnum(str, Enum):
```
jaheba, line 28:
```suggestion
class CardinalityLabel(str, Enum):
```

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -25,6 +25,11 @@
...
+    auto = "auto"
+    ignore = "ignore"
+
+
```
jaheba, line 32:
```suggestion
Cardinality = Union[List[int], CardinalityLabel]
```

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -275,10 +279,21 @@ def __init__(
         use_feat_static_real=False,
         use_feat_dynamic_real=False,
         use_feat_dynamic_cat=False,
-        cardinality="auto",
-        one_hot_encode: bool = True,  # should improve accuracy but will slow down model
+        cardinality: Union[List[int], str] = "auto",
```
jaheba, line 282:
```suggestion
        cardinality: Cardinality = CarindalityLabel.auto,
```

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -275,10 +279,21 @@ def __init__(
...
+        one_hot_encode: bool = True,  # Should improve accuracy but will slow down model
         **kwargs
     ):
+
+        if one_hot_encode:
+            assert cardinality != "ignore" or (
+                isinstance(cardinality, List)
+                and all(c > 0 for c in cardinality)
+            ), "You should set `one_hot_encode=True` if and only if cardinality is a valid list or not ignored"
```
jaheba: This doesn't look too right to me.

I understand `one_hot_encode` as an option which can be enabled, but it doesn't require that there are actually categorical features.

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -275,10 +279,21 @@ def __init__(
...
+
+        assert (
+            isinstance(cardinality, List) or cardinality in CardinalityEnum
+        ), "cardinality should be a list or set to 'auto' or 'ignore'"
+
```
jaheba:
```suggestion
```

This should be taken care of through the `validated`-decorator.

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -306,6 +341,41 @@ def _pre_transform(cls, time_series_window) -> Tuple:
...
+        encoded_chain = chain.from_iterable(encoded)
+        return list(encoded_chain)
+
+    def create_cardinalities(self, time_series):
+        if "feat_static_cat" not in time_series[0]:
+            return []
+        mat = [None] * len(time_series)
+        for i, elem in enumerate(time_series):
+            mat[i] = np.array(elem["feat_static_cat"])
+        mat = np.array(mat)
```
jaheba, line 367:
```suggestion
        mat = np.array([elem["feat_static_cat"] for elem in time_series], dtype=int)
```

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -306,6 +341,41 @@ def _pre_transform(cls, time_series_window) -> Tuple:
...
+        assert (mat[i].size == mat[0].size for i in range(len(mat)))
```
jaheba, line 368: There should be no need to check for this, since proper arrays have this as a requirement.

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -306,6 +341,41 @@ def _pre_transform(cls, time_series_window) -> Tuple:
...
+        ret = (
+            [
+                int(len(pd.unique(pd.Series(mat[:, i]))))
+                for i in range(len(mat[0]))
+            ]
+            if mat[0].size != 0
+            else []
+        )
+        return ret
```
jaheba, line 377: Wouldn't this do the same?

```suggestion
        return [len(set(xs)) for xs in mat.T]
```

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -336,11 +406,15 @@ def make_features(self, time_series: Dict, starting_index: int) -> List:
...
-        feat_static_cat = (
-            list(time_series["feat_static_cat"])
-            if self.use_feat_static_cat
-            else []
-        )
+        if self.cardinality:
+            feat_static_cat = (
+                self.encode_one_hot_all(time_series["feat_static_cat"])
+                if self.one_hot_encode
+                else list(time_series["feat_static_cat"])
```
jaheba, line 413: Shouldn't this be already a list?

src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -352,13 +426,18 @@ def make_features(self, time_series: Dict, starting_index: int) -> List:
...
+        del np_feat_static_cat
 
-        assert (not feat_static_cat) or all(
-            [(np.floor(elem) == elem) for elem in feat_dynamic_cat]
-        )  # asserts that the categorical features are encoded
+        np_feat_dynamic_cat = np.array(feat_dynamic_cat)
+        assert (not feat_dynamic_cat) or all(
+            np.floor(np_feat_dynamic_cat) == np_feat_dynamic_cat
+        )
+        del np_feat_dynamic_cat
```
jaheba, line 440: What's the purpose of this line?

pmohamma: This just ensures that all the dynamic categorical variables are label encoded and are not strings or any other form that a categorical variable can take

src/gluonts/model/rotbaum/_preprocess.py
pmohamma: I was under the impression that it could potentially be a numpy array as well

PR #982, opened 2020-08-10
Fix DeterministicOutput, add tests

lostella:
*Description of changes:* Fixes the `DeterministicOutput` class and add tests for it. This class is useful e.g. to implement inflated distributions (through a `MixtureDistributionOutput`).

cc @PascalIversen 

### DIFF
--- a/src/gluonts/mx/distribution/deterministic.py
+++ b/src/gluonts/mx/distribution/deterministic.py
@@ -18,10 +18,11 @@ from typing import Dict, List, Optional, Tuple
 
 # Third-party imports
 import numpy as np
+import mxnet as mx
 
 # First-party imports
 from gluonts.model.common import Tensor
-from gluonts.core.component import validated
+from gluonts.core.component import DType, validated
 from gluonts.support.util import erf, erfinv
 
 # Relative imports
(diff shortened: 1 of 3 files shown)


src/gluonts/mx/distribution/distribution_output.py
```diff
@@ -93,7 +93,7 @@ def dtype(self):
     def dtype(self, dtype: DType):
         self._dtype = dtype
 
-    def get_args_proj(self, prefix: Optional[str] = None) -> ArgProj:
+    def get_args_proj(self, prefix: Optional[str] = None) -> gluon.HybridBlock:
```
lostella, line 96: This annotation seems to me more appropriate

src/gluonts/mx/distribution/deterministic.py
```diff
@@ -88,7 +89,7 @@ def cdf(self, x):
         return is_greater_equal_or_both_nan
 
     def sample(
-        self, num_samples: Optional[int] = None, dtype=np.int32
```
lostella, line 91: @PascalIversen was there any reason for this to be `np.int32`?

PascalIversen: No. Good catch, thanks!

lostella:
> because the deterministic distribution returns the log probability and not the log density (which would be infinite).

right, there's a bit of an abuse of notation in all distribution types

PR #989, opened 2020-08-17
try to change the ts data into the tabular data

arianacai1997:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/support/autogluont_adapter.py
+++ b/src/gluonts/support/autogluont_adapter.py
@@ -1,0 +1,96 @@
+from autogluon import TabularPrediction as task
+from gluonts.dataset.repository.datasets import get_dataset
+from gluonts.dataset.util import to_pandas
+from pandas.tseries.holiday import USFederalHolidayCalendar as calendar
+import pandas as pd
+import matplotlib.pyplot as plt
+
+
+class Gluonts2Auto:
+    def __init__(self, dataset_name):
+        self.dataset = self.get_ds(dataset_name)
+
+    @staticmethod
+    def get_ds(dataset_name):
+        return get_dataset(dataset_name, regenerate=False)
+
+    def conversion(self):
+        train_series, test_series = to_pandas(next(iter(self.dataset.train))), to_pandas(
+            next(iter(self.dataset.test)))
+        train_series = self.sample_train(train_series)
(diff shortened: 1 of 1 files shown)


src/gluonts/support/autogluont_adapter.py
```diff
@@ -0,0 +1,96 @@
...
+        convert_type = {x: 'category' for x in df.columns.values[:-2]}
+        df = df.astype(convert_type)
+        return df
+
+if __name__ == '__main__':
+    g = Gluonts2Auto('electricity')
+    # train_data1, train_data2, test_data = g.conversion()
+    train_data, test_data = g.conversion()
+    label_column = 'target'
+    predictor = task.fit(train_data=train_data, label=label_column, problem_type='regression', output_directory='AutogluonModels/ag-a')
```
Innixma, line 76: You are not separating the train/val data by time, so it will be randomly selected. This is a problem for time-series data.

Additionally, you should try something like hyperparameters={'LR': {}} as a test, since linear regression can extrapolate, but most of AG's models cannot.

Finally, if you suspect the error is still unreasonable, try fitting a linear regression model or some kind of classical model outside of AutoGluon and comparing.

Innixma: Also, to confirm that the data is being processed by AutoGluon's feature generator correctly, please also call predictor.transform_features() after fitting to see the input data that the ML models receive.

arianacai1997: Hi Nick. Thanks for the comment! 
"You are not separating the train/val data by time, so it will be randomly selected. This is a problem for time-series data."
I think I got the data by time since they derive from dataset.train, dataset.test in the Gluonts model. Are you looking at the function "split_train", which is the one that planned for incremental training but never used?

Innixma: Inside of task.fit, train_data is split into train and validation for model training with early stopping. You need to instead pass train_data=train_data, tuning_data=validation_data in task.fit by first splitting train_data into train_data and validation_data where validation_data is later in time.

arianacai1997: Tons of thanks!

PR #995, opened 2020-08-20
Glide

jaheba:
*Issue #, if available:*

*Description of changes:*


src/gluonts/nursery/glide/pipeline.py
```diff
@@ -0,0 +1,105 @@
...
+        self.fn = fn
+        self.partitions = partitions
+
+    def __iter__(self):
+        yield from concat(
+            map(self.fn, partition) for partition in self.partitions
+        )
+
+
+class ParPipelineIterator:
```
lostella, line 48: I don’t get the name “pipeline” (neither for this nor for the module): an object of this class is what you get when you call `iter` on a `ParMap` object. Shouldn’t it be `ParMapIterator`?

src/gluonts/nursery/glide/pipeline.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class ParMap:
+    def __init__(self, fn, emitter, batch_size=10):
+        self.fn = fn
+        self.emitter = emitter
+        self.batch_size = batch_size
+
+    def __iter__(self):
+        queue = mp.Queue()
```
lostella, line 91: Wouldn’t it make sense to limit this queue? Say for example the emitter is infinite

lostella (commented):
This seems really cool and simple. One thing I’m wondering is how this plays out with “stateful” transformations, like an InstantSplitter which uses a stateful sampler: https://github.com/awslabs/gluon-ts/blob/978e82f8c6a07baee8ab6ea0262a0ec0e8b8cfb0/src/gluonts/transform/sampler.py#L85

(How is this PR breaking so many checks?)

lostella:
> One thing I’m wondering is how this plays out with “stateful” transformations, like an InstantSplitter which uses a stateful sampler

I guess whatever object state will be local to each process and depend on what sequence of task each process gets assigned

src/gluonts/nursery/glide/pipeline.py
jaheba: Yes, I used to call `ParMap` `Pipeline` but forgot to also rename this :(

src/gluonts/nursery/glide/pipeline.py
jaheba: Yes!

jaheba:
> > One thing I’m wondering is how this plays out with “stateful” transformations, like an InstantSplitter which uses a stateful sampler
> 
> I guess whatever object state will be local to each process and depend on what sequence of task each process gets assigned

Yes. I assume that wouldn't be different from the current approach. Essentially, we get `N` different instances and each one has its own state.

lostella (approved):

PR #1008, opened 2020-08-26
Enable distribution output in seq2seq

lostella:
*Issue #, if available:* Replaces #936, adds the option of configuring seq2seq models with a distribution output instead of specific quantiles.

Joint work with @YuntianYeAWS 

### DIFF
--- a/src/gluonts/model/seq2seq/_forking_estimator.py
+++ b/src/gluonts/model/seq2seq/_forking_estimator.py
@@ -51,7 +51,6 @@ from gluonts.transform import (
 from ._forking_network import (
     ForkingSeq2SeqNetworkBase,
     ForkingSeq2SeqTrainingNetwork,
-    ForkingSeq2SeqDistributionTrainingNetwork,
     ForkingSeq2SeqPredictionNetwork,
     ForkingSeq2SeqDistributionPredictionNetwork,
 )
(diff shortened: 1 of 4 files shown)


src/gluonts/model/seq2seq/_forking_estimator.py
```diff
@@ -369,51 +378,90 @@ def create_transformation(self) -> Transformation:
...
+            return RepresentableBlockPredictor(
+                input_transform=transformation,
+                prediction_net=prediction_network,
+                batch_size=self.trainer.batch_size,
+                freq=self.freq,
+                prediction_length=self.prediction_length,
+                ctx=self.trainer.ctx,
+                forecast_generator=DistributionForecastGenerator(
+                    self.distr_output
+                ),
```
AaronSpieler, line 466: We could factor generator out as a variable, and set it in the appropriate if else cast, and then just pass the variable to the forecast_generator. This way we have less code duplication, since both predictors are wrapped in a RepresentableBlockPredictor.

src/gluonts/model/seq2seq/_forking_network.py
```diff
@@ -232,6 +242,62 @@ def hybrid_forward(
...
+            feat_static_cat,
+            past_observed_values,
+        )
+        distr_args = self.distr_args_proj(dec_output)
+        distr = self.distr_output.distribution(distr_args, scale=scale)
+        loss = distr.loss(future_target)
+        weighted_loss = weighted_average(
+            F=F, x=loss, weights=future_observed_values, axis=1
+        )
+        return weighted_loss
```
AaronSpieler, line 298: Maybe this could be handled in an `if distribution ... else ...` since both return `weighted_loss `.

src/gluonts/model/seq2seq/_forking_estimator.py
lostella: addressed

src/gluonts/model/seq2seq/_forking_network.py
lostella: addressed

AaronSpieler (approved):
Neat!

PR #1012, opened 2020-08-28
Adding mean absolute quantile loss to avoid the case of dividing by 0…

dcmaddix:
… as a possible HPO metric

### Summary
- Added mean_absolute_QuantileLoss to avoid dividing by 0 case in HPO when the target array is all zeros
- Added flag to wQuantileLoss calculation that if totals["abs_target_sum"] == 0 to just divide by 1 and mean_wQuantileLoss will only output the numerator in this case and can still be used by HPO

### Testing
- Added test where the metrics are zero to test the devision
- Ran test_train_shell in shell/test_shell.py and outputs the mean metric in the logs:
 [2020-08-27 14:06:33] [INFO] gluonts.shell.train #test_score (local,    mean_absolute_QuantileLoss): 20.17107310626242

[2020-08-27 14:06:33] [INFO] gluonts.shell.train #test_score (local,    mean_wQuantileLoss): 0.0747076781713423

[2020-08-27 14:06:33] [INFO] gluonts.shell.train #test_score (local, MAE_Coverage): 0.2925925925925925`


src/gluonts/evaluation/_base.py
```diff
@@ -91,6 +89,7 @@ def __init__(
         self.seasonality = seasonality
         self.alpha = alpha
         self.calculate_owa = calculate_owa
+        self.zero_tol = 1e-8
```
AaronSpieler, line 92: Why exactly this number?
Isn't there in numpy something like `epsilon`: https://stackoverflow.com/questions/19141432/python-numpy-machine-epsilon

PR #1031, opened 2020-09-10
Generalised Pareto distribution

elenaehrlich:
Description of changes: Addition of the Generalised Pareto distribution


test/distribution/test_distribution_inference.py
```diff
@@ -26,7 +26,7 @@
 
 # First-party imports
 from gluonts.model.common import NPArrayLike
-from gluonts.mx.distribution.box_cox_transform import (
```
lostella, line 29: Looks like this shouldn’t be removed? Same later on in this module

src/gluonts/mx/distribution/genpareto.py
```diff
@@ -0,0 +1,176 @@
...
+    Generalised Pareto distribution.
+
+    Parameters
+    ----------
+    xi
+        Tensor containing the xi shape parameters, of shape `(*batch_shape, *event_shape)`.
+    beta
+        Tensor containing the beta scale parameters, of shape `(*batch_shape, *event_shape)`.
+    loc
+        Tensor containing the (for-now positive) location parameters, of shape `(*batch_shape, *event_shape)`.
```
lostella, line 48: Why for now positive? It seems like the corresponding `GenParetoOutput` class doesn’t constrain `loc`

src/gluonts/mx/distribution/genpareto.py
```diff
@@ -0,0 +1,176 @@
...
+        Returns
+        -------
+        Tuple[Tensor, Tensor, Tensor]:
+            Three squeezed tensors, of shape `(*batch_shape)`: both have entries mapped to the
+            positive orthant.
+        """
+        epsilon = np.finfo(cls._dtype).eps
+        xi = softplus(F, xi) + epsilon
+        beta = softplus(F, beta) + epsilon
+        # loc = softplus(F, loc) + epsilon
```
lostella, line 167: I guess this can just be removed?

lostella:
Looks good to me, thanks @elenaehrlich. One thing that is missing is the “joint test” for `log_prob` and `sample` that is there for the other distributions: you can see some examples here https://github.com/awslabs/gluon-ts/blob/master/test/distribution/test_distribution_inference.py

Essentially this generates a sample using the distribution, and then fits the parameter with SGD minimizing negative loglikelihood. Would something like that work for GenPareto as well?

src/gluonts/mx/distribution/genpareto.py
elenaehrlich: To your point about including a test for MLE via SGD, I dropped the inclusion of a location-shift parameter in order to permit this test's inclusion. MLE with a location-shift parameter requires Grimshaw's algorithm [https://amstat.tandfonline.com/doi/abs/10.1080/00401706.1993.10485040], and can perhaps be a future commit.

src/gluonts/mx/distribution/genpareto.py
elenaehrlich: yes, apologies

lostella (approved):
Looks good, thanks!

PR #1043, opened 2020-09-18
Move loss method from Distribution to DistributionOutput

vafl:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/model/canonical/_network.py
+++ b/src/gluonts/model/canonical/_network.py
@@ -111,8 +111,7 @@ class CanonicalTrainingNetwork(CanonicalNetworkBase):
             self.proj_distr_args(outputs), scale=target_scale
         )
 
-        loss = distr.loss(past_target)
-
+        loss = self.distr_output.loss(distr, past_target)
         return loss
 
 
--- a/src/gluonts/model/deepar/_network.py
+++ b/src/gluonts/model/deepar/_network.py
@@ -455,7 +455,7 @@ class DeepARTrainingNetwork(DeepARNetwork):
         )
 
         # (batch_size, seq_len)
-        loss = distr.loss(target)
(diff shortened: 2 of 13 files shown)


lostella:
This is potentially very breaking for anyone who might have custom models on the side. Maybe a smoother transition that keeps the `loss` method in the distribution classes, but deprecates it, would be better?

We could use a decorator to deprecate the `loss` method and print out a warning.

kashif:
I find this refactoring useful but one request I would have is to also move the `sample()` method to the distribution output. This would make it easy to implement certain distributions e.g. Implicit Quantile Networks which at the moment would have an unnatural coupling to the distribution and distribution output.

lostella:
> I find this refactoring useful but one request I would have is to also move the `sample()` method to the distribution output. This would make it easy to implement certain distributions e.g. Implicit Quantile Networks (Dabney et al) which at the moment would have an unnatural coupling to the distribution and distribution output.

I haven't look into the details of Implicit Quantile Networks, but doesn't it seem unnatural to move the `sample` method out of distribution classes, in all other cases?

kashif:
@lostella so i was thinking the models call the `loss()` and `sample()` of the corresponding distribution through the distribution output class (which by default will call the initialised distribution's  `self.distr.sample()` for sampling and `log_prob()` or `crps()` for the loss)... its probably a breaking change but hopefully alias help that.

The issue with implicit quantile network is that the distribution is implicitly defined via the output of its `arg_proj` network which means both during sampling and calculating the loss one needs to do a hack to call it again... whereas for this case it makes more sense to just have an implicit quantile network output layer (and no distribution at all...)

Does that make sense?

kashif:
@lostella actually kindly ignore my request... thinking about it more this current PR will not help my case where I don't have an actual distribution class... need to think more

PR #1047, opened 2020-09-21
Using pad_to_size function to remove duplicate code in pad_arrays

dcmaddix:
*Description of changes:*
- Moved `pad_arrays` to `util.py`
- Used `pad_to_size` function to remove duplicate code and return before calling `nd.pad` if no padding is needed.
- Moved `get_dataset` to `testutil/dummy_datasets`
- Moved `test_pad_arrays_axis` to `test/util.py`

### DIFF
--- a/src/gluonts/mx/batchify.py
+++ b/src/gluonts/mx/batchify.py
@@ -11,14 +11,15 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-from typing import List, Optional
+from typing import List, Optional, Union
+import functools
 
 import numpy as np
 import mxnet as mx
 
 from gluonts.core.component import DType
 from gluonts.dataset.common import DataBatch
-from gluonts.support.util import pad_arrays
+from gluonts.support.util import pad_to_size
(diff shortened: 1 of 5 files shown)


lostella (changes requested):
@dcmaddix it would be nice to keep the code that requires MXNet under the `gluonts.mx` subpackage, so that it's clear which imports path require MXNet being installed and which ones don't.

I guess `pad_arrays` could be part of `gluonts.mx.utils`?

dcmaddix:
> @dcmaddix it would be nice to keep the code that requires MXNet under the `gluonts.mx` subpackage, so that it's clear which imports path require MXNet being installed and which ones don't.
> 
> I guess `pad_arrays` could be part of `gluonts.mx.utils`?

yes, I moved it back to `mx.batchify`.  Thanks!

lostella (dismissed):
Looks good to me! I guess the number of test cases changed because of this PR, and this makes some other flaky tests fail consistently -- we should robustify those

lostella (approved):

PR #1051, opened 2020-09-22
PyTorch Predictor

PascalIversen:
*Description of changes:*

1. added a `PyTorchPredictor` class
2. added `gluonts.torch.support.util.copy_parameters` function

based on PR #1052


src/gluonts/model/predictor.py
```diff
@@ -73,7 +56,6 @@
 class Predictor:
     """
     Abstract class representing predictor objects.
-
```
lostella, line 76: Hmmm these changes are strange? Did `black` apply them? There's a few more blank lines removed further down

lostella (changes requested):
Thanks for this! I have only a few comments inline for now.

For the notebook: before we can merge it you should make sure to clear all cells, so that the state of the notebook is not checked into the repo (this makes the file much more lightweight in general)

src/gluonts/model/predictor.py
PascalIversen: Yes, `black` did that, and I don't know why... I checked that I have the right version (19.10b0).

lostella:
@PascalIversen one more comment: maybe it's better to split this into multiple PRs
1. moving the existing code,
2. introduces the new predictor type (this PR),
3. adding the example the notebook.

I'm checking how quick this can be done, but it seems that if you reset the branch to master and only add files outside of `src/gluonts/torch` you should quickly get item 1.

Let me know if you have time for this, otherwise I can take care of 1, and rebase the current PR on top of that.

src/gluonts/model/forecast_generator.py
```diff
@@ -19,6 +19,7 @@
 import numpy as np
 
 # First-party imports
+
```
jaheba, line 22:
```suggestion
```

src/gluonts/model/forecast_generator.py
```diff
@@ -31,6 +32,7 @@
 
 OutputTransform = Callable[[DataEntry, np.ndarray], np.ndarray]
 
+
```
jaheba, line 35:
```suggestion
```

src/gluonts/model/predictor.py
```diff
@@ -46,6 +46,7 @@
 
 from gluonts.model.forecast import Forecast
 
+
```
jaheba, line 49:
```suggestion
```

jaheba (approved):

src/gluonts/torch/support/util.py
```diff
@@ -0,0 +1,38 @@
...
+        Input network.
+    net_dest
+        Output network.
+    strict:
+        whether to strictly enforce that the keys
+        in :attr:`state_dict` match the keys returned by this module's
+        :meth:`~torch.nn.Module.state_dict` function. Default: ``True``
+    """
+
+    net_dest.load_state_dict(net_source.state_dict(), strict=strict)
```
lostella, line 38: As long as this can be done in such a simple way, maybe we don't need to wrap this into a utility function.

Comparing this to [the MXNet equivalent](https://github.com/awslabs/gluon-ts/blob/6287a4a38b549340ed9b49546e2e25c605c83e82/src/gluonts/support/util.py#L139-L171), I can see why we introduced one there

lostella (approved):

PR #1052, opened 2020-09-23
Refactoring forecast generators and predictors for framework independence

PascalIversen:
*Description of changes:*

1. made SampleForecastGenerator and QuantileForecastGenerator framework independent by adding a dispatching function
2. moved mxnet-dependent predictors/forecast generators to gluonts.mx.model...

This is related to the [PyTorchPredictor PR](https://github.com/awslabs/gluon-ts/pull/1051)


src/gluonts/model/forecast_generator.py
```diff
@@ -223,3 +154,16 @@ def __call__(
                     info=batch["info"][i] if "info" in batch else None,
                 )
             assert i + 1 == len(batch["forecast_start"])
+
+
+def DistributionForecastGenerator(*args, **kwargs):
```
jaheba, line 159: I wonder, maybe something like this https://www.python.org/dev/peps/pep-0549/ is a better way to do the deprecation warning.

jaheba: New proposal: We just re-import moved classes at their old locations and add an issue to deprecate these later.

That way we can merge this soon and thing about deprecation later.

lostella: > New proposal: We just re-import moved classes at their old locations and add an issue to deprecate these later.

I second this

PascalIversen: How would you do that concretely? Just importing the classes from the new location would introduce circular imports, since the mx classes depend on the generic classes.

lostella: I just tried it and it seems to be working: at the bottom of `src/gluonts/model/forecast_generator.py` I put

```python
from gluonts.mx.model.forecast_generator import DistributionForecastGenerator
```

and then I can do

```python
>>> from gluonts.model.forecast_generator import DistributionForecastGenerator
>>> DistributionForecastGenerator
<class 'gluonts.mx.model.forecast_generator.DistributionForecastGenerator'>
```

Similarly for `src/gluonts/model/predictor.py`, if I replace your definitions with

```python
from gluonts.mx.model.predictor import GluonPredictor, SymbolBlockPredictor, RepresentableBlockPredictor
```

I think the circular import issue shows up if instead you move these imports to the top of the modules.

PascalIversen: Ah! Thanks :)

src/gluonts/model/predictor.py
```diff
@@ -580,16 +232,50 @@ def _worker_loop(
...
+    )
+
+    warnings.warn(
+        "gluonts.model.predictor.RepresentableBlockPredictor is deprecated. Use gluonts.mx.model.predictor.RepresentableBlockPredictor instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+    return NewRepresentableBlockPredictor(*args, **kwargs)
+
+
```
lostella, line 271: These should be removed since they're imported below?

```suggestion
```

lostella (changes requested):

src/gluonts/model/predictor.py
PascalIversen: Yes thanks! Forget to remove that

lostella (approved):

jaheba (approved):

Issue #1056, opened 2020-09-28
NaN handling of distributions' log_prob() for samples outside distribution's support


elenaehrlich:
## Description
MixtureDistribution with components [Gaussian(..), Gamma(..)]) returns a NaN instead a real-value log-prob for x<=0 (see MWE-1). MixtureDistributionOutput with components [GaussianOutput(..), GammaOutput(..)]) results in a NaN gradient for x<=0 (see MWE-2).

## MWE-1
```python
import mxnet as mx
from scipy import stats
from gluonts.mx.distribution.mixture import *
from gluonts.mx.distribution.gaussian import *
from gluonts.mx.distribution.gamma import *

p = 0.5
mu, sigma = 0., 2.
alpha, beta = 0.9, 2.

# The log-prob of x=-1 should return a real-value and not NaN
x = -1.

scipy_logprob = np.log(p * stats.norm(mu, sigma).pdf(x) + (1-p) * stats.gamma(alpha, 1./beta).pdf(x))
print('scipy_logprob', scipy_logprob)

gaussian = Gaussian(mx.nd.array([mu]), mx.nd.array([sigma]))
gamma = Gamma(mx.nd.array([alpha]), mx.nd.array([beta]))
mixed_dist = MixtureDistribution(mx.nd.array([p,1-p]), [gaussian, gamma])
gluonts_logprob = mixed_dist.log_prob(mx.nd.array([x]))
print('gluonts_logprob', gluonts_logprob.asscalar())

assert np.abs(scipy_logprob-gluonts_logprob.asscalar())<1e-6
```

## Error message or code output
```AssertionError```

## MWE-2
```python
import mxnet as mx, numpy as np
from gluonts.gluonts_tqdm import tqdm
from gluonts.model.common import Tensor
from gluonts.mx.distribution.mixture import *
from gluonts.mx.distribution.mixture import *
from gluonts.mx.distribution.gaussian import *
from gluonts.mx.distribution.gamma import *


def fit_mixture_distribution(x: Tensor, mdo: MixtureDistributionOutput, variate_dimensionality: int = 1, epochs: Optional[int]=1_000):

    args_proj = mdo.get_args_proj()
    args_proj.initialize()
    args_proj.hybridize()

    input = mx.nd.ones((variate_dimensionality, 1))


    trainer = mx.gluon.Trainer(
        args_proj.collect_params(), "sgd", {"learning_rate": 0.02}
    )
[...]

PR #1057, opened 2020-09-28
Improved NaN handling of distributions' log_prob() for samples outside distribution's support

elenaehrlich:
*Issue* #1056

*Description of changes:*
Gamma's log_prob(x) now returns -inf for x<=0 instead of NaNs, similarly with genpareto's log_prob(x) for x<0. Wherever there are NaN in either of the F.where() conditional vectors, then F.where() returns NaN at that entry as well, due to its indicator function multiplication:  1*f(x) + np.nan*0 = nan, since np.nan*0 return nan. 
Therefore replacing log_prob(x) with log_prob(abs(x)) mitigates NaN returns and NaN gradients in cases of x<=0, without altering the value in cases of x>0. This is a known issue in pytorch as well https://github.com/pytorch/pytorch/issues/12986.

Two corresponding pytests are also added.


src/gluonts/mx/distribution/mixture.py
```diff
@@ -150,9 +150,10 @@ def sample(
 
         idx = F.random.multinomial(mixture_probs)
 
-        for _ in range(self.event_dim):
+        n_axes = len(samples_list[0].shape)
```
lostella, line 153: I think these changes should not be here, right?

lostella:
@elenaehrlich it looks like changes from a previous PR are shown here too?

elenaehrlich:
Cleaned up my local branches and opened new pull-request #1060

PR #1062, opened 2020-09-29
Added the GP predictive likelihood for the training

Francois-Aubet:
The currently the optimisation of the GP model's kernel hyper-parameters is done by maximising the likelihood of the training data, and not the predictive likelihood of the future data as one would be interested in forecasting.

In the following explanation we use:
- like_train : marginal likelihood of the training/past data : p(X_past | Y_past, k)
- like_pred : conditional predictive likelihood of the future data : p(X_future | Y_future, X_past, Y_past, k)

If the kernel is well specified, this is not a problem because the kernel hyper-parameters that maximise like_train are the same ones as the ones that maximise like_pred.

In practice however, the kernel is most likely to be miss specified (especially with our current choice of kernels). This means that the hyper-parameters that maximise like_pred, and so are the most suited for forecasting, are not the ones that maximise like_train.

For this reason I propose to add an option allowing to learn the kernel hyper parameters that maximise both.

I tested this change on some toy synthetic data, the new method seems to give a better forecasting performance (MSE).
I still need to test that on a bigger scale.


src/gluonts/model/gp_forecaster/_network.py
```diff
@@ -182,6 +186,8 @@ def hybrid_forward(
         past_target: Tensor,
         past_time_feat: Tensor,
         feat_static_cat: Tensor,
+        future_target: Tensor,
```
dcmaddix, line 189: Looks great.  Thanks for adding!

src/gluonts/model/gp_forecaster/_network.py
```diff
@@ -208,12 +219,20 @@ def hybrid_forward(
...
             float_type=self.float_type,
             max_iter_jitter=self.max_iter_jitter,
             jitter_method=self.jitter_method,
         )
-        return gp.log_prob(past_time_feat, past_target)
+        log_prob_data = gp.log_prob(past_time_feat, past_target)
+        log_prob_forecast = gp.log_prob_predictive(
+            past_time_feat, past_target, future_time_feat, future_target
+        )
+        if self.train_forecast:
```
dcmaddix, line 232: I will you can make this part a bit more succinct with
`return log_prob_data + self.train_forecast * log_prob_forecast`, since `self.train_forecast` is a boolean here this should work.

src/gluonts/model/gp_forecaster/gaussian_process.py
```diff
@@ -224,6 +224,59 @@ def log_prob(self, x_train: Tensor, y_train: Tensor) -> Tensor:
...
+            transpose_a=True,
+        ).squeeze(axis=-1)
+        predictive_covariance = self.kernel.kernel_matrix(
+            x_test, x_test
+        ) - self.F.linalg.gemm2(
+            lower_tri_solve, lower_tri_solve, transpose_a=True
+        )
+
+        l_test = -MultivariateGaussian(
+            predictive_mean, predictive_covariance,
```
dcmaddix, line 275: I think you could call `exact_inference` to compute the `predictive_mean` and `predictive_covariance` and just input them here so we don't need the duplicate code?

dcmaddix (changes requested):

src/gluonts/model/gp_forecaster/_network.py
dcmaddix: Actually, we don't need to compute `log_prob_predictive` if `self.train_forecast = True`.  so we can have:
` if self.train_forecast: return 
`

Francois-Aubet:
I compared the two methods on electricity_nips and traffic_nips. 

Evaluating on electricity_nips:
- Original method:
   'MSE': 6373341.03141, 'abs_error': 13501996.3605

- Using the predictive distribution:
   'MSE': 6526002.22909, 'abs_error': 13663573.2570

Evaluating on traffic_nips:
- Original method:
   'MSE': 0.002243303127143, 'abs_error': 5079.79117838

- Using the predictive distribution:
   'MSE': 0.002225792571758, 'abs_error': 5037.49058717

So, using the predictive distribution seems to be slightly better in one case and slightly worse in another.

I still have to investigate why it is worse on electricity.

dcmaddix (approved):

dcmaddix:
Thanks for running the experiments Francois and making the changes. I think we are good to merge but it looks like some of the tests were failing? Thanks!

Issue #1069, opened 2020-10-01
MQCNNEstimator behaves strangely results when constructed with distr_output


lostella:
## Description
I'm getting NaN loss, or inconsistent shapes in the forecast object, when training a `MQCNNEstimator` with the `distr_output` option instead of `quantiles`.

## To Reproduce
(Please provide minimal example of code snippet that reproduces the error. For existing examples, please provide link.)

```python
from gluonts.dataset.repository.datasets import get_dataset
from gluonts.distribution import StudentTOutput
from gluonts.model.seq2seq import MQCNNEstimator
from gluonts.trainer import Trainer

dataset = get_dataset("m4_hourly")
estimator = MQCNNEstimator(freq="H", prediction_length=1, distr_output=StudentTOutput(), trainer=Trainer(epochs=1, learning_rate=1e-4))
predictor = estimator.train(dataset.train)

for forecast in predictor.predict(dataset.test):
    print(forecast.distribution)
    break
```

## Error message or code output

Often times the above results in NaN loss at the first training iteration (this often happens when putting e.g. `prediction_length=24`, if not always). Otherwise I get the following prediction

```
gluonts.mx.distribution.transformed_distribution.TransformedDistribution(base_distribution=gluonts.mx.distribution.student_t.StudentT(F=None, mu=mxnet.nd.array([52.238128662109375], dtype=numpy.float32), nu=mxnet.nd.array([2.0], dtype=numpy.float32), sigma=mxnet.nd.array([302.1168518066406], dtype=numpy.float32)), transforms=[gluonts.mx.distribution.bijection.AffineTransformation(loc=mxnet.nd.array([[0.0]], dtype=numpy.float32), scale=mxnet.nd.array([[1.0]], dtype=numpy.float32))])
```

Note that the `loc` and `scale` of the `AffineTransformation` have one additional axis which shouldn't be there. (They should have the same shape as `mu`, `nu`, `sigma`.)

lostella:
The occurrence of nans is probably due to scaling being disabled by default; the shape issue is likely due to the scaler being initialized with `keepdims=True`

PR #1070, opened 2020-10-01
Fix scaling for MQ-(C|R)NN when distribution outputs are used

dcmaddix:
*Issue #, if available:*
https://github.com/awslabs/gluon-ts/issues/1069

*Description of changes:*
- Updated default `scaling` for large to `True` to be used with `DistrOutput`
- Left default for `scaling_decoder_dynamic_feature` to be `False`
- Pass `scaling_decoder_dynamic_feature` to training and prediction networks
- Fix the shape of `loc` and `scale`

### DIFF
--- a/src/gluonts/model/seq2seq/_forking_estimator.py
+++ b/src/gluonts/model/seq2seq/_forking_estimator.py
@@ -23,15 +23,14 @@ from gluonts.dataset.field_names import FieldName
 from gluonts.model.estimator import GluonEstimator
 from gluonts.model.forecast import Quantile
 from gluonts.model.forecast_generator import QuantileForecastGenerator
-from gluonts.mx.model.forecast_generator import DistributionForecastGenerator
 from gluonts.model.predictor import Predictor
-from gluonts.mx.model.predictor import RepresentableBlockPredictor
-
 from gluonts.mx.block.decoder import Seq2SeqDecoder
 from gluonts.mx.block.enc2dec import FutureFeatIntegratorEnc2Dec
 from gluonts.mx.block.encoder import Seq2SeqEncoder
 from gluonts.mx.block.quantile_output import QuantileOutput
 from gluonts.mx.distribution import DistributionOutput
+from gluonts.mx.model.forecast_generator import DistributionForecastGenerator
+from gluonts.mx.model.predictor import RepresentableBlockPredictor
 from gluonts.mx.trainer import Trainer
 from gluonts.support.util import copy_parameters
 from gluonts.time_feature import time_features_from_frequency_str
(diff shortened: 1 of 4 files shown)


lostella:
@dcmaddix thanks! I think scaling *must* be off when `quantile_output` is used, because apparently the output doesn’t get scaled back: so you don’t want to scale the input, otherwise for fixed model parameters and two series `ts` and `alpha*ts`, you will get the same predictions for any `alpha > 0`. To allow for scaling in this case, the output quantiles should be scaled back.

On the other hand, scaling appears to be necessary when using distributions, otherwise NaNs are all over the place unless data is nicely scaled (like all time series with the same magnitude or something).

So I ended up making the default value for `scaling` dependent on what is being used, which maintains backwards compatibility in the `quantile_output` case, and fixese the `distr_output` one. Plus I've fixed the shape issue described in #1069.

src/gluonts/model/seq2seq/_forking_network.py
```diff
@@ -77,7 +77,7 @@ def __init__(
         embedding_dimension: List[int],
         distr_output: Optional[DistributionOutput] = None,
         quantile_output: Optional[QuantileOutput] = None,
-        scaling: bool = False,
+        scaling: bool = True,
```
dcmaddix, line 80: Should we update the default for `scaling` back to `False` here?

dcmaddix:
> @dcmaddix thanks! I think scaling _must_ be off when `quantile_output` is used, because apparently the output doesn’t get scaled back: so you don’t want to scale the input, otherwise for fixed model parameters and two series `ts` and `alpha*ts`, you will get the same predictions for any `alpha > 0`. To allow for scaling in this case, the output quantiles should be scaled back.
> 
> On the other hand, scaling appears to be necessary when using distributions, otherwise NaNs are all over the place unless data is nicely scaled (like all time series with the same magnitude or something).
> 
> So I ended up making the default value for `scaling` dependent on what is being used, which maintains backwards compatibility in the `quantile_output` case, and fixese the `distr_output` one. Plus I've fixed the shape issue described in #1069.
> 
> Edit: I had to bump the test accuracy for the `distr_output` cases, since now the default scaling behaviour changed.

dcmaddix:
> @dcmaddix thanks! I think scaling _must_ be off when `quantile_output` is used, because apparently the output doesn’t get scaled back: so you don’t want to scale the input, otherwise for fixed model parameters and two series `ts` and `alpha*ts`, you will get the same predictions for any `alpha > 0`. To allow for scaling in this case, the output quantiles should be scaled back.
> 
> On the other hand, scaling appears to be necessary when using distributions, otherwise NaNs are all over the place unless data is nicely scaled (like all time series with the same magnitude or something).
> 
> So I ended up making the default value for `scaling` dependent on what is being used, which maintains backwards compatibility in the `quantile_output` case, and fixese the `distr_output` one. Plus I've fixed the shape issue described in #1069.
> 
> Edit: I had to bump the test accuracy for the `distr_output` cases, since now the default scaling behaviour changed.

Looks great, thanks for updating the default to cover both cases! :)

src/gluonts/model/seq2seq/_forking_estimator.py
```diff
@@ -154,7 +155,7 @@ def __init__(
         enable_encoder_dynamic_feature: bool = True,
         enable_decoder_dynamic_feature: bool = True,
         trainer: Trainer = Trainer(),
-        scaling: bool = False,
+        scaling: Optional[bool] = None,
         scaling_decoder_dynamic_feature: bool = False,
```
dcmaddix, line 159: Do we want to turn on scaling of the dynamic features or this should be unrelated to the distribution change right?

src/gluonts/model/seq2seq/_forking_network.py
lostella: Right!

src/gluonts/model/seq2seq/_forking_estimator.py
lostella: I think this is unrelated maybe, we could do it but maybe as a separate story

lostella (approved):

PR #1073, opened 2020-10-02
Update requirement to mxnet~=1.7

lostella:
*Description of changes:* MXNet 1.6 had an annoying [issue](https://github.com/apache/incubator-mxnet/issues/17914) on GPU when hybridizing, which forced some [unpleasant fixes](https://github.com/awslabs/gluon-ts/pull/728). However we didn't fix everything, and the issue might be related to [other observed strange behaviours](https://github.com/awslabs/gluon-ts/issues/833): before we go an pollute the code with undesirable, very verbose expression in place of `+`, `-`, `*`, and `/`, maybe it's better to recommend an updated MXNet version.


jaheba:
Also, do we still need `--upgrade` option?

lostella:
> Also, do we still need `--upgrade` option?

Well in case you already have either package, I guess. For example having mxnet pre-1.6 would not work

jaheba (approved):

PR #1074, opened 2020-10-05
Add piecewise linear quantile function output with fixed knots

lostella:
*Description of changes:* This adds the extension to the `PiecewiseLinearOutput` class which uses a fixed set of interpolation knots. Mainly authored by @canerturkmen 


canerturkmen (approved):
lgtm assuming windows test fail is OK.

Issue #1066, opened 2020-09-30
Gluon-ts DeepAR not able to learn a Gamma-Gaussian MixtureDistributionOutput() for x_t \in Reals


elenaehrlich:
## Description
Gluon-ts MixtureDistributionOutput() of Gamma-Gaussian is able to learn the component weights and arguments for real-valued data. Why is Gluon-ts DeepAR not similarly able to learn the component weights and arguments of a Gamma-Gaussian MixtureDistributionOutput() from the same data? 

## To Reproduce
```python
from gluonts.mx.distribution.mixture import *
from gluonts.mx.distribution.gaussian import *
from gluonts.mx.distribution.gamma import *
import numpy as np
from gluonts.dataset import common
from gluonts.model import deepar
from gluonts.mx.trainer import Trainer
from gluonts.model.common import Tensor, NPArrayLike
from gluonts.gluonts_tqdm import tqdm
import gluonts
print(f'gluonts {gluonts.__version__}')


##########################################
# Function to learn MLE
def fit_mixture_distribution(x: Tensor, mdo: MixtureDistributionOutput, variate_dimensionality: int = 1, epochs: Optional[int]=1_000):

    args_proj = mdo.get_args_proj()
    args_proj.initialize()
    args_proj.hybridize()

    input = mx.nd.ones((variate_dimensionality, 1))


    trainer = mx.gluon.Trainer(
        args_proj.collect_params(), "sgd", {"learning_rate": 0.02}
    )
    print('trainer.learning_rate',trainer.learning_rate)

    t = tqdm(list(range(epochs)))
    for _ in t:
        with mx.autograd.record():
            distr_args = args_proj(input)
            d = mdo.distribution(distr_args)
            loss = d.loss(x).mean()
        loss.backward()
        loss_value = loss.asnumpy()
        t.set_postfix({"loss": loss_value})
        trainer.step(1)

    distr_args = args_proj(input)
    d = mdo.distribution(distr_args)
    return d

##########################################
# Simulate data 
mx.random.seed(1)
np.random.seed(1)
num_samples = 1000
p1 = 0.2; p2 = 1.0 - p1
mu1 = 0.0; sigma1 = 0.2
alpha = 9.; beta = 2.0
mixture = MixtureDistribution(mx.nd.array([[p1, p2]]), [Gaussian(mx.nd.array([mu1]), mx.nd.array([sigma1])), Gamma(mx.nd.array([alpha]), mx.nd.array([beta]))])
[...]

PR #1078, opened 2020-10-06
fix Gamma distribution's NaN gradients for zero inputs

PascalIversen:
*Issue #, if available:*
fix for #1066

*Description of changes:*
Added masking to prevent NaN gradients of the `Gamma` for `x=0`.
This is important for fitting mixture distributions such as: 
```python
mdo = MixtureDistributionOutput([GaussianOutput(), GammaOutput()]).
```
Example code:
```python
import mxnet as mx
from gluonts.mx.distribution import Gamma

a =mx.nd.array([1])
b =mx.nd.array([1])

a.attach_grad()
b.attach_grad()

with mx.autograd.record():
    loss = mx.nd.mean(Gamma(a,b).loss(x=mx.nd.array([0])))
    loss.backward()
 

print(a.grad)
print(b.grad)
```
will return
```python
[0.]
<NDArray 1 @cpu(0)>

[0.]
<NDArray 1 @cpu(0)>
```

instead of
```python
[nan]
<NDArray 1 @cpu(0)>

[0.]
<NDArray 1 @cpu(0)>

```

### DIFF
--- a/src/gluonts/mx/distribution/gamma.py
+++ b/src/gluonts/mx/distribution/gamma.py
@@ -84,9 +84,12 @@ class Gamma(Distribution):
         altering the value in cases of x>0. 
         This is a known issue in pytorch as well https://github.com/pytorch/pytorch/issues/12986.
         """
+        # mask zeros to prevent NaN gradients for x==0
+        x_masked = F.where(x == 0, x.ones_like() * 0.5, x)
+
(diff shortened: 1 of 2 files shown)


PascalIversen:
I added a test case which fails on master but passes on this branch.

lostella (approved):

PR #1082, opened 2020-10-12
initial pytorch distribution output class

kashif:
Here is the initial PR to start adding the distributional output modules in pytorch.


lostella:
Hey @kashif, thank you for this! A couple of things seem to break the checks:
* Style checks complain that the header is missing in one of the files, you can use the script in https://github.com/awslabs/gluon-ts/blob/master/.devtools/license to check/fix the headers across the code base
* Tests are failing because of the `test/torch/modules/test_distribution_inference.py` module, whose file has the same name as `test/distribution/test_distribution_inference.py`. Apparently pytest doesn't like this, one thing we could do is renaming the test modules as `test_mx_distribution_inference` and `test_torch_distribution_inference`

I'm not sure why the doc build is failing, but maybe fixing everything else will unlock that too

kashif:
thanks @lostella yes the docs are failing due to the same issue that there is already an `ArgProj` class defined... i was gonna investigate and fix these issues

kashif:
@lostella perhaps i missed to add the pytorch path for the correct workflow files... if you could kindly add them I can test again

lostella (approved):
Thanks @kashif, looks good to me!

Issue #1030, opened 2020-09-10
Usage of DateSplitter with max_history


MaximilianPavon:
Hi,
I am trying to use the `DateSplitter` together with `max_history` in order to make the test data shorter and would love to get some help.

The documentation [here](https://gluon-ts.mxnet.io/api/gluonts/gluonts.dataset.split.splitter.html) says the following about `max_history`:
```
If given, all entries in the test-set have a max-length of max_history. 
This can be sued to produce smaller file-sizes.
```

But when I set this value to e.g. 3 weeks of hourly data (`max_history = 504`) I am getting AssertionErrors from [this](https://github.com/awslabs/gluon-ts/blob/18ab342a3abd7cfb1b6460dacfffd42b64fbefde/src/gluonts/dataset/split/splitter.py#L228) line, here is the relevant code block:
```python
def split(self, items: List[DataEntry]) -> TrainTestSplit:
    split = TrainTestSplit()

    for item in map(TimeSeriesSlice.from_data_entry, items):

        train = self._train_slice(item)
        test = self._trim_history(self._test_slice(item))

        split._add_train_slice(train)

        assert len(test) - len(train) >= getattr(self, "prediction_length")
        split._add_test_slice(test)

    return split
```

So the length of the test slice minus the length of the train slice is not greater or equal to the defined prediction length.

When debugging this code line by line, the problem seems to originate from the function `_trim_history()`, implemented a few [lines](https://github.com/awslabs/gluon-ts/blob/18ab342a3abd7cfb1b6460dacfffd42b64fbefde/src/gluonts/dataset/split/splitter.py#L212-L216) above:
```python
def _trim_history(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
    if getattr(self, "max_history") is not None:
        return item[: -getattr(self, "max_history")]
    else:
        return item
```

Currently `_trim_history()` **disregards** the last `max_history` data points from the test slice. Considering that the test slice is exactly `prediction_length` data points longer than the training slice, this is bound to run into the mentioned AssertionError.

Is this the expected behaviour of this parameter or am I doing something wrong?

Here is some example code to replicate this:
```python
import gluonts
from gluonts.dataset.field_names import FieldName
from gluonts.dataset.split import DateSplitter
import numpy as np
import pandas as pd

print(f"gluonts version: {gluonts.__version__}")  # gluonts version: 0.5.2

prediction_length = 72

start_date = pd.Timestamp("2020-01-01", freq="1H")
[...]

PR #1085, opened 2020-10-13
Fix dataset.splitter max_history argument

PascalIversen:
*Issue #, if available:*
#1030 

*Description of changes:*

I fixed a bug concerning the `max_history` argument of the dataset.splitter tools and added tests.
(Based on PR #1083)


src/gluonts/dataset/split/splitter.py
```diff
@@ -224,8 +225,10 @@ def split(self, items: List[DataEntry]) -> TrainTestSplit:
             test = self._trim_history(self._test_slice(item))
 
             split._add_train_slice(train)
-
-            assert len(test) - len(train) >= getattr(self, "prediction_length")
+            if (getattr(self, "max_history")) is None:
+                assert len(test) - len(train) >= getattr(
+                    self, "prediction_length"
+                )
```
lostella, line 231: Why is the guard needed here? Doesn't the condition need to hold in any case?

PascalIversen: As I understand it, the reason to use the max_split parameter is to shorten the test set.
For some models it will be enough to use `max_history = context_length+prediction_length`. In that case the test set has length `context_length+prediction_length`, while the train set might be much larger.

lostella: > What always holds is something like end_date_train+pd.Timedelta(prediction_length) == end_date_test

Or maybe something like

`end_date_train + prediction_length * end_date_train.freq <= end_date_test`

lostella: I think we should keep a meaningful assertion without any condition, because this is essentially checking whether it makes sense to do backtesting on the test data, for the given prediction length, given that you trained on this training data

lostella (changes requested):
Thanks for this @PascalIversen! I think a similar assertion logic should go in the `rolling_split` method, in addition to `split`.

Maybe we can add tests for that method too?

lostella (approved):

PR #1086, opened 2020-10-15
PyTorchPredictor serde

lostella:
*Description of changes:* This adds support for serialization of `PyTorchPredictor` objects


src/gluonts/torch/model/predictor.py
```diff
@@ -83,3 +79,58 @@ def predict(
...
+            print(dump_json(self.prediction_net), file=fp)
+        torch.save(
+            self.prediction_net.state_dict(), path / "prediction_net_state"
+        )
+
+        # serialize transformation chain
+        with (path / "input_transform.json").open("w") as fp:
+            print(dump_json(self.input_transform), file=fp)
+
+        # FIXME: also needs to serialize the output_transform
```
lostella, line 97: @vafl do you remember what the idea behind `output_transform` was? I don't think this is used anywhere (and also MXNet-based predictors have this TODO, so they don't serialize it)

jaheba:
Not sure, but shouldn't we test if the Predictor can be used with `serde.encode / decode`?

lostella:
> Not sure, but shouldn't we test if the Predictor can be used with `serde.encode / decode`?

Also not sure. I can try, for sure predictors rely on the serialize and deserialize methods

src/gluonts/torch/model/predictor.py
mbohlkeschneider: It is some sort of invertible transformation that can be put into the transformation chain. This will not yield the correct jacobian, but some transformations are hard to implement inside the graph (for instance, the empirical cdf of `GPVAR`).

lostella: Ok -- I think I left this FIXME here since the MXNet-based predictor classes have this

mbohlkeschneider: Sounds good.

mbohlkeschneider (approved):

mbohlkeschneider (approved):

PR #1087, opened 2020-10-15
include callback in trainer and example for warm starting

vafl:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/mx/trainer/_base.py
+++ b/src/gluonts/mx/trainer/_base.py
@@ -99,9 +99,9 @@ class Trainer:
         Initializer of the weights of the network (default: "xavier").
     hybridize
         If set to true the network will be hybridized before training
-    after_initialize_cb
+    post_initialize_cb
         An optional callback function. If provided the function will be called with the
-        initialized network `after_initialize_cb(net)` before the training starts.
+        initialized network `post_initialize_cb(net)` before the training starts.
         This callback can be used to e.g. overwrite parameters for warm starting, to freeze some
         of the network parameters etc.
     """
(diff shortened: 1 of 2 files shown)


src/gluonts/mx/trainer/_base.py
```diff
@@ -198,6 +206,13 @@ def base_path() -> str:
                 static_shape=True,
             ):
                 batch_size = train_iter.batch_size
+                sample_batch = next(iter(train_iter))
```
lostella, line 209: Could it be that consuming a batch here leaves the data loader modified just enough to change some results in tests? The tests are definitely to blame, but maybe this surfaces an undesired property of the data loader?

src/gluonts/mx/trainer/_base.py
```diff
@@ -98,6 +98,12 @@ class Trainer:
     init
         Initializer of the weights of the network (default: "xavier").
     hybridize
+        If set to true the network will be hybridized before training
+    after_initialize_cb
```
jaheba, line 102: Maybe `post_initialize` is a nicer name? Personally, I'm not a fan of the `_cb` prefix.

src/gluonts/mx/trainer/_base.py
vafl: yeah, maybe. 

It could be a problem with the dataloader or perhaps the tests are so brittle that small changes (e.g. shuffling the lines in the files) breaks things. I'll try to debug this.

lostella: In the data loader some internal iterator is constructed at construction time, when that should probably happen at `iter` time, I'll also look into that

vafl: Yeah, I think this is actually intended behaviour, since we want the next epoch to continue where the previous ended. The failing test is still strange though. I can't reproduce it on my laptop or a docker image. 

For now I changed things to not pull an extra batch for this.

lostella (approved):

PR #1093, opened 2020-10-16
Fix a few bugs about tensor shapes in default values for TFT implementation

### LINKED ISSUE #1075: Provide working examples for Temporal Fusion Transformer
# Description

Referencing to this issue https://github.com/awslabs/gluon-ts/issues/1013#issuecomment-687575025, I am trying to use the [Temporal Fusion Transformer](https://github.com/awslabs/gluon-ts/tree/50bc3835f50f73deed26d90e6b1226c78d4fe308/src/gluonts/model/tft) in GluonTS. 

@dcmaddix & @Gandor26: can you please provide a detailed example on how to use the TFT, also with examples on how to use

-  `time_features:List[gluonts.time_feature._base.TimeFeature]`
-  `static_cardinalities:Dict[str, int]`
- `dynamic_cardinalities:Dict[str, int]`
- `static_feature_dims:Dict[str, int]`
- `dynamic_feature_dims:Dict[str, int]`
- `past_dynamic_features:List[str]`

and add it to https://github.com/awslabs/gluon-ts/tree/master/evaluations.

https://pytorch-forecasting.readthedocs.io/en/stable/tutorials/stallion.html is a really exhaustive tutorial on how to use TFT. Would be great if you can also provide it for the GluonTS implementation.

Thanks

Gandor26:
*Issue #, if available:*
Fixes #1075 

*Description of changes:*
Found some incompatibility of tensor shapes when features are not provided. Fixed default value.


lostella:
Hey @Gandor26, thanks for this! It looks like tests for this model are completely missing, maybe we could add some that would have caught the issue? You can take a look at how "smoke tests" are done for other models here: https://github.com/awslabs/gluon-ts/tree/master/test/model, see for example the test for the `SimpleFeedForwardEstimator` https://github.com/awslabs/gluon-ts/blob/master/test/model/simple_feedforward/test_model.py

src/gluonts/model/tft/_estimator.py
```diff
@@ -40,13 +40,23 @@
...
+def _default_feat_args(dims_or_cardinalities: List[int]):
+    if dims_or_cardinalities:
+        return dims_or_cardinalities
+    return [1]
+
+
+
 class TemporalFusionTransformerEstimator(GluonEstimator):
     @validated()
     def __init__(
```
lostella, line 62: Why is `context_length` required and `prediction_length` optional? Shouldn't it be the opposite?

lostella (changes requested):
I've looked into this branch a bit to see if similar tests to other models could pass on `TemporalFusionTransformerEstimator`. Ideally, the following minimal example should be able to run, both with `hybridize=False` as well as `hybridize=True`.

```python
from gluonts.model.tft import TemporalFusionTransformerEstimator

estimator = TemporalFusionTransformerEstimator(freq="H", prediction_length=24, trainer=Trainer(epochs=1, hybridize=False))

for forecast in predictor.predict(dataset.test):
    print(forecast.mean)
    break
```

When running this, I encountered a few issues:
* the networks rely on `+`, `-`, `*` and `/` to do broadcasting, which doesn't seem to be the case when `hybridize=True`. You will have to change those for `F.broadcast_add`. `F.broadcast_minus`, `F.broadcast_mul` and `F.broadcast_div`, respectively.
* as commented inline, I think the constructor has `context_length` and `prediction_length` swapped (the latter shouldn't be optional)
* I got an error in `gluonts/model/tft/_network.py`, line 365, in _preprocess: `MXNetError: Check failed: (*begin < *end): Invalid begin, end, get begin=24, end=None`

lostella:
@Gandor26 I believe the issue might be in the `BroadcastTo` transformation, which always broadcasts to the same time length as the `"target"` field. However, dynamic features might need to be `prediction_length` longer than `"target"` at inference time.

src/gluonts/model/tft/_estimator.py
Gandor26: You are right. It should be opposite.

Gandor26:
@lostella Thanks for your suggestions! You just pointed every issue out precisely. The fixes have been updated and the the (minimum) smoke test is also added.

test/model/tft/test_model.py
```diff
@@ -0,0 +1,37 @@
...
+def hyperparameters():
+    return dict(ctx="cpu", epochs=1, learning_rate=1e-3,)
+
+
+@pytest.mark.parametrize("hybridize", [True, False])
+def test_accuracy(accuracy_test, hyperparameters, hybridize):
+    hyperparameters.update(num_batches_per_epoch=200, hybridize=hybridize)
+    accuracy_test(
+        TemporalFusionTransformerEstimator, hyperparameters, accuracy=0.2
+    )
```
lostella, line 29: In this test, all other models are run with very minimal settings (like small number of nodes or layers, and short-ish epochs) so as not to take too long. This test here appears to be timing out (we have a 30s timeout on each test case I believe): maybe the settings here need to be adjusted?

lostella:
Also, I'm not sure why the docs build is failing, but maybe you can update the PR branch to test it together with the latest changes from mainline. We can also figure out the docs build later, otherwise.

lostella (approved):
Looks good! Thank you @Gandor26!

PR #1104, opened 2020-10-23
Ignore NaN values during training and throw a warning (training got stuck before)

prolearner:
### Issue
Before this change, the parameters of the model where updated even when the loss was NaN. The check for the NaN value was only at the end of the epoch and resulted in a warning. An error was thrown only if a NaN loss value was found in the first epoch with a suggestion of reducing the initial learning rate, otherwise the training would continue resulting in undesired behaviour, like training getting stuck and freezing the CPU, which I experienced firsthand.

### Change
I changed the code to throw a warning and ignore NaNs during training. With this change, the training does not get stuck anymore, although the previously thrown error now triggers only when ALL the batches of the first epoch give NaNs.

### Comments
I think there should probably be a parameter to select the appropriate behaviour when encountering NaNs during training (warning or error). It might be that some samples taken from the time-series are just unlucky and ignoring them gives still good results. This is what I experienced in my experimental setup.

### DIFF
--- a/src/gluonts/mx/trainer/_base.py
+++ b/src/gluonts/mx/trainer/_base.py
@@ -275,27 +275,25 @@ class Trainer:
                                 else:
                                     loss = output
 
-
                             if not np.isfinite(ndarray.sum(loss).asscalar()):
                                 logger.warning(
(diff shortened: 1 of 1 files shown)


vafl (approved):
Looks good. Thanks!

PR #1105, opened 2020-10-25
Estimator and predictor for autogluon tabular

arianacai1997:
*Issue #, if available:*
I tried to use time series data (m4_weekly in the code) in autogluon. After fitting and predicting with autogluon models, I converted the prediction back to gluonts SampleForecast format for evaluation. 
*Description of changes:*
Added this part of the code


src/gluonts/nursery/autogluon_for_single_ts.py
```diff
@@ -0,0 +1,131 @@
...
+test = list(ds.test)[0]
+series = to_pandas(test)
+df = g.parse_series(series)
+task_tr, task_val, task_ts = g.split_df(df)
+label_column = 'target'
+problem_type = 'regression'
+predictor = task.fit(train_data=task_tr, tuning_data=task_val, label=label_column, problem_type=problem_type,
+                     output_directory='AutogluonModels/ag-a') #, presets='best_quality'
+# print(predictor.fit_summary())
+print(g.evaluate_it(predictor, task_ts))
```
jwmueller, line 131: I think this should all be wrapped into a new GluonTS model called `AutoGluonPredictor` or something like that. Users should be able to use this new GluonTS model just like they can use any other GluonTS models. They should also be able to pass in any kwargs they wish into AutoGluon's task.fit() call.

@lovvge @dcmaddix  Please chime in here on the best way to achieve an integrated API!

jwmueller: your class can be called `AutogluonEstimator` and should inherit `Estimator` class:

https://github.com/awslabs/gluon-ts/blob/13eb1f5396240232ddc6ea04ad37eb8fbf3af150/src/gluonts/model/estimator.py#L37

test/nursery/autogluon_for_single_ts.py
```diff
@@ -0,0 +1,131 @@
...
+Autogluon task fit problem_type
+num_samples:
+Number of samples to draw on the model when evaluating.
+"""
+
+    def __init__(self, dataset_name, problem_type, num_samples, context_length=None):
+        self.dataset = get_dataset(dataset_name, regenerate=False)
+        self.problem_type = problem_type
+        self.num_samples = num_samples
+        self.context_length = context_length if context_length else self.dataset.metadata.prediction_length
```
lostella, line 32: Is this the same as the other module in this PR? I think this can be removed

src/gluonts/nursery/autogluon_for_single_ts.py
```diff
@@ -0,0 +1,131 @@
...
+from gluonts.evaluation.backtest import make_evaluation_predictions
+from gluonts.model.forecast import SampleForecast
+from gluonts.dataset.common import ListDataset
+import json
+import numpy as np
+import matplotlib.pyplot as plt
+import ssl
+ssl._create_default_https_context = ssl._create_unverified_context
+
+class Gluonts3Auto:
```
lostella, line 17: Maybe minor: why this name choice?

src/gluonts/nursery/autogluon_for_single_ts.py
lostella: I think one could do both:
1. for training on a single time series "on the fly" and directly getting predictions back, the functionality can be wrapped into a class that extends `Predictor` from GluonTS
2. if we want to also have offline training (i.e. train on a bunch of time series and get a regression model to be used in the future for predictions) then that should be done by extending `Estimator`

I think option 1. is simpler to start with

lostella:
Hey @arianacai1997 I think you could wrap the functionality into a class extending `Predictor`, or better `RepresentablePredictor`. You can find an example here: https://github.com/awslabs/gluon-ts/blob/13eb1f5396240232ddc6ea04ad37eb8fbf3af150/src/gluonts/model/predictor.py#L152

Essentially you would only need to implement `__init__` and `predict_item`. The latter should do the whole procedure of setting up features and formatting the `DataFrame` training a regression model, and then producing and returning the forecast: you can split this into sub-procedures to keep the code clean and organized, but in the end `predict_item` is where everything should happen

jwmueller:
> Hey @arianacai1997 I think you could wrap the functionality into a class extending `Predictor`, or better `RepresentablePredictor`.

@lostella  What's the advantage/disadvantage of this vs having the AutoGluon-models be a subclass of `Estimator`? Shouldn't we provide a nice way for users to easily fit AutoGluon-models for every time series in their dataset? (especially if we want to later extend AutoGluon to predict all time series simultaneously using info from all the other series)

lostella:
@jwmueller I think we can do both, reusing most of the code. The way I see it one could have:

1. a `TabularLocalPredictor` class that, upon calling the `predict` method will, for each time series (i) set up the dataframe with features/target (ii) train a model using AutoGluon tabular (iii) predict
2. a `TabularPretrainedPredictor` that is initialized with a trained model for tabular tasks, and in `predict` only does steps (i) for the features in the prediction range, and naturally (iii)
3. a `TabularEstimator` that constructs the features/target dataframe for a collection of time series, train a tabular model, and return a `TabularPretrainedPredictor`

All of these of course can rely on the same code for feature processing, which could actually be used to interface to any package for regression.

But I may be missing something, especially on the AutoGluon side, so correct me if I'm wrong @jwmueller

jwmueller:
> @jwmueller I think we can do both, reusing most of the code. The way I see it one could have:
> 
> 1. a `TabularLocalPredictor` class that, upon calling the `predict` method will, for each time series (i) set up the dataframe with features/target (ii) train a model using AutoGluon tabular (iii) predict
> 2. a `TabularPretrainedPredictor` that is initialized with a trained model for tabular tasks, and in `predict` only does steps (i) for the features in the prediction range, and naturally (iii)
> 3. a `TabularEstimator` that constructs the features/target dataframe for a collection of time series, train a tabular model, and return a `TabularPretrainedPredictor`

My main hope is just that this PR can serve as a sandbox for later determining the actual best way to utilize AutoGluon for forecasting. If this PR were just a `TabularLocalPredictor`, then do you think it will require nontrivial work to convert it to a `TabularEstimator` later on?  If so, then I'm in favor of just having this PR be a `TabularEstimator` now, so that we can easily in the future explore how to better feed cross-time-series information into AutoGluon. 

What's the advantage of just keeping this PR to a `TabularLocalPredictor`? My (limited) understanding is this PR could for now be a `TabularEstimator` that just happens to naively fit models for each time-series independently (for now only,  where the fact that these models for now ignore cross-times-series info would just be hidden from the user). Or is creating a new `TabularEstimator` much harder than a new `TabularLocalPredictor`?

Once you've formatted your time-series data into a single pandas Dataframe with one column containing the target-values to predict and the other columns containing the information to condition these predictions on (whether from one time-series or multiple), then the best way to think about using AutoGluon is just like scikit-learn's RandomForestRegressor (i.e. it just does regression without many important hyperparameters or need for any data processing).   AutoGluon provides a simple `fit()` method that does the data-processing + predictor-training on a given Dataframe, and also a simple `predict()` method that does the data-processing + predictor-inference for a new Dataframe.

lostella:
> Or is creating a new TabularEstimator much harder than a new TabularLocalPredictor?

The thing is that a new type extending `Predictor` would be needed anyway: an estimator returns a predictor when `train` is called, and the predictor being returned here should wrap an Autogluon model.

To clarify, the estimator/predictor interface is to be intended as follows:
* `Estimator.train` takes a collection of time series, and learns a single _common_ model for them, that is a model that is good (hopefully) at predicting each of them; the learned model is returned within a predictor.
* `Predictor.predict` takes a collection of time series, and produces a forecast for each one of them _independently_ using the model it implements.

The latter point is important: to take into account for interdependence between time series, we should go into multivariate models, but as you mentioned this is out of scope for the moment.

The simplest thing to do is probably to implement item 2 and 3 from my previous comment (ie implement `TabularEstimator` that yields a `TabularPredictor` when trained), while the local model (item 1) we will get for free using [this](https://github.com/awslabs/gluon-ts/blob/13eb1f5396240232ddc6ea04ad37eb8fbf3af150/src/gluonts/model/predictor.py#L376).

Note that the `TabularEstimator` needs to be trainable on a collection of time series: my understanding is that we can just stack the single dataframes obtained from each time series, to get a single one, and give that to Autogluon.

jwmueller:
Implementing a `TabularEstimator` that returns a `TabularPredictor` when trained sounds like a good plan to me. My main concern was simply that we do not commit to an API that later limits us from using AutoGluon in a powerful multivariate fashion, but it seems you've also already considered how we'll later be able to extend this same TabularEstimator+Predictor to multivariate time series modeling, so sounds good to me! My vision is basically to hide (at an API level)  the fact that the current TabularEstimator+Predictor is not utilizing any cross-time-series information, such that we can later use the same API to include cross-time-series info in the AutoGluon models. 

Re: "my understanding is that we can just stack the single dataframes obtained from each time series, to get a single one, and give that to Autogluon."
We can explore this more in the future but I don't find this particularly promising (because it requires AutoGluon to learn a super strong interaction between the time-series ID and every other variable). I think it's better for now to just maintain separate AutoGluon predictors for each time series in the collection (which is what @arianacai1997 has mainly explored so far, even though it's less memory/disk efficient). These per-series AutoGluon predictors would then be wrapped together into a single GluonTS `TabularEstimator/TabularPredictor` object, such that they appear as a single object to users.


So `fit()` (or `train()`) should be called on all time-series simultaneously in this API, as should `predict()`, and internally we can for now then just do separate training of a collection of independent models for each individual series which are stored together as part of a single `Predictor` object.

I envision the GluonTS API might look something like this: 
```
def TabularEstimator.train(timeseries_collection, gluonts_kwargs, autogluon_kwargs):
    predictor = TabularPredictor(gluonts_kwargs)  # initialize GluonTS object to use for inference
    for each timeseries in timeseries_collection:
        series_id = getSeriesName(timeseries,timeseries_collection)  # identifier for this series
        series_df = convertTStoDataFrame(timeseries, gluonts_kwargs)  # pandas Dataframe representation of the series used for AutoGluon-training
        output_directory = autogluon_kwargs['output_directory'] + '/'+series_id  # where model objects are saved on disk
        autogluon_model = autogluon.TabularPrediction.fit(series_df, output_directory=output_directory, autogluon_kwargs)
        predictor.models[series_id] = autogluon_model
    return predictor
```

``` def TabularPredictor.predict(timeseries_collection, gluonts_kwargs, autogluon_kwargs):
        predictions = []  # appropriate matrix/object to store predictions for all time-series
        for each timeseries in timeseries_collection:
            series_id = getSeriesName(timeseries,timeseries_collection)
            series_df = convertTStoDataFrame(timeseries, gluonts_kwargs)
            autogluon_model = predictor.models[series_id]
            predictions[series_id] = autogluon_model.predict(series_df)
    return predictions
```

@lostella @arianacai1997 What do you think?

src/gluonts/nursery/autogluon_for_single_ts.py
arianacai1997: Sorry. I used to name it as Gluonts2Auto. But after several modifications on the code, I have written Gluonts3Auto, Gluonts4Auto...

test/nursery/autogluon_for_single_ts.py
arianacai1997: Do you mean the __init__? I did not see the duplication with other segments thought.

lostella: I mean the files `src/gluonts/nursery/autogluon_for_single_ts.py` and `test/nursery/autogluon_for_single_ts.py`, they appear identical to me

lostella:
> I think it's better for now to just maintain separate AutoGluon predictors for each time series in the collection

@jwmueller I would suggest not doing that directly, since keeping an implicit mapping between models and time series that they refer to can lead to confusion. If one wants to train an estimator on a single time series and get the ad-hoc predictor out of it, one can do it manually: that's not a pattern which is addressed in GluonTS yet, and if it's needed we can implement it in general.

In pseudocode, what I think should be done here is the following:

```python
class TabularPredictor(Predictor):
    def __init__(self, model):
        self.model = model

    def predict(self, dataset):
        # for each time series in dataset
        #     get dataframe with features for the prediction range
        #     use self.model to predict target
        #     yield forecast object


class TabularEstimator(Estimator):
    def train(dataset):
        # for each time series in dataset
        #     get dataframe with features for the time series
        # stack dataframe into a single one
        # train autogluon model
        # return TabularPredictor(model)
```

jwmueller:
@lostella 

>> keeping an implicit mapping between models and time series that they #refer to can lead to confusion.

Not sure why this would lead to confusion for users if it's internally hidden from them? I'm not suggesting the user needs to know anything about this implicit mapping, just the GluonTS predictor internally maintains it via the IDs of each timeseries in the collection (these IDs anyway need to be 100% consistent in the data in order for your proposal with all dataframes stacked together to work).

Regardless I'm happy to explore your proposal (with 1 autogluon-model simultaneously fit for all time series, with the series-ID as a feature in the Dataframe passed to autogluon). The only issues are: 
(1) I think it may lead to lower accuracy (because autogluon needs to learn how to model incredibly strong variable-interactions with this series-ID feature).
(2) @arianacai1997 never experimented with this strategy (to my knowledge), so the results for this remain uncertain, and thus this PR could be wasted effort if the resulting autogluon-model is generally inaccurate.

Do you think we should first benchmark your proposed strategy and then only later revisit the PR once we've established it can produce comparable accuracy to the per-time-series autogluon model strategy? 
Another issue with this plan is (I believe) @arianacai1997 doesn't have much time left on this project.

arianacai1997:
As I have finished running through the whole m4_weekly dataset, here is a list of evaluation results:
- autogluon_lag (autoregression): 
   **MSE: 427940.95127**; 
   wQuantileloss: 0.05947 
- autogluon_lagn: 
   MSE: 403349.02096; 
   wQuantileloss: 0.06259  
- autogluon: 
   MSE: 3238752.41979; 
   wQuantileloss: 0.12463  
- deep ar: 
   MSE: 1424644.78122 
   wQuantileloss: 0.1011 
- mqcnn: 
   MSE: nan 
   **wQuantileloss: 0.03122** 
- simple feed forward: 
   MSE: 1198418.19879 
   wQuantileloss: 0.10706

test/nursery/autogluon_for_single_ts.py
```diff
@@ -0,0 +1,120 @@
...
+from gluonts.model.predictor import Localizer
+from gluonts.model.estimator import Estimator
+
+if TYPE_CHECKING:  # avoid circular import
+    from gluonts.model.estimator import Estimator  # noqa
+
+OutputTransform = Callable[[DataEntry, np.ndarray], np.ndarray]
+
+
+class TabularPredictor(Predictor):
```
lostella, line 29: why is the current file in `test/nursery/`? looks like this is what should go in `src/nursery`

test/nursery/autogluon_for_single_ts.py
```diff
@@ -0,0 +1,120 @@
...
+    def __init__(self,
+                 ag_model,
+                 freq: str,
+                 prediction_length: int,
+                 ) -> None:
```
lostella, line 34: please run
```bash
pip install -U requirements/requirements-setup.txt
./dev_setup.sh
```

and then

```bash
black src test
```

this will format the code and set a commit hook that will check formatting every time you commit

test/nursery/autogluon_for_single_ts.py
```diff
@@ -0,0 +1,120 @@
...
+            year_idx = series.index.year
+            target = series.values
+            cal = calendar()
+            holidays = cal.holidays(start=series.index.min(), end=series.index.max())
+            df = pd.DataFrame(
+                zip(year_idx, month_of_year, day_of_week, hour_of_day, series.index.isin(holidays), target),
+                columns=['year_idx', 'month_of_year', 'day_of_week', 'hour_of_day', 'holiday', 'target'])
+            convert_type = {x: 'category' for x in df.columns.values[:4]}
+            df = df.astype(convert_type)
+            return df
```
lostella, line 95: Is this the same as the one defined in `TabularPredictor`? Then it's worth defining this on its own, and use it twice

test/nursery/autogluon_for_single_ts.py
```diff
@@ -0,0 +1,120 @@
...
+                 ag_model,
+                 freq: str,
+                 prediction_length: int,
+                 ) -> None:
+        self.ag_model = ag_model  # task?
+        self.freq = freq
+        self.prediction_length = prediction_length
+
+    def predict(self, dataset: Iterable[Dict], lag=0, model=None, as_pandas=False) -> Iterator[SampleForecast]:
+        def get_prediction_dataframe(series, lag=0, model=None, as_pandas=False):
```
lostella, line 40: make sure that you annotate all input arguments and return types, wherever possible

test/nursery/autogluon_for_single_ts.py
```diff
@@ -0,0 +1,120 @@
...
+    for i, entry in enumerate(list(dataset.test)):
+        series.append(to_pandas(entry))
+    return series
+
+
+if __name__ == '__main__':
+    from gluonts.dataset.repository.datasets import get_dataset
+
+    dataset = get_dataset("m4_weekly", regenerate=False)
+    series_ls = stack_ts(dataset)
```
lostella, line 117: it doesn't look like this is used, so also `stack_ts` is maybe not needed?

lostella (changes requested):

test/nursery/autogluon_for_single_ts.py
arianacai1997: Yes. I will delete it.

test/nursery/autogluon_for_single_ts.py
arianacai1997: Yes they are the same. Do you mean specify a function outside the classes?

src/gluonts/nursery/autogluon_tabular.py
```diff
@@ -0,0 +1,118 @@
+from gluonts.model.predictor import Predictor
```
lostella, line 1: The header for both files is missing; you can add it by running

```
.devtools/license fix src test
```

and then committing and pushing the resulting changes

test/test_transform.py
```diff
@@ -111,27 +111,6 @@ def aligned_with(date_str, freq):
         )
 
 
-def test_add_method():
```
lostella, line 114: Aren't these changes unrelated to this PR? Could you revert them?

src/gluonts/nursery/autogluon_tabular.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class TabularPredictor(Predictor):
+    def __init__(self, ag_model, freq: str, prediction_length: int,) -> None:
+        self.ag_model = ag_model  # task?
+        self.freq = freq
+        self.prediction_length = prediction_length
+
+    def predict(
+        self, dataset: Iterable[Dict], model=None, as_pandas=False
```
lostella, line 69: What is the purpose of `model` and `as_pandas`? They seem unused, so maybe you can remove them

src/gluonts/nursery/autogluon_tabular.py
```diff
@@ -0,0 +1,118 @@
...
+    ) -> Iterator[SampleForecast]:
+        for idx in range(len(list(dataset))):
+            ts = to_pandas(list(dataset)[idx])
+            df = get_prediction_dataframe(ts)
+            output = self.ag_model.predict(df)
+            yield self.to_forecast(dataset, output, idx)
```
lostella, line 75: You can simplify this as follows

```python
for entry in dataset:
    ts = to_pandas(entry)
    df = get_prediction_dataframe(ts)
    output = self.ag_model.predict(df)
    yield self.to_forecast(output, start_timestamp, entry[FieldName.ITEM_ID])
```

for which you will need to compute `start_timestamp` here, and also import the following

```
```

src/gluonts/nursery/autogluon_tabular.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def to_forecast(
+        self, gluonts_test: Dataset, y_pred, i: int
```
lostella, line 78: Why is a `Dataset` needed here? I think this function should just convert the single output of the tabular task, and convert it to a `Forecast` object. For that, it will need as input arguments the following:
* the output of the tabular task itself
* the timestamp of the forecast start
* the `"item_id"` field (optional, can be `None`)

src/gluonts/nursery/autogluon_tabular.py
```diff
@@ -0,0 +1,118 @@
...
+    ) -> Iterator[SampleForecast]:
+        test = [list(gluonts_test)[i]]
+        forecasts = []
+        t = test[i]
+        start = to_pandas(t).index[-self.prediction_length]
```
lostella, line 83: The start timestamp of the forecast should be the next timestamp after the time series end. There is no need to remove `prediction_length` here, we're not backtesting: we are given a time series, and we want a prediction of its future

test/test_autogluon_tabular.py
```diff
@@ -0,0 +1,80 @@
...
+    predictor = LocalTabularPredictor(freq="W-SUN", prediction_length=1,)
+    forecasts = predictor.predict(dataset)
+
+    # (optional) do evaluation
+    if verbose:
+        evaluator = Evaluator(quantiles=[0.5])
+        tss = change_test(dataset, 10, "W-SUN")
+        agg_metrics, item_metrics = evaluator(iter(tss), iter(list(forecasts)))
+        print(agg_metrics)
+    return list(forecasts)
```
lostella, line 61: There's no need to test evaluation, we need to test that the forecasts are as expected. Here you should iterate the `forecasts` and `assert` at least that:
* their start timestamp is as expected
* their time length is as expected

test/test_autogluon_tabular.py
```diff
@@ -0,0 +1,80 @@
...
+                "start": pd.Timestamp("1750-01-04 00:00:00", freq="W-SUN"),
+                "target": np.array(
+                    [1049.2, 1057.91, 1049.88, 35590.55, 34036.95, 34366.95],
+                ),
+            },
+        ],
+        freq="W-SUN",
+    )
+
+    predictor = LocalTabularPredictor(freq="W-SUN", prediction_length=1,)
```
lostella, line 52: You can try with different `prediction_length` settings here, by looping through different values

test/test_autogluon_tabular.py
```diff
@@ -0,0 +1,80 @@
...
+        index = pd.date_range(
+            start=data["start"], freq=freq, periods=data["target"].shape[-1],
+        )
+        data["ts"] = pd.DataFrame(index=index, data=data["target"].transpose())
+        yield data
+
+
+def change_test(dataset, prediction_length, freq):
+    for data_entry in convert_df(iter(list(dataset)), freq):
+        yield data_entry["ts"][-prediction_length:]
```
lostella, line 76: As a consequence of the above, these are not needed

lostella (changes requested):

src/gluonts/nursery/autogluon_tabular.py
```diff
@@ -0,0 +1,118 @@
...
+        self.task = task
+        self.freq = freq
+        self.prediction_length = prediction_length
+
+    def train(self, training_data: Dataset) -> TabularPredictor:
+        # every time there is only one time series passed
+        # list(training_data)[0] is essentially getting the only time series
+        df = get_prediction_dataframe(
+            to_pandas(list(training_data)[0])[: -self.prediction_length]
+        )
```
lostella, line 112: The whole `training_data` should be used for training here, no slicing.

Also, check my [previous comment](https://github.com/awslabs/gluon-ts/pull/1105#issuecomment-723858720): each time series from the training data should be turned into a dataframe, and these should be stacked into a single one to be passed to `self.task.fit`

src/gluonts/nursery/autogluon_tabular.py
arianacai1997: added model and as_pandas to predict as the parameters. I have added them to the predict. It is described in the document: 
modelstr (optional)

> modelstr (optional)
>The name of the model to get predictions from. Defaults to None, which uses the highest scoring model on the validation set. Valid models are listed in this predictor by calling predictor.get_model_names()
> 
> as_pandasbool (optional)
> Whether to return the output as a pandas Series (True) or numpy array (False)

src/gluonts/nursery/autogluon_tabular.py
arianacai1997: Since we are predicting the future values starting from the next timestamp after the given time series, I think the we should pass the future timestamps into prediction like: `self.ag_model.predict(future_df)`. Is this correct?

src/gluonts/nursery/autogluon_tabular.py
arianacai1997:
```
predictor = LocalTabularPredictor(...)
forecasts = predictor.predict(inference_dataset)
```

> equivalent to:
> for ts in inference_dataset:
>      predictor_single = tabular_estimator.train([ts])
>      yield from predictor_single.predict([ts])

There is a localizer in the for loop. So actually only one time series is passed into train each time. Do you mean I should do a "fake" stacking to prevent the case that a Localizer is not used?

lostella: > Do you mean I should do a "fake" stacking to prevent the case that a Localizer is not used?

Exactly. One can in principle use the estimator with more than one time series to learn a global model, provided that the right features are used: that's not the case now, and so we use the localizer trick, but this way the estimator is multiple-time-series-ready.

src/gluonts/nursery/autogluon_tabular.py
lostella: yes, correct!

src/gluonts/nursery/autogluon_tabular.py
lostella: Understood. However it doesn't seem like `as_pandas` would be useful, right? It's more of an internal detail, since the output of the Autogluon task's `predict` method will be anyway converted into a `SampleForecast` object. Or am I missing something?

arianacai1997: Yes. If we just want `SampleForecast`, these parameters can be deleted.

lostella (changes requested):
There are some changes to `test_transform.py`, can you make sure to revert them? I don't think they are needed, right?

Also the tests are failing because of black, you can fix it with `black src test` and committing the changes.

Other tests are failing because Autogluon is not installed in the workflows: you should add file say `requirements-extras-autogluon.txt` to [this folder](https://github.com/awslabs/gluon-ts/tree/master/requirements) and make sure it's used in the workflows [here](https://github.com/awslabs/gluon-ts/tree/master/.github/workflows), see for example [this line](https://github.com/awslabs/gluon-ts/blob/260fd5cf93d434eeb88369232b4d2275f9a5c0e4/.github/workflows/tests.yml#L30)

dcmaddix:
To fix the style check, run `black gluon-ts/gluon-ts/test/test_autogluon_tabular.py` and `black gluon-ts/src/gluonts/nursery/autogluon_tabular.py` and then update the cr.

dcmaddix (commented):
Add to requirements file and fix style check using black.

lostella (approved):
Looks good! Thanks!

PR #1121, opened 2020-10-30
Added post epoch callback for training

trunghlt:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/mx/trainer/_base.py
+++ b/src/gluonts/mx/trainer/_base.py
@@ -105,6 +105,9 @@ class Trainer:
         initialized network `post_initialize_cb(net)` before the training starts.
         This callback can be used to e.g. overwrite parameters for warm starting, to freeze some
         of the network parameters etc.
+    post_epoch_callback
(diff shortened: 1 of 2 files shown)


src/gluonts/mx/trainer/_base.py
```diff
@@ -126,6 +129,7 @@ def __init__(
             AveragingStrategy, IterationAveragingStrategy
         ] = SelectNBestMean(num_models=1),
         post_initialize_cb: Optional[Callable[[mx.gluon.Block], None]] = None,
+        post_epoch_callback: Optional[Callable[[int, float], None]] = None,
```
lostella, line 132: I think your naming (`_callback`) is better than the one for the callback which was there already (`_cb`). Maybe you could uniform both using the `_callback` naming, as part of this PR?

lostella:
@trunghlt There is ongoing work to provide the `Trainer` class with a richer set of callbacks, see #1168: I think the callback you propose fits there, so feel free to comment there on the various choices if you want to

PR #1122, opened 2020-10-30
Move itertools to dedicated module, add caching utility

lostella:
* Move iterator utilities to a dedicated itertools module
* Remove `shuffle` (not used) introduce `pseudo_shuffle` instead (this is based on the mechanism we're using in the data loader, which is just as general)
* Add the `cache` util that caches an iterable so that the second pass through it is faster. This can be used to cache datasets or transformed datasets (as long as their iteration is deterministic, i.e. no instance splitting or other random transformations).

### DIFF
--- a/src/gluonts/itertools.py
+++ b/src/gluonts/itertools.py
@@ -18,7 +18,7 @@ import random
 T = TypeVar("T")
 
 
-def cycle(it):
+def cyclic(it):
     """Like `itertools.cycle`, but does not store the data."""
 
     while True:
@@ -43,7 +43,7 @@ def batcher(iterable: Iterable[T], batch_size: int) -> Iterator[List[T]]:
     return iter(get_batch, [])
 
 
-class cache(Iterable):
+class cached(Iterable):
     """
     An iterable wrapper, which caches values in a list the first time it is iterated.
 
@@ -69,7 +69,7 @@ class cache(Iterable):
             yield from self.cache
 
 
-def pseudo_shuffle(iterator: Iterator, shuffle_buffer_length: int):
+def pseudo_shuffled(iterator: Iterator, shuffle_buffer_length: int):
     """
     An iterator that yields item from a given iterator in a pseudo-shuffled order.
     """
--- a/src/gluonts/itertools.py
+++ b/src/gluonts/itertools.py
@@ -1,3 +1,16 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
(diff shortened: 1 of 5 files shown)


src/gluonts/itertools.py
```diff
@@ -0,0 +1,115 @@
...
+    it: Iterator[T] = iter(iterable)
+
+    def get_batch():
+        return list(take(it, batch_size))
+
+    # has an empty list so that we have a 2D array for sure
+    return iter(get_batch, [])
+
+
+class _CachedIterable(Iterable):
```
jaheba, line 58: Couldn't we just do `class cache`?

src/gluonts/itertools.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def cycle(it):
+    """Like `itertools.cycle`, but does not store the data."""
+
+    while True:
+        yield from it
+
+
+def take(iterable: Iterable[T], n: int) -> Iterator[T]:
```
jaheba, line 28: Since we now have `toolz` as a dependency, we could just use `toolz.itertoolz.take` instead.

src/gluonts/itertools.py
```diff
@@ -0,0 +1,115 @@
...
+    return _CachedIterable(iterable)
+
+
+class _PseudoShuffledIterator(Iterator):
+    """
+    An iterator that yields item from a wrapped iterator in a pseudo-shuffled order.
+    """
+
+    def __init__(self, iterator: Iterator, shuffle_buffer_length: int):
+        self.shuffle_buffer: List = []
```
jaheba, line 83:
```suggestion
        self.shuffle_buffer: list = []
```

src/gluonts/itertools.py
```diff
@@ -0,0 +1,115 @@
...
+        self.shuffle_buffer_length = shuffle_buffer_length
+        self.iterator = iterator
+
+    def __next__(self):
+        # If the buffer is empty, fill the buffer first.
+        if not self.shuffle_buffer:
+            self.shuffle_buffer = list(
+                itertools.islice(self.iterator, self.shuffle_buffer_length)
```
jaheba, line 91:
```suggestion
                take(self.iterator, self.shuffle_buffer_length)
```

test/test_itertools.py
```diff
@@ -16,15 +16,13 @@
 import pytest
 
 from gluonts.dataset.artificial import constant_dataset
-from gluonts.dataset.loader import PseudoShuffledIterator
+from gluonts.dataset.loader import pseudo_shuffle
```
jaheba, line 19: Why import it from `loader`?

src/gluonts/itertools.py
lostella: we could, I just didn't want to break the sacred tradition of upper-case initials for class names

src/gluonts/itertools.py
lostella: nice!

lostella: actually, what's wrong with diretly using `itertools.islice`?

test/test_itertools.py
lostella: forgot to edit; fixing it

src/gluonts/itertools.py
jaheba: It's less descriptive. Functionally they are the same :)

src/gluonts/itertools.py
jaheba: The exception that proves the rule!

jaheba (approved):

Issue #1106, opened 2020-10-26
Is it possible to add a custom evaluation metric?


kaijennissen:
Is it possible to add a custom evaluation metric?
I am working on an intermittent demand forecast problem and found the [MAAPE](https://www.sciencedirect.com/science/article/pii/S0169207016000121) to be a metric supperior to MAPE and MASE for comparison of point predictions.

PascalIversen:
Hey @kaijennissen,
you can refer to the code @verdimrc provided [here](https://github.com/awslabs/gluon-ts/pull/732/files).
I think for the MAAPE this would be (please double-check the math):
```python
import json
from typing import Dict, Tuple, Union
import numpy as np
import pandas as pd
from gluonts.evaluation import Evaluator
from gluonts.model.forecast import Forecast

from gluonts.dataset.repository.datasets import get_dataset

from gluonts.evaluation.backtest import make_evaluation_predictions
from gluonts.model.simple_feedforward import SimpleFeedForwardEstimator
from gluonts.mx.trainer import Trainer



class MyEvaluator(Evaluator):

    @staticmethod
    def maape(target, forecast):
        denominator = np.abs(target)
        flag = denominator == 0

        maape = np.mean(
            np.arctan((np.abs(target - forecast) * (1 - flag)) / (denominator + flag))
        )
        return maape

    def get_metrics_per_ts(
        self, time_series: Union[pd.Series, pd.DataFrame], forecast: Forecast
    ) -> Dict[str, Union[float, str, None]]:
        metrics = super().get_metrics_per_ts(time_series, forecast)

        pred_target = np.array(self.extract_pred_target(time_series, forecast))
        pred_target = np.ma.masked_invalid(pred_target)
        median_fcst = forecast.quantile(0.5)

        metrics["MAAPE"] = self.maape(
            pred_target, median_fcst
        ) 
        return metrics

    def get_aggregate_metrics(
        self, metric_per_ts: pd.DataFrame
    ) -> Tuple[Dict[str, float], pd.DataFrame]:
        totals, metric_per_ts = super().get_aggregate_metrics(metric_per_ts)

        agg_funs = {
            "MAAPE": "mean",
        }
        assert (
            set(metric_per_ts.columns) >= agg_funs.keys()
        ), "Some of the requested item metrics are missing."
        my_totals = {
            key: metric_per_ts[key].agg(agg) for key, agg in agg_funs.items()
        }

        totals.update(my_totals)
        return totals, metric_per_ts



[...]

PR #1125, opened 2020-10-31
Fixes awslabs/gluon-ts#1106

kaijennissen:
#1106

Added custom_eval_fn argument to Evaluator/MultivariateEvaluator class

### DIFF
--- a/src/gluonts/evaluation/_base.py
+++ b/src/gluonts/evaluation/_base.py
@@ -26,6 +26,7 @@ from typing import (
     Optional,
     Tuple,
     Union,
+    TypedDict,
 )
 
 # Third-party imports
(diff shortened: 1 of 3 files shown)


kaijennissen:
@lostella `pytest test/model/tft/test_model.py` sometimes fails due to a timeout error even on `master`. Is there a way I can fix this?

lostella:
I think you can mark it as skipped, and we can figure it out separately. To do that, add the following decorator to the test that times out:

```python 
@pytest.mark.skip("This times out frequently.")
```

src/gluonts/evaluation/_base.py
```diff
@@ -64,6 +64,16 @@ class Evaluator:
         which is computationally expensive to evaluate and thus slows
         down the evaluation process considerably.
         By default False.
+    Option to include custom evaluation metrics. Expected input is
```
lostella, line 67:
```suggestion
    custom_eval_fn
        Option to include custom evaluation metrics. Expected input is
```

src/gluonts/evaluation/_base.py
```diff
@@ -292,6 +304,26 @@ def get_metrics_per_ts(
...
+                    val = {
+                        k: eval_fn(
+                            pred_target,
+                            eval(
+                                fcst_type + "_fcst",
+                                {
+                                    "median_fcst": median_fcst,
+                                    "mean_fcst": mean_fcst,
+                                },
+                            ),
```
lostella, line 319: Since it's just two values, maybe it is better to use an `if` here. We can then rework this whenever needed

src/gluonts/evaluation/_base.py
```diff
@@ -340,13 +372,23 @@ def get_aggregate_metrics(
             "OWA": "mean",
             "MSIS": "mean",
         }
+
+        if self.custom_eval_fn is not None:
+            for k, (_, agg_type, _) in self.custom_eval_fn.items():
+                try:
+                    custom_agg = {k: agg_type}
```
lostella, line 379: I'm not sure I understand the need for the `try` block here, what could go wrong?

lostella (changes requested):
Hey @kaijennissen thanks for this! Looks good overall, I just have some comments inline

lostella (approved):

Issue #467, opened 2019-11-20
Missing distributions


lostella:
Some of the distributions which are currently missing are:
- [x] [Gamma](https://en.wikipedia.org/wiki/Gamma_distribution)
- [x] [Beta](https://en.wikipedia.org/wiki/Beta_distribution) (this can be implemented ad hoc, or can use Dirichlet internally, since it's a special case)
- [ ] [Inflated Beta](https://arxiv.org/pdf/0705.0700.pdf)

jaheba:
Help wanted?

lostella:
@jaheba indeed

AaronSpieler:
Inflated Beta cannot be used because it has a non continuous PDF.

AaronSpieler:
Everything addressed.

AaronSpieler:
The PDF of the inflated Beta is separable, so the parameters can be optimised independently, which would allow us to implement it as a mixture of distributions.

lostella:
Once #982 is merged, the inflated distributions should be a piece of 🍰

PascalIversen:
Since currently, log_prob returns the pmf for discrete distributions and the pdf for continuous distributions, we can not mix discrete and continuous distributions if they have overlapping support.
We can not implement the `InflatedBeta`  and `InflatedGamma` as a naive `MixtureDistribution` between `Beta`/`Gamma` and the `Deterministic` because the distributions have overlapping support for some parameter ranges. (E.g. the beta distribution has infinite density at zero for alpha=0.5, beta=0.5)

lostella:
@PascalIversen regarding Beta, the authors of https://www.researchgate.net/profile/Silvia_Ferrari/publication/1889063_Inflated_Beta_Distributions/links/09e41505c6e8562e88000000/Inflated-Beta-Distributions.pdf?origin=publication_detail don't seem to treat those cases differently (see eq 2 and fig 1). Maybe you can look into that, although they use a slightly different parametrization.

PascalIversen:
Right, I see. Then the problem is not what I mentioned but rather a inconsistency/bug in the log_prob of the Beta. For alpha, beta < 1 it returns inf for 0 and 1 and not -inf.

lostella:
In any case, eq (2) in the above paper suggests that the mixture density is in fact to be computed by cases, which is what @vafl suggested in #982

PR #1130, opened 2020-11-02
#467 Zero Inflated Poisson Distribution

kaijennissen:
#467

Added Zero Inflated Poisson Distribution


kashif:
@kaijennissen you can use my zero inflated class and port that over, then this general class will allow for ZIP as well as ZINB and other distributions. What do you think? https://github.com/zalandoresearch/pytorch-ts/blob/master/pts/distributions/zero_inflated.py

PascalIversen:
Hey @kaijennissen, thanks for looking into this! I think we could also implement the ZIP it like this in the poisson.py:
```python
class ZeroInflatedPoissonOutput(MixtureDistributionOutput):
    def __init__(self):
        super().__init__([PoissonOutput(), DeterministicOutput(0)])
``` 
I would say defining the distribution explicitly is not necessarily needed as it can be accessed with `ZeroInflatedPoissonOutput().distribution()`. Keeping the test you wrote would still be great. What do you think of this approach?

kaijennissen:
@PascalIversen I'll move the code to `poisson.py` and and also inherit from `MixtureDistributionOutput`.
But I have a question regarding the use of `ZeroInflatedPoissonOutput().distribution()`. Does that mean we do not need a `ZeroInflatedPoisson` class? Or just wrap the class around `ZeroInflatedPoissonOutput().distribution()`?

PascalIversen:
Thanks @kaijennissen. Yes I think we would not need it. The main purpose of the Distribution classes is to define how sampling and the   log likelihood works, and that's implicitly defined trough the MixtureDistribution. So I'd say it is fine to just define the Output class.

lostella:
But it looks like @kaijennissen explicitly redefined the `log_prob` method, apart from extending `MixtureDistribution`: my understanding is that this is needed for the same reasons for which we have this in the case of [inflated beta](https://github.com/awslabs/gluon-ts/blob/13eb1f5396240232ddc6ea04ad37eb8fbf3af150/src/gluonts/mx/distribution/inflated_beta.py#L78)? If so, then the `DistributionOutput` should use `ZeroInflatedPoisson`, right @PascalIversen?

I think what @kashif is proposing is to tackle all zero inflated cases at once, given that there's so much in common. Then it probably makes sense to consolidate Poisson, Beta, NB.

PascalIversen:
The inflated beta distribution is special because the beta is not defined on {0,1}. I don't think the zero inflated poisson will result in any nan gradients, since the poisson distribution supports zeros. (I think I checked that once, I guess it makes sense to also add a test for the gradient at the edge case x=0)

PascalIversen:
No nan gradients:

```python
from gluonts.mx.distribution.mixture import MixtureDistributionOutput
from gluonts.mx.distribution.deterministic import DeterministicOutput
from gluonts.mx.distribution.poisson import PoissonOutput
import mxnet as mx

class ZeroInflatedPoissonOutput(MixtureDistributionOutput):
    def __init__(self):
        super().__init__([PoissonOutput(), DeterministicOutput(0)])
        
zip_distr = ZeroInflatedPoissonOutput().distribution(distr_args=[mx.nd.array([[0.4, 0.6]]), mx.nd.array([3.]), mx.nd.array([0.])])        

inputs = mx.nd.array([0., 1.])
inputs.attach_grad()
with mx.autograd.record():
    result = zip_distr.log_prob(inputs)
result.backward()

print(result)
inputs.grad

[-0.4781732 -2.8176785]
<NDArray 2 @cpu(0)>

[0.05383615 0.675828  ]
<NDArray 2 @cpu(0)>
```



Maybe instead of making a superclass for inflated distributions we could think about changing the `MixtureDistribution`'s log_prob to account for mixtures of distributions with non-overlapping support, like the inflated beta. The problem boils down to:

```python
import mxnet as mx
inputs = mx.nd.array([0., 0.5])
inputs.attach_grad()
with mx.autograd.record():
    result = mx.nd.log(inputs)
    result = mx.nd.where(inputs != 0., result, result.zeros_like())
result.backward()
inputs.grad

[nan  2.]
<NDArray 2 @cpu(0)>
```

I wondered if that is a bug, but torch exhibits the same behaviour:
```python
import torch
inputs = torch.tensor([0., 0.5], requires_grad=True)
result = torch.log(inputs)
result =torch.where(inputs != 0., result, torch.tensor([0., 0.]))
result = result.mean()
result.backward()
inputs.grad

tensor([nan, 1.])
```
but in torch one could use an if else statement to circumvent this.

kaijennissen:
Then I'll also add the inflated NB using the solution suggested by @PascalIversen. For the `InflatedGamma` I would also expect `inf`'s / `nan`'s and therefore use the solution of the `ÌnflatedBeta`.

kaijennissen:
@PascalIversen Wouldn't we expect the gradient of the ZIP to be zero for `x=0`? Since x is not part of the pmf for the zero case.
https://en.wikipedia.org/wiki/Zero-inflated_model

PascalIversen:
x is part of the PMF, it just simplifies since 0!=1. 
Thats the gradient we are searching:
image

which is 
image.

kaijennissen:
Of couse your're right. I just looked at the simplified equation. Than it's good that we talked about this, since the gradient produced by my method is zero (Possibly due to the use of  `Deterministic(0)` inside the mixture?).

PascalIversen:
Interesting, I think it is not because of the Deterministic(0) (I am using that too), but in your implementation changing:
```python
# log_prob of zeros
        zero_log_like = self.zero_probability + (
            1 - self.zero_probability
        ) * F.exp(-self.rate) 
```
to
```python
# log_prob of zeros
        zero_log_like = self.zero_probability + (
            1 - self.zero_probability
        ) * self.poisson_distribution.log_prob(inputs)
```
would yield the same log_prob but the correct gradient :-D.

I am wondering why the maximum likelihood parameter inference test you implemented worked anyway....

test/distribution/test_mx_distribution_inference.py
```diff
@@ -1144,3 +1150,103 @@ def test_genpareto_likelihood(xi: float, beta: float, hybridize: bool) -> None:
...
+@pytest.mark.parametrize("hybridize", [False, True])
+def test_inflated_neg_binomial_likelihood(
+    mu_alpha: float,
+    zero_probability: float,
+    hybridize: bool,
+) -> None:
+    """
+    Test to check that maximizing the likelihood recovers the parameters
+    """
+    mu, alpha = mu_alpha
```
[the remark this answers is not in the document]
kaijennissen: fixed.

test/distribution/test_distribution_sampling.py
```diff
@@ -299,3 +300,109 @@ def test_inflated_beta_sampling(
...
+    )
+
+
+# @pytest.mark.parametrize("mu_alpha", [(5.0, 0.5)])
+# @pytest.mark.parametrize("zero_probability", [0.0, 0.3, 0.9])
+# def test_inflated_neg_binomial_sampling(
+#     mu_alpha: float,
+#     zero_probability: float,
+# ):
+#     mu, alpha = mu_alpha
```
lostella, line 363: @kaijennissen is this test meant to run? If not, it's maybe better to decomment it and `@pytest.mark.skip(...)`-it

Other than that, the PR looks good to me :+1:

kaijennissen: @lostella I was not sure whether this test is necessary or not since the `ZeroInflatedNegativeBinomialOutput` is a subclass of `MixtureDistributionOutput` which has its own tests. What do you think?

lostella: I think it's fine not to have it, but then maybe you could just remove the commented lines (just to avoid confusion, no big deal)

PascalIversen:
The `test_inflated_neg_binomial_likelihood` seems to fail on windows. I actually experienced before that maximum likelihood inference with an inflated negative binomial is quite delicate. Could be that the optimization gets stuck in some local minimum. 
@kaijennissen, could you try using more epochs for the maximum likelihood inference test of the inflated negative binomial?

If this does not help we could think about explicitly testing the log_prob() and its gradients instead.

test/distribution/test_distribution_sampling.py
kaijennissen: Removed.

kashif:
@PascalIversen indeed the ZIB and ZINB type distributions when parametrised by probs are a bit unstable in my experience... on the pytorch side i did it via logits (apart from the gate) but there is fully logit based parametrization now in pyro which could be used but my current version using logits for Poisson or NB is working without issue...

PascalIversen:
Thanks @kashif, interesting point. We could think about basing the NB and maybe Poisson, Binomial on logits. This could stabelize the ZINB inference. (Out of scope of this PR though I think)

test/distribution/test_mx_distribution_inference.py
```diff
@@ -150,6 +153,8 @@ def maximum_likelihood_estimate_sgd(
                 if not hybridize:
                     assert loss.shape == distr.batch_shape
             loss.backward()
+            # import pdb
+            # pdb.set_trace()
```
PascalIversen, line 157: This change was probably not intended? I think after removing this we can merge.

PascalIversen (changes requested):

test/distribution/test_mx_distribution_inference.py
kaijennissen: Removed.

PascalIversen (approved):

PascalIversen:
The tests seem stable now. If that changes or the distributions do not perform well with time series models, we should implement the suggested changes by @kashif for the negative binomial.

Thank you a lot for contributing this @kaijennissen!

PR #1134, opened 2020-11-06
Add a MinMax scaler

pablo2909:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/mx/block/scaler.py
+++ b/src/gluonts/mx/block/scaler.py
@@ -207,12 +207,17 @@ class MinMax(Scaler):
             shape (N, T, C) or (N, C, T) scaled along the specified axis. 
 
         """
+        axis_zero = mx.nd.prod(data==data.zeros_like(), self.axis, keepdims = True) # Along the specified axis, which array are always at zero
+        axis_zero = mx.nd.broadcast_to(axis_zero, shape=data.shape) # Broadcast it to the shape of data
 
-        min_val = data.min(axis = self.axis, keepdims=True)
-        max_val = data.max(axis = self.axis, keepdims=True)
-        scaled_data = (data-min_val)/(max_val-min_val)
+        min_val = mx.nd.where(1-observed_indicator, mx.nd.broadcast_to(data.max(keepdims=True), shape=data.shape), data).min(axis=self.axis, keepdims=True) # return the min value along specified axis while ignoring value according to observed_indicator
+        max_val = mx.nd.where(1-observed_indicator, mx.nd.broadcast_to(data.min(keepdims=True), shape=data.shape), data).max(axis=self.axis, keepdims=True) # return the max value along specified axis while ignoring value according to observed_indicator
 
-        return mx.nd.where(scaled_data!=scaled_data, scaled_data.ones_like(), scaled_data) #Clip the Nan values to one. 
+        scaled_data = (data-min_val)/(max_val-min_val) # Rescale
+        scaled_data = mx.nd.where(axis_zero, scaled_data.zeros_like(), scaled_data) #Clip Nan values to zero if the data was equal to zero along specified axis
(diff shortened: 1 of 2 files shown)


src/gluonts/mx/block/scaler.py
```diff
@@ -174,6 +175,44 @@ def compute_scale(
...
+        Parameters
+        ----------
+        F
+            A module that can either refer to the Symbol API or the NDArray
+            API in MXNet.
+
+        data
+            tensor containing the data to be scaled.
+
+        observed_indicator
```
mbohlkeschneider, line 199: I think we need to take into account the observed indicator. NaN values will be replaced with dummy values (0 by default), which will impact the min/max values. Thus, the target values should be masked along the observed indicators to exclude them.

src/gluonts/mx/block/scaler.py
```diff
@@ -174,6 +175,44 @@ def compute_scale(
...
+        Returns
+        -------
+        Tensor
+            shape (N, T, C) or (N, C, T) scaled along the specified axis. 
+
+        """
+
+        min_val = data.min(axis = self.axis, keepdims=True)
+        max_val = data.max(axis = self.axis, keepdims=True)
+        scaled_data = (data-min_val)/(max_val-min_val)
```
mbohlkeschneider, line 213: What happens if `max_val == min_val` in this case? Will this break?

src/gluonts/mx/block/scaler.py
```diff
@@ -174,6 +175,44 @@ def compute_scale(
...
+
+        return mx.nd.where(scaled_data!=scaled_data, scaled_data.ones_like(), scaled_data) #Clip the Nan values to one.
```
mbohlkeschneider, line 215: I think we should distinguish the cases: all data 0 and all data the same value. Also what does `scaled_data!=scaled_data` do?

mbohlkeschneider (changes requested):
Thank you! Can you add some tests?

src/gluonts/mx/block/scaler.py
pablo2909: In that case it may produce nan values which are handled after by the mx.nd.where(...)

mbohlkeschneider: I see. There seem to be at least three cases that we might want to handle differenty:
1) All NaNs in target
2) All zeroes
3) All the same value.

src/gluonts/mx/block/scaler.py
pablo2909: Ok, I will change so that if all data 0 it returns 0 after scaling.
`scaled_data!=scaled_data` returns a tensor of the same shape as `scaled_data` with 1 at the location of nan values, thus telling us where the nan values are.

mbohlkeschneider: Maybe this warrants a comment because it is not intuitive...

src/gluonts/mx/block/scaler.py
pablo2909: Ok, I will replace the value of the scaled target with 0 where observed_indicator tells us to.

mbohlkeschneider: Not sure we mean the same thing. Consider this case: 

```
target = [1, 2, 1, 0, 3]
observed_indicator = [1, 1, 1, 0, 1] 
```
then, the `min` value should be 1 (not 0) and the `max` value 3 (the target value with NaN will already be replaced with zero earlier in the transformation chain.

mbohlkeschneider (approved):
I now realize that the interface makes no sense for a min-max scaler. We need to actually return the min value and the scale (corresponding to the loc and scale parameters in the `AffineTransformation` to properly use it downstream). I think we need to refactor the `Scaler` to incorporate this properly. I'll open a new issue for it.

pablo2909:
Added MinMax scaler, as is it can't be used in a framework.

Issue #1135, opened 2020-11-06
MASE metric not implemented per definition


arangatang:
## Description
The current implementation of the MASE metric calculation is not true to the definition of the metric. Our implementation is discontinuous and plain wrong in some scenarios. 

For example when forecasting on the constant dataset using some estimator, the current implementation reports a MASE of 0 which is impossible as the naive seasonal forecaster is optimal in this scenario. 

## To Reproduce
(Please provide minimal example of code snippet that reproduces the error. For existing examples, please provide link.)

```python
from gluonts.dataset.repository.datasets import get_dataset
from gluonts.evaluation.backtest import backtest_metrics
from gluonts.model.deepar import DeepAREstimator
from gluonts.mx.trainer import Trainer

ds = get_dataset("constant")
estimator = DeepAREstimator(
    freq=ds.metadata.freq,
    prediction_length=ds.metadata.prediction_length,
    trainer=Trainer(epochs=1),
)
pred = estimator.train(ds.train)
agg, _ = backtest_metrics(ds.test, pred)

print(agg['MASE'])
```

## Error message or code output

```
0.0 # it should be more than or equal to 1
```

arangatang:
I suggest we add two versions: 

1. one which is the MASE metric as per its definition, thus allowing MASE to reach Infinity. 
2. one which is a continuous version of what we have

arangatang:
Closing this as #1409 addresses this.

PR #1136, opened 2020-11-06
Metric flag fixes

arangatang:
*Issue #, if available:*
#1135

*Description of changes:*
Made existing MASE implementation continuous, also added a *vanilla* version of MASE following the definition. 

### DIFF
--- a/src/gluonts/evaluation/_base.py
+++ b/src/gluonts/evaluation/_base.py
@@ -300,6 +300,9 @@ class Evaluator:
             "abs_target_mean": self.abs_target_mean(pred_target),
             "seasonal_error": seasonal_error,
             "MASE": self.mase(pred_target, median_fcst, seasonal_error),
+            "MASE_vanilla": self.vanilla_mase(
+                pred_target, median_fcst, seasonal_error
+            ),
             "MAPE": self.mape(pred_target, median_fcst),
             "sMAPE": self.smape(pred_target, median_fcst),
             "OWA": np.nan,  # by default not calculated
@@ -373,6 +376,7 @@ class Evaluator:
             "abs_target_mean": "mean",
             "seasonal_error": "mean",
             "MASE": "mean",
+            "MASE_vanilla": "mean",
             "MAPE": "mean",
             "sMAPE": "mean",
             "OWA": "mean",
@@ -459,7 +463,7 @@ class Evaluator:
         return np.mean((target < quantile_forecast))
 
     @staticmethod
-    def mase(target, forecast, seasonal_error):
+    def vanilla_mase(target, forecast, seasonal_error):
         r"""
         .. math::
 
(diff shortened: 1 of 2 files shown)


arangatang:
Im going to look into the errors in a little bit.

src/gluonts/evaluation/_base.py
```diff
@@ -458,18 +456,30 @@ def quantile_loss(target, quantile_forecast, q):
...
         .. math::
 
             mase = mean(|Y - Y_hat|) / seasonal_error
 
         https://www.m4.unic.ac.cy/wp-content/uploads/2018/03/M4-Competitors-Guide.pdf
         """
-        flag = seasonal_error <= Evaluator.zero_tol
-        return (np.mean(np.abs(target - forecast)) * (1 - flag)) / (
-            seasonal_error + flag
+        return np.mean(np.abs(target - forecast)) / max(
```
rshyamsundar, line 481: In principle this is correct, but this would blow up the overall `MASE` if one time series in the data set has a very small seasonal error (e.g., constant time series). It does not reflect what we actually want because ideally such time series should be removed before evaluating the whole data set and we usually don’t that explicitly (e.g., `electricity` has one such time series).

The best thing here is to keep the current version of `MASE` as it as and introduce `MASE_strict` and have your version there. Keeping track of both these metrics during evaluations will tell us if the data set has any time series that is “bad” for `MASE` and choose either of them accordingly.

lostella: > It does not reflect what we actually want because ideally such time series should be removed before evaluating the whole data set and we usually don’t that explicitly

That's a good point: those are the cases where your evaluation might be skewed by very simple time series, and also comparisons between models (and based on other metrics) should be looked into with suspicion

arangatang: Alrighty, thank you for the feedback @rshyamsundar. I'll implement this `MASE strict` version then. 
Would it make sense to implement similar alternative metrics for the other metrics which currently are implemented using the flag approach? (i.e. `MAPE_strict`, `sMAPE_strict` etc)

src/gluonts/evaluation/_base.py
```diff
@@ -300,6 +299,9 @@ def get_metrics_per_ts(
             "abs_target_mean": self.abs_target_mean(pred_target),
             "seasonal_error": seasonal_error,
             "MASE": self.mase(pred_target, median_fcst, seasonal_error),
+            "MASE_strict": self.mase_strict(
+                pred_target, median_fcst, seasonal_error
+            ),
```
lostella, line 304: Could we actually have just one implementation of the `mase` method maybe, i.e. the one from the book, and handle any adjustment here? If I'm not wrong, the adjusted MASE we have implemented just makes MASE = 0 whenever the seasonal error is too small, so this could be done here with an `if` statement

arangatang: Sure, this makes sense, however what should we present as MASE then? The custom version or the original one?

arangatang: That should work and i've implemented this in the latest commit

arangatang: Probably we would want to do this to all of the other metrics with similar flag handling?

arangatang:
Closing this as #1409 addresses this.

PR #1137, opened 2020-11-06
Masking edge case fix

benidis:
The missing value imputations methods break if the window is all `nan` values. 

### DIFF
--- a/test/test_transform.py
+++ b/test/test_transform.py
@@ -898,6 +898,7 @@ def test_AddObservedIndicator():
     array_values = [
         np.array([np.nan, 1.0, 1.0, np.nan, 2.0, np.nan, 1.0, np.nan]),
         np.array([np.nan]),
+        np.array([np.nan, np.nan, np.nan, np.nan, np.nan]),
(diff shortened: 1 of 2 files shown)


mbohlkeschneider (commented):
Thank you! Maybe add a test for that edge case?

lostella:
> Maybe add a test for that edge case?

+1, otherwise looks good, thanks for spotting this!

test/test_transform.py
```diff
@@ -898,6 +898,7 @@ def test_AddObservedIndicator():
     array_values = [
         np.array([np.nan, 1.0, 1.0, np.nan, 2.0, np.nan, 1.0, np.nan]),
         np.array([np.nan]),
```
lostella, line 900: So there was a test with all nan, but it also had just one value!

lostella (approved):

test/test_transform.py
benidis: Yeah because having a single missing values is another edge case!

Issue #1133, opened 2020-11-05
GluonPredictor should accept a Dataset as argument to `as_symbol_block_predictor`


lostella:
## Description

[This method](https://github.com/awslabs/gluon-ts/blob/13eb1f5396240232ddc6ea04ad37eb8fbf3af150/src/gluonts/mx/model/predictor.py#L136-L138) is not really easy to use, since coming up with a batch requires applying transformations and so on. The predictor itself knows how to do it, so the method could get a `Dataset` instead (e.g. the one that one wants to get predictions out of) and extract a single batch out of it automatically.

cc @timforby

timforby:
For now I have this quick little hack that __seems__ to work:
```python
    def as_symbol_block_predictor(
        self, dataset: Dataset
    ) -> SymbolBlockPredictor:
        inference_data_loader = InferenceDataLoader(
            dataset,
            transform=self.input_transform,
            batch_size=self.batch_size,
            stack_fn=partial(batchify, ctx=self.ctx, dtype=self.dtype)
        )
        batch = next(iter(inference_data_loader))
        ...
```

lostella:
that's pretty much the change I was thinking about, I think

lostella:
@timforby could you open a PR for this?

PR #1140, opened 2020-11-10
added dataset as argument to as_symbol_block_predictor

timforby:
*Issue #, if available:*  #1133

*Description of changes:*
Added argument `dataset` to all `as_symbol_block_predictor` functions.
Did not want to remove `batch` argument to allow for backward compatibility.

### DIFF
--- a/src/gluonts/model/rotbaum/_predictor.py
+++ b/src/gluonts/model/rotbaum/_predictor.py
@@ -282,7 +282,7 @@ class TreePredictor(GluonPredictor):
             return load_json(fp.read())
 
     def as_symbol_block_predictor(
-        self, batch: DataBatch = None, dataset: Dataset = None
+        self, batch: Optional[DataBatch] = None, dataset: Optional[Dataset] = None
     ) -> "SymbolBlockPredictor":
         return None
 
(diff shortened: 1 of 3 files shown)


src/gluonts/model/rotbaum/_predictor.py
```diff
@@ -282,7 +282,7 @@ def deserialize(
             return load_json(fp.read())
 
     def as_symbol_block_predictor(
-        self, batch: DataBatch
+        self, batch: DataBatch = None, dataset: Dataset = None
```
lostella, line 285: If these are allowed to be `None`, then the proper annotation is `Optional[DataBatch]` and `Optional[Dataset]`

You may need to import `Optional` from `typing` for this

(Similarly elsewhere)

lostella (changes requested):
Looks good to me, thanks for this! I just have a minor comment

src/gluonts/model/rotbaum/_predictor.py
timforby: Oh right! thanks!

lostella (approved):
Thanks, looks great!

PR #1145, opened 2020-11-11
Added quantile function based on cdf bisection

benidis:
- Added quantile function based on cdf bisection.
- Added a test that ensures we get the correct quantiles.

This method is used as the default in the `Distribution` class unless it is overwritten by the corresponding subclasses. It requires a cdf method. 

### DIFF
--- a/src/gluonts/mx/distribution/binned.py
+++ b/src/gluonts/mx/distribution/binned.py
@@ -71,7 +71,7 @@ class Binned(Distribution):
         return getF(self.bin_log_probs)
 
     @property
-    def support(self) -> Tuple[Tensor, Tensor]:
+    def support_min_max(self) -> Tuple[Tensor, Tensor]:
         F = self.F
         return (
             F.broadcast_minimum(
--- a/src/gluonts/mx/distribution/binned.py
+++ b/src/gluonts/mx/distribution/binned.py
@@ -24,7 +24,7 @@ from gluonts.core.component import validated
 from gluonts.model.common import Tensor
 
 # Relative imports
-from .distribution import Distribution, _sample_multiple, getF
+from .distribution import Distribution, _sample_multiple, getF, MAX_SUPPORT_VAL
 from .distribution_output import DistributionOutput
 
 
@@ -69,6 +69,22 @@ class Binned(Distribution):
     @property
     def F(self):
         return getF(self.bin_log_probs)
+
+    @property
+    def support(self) -> Tuple[Tensor, Tensor]:
+        F = self.F
+        return (
+            F.broadcast_minimum(
+                F.zeros(self.batch_shape),
+                F.sign(F.min(self.bin_centers, axis=-1)),
+            )
+            * MAX_SUPPORT_VAL,
+            F.broadcast_maximum(
+                F.zeros(self.batch_shape),
+                F.sign(F.max(self.bin_centers, axis=-1)),
+            )
+            * MAX_SUPPORT_VAL,
+        )
 
     @staticmethod
     def _compute_edges(F, bin_centers: Tensor) -> Tensor:
--- a/src/gluonts/mx/distribution/distribution.py
+++ b/src/gluonts/mx/distribution/distribution.py
@@ -18,6 +18,10 @@ from typing import Any, List, Optional, Tuple
 import mxnet as mx
 import numpy as np
 from mxnet import autograd
+from scipy import optimize
+from itertools import product
+import functools
+import multiprocessing
 
 # First-party imports
 from gluonts.model.common import Tensor
@@ -80,6 +84,10 @@ class Distribution:
 
     arg_names: Tuple
     is_reparameterizable = False
+
+    @property
+    def F(self):
+        raise NotImplementedError()
 
     def log_prob(self, x: Tensor) -> Tensor:
         r"""
@@ -275,6 +283,14 @@ class Distribution:
         """
         raise NotImplementedError()
 
+    def _bisect_cdf(self, level, idx):
+        F = self.F
+
+        def get_quantile(x):
+            return self[idx].cdf(F.array([x])).asnumpy() - level.asnumpy()
+
+        return optimize.bisect(get_quantile, -1e16, 1e16, xtol=1e-6, rtol=1e-6)
+
     def quantile(self, level: Tensor) -> Tensor:
         r"""
 
@@ -296,7 +312,20 @@ class Distribution:
 
             where DISTRIBUTION_SHAPE is the shape of the underlying distribution.
         """
-        raise NotImplementedError()
+
+        F = self.F
+        res = F.empty((len(level), *self.batch_shape))
+        for i, l in enumerate(level):
+            partial_bisect = functools.partial(self._bisect_cdf, l)
+
+            pool = multiprocessing.Pool()
+            out = pool.map(
+                partial_bisect, product(*[range(x) for x in self.batch_shape])
+            )
+
+            res[i, :] = F.array(out).reshape(self.batch_shape)
+
+        return res
 
     def __getitem__(self, item):
         sliced_distr = self.__class__(
--- a/src/gluonts/mx/distribution/distribution.py
+++ b/src/gluonts/mx/distribution/distribution.py
@@ -85,6 +85,10 @@ class Distribution:
     def F(self):
         raise NotImplementedError()
 
+    @property
+    def domain(self) -> Tuple[float, float]:
+        raise NotImplementedError()
+
     def log_prob(self, x: Tensor) -> Tensor:
         r"""
         Compute the log-density of the distribution at `x`.
@@ -302,26 +306,22 @@ class Distribution:
         """
 
         def _tensor_cdf_bisection(
-            level: Tensor, ub=1e16, lb=-1e16, tol=1e-6, max_iter=100
+            level: Tensor,
+            domain_lb: float,
+            domain_ub: float,
+            tol=1e-6,
+            max_iter=100,
         ) -> Tensor:
             r"""
             Returns a Tensor of shape (len(level), *batch_size) with the corresponding quantiles.
             """
-            upper_bound = F.ones((len(level), *self.batch_shape)) * ub
-            lower_bound = F.ones((len(level), *self.batch_shape)) * lb
+            upper_bound = F.ones((len(level), *self.batch_shape)) * domain_ub
+            lower_bound = F.ones((len(level), *self.batch_shape)) * domain_lb
 
             for _ in range(self.all_dim):
                 level = level.expand_dims(axis=-1)
 
-            try:
-                q = F.broadcast_like(
-                    self.mean.expand_dims(axis=0),
-                    level,
-                    lhs_axes=0,
-                    rhs_axes=0,
-                )
-            except:
-                q = F.zeros((len(level), *self.batch_shape)) + 1e-6
+            q = 0.5 * F.broadcast_add(upper_bound, lower_bound)
             val = self.cdf(q) - level
 
             cnt = 0
@@ -346,8 +346,14 @@ class Distribution:
                 cnt += 1
             return q
 
+        try:
+            domain_lb, domain_ub = self.domain
+        except:
+            # default to R+ if not defined
+            domain_lb, domain_ub = (0.0, 1e16)
+
         F = self.F
-        return _tensor_cdf_bisection(level)
+        return _tensor_cdf_bisection(level, domain_lb, domain_ub)
 
     def __getitem__(self, item):
         sliced_distr = self.__class__(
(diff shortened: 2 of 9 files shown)


lostella:
I suspect it would be simpler (and maybe more efficient?) to roll our own vectorized version of bisect, that does all distribution dimensions & quantile levels at the same time. Should require some broadcast magic, but otherwise I don't see many risks given that bisection is very simple.

Thoughts @benidis @jgasthaus @vafl ?

PascalIversen:
Seems like the tests are failing because maximum likelihood inference of some zero inflated distributions which were recently added are unstable. I will address this in another PR.

src/gluonts/mx/distribution/mixture.py
```diff
@@ -77,12 +77,18 @@ def __init__(
                                     the axis)."""
 
             expected_shape = self.batch_shape + (len(components),)
-            assert len(expected_shape) == len(mixture_probs.shape), (
+
+            # fix edge case: if batch_shape == (1,) the mixture_probs shape is squeezed to (k,)
+            # reshape it to (1, k)
+            if self.batch_shape == (1,):
```
PascalIversen, line 83: When does this squeeze happen? Do you want to account for invalid user input for that edge case? I think the documentation is quite clear about the shape and in any case I added the assertion to make sure that the mixture_probs have the correct shape.
For a batch shape of (,1) this is the correct way of defining the mixture_probs:
`mixture_probs=mx.nd.array([[0.1, 0.9]]) `
while this is not:
`mixture_probs=mx.nd.array([0.1, 0.9]) `

Also this is currently in the 
`if not isinstance(mixture_probs, mx.sym.Symbol):`
block, which probably is not intended? The block is there because the assertions do not work with mx.sym.Symbol mode.

benidis: I think this squeeze is happening when we get a subset of the batch_shape distribution and is due to this:
https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/mx/distribution/distribution.py#L43

Assuming a 2D batch_shape, If I cut just a single distribution, i.e., `distr[i, j]`, then the `mixture_probs` will be of shape `(k,)` and not `(1,k)`. I guess when you define this distribution there is no issue but when you cut it there is.

I will move it outside of the block. Good catch, thanks!

benidis: I actually moved the fix to `__getitem__` which is the method that is not consistent. I guess this addresses both of your comments simultaneously.

PascalIversen: I just thought about this way of addressing it, too. Makes sense! 👍

vafl:
> I suspect it would be simpler (and maybe more efficient?) to roll our own vectorized version of bisect, that does all distribution dimensions & quantile levels at the same time. Should require some broadcast magic, but otherwise I don't see many risks given that bisection is very simple.
> 
> Thoughts @benidis @jgasthaus @vafl ?

Yes, I think that would be much faster and it works well with bisect since the accuracy is the same for all dimensions. You could probably even write it as a static mxnet graph.

jgasthaus:
> > I suspect it would be simpler (and maybe more efficient?) to roll our own vectorized version of bisect, that does all distribution dimensions & quantile levels at the same time. Should require some broadcast magic, but otherwise I don't see many risks given that bisection is very simple.
> > Thoughts @benidis @jgasthaus @vafl ?
> 
> Yes, I think that would be much faster and it works well with bisect since the accuracy is the same for all dimensions. You could probably even write it as a static mxnet graph.

Yeah, maybe we can even get the gradients to flow. :)

If this is blocking, maybe we can use the current version only on a specific instance of the mixture distribution and then work on a more generic fast implementation?

benidis:
> > > I suspect it would be simpler (and maybe more efficient?) to roll our own vectorized version of bisect, that does all distribution dimensions & quantile levels at the same time. Should require some broadcast magic, but otherwise I don't see many risks given that bisection is very simple.
> > > Thoughts @benidis @jgasthaus @vafl ?
> > 
> > 
> > Yes, I think that would be much faster and it works well with bisect since the accuracy is the same for all dimensions. You could probably even write it as a static mxnet graph.
> 
> Yeah, maybe we can even get the gradients to flow. :)
> 
> If this is blocking, maybe we can use the current version only on a specific instance of the mixture distribution and then work on a more generic fast implementation?

Fixed this with a new version. I implemented a tensor bisection. It is much faster.

test/distribution/test_default_quantile_method.py
```diff
@@ -0,0 +1,49 @@
...
+# Third-party imports
+import numpy as np
+import mxnet as mx
+
+# First-party imports
+from gluonts.mx.distribution.gaussian import Gaussian
+from gluonts.mx.distribution.mixture import MixtureDistribution
+
+
+def test_quantile() -> None:
```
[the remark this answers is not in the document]
benidis: I did this test because with the mixture of two identical Gaussians we know the quantiles of the mixture which should be the quantiles of the Gaussian components. However, the quantiles of the single gaussian are computed with the quantile method of the Gaussian distribution while the quantiles of the mixture are computed with the cdf bisection method. This is a sanity check that this method gives proper results (or at least that it does not fail).

I will try to add also the tests you suggested.

src/gluonts/mx/distribution/binned.py
```diff
@@ -70,6 +70,22 @@ def __init__(
     def F(self):
         return getF(self.bin_log_probs)
 
+    @property
+    def support(self) -> Tuple[Tensor, Tensor]:
+        F = self.F
+        return (
+            F.broadcast_minimum(
```
vafl, line 77: Naively, I would have thought this is `[min(bin_edges), max(bin_edges)]`. 
Is this due to this last catch-all bin?

src/gluonts/mx/distribution/distribution.py
```diff
@@ -296,7 +307,66 @@ def quantile(self, level: Tensor) -> Tensor:
...
+                    lhs_axes=0,
+                    rhs_axes=0,
+                )
+                lower_bound = F.broadcast_like(
+                    support_ub.expand_dims(axis=0),
+                    level,
+                    lhs_axes=0,
+                    rhs_axes=0,
+                )
+            except:
```
vafl, line 331: Maybe catch only the `NotImplementedError` here.

src/gluonts/mx/distribution/transformed_distribution.py
```diff
@@ -51,6 +51,10 @@ def __init__(
         self._event_shape: Optional[Tuple] = None
         self._batch_shape: Optional[Tuple] = None
 
+    @property
+    def support(self) -> Tuple[Tensor, Tensor]:
+        return self.base_distribution.support
```
vafl, line 56: I think this should be something like:

```python
lb, ub = self.base_distribution.support
for t in self.transforms:
  lb = t.f(lb)
  ub = t.f(ub)
return lb, ub
```

src/gluonts/mx/distribution/binned.py
benidis: Yes exactly.

src/gluonts/mx/distribution/transformed_distribution.py
benidis: This was my original implementation but there was a minor issue with that so I reverted it to the version I have now. 

The issue is that the support in most of the distributions is defined in the "base" distribution part and in theory the affine transformation should be applied to it as well to scale it appropriately. However, in the binned distribution there is effectively no base distribution and the support is defined on the already scaled version. This causes a discrepancy when we have for example a mixture since the final support of distributions that in theory should have the same support does not match since in one case the affine transformation is applied and in the other case not. 

Also, since binned uses this support range for the scaled version, it raises the question if we should actually scale the support or not. For example for a gaussian distribution the support is the real line regardless of the scale, so here I basically make the assumption that the support range that such distributions have should not be scaled since it resembles a "fake infinity". 

For all the above I decided to just propagate the support as is to the transformed distribution, making implicitly the assumption that the limits of the support (e.g. `1e16`) apply to both base and scaled distributions since they are effectively equivalent to the real line. If I was using for example `np.fifo(np.float64).max` as limit, scaling would be effectively useless. I made this assumption but for lower limits.

Note that this assumption is problematic for other cases like uniform (this is the only example I could think of) where the support should be actually scaled since it does not extend to "fake infinity". However, I could not think of a clean solution for everything so I would appreciate feedback here. Should we actually set it to infinity and cut it to a proper value when need to use it?

src/gluonts/mx/distribution/binned.py
```diff
@@ -70,6 +70,22 @@ def __init__(
     def F(self):
         return getF(self.bin_log_probs)
 
+    @property
+    def support(self) -> Tuple[Tensor, Tensor]:
```
lostella, line 74: I think a better name for this would be `support_min_max` or `support_inf_sup`, given that for example [mxnet](https://github.com/apache/incubator-mxnet/blob/66488662ea6d33b8aeb64b632ea45194181f794a/python/mxnet/gluon/probability/distributions/gamma.py#L46) and [pytorch](https://github.com/pytorch/pytorch/blob/eb49dabe925b2fa32897dae771f831cb056aef65/torch/distributions/pareto.py#L48-L50) use `support` to provide the actual support

src/gluonts/mx/distribution/binned.py
vafl: If the mixture approach works well, we could maybe simplify the binned distribution to have finite support and remove this catch-all bin.

benidis: But then the binned would not be standalone... probably we will need to add this as an option.

src/gluonts/mx/distribution/distribution.py
```diff
@@ -296,7 +307,66 @@ def quantile(self, level: Tensor) -> Tensor:
 
             where DISTRIBUTION_SHAPE is the shape of the underlying distribution.
         """
-        raise NotImplementedError()
+
+        def _tensor_cdf_bisection(
+            level: Tensor, tol=1e-6, max_iter=120,
+        ) -> Tensor:
```
lostella, line 313: This can be made a method of the class, maybe

src/gluonts/mx/distribution/transformed_distribution.py
lostella: I'm not sure I understand the issue here, but I think this might be related to https://github.com/awslabs/gluon-ts/issues/1050 (which I think would simplify a few things).

> This causes a discrepancy when we have for example a mixture since the final support of distributions that in theory should have the same support does not match since in one case the affine transformation is applied and in the other case not.

Why should they have the same support?

lostella: @vafl I think it's hard to get the support for transformed variables under general transformations: even if one is interested in the inf and sup of the support, you probably need to restrict yourself to monotonic transformations. Or am I missing something? This is why I believe that focusing on affine transformations might be better

lostella: on the other hand, bijections are fine (as long as they are continuous)

src/gluonts/mx/distribution/transformed_distribution.py
```diff
@@ -51,6 +51,14 @@ def __init__(
         self._event_shape: Optional[Tuple] = None
         self._batch_shape: Optional[Tuple] = None
 
+    @property
+    def support_min_max(self) -> Tuple[Tensor, Tensor]:
+        lb, ub = self.base_distribution.support_min_max
+        for t in self.transforms:
+            lb = t.f(lb)
+            ub = t.f(ub)
```
lostella, line 59: Depending on whether the transformation is increasing or decreasing, one needs to invert upper and lower bound here I think. You can test this by using an affine transformation with negative scaling.

lostella (changes requested):

src/gluonts/mx/distribution/mixture.py
```diff
@@ -77,12 +83,12 @@ def __init__(
...
 
             expected_shape = self.batch_shape + (len(components),)
-            assert len(expected_shape) == len(mixture_probs.shape), (
+            assert len(expected_shape) == len(self.mixture_probs.shape), (
                 assertion_message
                 + " Maybe you need to expand the shape of mixture_probs at the zeroth axis."
             )
             for expected_dim, given_dim in zip(
-                expected_shape, mixture_probs.shape
+                expected_shape, self.mixture_probs.shape
```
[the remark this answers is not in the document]
benidis: Yes this change was needed since the fix in `__getitem__` was initially here. It is irrelevant now I guess.

src/gluonts/mx/distribution/transformed_distribution.py
benidis: True. This assumes increasing, or given that at this point there are only affine transformations, that the scale is positive. Not sure if we can get this info though from the transformation in general unless we know it is affine. The obvious solution is to compare lb and ub after each transformation and do something like this:
```
_lb = t.f(lb)
_ub = t.f(ub)
lb = F.minimum(_lb, _ub)
ub = F.maximum(_lb, _ub)
```
which is transformation agnostic. Any better ideas or is there any particular way you want this done?

lostella (approved):

PR #1153, opened 2020-11-18
introduce functional api for data generation recipes

vafl:
*Issue #, if available:*

*Description of changes:*
Introduce a functional api for data generation recipes. I.e. you can write

```python
stddev = rcp.RandomUniform(low=0, high=1, shape=(1,))
x = rcp.RandomGaussian(stddev=stddev)
evaluate(dict(x=x), length=...)
```

Also add support for multiple return values (currently only used in in a test).

```python
a, b = SomeOperator(...)

evaluate(dict(
  a=a
))
```

This change is fully backwards compatible for now...

### DIFF
--- a/docs/examples/synthetic_data_generation_tutorial/tutorial.md
+++ b/docs/examples/synthetic_data_generation_tutorial/tutorial.md
@@ -1,5 +1,6 @@
 # Synthetic Data Generation Tutorial
 
+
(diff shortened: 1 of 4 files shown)


jgasthaus (approved):
Nice! I think this makes the API simpler and more approachable!

PR #1159, opened 2020-11-20
Fix edge case for `cyclic`

### LINKED ISSUE #1157: Multiprocessing hangs when num_workers > len(dataset)
## Description
I'm trying to serialize a predictor trained on multiple cores. When calling the `serialize` method nothing happens.
Running the same code, but without specifying `num_workers`, it works as expected.

## To Reproduce

```python
from pathlib import Path
from typing import Optional

from gluonts.dataset.multivariate_grouper import MultivariateGrouper
from gluonts.dataset.common import TrainDatasets
from gluonts.model.gpvar import GPVAREstimator


def load_multivariate_dataset(dataset_name: str, target_dim: Optional[int] = None):
    ds = get_dataset(dataset_name)

    if target_dim is None:
        target_dim = len(ds.train)

    grouper = MultivariateGrouper(max_target_dim=target_dim)

    meta = ds.metadata
    meta.feat_static_cat[0].cardinality = target_dim

    return (TrainDatasets(
        metadata=meta,
        train=grouper(ds.train),
        test=grouper(ds.test)
    ), target_dim)


ds, target_dim = load_multivariate_dataset("exchange_rate")
metadata = ds.metadata

estimator = GPVAREstimator(
    prediction_length=metadata.prediction_length,
    freq=metadata.freq,
    target_dim=target_dim,
    trainer=Trainer(
        epochs=2,
        num_batches_per_epoch=10,
        batch_size=8,
    ),
)

predictor = estimator.train(training_data=ds.train, num_workers=2)

predictor.serialize(Path("/tmp"))

```

Nothing happens.

## Environment
- Operating system: Mac OSX 10.15.7
- Python version: 3.6.12
- GluonTS version: 0.6.0
- MXNet version: 1.7.0post1

lostella:
*Issue #, if available:* fixes #1157 

*Description of changes:* `cyclic` would result in an infinite loop when given an empty iterable, causing worker processes to hang when `num_workers > len(dataset)`. This fixes it and adds a test.

### DIFF
--- a/src/gluonts/itertools.py
+++ b/src/gluonts/itertools.py
@@ -21,8 +21,13 @@ T = TypeVar("T")
 def cyclic(it):
     """Like `itertools.cycle`, but does not store the data."""
 
+    at_least_one = False
     while True:
-        yield from it
+        for el in it:
+            at_least_one = True
+            yield el
+        if not at_least_one:
+            break
 
 
 def batcher(iterable: Iterable[T], batch_size: int) -> Iterator[List[T]]:
--- a/test/test_itertools.py
+++ b/test/test_itertools.py
@@ -11,12 +11,22 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-from typing import Iterable
+from typing import Iterable, List
+import itertools
 
 import pytest
 
 from gluonts.dataset.artificial import constant_dataset
-from gluonts.itertools import pseudo_shuffled
+from gluonts.itertools import pseudo_shuffled, cyclic
+
+
+@pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


lostella:
@jaheba I think it would be nice to backport this to 0.6.x: however, this would require backporting #1122, which is slightly breaking. What do you think?

jaheba:
Do understand this right, the issue is that `next(cycle([])` never returns?

jaheba:
How about this:

```python
def cycle(iterable):
    while True:
        it = iter(iterable)
        yield next(it)
        yield from it
```

lostella:
> Do understand this right, the issue is that next(cycle([]) never returns?

Yes, it's actually `next(iter(cyclic([])))`

jaheba (approved):

PR #1161, opened 2020-11-20
Add AffineTransformedDistribution

### LINKED ISSUE #1050: Add AffineTransformedDistribution type
## Description

Currently, `DistributionOutput` classes produce a `TransformedDistribution` when the `distribution` method is invoked with `loc` and `scale`. However, we cannot get mean and variance out of such object:

```python
import mxnet as mx
from gluonts.mx.distribution import GaussianOutput

distr_output = GaussianOutput()

args = mx.nd.array([0.0]), mx.nd.array([1.0])
loc = mx.nd.array([42.0])
scale = mx.nd.array([100.0])
distr = distr_output.distribution(args, loc, scale)

distr.mean  # FAILS
distr.variance  # FAILS
```

One way to solve this is to have a specialized `AffineTransformedDistribution` class where mean and variance (and anything else that can be computed in this case) are exposed.

lostella:
*Issue #, if available:* fixes #1050 

*Description of changes:* adds an AffineTransformedDistribution subtype of TransformedDistribution, which allows computing other things like mean and variance (and potentially CRPS and more). Modify distribution output classes to use this when given loc and scale, and expand tests to cover mean and variance. Fix bugs in some distributions along the way (Dirichlet, DirichletMultinomial, MixtureDistribution).


benidis:
Thanks Lorenzo! Overall seems ok to me. 

One general comment that does not apply directly to this pr but touches the concept of transformed distributions is the hard coded restriction we have for affine transformations. We can always default to this but shouldn't we have the option to select or use a custom one? I have in mind a general normalizing flow for example. It would be great if we were able to define the transformation when we define the `DistributionOutput`. For example, assuming a Gaussian, we can go from this:
```
distr_out=GaussianOutput()  # implicit AffineTransformedDistribution
```
to this:
```
distr_out=GaussianOutput(transform=MyFlow). 
```
Of course any transformation should have a predefined behaviour/api that matches and we can learn its parameters as we learn the parameters of any other projection layer. I am not sure what is the approach that @rshyamsundar followed here for NKF and if this matches...

benidis (dismissed):

lostella:
@benidis that's a valid point, I think we should work on something to allow models to output "more complex" distributions (unless someone has already worked something out there), but maybe wrapping is not the best way to do it. Also, in your example
```
distr_out=GaussianOutput()  # implicit AffineTransformedDistribution
```
I think there is nothing implicit about affine transformations here: when affine transformations enter the game is when calling
```
distr_out.distribution(...)
```

benidis (approved):

PR #1165, opened 2020-11-23
Added pickle_dataset.

jaheba:
*Issue #, if available:*

*Description of changes:*


theadroit-m (approved):
LGTM

For our usecase, we most likely would be generating the the PickledDataset ourself. So the batch size could be different (in terms of 250?). How would a batch size of 250 affect GluonTS?

PR #1169, opened 2020-11-24
include m3 dataset

vafl:
*Issue #, if available:*

*Description of changes:*
Include m3 dataset.

### DIFF
--- a/src/gluonts/dataset/repository/_m3.py
+++ b/src/gluonts/dataset/repository/_m3.py
@@ -1,0 +1,156 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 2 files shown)


src/gluonts/dataset/repository/_m3.py
```diff
@@ -0,0 +1,156 @@
...
+
+def generate_m3_dataset(dataset_path: Path, m3_freq: str):
+    from gluonts.dataset.repository.datasets import default_dataset_path
+
+    m3_xls_path = default_dataset_path / "M3C.xls"
+    if not os.path.exists(m3_xls_path):
+        raise RuntimeError(
+            f"The m3 data is available at https://forecasters.org/resources/time-series-data/m3-competition/ "
+            f"Please download the file and copy the files to this location: {m3_xls_path}"
+        )
```
lostella, line 63: Couldn't the original data be downloaded automatically?

lostella: This is what happens for M4 at least: https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/dataset/repository/_m4.py

vafl: The server is blocking automated requests. We can spoof the user agent, but that seems a bit too dodgy for including it in the package.

vafl: Ideally, we would host the file ourselves, but there is no licence on the website. It is however also included in this gpl3 package: https://github.com/robjhyndman/Mcomp/tree/master/data

lostella (approved):

PR #1175, opened 2020-11-26
introduce quarterly and yearly base frequency

vafl:
and use them for m3 and m4

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/time_feature/lag.py
+++ b/src/gluonts/time_feature/lag.py
@@ -85,9 +85,14 @@ def get_lags_for_frequency(
     # multiple, granularity = get_granularity(freq_str)
     offset = to_offset(freq_str)
 
-    if offset.name == "M":
+    if offset.name == "A-DEC":  # Y
+        lags = []
+    elif offset.name == "Q-DEC":  # Q
+        assert offset.n == 1
+        lags = _make_lags_for_month(offset.n * 3.0)
+    elif offset.name == "M":
         lags = _make_lags_for_month(offset.n)
-    elif offset.name == "W-SUN":
+    elif offset.name == "W-SUN":  # W
         lags = _make_lags_for_week(offset.n)
     elif offset.name == "D":
         lags = _make_lags_for_day(offset.n) + _make_lags_for_week(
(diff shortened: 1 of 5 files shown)


src/gluonts/time_feature/lag.py
```diff
@@ -85,9 +85,14 @@ def _make_lags_for_month(multiple, num_cycles=3):
     # multiple, granularity = get_granularity(freq_str)
     offset = to_offset(freq_str)
 
-    if offset.name == "M":
+    if offset.name == "A-DEC":  # Y
+        lags = []
+    elif offset.name == "Q-DEC":  # Q
+        assert offset.n == 1
```
jaheba, line 91: Can we add a message here?

src/gluonts/time_feature/lag.py
```diff
@@ -85,9 +85,14 @@ def _make_lags_for_month(multiple, num_cycles=3):
     # multiple, granularity = get_granularity(freq_str)
     offset = to_offset(freq_str)
 
-    if offset.name == "M":
+    if offset.name == "A-DEC":  # Y
```
jaheba, line 88: Can we have make the comment a bit more expressive? I assume these are aliases for pandas frequencies?

jaheba (commented):
Should we also add the new frequencies to `get_seasonality` in `time_feature`?

src/gluonts/time_feature/lag.py
vafl: You mean a message saying that we don't use features for years?
I think that is just the correct behavior. There are no useful features for years I think

src/gluonts/time_feature/lag.py
vafl: Yeah its just the normalized frequency string (like we already use for W-SUN)

src/gluonts/time_feature/lag.py
jaheba: Sorry, I meant to the assert statement.

jaheba (approved):

PR #1178, opened 2020-11-27
Remove input_names argument to trainer __call__

lostella:
*Description of changes:* This makes the `Trainer` interface easier to use e.g. for interactive usage.

Note: while this is a breaking change
1. existing estimators in the package are unaffected
2. any custom estimator that simply inherit from `GluonEstimator` (and do not customize how the `Trainer` is used) will also continue to work as before
3. previously trained and serialized predictors will continue to work (no change in how predictors work)

### DIFF
--- a/src/gluonts/model/wavenet/_estimator.py
+++ b/src/gluonts/model/wavenet/_estimator.py
@@ -51,6 +51,7 @@ from gluonts.transform import (
     SetFieldIfNotPresent,
     SimpleTransformation,
     VstackFeatures,
+    SelectFields,
 )
 from gluonts.mx.batchify import batchify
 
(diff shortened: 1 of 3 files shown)


jaheba (dismissed):
I for one welcome this addition.

vafl (dismissed):

src/gluonts/model/wavenet/_estimator.py
```diff
@@ -275,9 +276,18 @@ def train(
...
+            params = self._get_wavenet_args(bin_centers)
+            params.update(pred_length=self.train_window_length)
+            trained_net = WaveNet(**params)
+
+        input_names = get_hybrid_forward_input_names(trained_net)
+
         training_data_loader = TrainDataLoader(
             dataset=training_data,
-            transform=transformation,
+            transform=transformation + SelectFields(input_names),
```
mbohlkeschneider, line 290: This now hides what a batch looks like. I understand that that is the point, but maybe that should be mentioned somehow?

vafl:
LGTM

src/gluonts/mx/trainer/_base.py
```diff
@@ -251,20 +250,18 @@ def loop(
...
+                        for batch_no, batch in enumerate(it, start=1):
                             if self.halt:
                                 break
 
-                            inputs = [data_entry[k] for k in input_names]
-
                             if first_forward:
                                 first_forward = False
-                                _ = net(*inputs)
+                                _ = net(*batch.values())
```
mbohlkeschneider, line 259: We now hide what a batch is, maybe we want to add a line here in the trainer that it is a dictionary with different fields?

mbohlkeschneider (approved):
Looks great. Thank you!

src/gluonts/mx/trainer/_base.py
lostella: yes, makes sense to have a comment for clarity I think

src/gluonts/model/wavenet/_estimator.py
lostella: I'm not sure how this can be clarified though: this was already a bit implicit, and now we just have one more step at this level (instead of in the trainer). Maybe to understand: in what way were things more explicit before?

mbohlkeschneider (approved):

jaheba (approved):

PR #1206, opened 2020-12-08
Simplify data loader, deprecate Trainer's batch_size

lostella:
*Description of changes:*
1. Simplify data loader into a single iterable `DataLoader` type, that does one thing. Legacy `TrainDataLoader`, `ValidationDataLoader`, `InferenceDataLoader` are kept as "alternative constructors" for backward compatibility, they just decorate `DataLoader`'s input/output sequence with very general purpose itertools, in order to get the desired behaviour.
2. Turn some itertools from generators into proper iterable types, so as to make them pickleable, which is needed for multiprocessing.
3. Re-introduce a (optional) `num_batches_per_epoch` argument to the `TrainDataLoader` (removed in #1202) to enable making the data loader of finite length. This allows using this otherwise-infinite-length iterator with other training loops (e.g. PyTorch Lightning) that don't do any slicing on their side, which is usually the case. It's about being general and not making any assumptions on who's going to use this component, that's all.
4. Deprecate the `batch_size` argument to the `Trainer` class. For what it's needed there, `batch_size` can be inferred from the loss shape, otherwise this attribute was simply used from the outside to configure the data loader. This allows using any sequence of batches as training data (as opposed to an object that has the `batch_size` attribute...).
5. Add the `batch_size` attribute and constructor argument to `GluonEstimator` types, and use this to configure data loaders. The default value for this is set to 32, as it was in the `Trainer` class.

Also adding/updating docstrings, documentation sections, and adding more test cases for itertools.

### DIFF
--- a/src/gluonts/dataset/loader.py
+++ b/src/gluonts/dataset/loader.py
@@ -24,7 +24,7 @@ from typing import Callable, Iterable, Iterator, List, Optional
 
 from gluonts.dataset.common import DataBatch, DataEntry, Dataset
 from gluonts.dataset.util import MPWorkerInfo
-from gluonts.itertools import batcher, cyclic, pseudo_shuffled
+from gluonts.itertools import batcher, cyclic, iterable_slice, pseudo_shuffled
 from gluonts.transform import Transformation, TransformedDataset
 
 logger = logging.getLogger(__name__)
@@ -201,6 +201,7 @@ def TrainDataLoader(
     transform: Transformation,
     batch_size: int,
     stack_fn: Callable,
+    num_batches_per_epoch: Optional[int] = None,
     num_workers: Optional[int] = None,
     num_prefetch: Optional[int] = None,
     shuffle_buffer_length: Optional[int] = None,
@@ -216,15 +217,18 @@ def TrainDataLoader(
         if shuffle_buffer_length is not None
         else transformed_dataset
     )
-    return iter(
-        DataLoader(
-            data_iterable=data_iterable,
-            batch_size=batch_size,
-            stack_fn=stack_fn,
-            num_workers=num_workers,
-            num_prefetch=num_prefetch,
-            decode_fn=decode_fn,
-        )
+    data_loader = DataLoader(
+        data_iterable=data_iterable,
+        batch_size=batch_size,
+        stack_fn=stack_fn,
+        num_workers=num_workers,
+        num_prefetch=num_prefetch,
+        decode_fn=decode_fn,
+    )
+    return (
+        iter(data_loader)
+        if num_batches_per_epoch is None
+        else iterable_slice(iter(data_loader), num_batches_per_epoch)
     )
 
 
--- a/src/gluonts/dataset/loader.py
+++ b/src/gluonts/dataset/loader.py
@@ -30,37 +30,14 @@ from gluonts.transform import Transformation, TransformedDataset
 logger = logging.getLogger(__name__)
 
 
-def construct_training_iterator(
-    dataset: Dataset,
-    *,
-    transform: Transformation,
-    shuffle_buffer_length: Optional[int] = None,
-) -> Iterator[DataEntry]:
-    transformed_dataset = TransformedDataset(
-        cyclic(dataset),
-        transform,
-        is_train=True,
-    )
-
-    if shuffle_buffer_length is None:
-        return iter(transformed_dataset)
-    else:
-        return pseudo_shuffled(
-            iter(transformed_dataset),
-            shuffle_buffer_length=shuffle_buffer_length,
-        )
-
-
 class MultiProcessBatcher(Iterator):
     def __init__(
         self,
         dataset: Dataset,
-        transform: Transformation,
         batch_size: int,
         stack_fn: Callable,
         num_workers: int,
         max_queue_size: Optional[int] = None,
-        shuffle_buffer_length: Optional[int] = None,
         decode_fn: Callable = lambda x: x,
     ):
         assert num_workers >= 1
@@ -89,8 +66,6 @@ class MultiProcessBatcher(Iterator):
                     worker_id,
                     num_workers,
                     dataset,
-                    transform,
-                    shuffle_buffer_length,
                     batch_size,
                     stack_fn,
                     self.batch_queue,
@@ -108,8 +83,6 @@ class MultiProcessBatcher(Iterator):
         worker_id: int,
         num_workers: int,
         dataset,
-        transform,
-        shuffle_buffer_length: int,
         batch_size: int,
         stack_fn: Callable,
         batch_queue: mp.Queue,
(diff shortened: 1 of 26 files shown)


jaheba (dismissed):

src/gluonts/dataset/loader.py
```diff
@@ -194,99 +157,110 @@ def win32_guard(num_workers: Optional[int]) -> Optional[int]:
...
+    )
+    data_loader = DataLoader(
+        data_iterable=data_iterable,
+        batch_size=batch_size,
+        stack_fn=stack_fn,
+        num_workers=num_workers,
+        num_prefetch=num_prefetch,
+        decode_fn=decode_fn,
+    )
+    return (
```
vafl, line 228: This looked like a bug at first (I thought you could only iterate once over the `TrainDataLoader`), because it's a bit unexpected that `iterable_slice` is a class. Should we maybe make all classes camel case?

Also one should probably return the `data_loader` instance instead of `iter(data_loader)`?

src/gluonts/mx/model/estimator.py
```diff
@@ -45,9 +45,17 @@ class GluonEstimator(Estimator):
 
     @validated()
     def __init__(
-        self, trainer: Trainer, lead_time: int = 0, dtype: DType = np.float32
+        self,
+        trainer: Trainer,
+        batch_size: int = 32,
```
vafl, line 50: Maybe we can ensure that estimators written by someone else doesn't accidentally calls the super with the old signature and sets the batch_size with the `lead_time` value. Its a bit contrived, but perhaps we can ensure that the caller uses kwargs or something?

vafl (commented):
Looks nice overall! A few minor comments.

I was wondering, are we running automatic performance tests for the data loader?
I remember we had a few regressions before when changing even small things in the iterators that resulted in e.g. accidentally caching the whole dataset in memory.

src/gluonts/dataset/loader.py
lostella: > Should we maybe make all classes camel case?

Yes, why not

> Also one should probably return the data_loader instance instead of iter(data_loader)?

The thing is, that `DataLoader` starts from scratch every time it’s iterated over. While this is fine for validation and prediction, at training time we want it to keep going from where it left instead: with `iter` we’re exactly returning such a “stateful” object.

lostella:
> I was wondering, are we running automatic performance tests for the data loader?

@vafl we had something like that at some point, but it was unstable and would fail at times, of course.

But maybe I can run some benchmarks against e.g. the latest stable release to compare.

src/gluonts/dataset/loader.py
lostella: The change from generators to classes was necessary since otherwise these itertools wouldn’t be pickleable, which would make them useless with multiprocessing when “spawn” is used.

Before we were just passing everything down to the processes, which would then construct the generators themselves, but I think this way it’s much more elegant.

lostella:
After some very quick inspection, time-wise these changes look fine, here are 10 epochs of DeepAREstimator on "electricity", using 4 workers on CPU.

On gluonts==0.6.4:

```
100%|██████████| 50/50 [00:11<00:00,  4.52it/s, epoch=1/10, avg_epoch_loss=6.32]
100%|██████████| 50/50 [00:05<00:00,  8.47it/s, epoch=2/10, avg_epoch_loss=5.87]
100%|██████████| 50/50 [00:05<00:00,  8.34it/s, epoch=3/10, avg_epoch_loss=5.71]
100%|██████████| 50/50 [00:05<00:00,  8.33it/s, epoch=4/10, avg_epoch_loss=5.7]
100%|██████████| 50/50 [00:05<00:00,  8.40it/s, epoch=5/10, avg_epoch_loss=5.54]
100%|██████████| 50/50 [00:05<00:00,  8.43it/s, epoch=6/10, avg_epoch_loss=5.53]
100%|██████████| 50/50 [00:06<00:00,  8.15it/s, epoch=7/10, avg_epoch_loss=5.52]
100%|██████████| 50/50 [00:06<00:00,  8.23it/s, epoch=8/10, avg_epoch_loss=5.46]
100%|██████████| 50/50 [00:05<00:00,  8.34it/s, epoch=9/10, avg_epoch_loss=5.47]
100%|██████████| 50/50 [00:05<00:00,  8.43it/s, epoch=10/10, avg_epoch_loss=5.46]
```

On this branch:

```
100%|██████████| 50/50 [00:12<00:00,  4.10it/s, epoch=1/10, avg_epoch_loss=6.5] 
100%|██████████| 50/50 [00:06<00:00,  8.09it/s, epoch=2/10, avg_epoch_loss=5.86]
100%|██████████| 50/50 [00:06<00:00,  8.27it/s, epoch=3/10, avg_epoch_loss=5.78]
100%|██████████| 50/50 [00:05<00:00,  8.46it/s, epoch=4/10, avg_epoch_loss=5.68]
100%|██████████| 50/50 [00:06<00:00,  7.96it/s, epoch=5/10, avg_epoch_loss=5.56]
100%|██████████| 50/50 [00:05<00:00,  8.37it/s, epoch=6/10, avg_epoch_loss=5.57]
100%|██████████| 50/50 [00:05<00:00,  8.54it/s, epoch=7/10, avg_epoch_loss=5.58]
100%|██████████| 50/50 [00:05<00:00,  8.52it/s, epoch=8/10, avg_epoch_loss=5.49]
100%|██████████| 50/50 [00:05<00:00,  8.54it/s, epoch=9/10, avg_epoch_loss=5.41]
100%|██████████| 50/50 [00:05<00:00,  8.56it/s, epoch=10/10, avg_epoch_loss=5.42]
```

src/gluonts/dataset/loader.py
vafl: > > Should we maybe make all classes camel case?
> 
> Yes, why not
> 
> > Also one should probably return the data_loader instance instead of iter(data_loader)?
> 
> The thing is, that `DataLoader` starts from scratch every time it’s iterated over. While this is fine for validation and prediction, at training time we want it to keep going from where it left instead: with `iter` we’re exactly returning such a “stateful” object.
> 
> So essentially, without `iter` you will not go over the entire dataset in case an epoch is too short to cover it, while `iter` allows it.

Mmh, but this is only the case when `num_batches_per_epoch` is set, right? Otherwise you always loop over the entire dataset. If you return `iter(data_loader)` how do you go over the dataset again in the next epoch? 
(I'm probably missing something)

lostella: > Mmh, but this is only the case when `num_batches_per_epoch` is set, right?

I think currently that has to be set: `num_batches_per_epoch: int = 50`

vafl: Ok, so then I guess we should not have the `if num_batches_per_epoch is None` case.

lostella: > Ok, so then I guess we should not have the `if num_batches_per_epoch is None` case.

Sorry: I meant `num_batches_per_epoch: int = 50` in the `Trainer` class. Here, the sequence is infinite if you don't set `num_batches_per_epoch`, in which case the caller should limit the "chunks" it uses on its side (this is what the `Trainer` does)

vafl (approved):

StanislasGuinel:
I made a PR #1645 to fix the `batch_size` attribute of the `MQCNNEstimator` and `MQRNNEstimator`.

PR #1216, opened 2020-12-10
Use of constrained types instead of asserts.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/core/ty.py
+++ b/src/gluonts/core/ty.py
@@ -100,3 +100,10 @@ def checked(fn):
     fn_wrapper.__checked__ = checked_args
 
     return fn_wrapper
+
+
+PositiveInt = pydantic.PositiveInt
+PositiveInt0 = pydantic.conint(ge=0)
+
+PositiveFloat = pydantic.PositiveFloat
+PositiveFloat0 = pydantic.confloat(ge=0)
--- a/src/gluonts/model/deepar/_estimator.py
+++ b/src/gluonts/model/deepar/_estimator.py
@@ -16,6 +16,7 @@ from typing import List, Optional
 
 import numpy as np
 from mxnet.gluon import HybridBlock
+from pydantic import BaseModel
 
 from gluonts.core import ty
 from gluonts.core.component import DType, validated
@@ -61,7 +62,7 @@ class DropoutcellType(str, Enum):
     VariationalZoneoutCell = "VariationalZoneoutCell"
 
 
-class DeepAREstimator(GluonEstimator):
+class DeepAREstimator(BaseModel, GluonEstimator):
     """
     Construct a DeepAR estimator.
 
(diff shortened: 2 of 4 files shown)


src/gluonts/core/ty.py
```diff
@@ -100,3 +100,10 @@ def fn_wrapper(*args, **kwargs):
     fn_wrapper.__checked__ = checked_args
 
     return fn_wrapper
+
+
+PositiveInt = pydantic.PositiveInt
+PositiveInt0 = pydantic.conint(ge=0)
```
lostella, line 106: `NonnegativeInt` maybe?

jaheba: Let me mark this as a draft :).

lostella:
One advantage of this is that we can get validation in the network constructors too, without duplicating hundreds of lines. This is useful in case one wants to directly handle networks, without going through any estimator class

vafl:
I think we had this at some point (or at least considered doing it). If I remember correctly, we decided against it, because  many IDEs like intellij will show a warning when you pass an integer and don't wrap it in `PositiveInt` explicitly and it didn't really bring that much benefit. We can of course reconsider, but I think the warnings issue would be quite annoying.

jaheba:
> I think we had this at some point (or at least considered doing it). If I remember correctly, we decided against it, because many IDEs like intellij will show a warning when you pass an integer and don't wrap it in `PositiveInt` explicitly and it didn't really bring that much benefit. We can of course reconsider, but I think the warnings issue would be quite annoying.
> 

I really dislike this argument: Some IDE is dictating how to write code, which I feel not comfortable with.

This is a problem of the IDE and not the code.

> Also mypy will probably complain when you pass an `int` to a `PositiveInt` argument.

No, this part works.

vafl:
> I really dislike this argument: Some IDE is dictating how to write code, which I feel not comfortable with.
> This is a problem of the IDE and not the code.

Yeah, I agree its not ideal. I think for pycharm the issue may be this: https://youtrack.jetbrains.com/issue/PY-40569
On the other hand the manual assert is only a bit more verbose. 

>> Also mypy will probably complain when you pass an int to a PositiveInt argument.

> No, this part works.

Ah, cool. Is that because of the `validated` wrapper magic?

jaheba:
> Ah, cool. Is that because of the validated wrapper magic?

There are stub-files (ending in `.pyi`), which take preference over the `.py` file when type-checking:

```
PositiveInt = int
PositiveInt0 = int

PositiveFloat = float
PositiveFloat0 = float
```

That way, mypy assumes that for example `PositiveInt` is just `int`.

vafl:
Ok, nice. Maybe that works with pycharm, too then. I think we tried with some more naive option.

jaheba:
> Ok, nice. Maybe that works with pycharm, too then. I think we tried with some more naive option.

According to the documentation, there is support for stub files in pycharm: https://www.jetbrains.com/help/pycharm/stubs.html

src/gluonts/model/deepar/_estimator.py
```diff
@@ -127,118 +137,63 @@ class DeepAREstimator(GluonEstimator):
...
-            if lags_seq is not None
-            else get_lags_for_frequency(freq_str=freq)
-        )
-        self.time_features = (
-            time_features
-            if time_features is not None
-            else time_features_from_frequency_str(self.freq)
-        )
+    def __init__(self, **fields):
+        BaseModel.__init__(self, **fields)
```
lostella, line 169: Can one pass anything here to the constructor? Or just the fields declared in the class?

jaheba: You could pass anything there, but only the fields which are declared are handled.

PR #1221, opened 2020-12-15
DeepState: Compute log-density exactly when scaling transformation is…

rshyamsundar:
… used.

*Issue #, if available:*

*Description of changes:*
Log-density for deep state model is off by a constant factor when scaling is used. Other models don't have such problems. Fixing it so that log-density on test data can be used as an evaluation metric.

### DIFF
--- a/src/gluonts/mx/distribution/lds.py
+++ b/src/gluonts/mx/distribution/lds.py
@@ -229,6 +229,12 @@ class LDS(Distribution):
         # TODO: Based on form of the prior decide to do either filtering
         #   or residual-sum-of-squares
         log_p, final_mean, final_cov = self.kalman_filter(x, observed)
+        if scale is not None:
+            # log_abs_det_jac: sum over all the output dimensions.
+            ladj = self.F.sum(self.F.log(scale), axis=-1, keepdims=True)
+
+            # subtract `ladj` for every time step.
+            log_p = self.F.broadcast_sub(log_p, ladj)
         return log_p, final_mean, final_cov
 
     def kalman_filter(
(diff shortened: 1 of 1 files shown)


src/gluonts/mx/distribution/lds.py
```diff
@@ -229,6 +229,13 @@ def log_prob(
         # TODO: Based on form of the prior decide to do either filtering
         #   or residual-sum-of-squares
         log_p, final_mean, final_cov = self.kalman_filter(x, observed)
+        if scale is not None:
+            F = self.F
+            # log_abs_det_jac: sum over all output dimensions.
+            ladj = -F.sum(F.log(F.abs(scale)), axis=-1, keepdims=True)
```
lostella, line 235: The fact that we simply sum over the output dimensions implies that the scaling is diagonal, right?

I would also ask you to double check the minus here (this agrees with `TransformedDistribution`, and you've probably done this so many times, but just to be sure)

lostella (approved):

src/gluonts/mx/distribution/lds.py
rshyamsundar: Yes (about scale being diagonal), in the preceding statement the target is divided by scales independently. The scaling transformation and the Jacobian adjustment are done locally there.

About the minus sign: yes, I double-checked. I always imagine this simple example of uniform distribution in [0, 100] where the log-density is `- log 100` and if you divide by 100, the transformed log-density is 0. So we need to subtract `log 100` to get the original.

lostella: That's a good trick to remember it!

Issue #1200, opened 2020-12-04
How to use SeasonalityISSM in the DeepState Model?


YangZeqin:
I want to use the "m4_hourly" dataset to learn the DeepState Model,but when I set "issm = SeasonalityISSM(7)", I'm having the following problem.

Could you tell me how to solve it?

```python
from gluonts.dataset.repository.datasets import get_dataset
from gluonts.model.deepstate import DeepStateEstimator
from gluonts.model.deepstate.issm import SeasonalityISSM
from gluonts.mx.trainer import Trainer

def load_dataset():
    dataset = get_dataset("m4_hourly",regenerate=False)
    return dataset

def create_estimator(dataset):
    estimator = DeepStateEstimator(
        cardinality=[414],
        freq = dataset.metadata.freq,
        prediction_length = dataset.metadata.prediction_length,
        issm = SeasonalityISSM(7),
        trainer = Trainer(
            ctx = "cpu",
            epochs = 1,
            learning_rate = 1e-3,
            num_batches_per_epoch = 100,
            hybridize=False,
        )
    )

    return estimator

if __name__ == "__main__":
    dataset = load_dataset()
    estimator = create_estimator(dataset)
    predictor = estimator.train(dataset.train)
```

```
Traceback (most recent call last):
  File "D:/python/DeepSSM/deep ISSM.py", line 53, in <module>
    predictor = estimator.train(dataset.train)
  File "D:\python\DeepSSM\gluonts\model\estimator.py", line 254, in train
    training_data, validation_data, num_workers, num_prefetch, **kwargs
  File "D:\python\DeepSSM\gluonts\model\estimator.py", line 233, in train_model
    validation_iter=validation_data_loader,
  File "D:\python\DeepSSM\gluonts\trainer\_base.py", line 308, in __call__
    epoch_loss = loop(epoch_no, train_iter)
  File "D:\python\DeepSSM\gluonts\trainer\_base.py", line 242, in loop
    output = net(*inputs)
  File "C:\Anaconda\envs\time_series\lib\site-packages\mxnet\gluon\block.py", line 540, in __call__
    out = self.forward(*args)
  File "C:\Anaconda\envs\time_series\lib\site-packages\mxnet\gluon\block.py", line 917, in forward
    return self.hybrid_forward(ndarray, x, *args, **params)
  File "D:\python\DeepSSM\gluonts\model\deepstate\_network.py", line 193, in hybrid_forward
    length=self.past_length,
[...]

PR #1223, opened 2020-12-16
Improve ISSM interface

lostella:
*Issue #, if available:* addresses #1200 

*Description of changes:* Using custom ISSMs in DeepState (i.e. other than the default one for the given frequency), does not work because the time features from which coefficients are constructed always fall back to the default ones. This PR changes this, by coupling each ISSM with the time feature that is needed to derive its coefficients.

Tests are added for the ISSM classes, which were not there before.

### DIFF
--- a/test/model/deepstate/test_issm.py
+++ b/test/model/deepstate/test_issm.py
@@ -184,8 +184,8 @@ def test_seasonality_issm_h():
 
     season_indices = [[0] * 10, [0] * 2 + [1] * 8]
 
-    for item in range(1, 2):
-        for time in range(2, 10):
+    for item in range(2):
+        for time in range(10):
             season_indicator = mx.nd.one_hot(
                 mx.nd.array([season_indices[item][time]]), 12
             )
@@ -253,8 +253,8 @@ def test_composite_issm_h():
 
     season_indices_moy = [[0] * 10, [0] * 2 + [1] * 8]
 
-    for item in range(1, 2):
-        for time in range(2, 10):
+    for item in range(2):
+        for time in range(10):
             expected_coeff = mx.nd.concat(
                 mx.nd.ones((1, 1)),
                 mx.nd.one_hot(
@@ -328,8 +328,8 @@ def test_composite_issm_h_default():
 
     season_indices_dow = [[2] * 3 + [3] * 7, [3] * 10]
 
-    for item in range(1, 2):
-        for time in range(2, 10):
+    for item in range(2):
+        for time in range(10):
             expected_coeff = mx.nd.concat(
                 mx.nd.ones((1, 2)),
                 mx.nd.one_hot(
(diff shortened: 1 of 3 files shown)


test/model/deepstate/test_issm.py
```diff
@@ -0,0 +1,269 @@
...
+        10,
+        issm.latent_dim(),
+        issm.latent_dim(),
+    )
+    assert innovation_coeff.shape == (2, 10, issm.latent_dim())
+
+
+def test_seasonality_issm_h():
+    issm = SeasonalityISSM(
+        num_seasons=12, time_feature=MonthOfYear(normalized=False)
```
lostella, line 130: The fact that `normalized=False` must be passed here is a bit annoying; as an alternative, we could either have alternative constructors that don't allow for normalization (eg `MonthOfYearIndex`) or just pass the feature *type* here, and construct the object internally.

Thoughts @rshyamsundar?

rshyamsundar (approved):

test/model/deepstate/test_issm.py
rshyamsundar: yea.. passing only type and constructing the object internally as you did in the latest revision is better...

test/model/deepstate/test_issm.py
```diff
@@ -0,0 +1,369 @@
...
+        )
+    )
+
+    emission_coeff, transition_coeff, innovation_coeff = issm.get_issm_coeff(
+        features
+    )
+
+    season_indices = [[0] * 10, [0] * 2 + [1] * 8]
+
+    for item in range(1, 2):
```
rshyamsundar, line 187: You are checking here for only one item?

test/model/deepstate/test_issm.py
```diff
@@ -0,0 +1,369 @@
...
+        for time in range(2, 10):
+            season_indicator = mx.nd.one_hot(
+                mx.nd.array([season_indices[item][time]]), 12
```
rshyamsundar, line 190: Maybe I am missing something: `mx.nd.array([season_indices[item][time]])` is always `[1]` for this range of `item` and `time` values and you always create this one_hot array: `[[0. 1. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]]`?

rshyamsundar:
Also maybe can you run `benchmark_m4.py` with deep state estimator? Just to see there are no serious differences..

test/model/deepstate/test_issm.py
lostella: right, that was for debugging, I'll revert it

test/model/deepstate/test_issm.py
lostella: on `item=0` the season index is always `0`, on `item=1` the season index changes from `0` to `1` after two time steps

rshyamsundar (approved):

lostella:
@rshyamsundar 

```
      dataset  mean_wQuantileLoss     sMAPE
0   m4_hourly            0.040880  0.134823
1    m4_daily            0.028031  0.036918
2   m4_weekly            0.046093  0.077881
3  m4_monthly            0.117584  0.142290
```

PR #1224, opened 2020-12-16
Normalizing Kalman Filter (local/univariate version)

rshyamsundar:
*Issue #, if available:*

*Description of changes:*
Adding Normalizing Kalman Filter (univariate version). This is done by including additional `output_transform` argument in `DeepStateEstimator`.

Currently updated only the smoke test to verify these changes. Will add more tests after getting feedback from Emmanuel.

### DIFF
--- a/src/gluonts/model/deepstate/_estimator.py
+++ b/src/gluonts/model/deepstate/_estimator.py
@@ -22,6 +22,7 @@ from gluonts.dataset.field_names import FieldName
 from gluonts.model.deepstate.issm import ISSM, CompositeISSM
 from gluonts.mx.model.estimator import GluonEstimator
 from gluonts.model.predictor import Predictor
+from gluonts.mx.distribution.bijection import ComposedBijectionHybridBlock
 from gluonts.mx.distribution.lds import ParameterBounds
 from gluonts.mx.model.predictor import RepresentableBlockPredictor
 from gluonts.mx.trainer import Trainer
@@ -88,6 +89,10 @@ class DeepStateEstimator(GluonEstimator):
     add_trend
         Flag to indicate whether to include trend component in the
         state space model
+    output_transform
+        Specifies the transformation to be applied to the target time series.
+        This is a bijection with learnable parameters (e.g., a normalizing flow) and should be an instance of
+        `ComposedBijectionHybridBlock`.
     past_length
         This is the length of the training time series;
         i.e., number of steps to unroll the RNN for before computing
@@ -154,6 +159,7 @@ class DeepStateEstimator(GluonEstimator):
         prediction_length: int,
         cardinality: List[int],
         add_trend: bool = False,
+        output_transform: Optional[ComposedBijectionHybridBlock] = None,
         past_length: Optional[int] = None,
         num_periods_to_train: int = 4,
         trainer: Trainer = Trainer(
@@ -210,6 +216,7 @@ class DeepStateEstimator(GluonEstimator):
         )
         self.prediction_length = prediction_length
         self.add_trend = add_trend
+        self.output_transform = output_transform
         self.num_layers = num_layers
         self.num_cells = num_cells
         self.cell_type = cell_type
@@ -321,6 +328,7 @@ class DeepStateEstimator(GluonEstimator):
             num_layers=self.num_layers,
             num_cells=self.num_cells,
             cell_type=self.cell_type,
+            output_transform=self.output_transform,
             past_length=self.past_length,
             prediction_length=self.prediction_length,
             issm=self.issm,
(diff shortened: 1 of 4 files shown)


src/gluonts/mx/distribution/lds.py
```diff
@@ -211,33 +213,105 @@ def log_prob(
...
+
+                # We need the `ladj` of the inverse transform but we compute `ladj` of the forward transform which only
+                # differs by sign. Shape (batch_size, seq_length)
+                ladj = -output_transform.log_abs_det_jac(x, x_transformed)
+            return x_transformed, ladj
+
+        x_scaled, ladj_scaling = _scale(x)
+        x_scaled_transformed, ladj_transform = _output_transform(x_scaled)
+
+        if scale is not None and output_transform is not None:
```
benidis, line 303: Minor comment:  Here you could return `ladj_scaling` and  `ladj_transform` as `F.zeros_like()` instead of `None` so you can add them. This would make the if else more compact to something like
```
ladj = F.broadcast_add(ladj_scaling, ladj_transform) if scale is not None or output_transform is not None else None
```
 I don't have a strong opinion on what is the best approach though.

benidis (commented):
Thanks Syama, looks nice overall. It was fairly straightforward so I just included a minor comment that you can safely ignore.

src/gluonts/mx/distribution/lds.py
```diff
@@ -211,33 +213,105 @@ def log_prob(
...
+        F = self.F
+
+        def _scale(x):
+            x_scaled, ladj = x, None
+            if scale is not None:
+                # We apply scaling before transforming the target by the given output transformation.
+                x_scaled = F.broadcast_div(x, scale.expand_dims(axis=1))
+
+                # log_abs_det_jac: sum over all output dimensions. Shape (batch_size, 1)
+                ladj = -F.sum(F.log(F.abs(scale)), axis=-1, keepdims=True)
```
emited, line 287: This could probably be done directly by using the AffineTransformation bijection which would handle the scale directly. One could also use a ComposedBijection object to compose the scale and the output_transform bijections.

src/gluonts/mx/distribution/lds.py
```diff
@@ -211,33 +213,105 @@ def log_prob(
...
+                # We need the `ladj` of the inverse transform but we compute `ladj` of the forward transform which only
+                # differs by sign. Shape (batch_size, seq_length)
+                ladj = -output_transform.log_abs_det_jac(x, x_transformed)
+            return x_transformed, ladj
+
+        x_scaled, ladj_scaling = _scale(x)
+        x_scaled_transformed, ladj_transform = _output_transform(x_scaled)
+
+        if scale is not None and output_transform is not None:
+            # Sum `ladj_scaling` over all time steps.
```
emited, line 304: this code would no longer be useful using the composed transforms, see comment above

src/gluonts/model/deepstate/_estimator.py
```diff
@@ -154,6 +159,7 @@ def __init__(
         prediction_length: int,
         cardinality: List[int],
         add_trend: bool = False,
+        output_transform: Optional[ComposedBijectionHybridBlock] = None,
```
emited, line 162: As the normalizing flow bijections are inherently multivariate, for the Local-NKF, one would need to wrap these in a LocalBijection class which applies a local flow to each observations implicitly, and Jacobian term is handled appropriately (this has already been coded and should be added in this PR, I can share this with you)

lostella: I would use the most general annotation for this argument, if possible: couldn't one use a generic `Bijection` here?

src/gluonts/model/deepstate/_network.py
```diff
@@ -31,6 +32,7 @@ def __init__(
         num_layers: int,
         num_cells: int,
         cell_type: str,
+        output_transform: Optional[ComposedBijectionHybridBlock],
```
lostella, line 35: Same as for the `DeepStateEstimator`: couldn't one use whatever `Bijection` here in principle?

src/gluonts/mx/distribution/lds.py
```diff
@@ -198,6 +199,7 @@ def log_prob(
         self,
         x: Tensor,
         scale: Optional[Tensor] = None,
+        output_transform: Optional[ComposedBijectionHybridBlock] = None,
```
lostella, line 202: Same as for the `DeepStateEstimator`: couldn't one use whatever `Bijection` here in principle?

lostella (commented):
Would it make sense to transform `LDS` from the outside, instead of passing in the transformation?

kashif:
@rshyamsundar would you have the `CRPS-SUM` test set metrics for this method on the datasets you considered somewhere for me? Thanks in advance.

Issue #1211, opened 2020-12-09
ZeroAndOneInflatedBetaOutput: module 'mxnet.symbol' has no attribute 'logical_or'


deejlucas:
I originally thought this was a typo, but I tried to make the change to np.logical_or locally, and that did not fix the problem. Please let me know if more information would be helpful

In gluonts 0.6.3, I am submitting the following:
```
estimator_4 = deepar.DeepAREstimator(
    freq="W", prediction_length=4, trainer=trainer_1, distr_output=ZeroAndOneInflatedBetaOutput(),
    context_length=26, num_cells=104, dropout_rate=0.1)
predictor_4 = estimator_4.train(training_data=data_train_beta)
```
This produces the error referenced in the title with the following stack trace:
```
  0%|          | 0/50 [00:00<?, ?it/s]
learning rate from ``lr_scheduler`` has been overwritten by ``learning_rate`` in optimizer.
  0%|          | 0/50 [00:01<?, ?it/s]
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-24-94b4fecb46f1> in <module>
      2     freq="W", prediction_length=4, trainer=trainer_1, distr_output=ZeroAndOneInflatedBetaOutput(),
      3     context_length=26, num_cells=104, dropout_rate=0.1)
----> 4 predictor_4 = estimator_4.train(training_data=data_train_beta)

~/.conda/envs/cuda_test/lib/python3.6/site-packages/gluonts/model/estimator.py in train(self, training_data, validation_data, num_workers, num_prefetch, shuffle_buffer_length, **kwargs)
    274             num_prefetch,
    275             shuffle_buffer_length,
--> 276             **kwargs,
    277         ).predictor

~/.conda/envs/cuda_test/lib/python3.6/site-packages/gluonts/model/estimator.py in train_model(self, training_data, validation_data, num_workers, num_prefetch, shuffle_buffer_length, **kwargs)
    247             input_names=get_hybrid_forward_input_names(trained_net),
    248             train_iter=training_data_loader,
--> 249             validation_iter=validation_data_loader,
    250         )
    251 

~/.conda/envs/cuda_test/lib/python3.6/site-packages/gluonts/mx/trainer/_base.py in __call__(self, net, input_names, train_iter, validation_iter)
    343                     )
    344 
--> 345                     epoch_loss = loop(epoch_no, train_iter)
    346                     if is_validation_available:
    347                         epoch_loss = loop(

[...]

PR #1226, opened 2020-12-17
Use broadcast_logical_or in inflated_beta

deejlucas:
*Issue #, if available:* addresses [#1211](https://github.com/awslabs/gluon-ts/issues/1211)

*Description of changes:* The various inflated beta distribution classes used `F.logical_or` to mask ones and zeros (for which the loss function is undefined), but there is no `logical_or` method for `mxnet.symbol`.

We now use `F.broadcast_logical_or` which exists as a method of both `mxnet.ndarray` and `mxnet.symbol`, either of which can be imported as `F`.

### DIFF
--- a/test/distribution/test_inflated_beta.py
+++ b/test/distribution/test_inflated_beta.py
@@ -1,0 +1,52 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+import pandas as pd
+import pytest
+
+from gluonts.dataset import common
+from gluonts.model import deepar
+from gluonts.mx.distribution.inflated_beta import ZeroAndOneInflatedBetaOutput
+from gluonts.mx.trainer import Trainer
+
+
+@pytest.mark.parametrize("hybridize", [False, True])
+def test_symbol_and_array(hybridize: bool):
+    # Tests for cases like the one presented in issue 1211, in which the Inflated
+    # Beta outputs used a method only available to arrays and not to symbols.
+    # We simply go through a short training to ensure no exceptions are raised.
+    data = [
+        {
+            "target": [0, 0.0460043, 0.263906, 0.4103112, 1],
+            "start": pd.to_datetime("1999-01-04"),
+        },
+        {
+            "target": [1, 0.65815564, 0.44982578, 0.58875054, 0],
+            "start": pd.to_datetime("1999-01-04"),
+        },
+    ]
+    dataset = common.ListDataset(data, freq="W-MON", one_dim_target=True)
+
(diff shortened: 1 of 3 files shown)


lostella:
Nice! Did you verify that this works? You could also add a test, that just sets up an estimator with the inflated distribution output, and trains it for one batch on some dummy data in [0, 1], with hybridize both True/False

deejlucas:
@lostella I checked that it works. I'll add a test too and push another commit

deejlucas:
@lostella I had my dev setup wrong when I made my earlier commits, so not all checks were being made. Now that I've corrected this, I'm getting the following:
```
src/gluonts/model/deepstate/issm.py:109: error: NotImplemented? not callable
src/gluonts/model/deepstate/issm.py:112: error: NotImplemented? not callable
src/gluonts/model/deepstate/issm.py:115: error: NotImplemented? not callable
src/gluonts/model/deepstate/issm.py:118: error: NotImplemented? not callable
src/gluonts/model/deepstate/issm.py:121: error: NotImplemented? not callable
Found 5 errors in 1 file (checked 204 source files)
```

It would be easy enough to replace these with `NotImplementedError`, but it looks like they've been this way for a while, so I'm wondering if I have something configured incorrectly here. I've changed them for now, just so that I can get the commit in, but please let me know if they should be reverted.

test/distribution/test_inflated_beta.py
```diff
@@ -0,0 +1,50 @@
...
+            "target": [1, 0.65815564, 0.44982578, 0.58875054, 0],
+            "start": pd.to_datetime("1999-01-04"),
+        },
+    ]
+    data = common.ListDataset(data, freq="W-MON", one_dim_target=True)
+
+    trainer_hybrid = Trainer(ctx=mx.context.gpu(), epochs=1, hybridize=True)
+
+    trainer_not_hybrid = Trainer(
+        ctx=mx.context.gpu(), epochs=1, hybridize=False
```
lostella, line 29: I would set the context to `cpu` here (or better leave it to default) since that's the only one available on the instances running the tests on GitHub I think.

You can also set `num_batches_per_epoch=2` to make the test very fast. I think the point here is to just to have a smoke test (just check that it runs).

test/distribution/test_inflated_beta.py
```diff
@@ -0,0 +1,50 @@
+import mxnet as mx
+import pandas as pd
+
+from gluonts.dataset import common
+from gluonts.model import deepar
+from gluonts.mx.distribution.inflated_beta import ZeroAndOneInflatedBetaOutput
+from gluonts.mx.trainer import Trainer
+
+
+def test_symbol_and_array():
```
lostella, line 10: You could have a parameter `hybridize: bool` here, and make two test cases with

```python
import pytest

@pytest.mark.parametrize("hybridize", [False, True])
def test_symbol_and_array(hybridize: bool):
    ...
```

and further below only construct 1 trainer/estimator, using this argument to set hybridization on/off

test/distribution/test_inflated_beta.py
```diff
@@ -0,0 +1,50 @@
+import mxnet as mx
```
lostella, line 1: The header with license notice is missing, you can fix it by running

```
.devtools/license fix test
```

which should only affect the present file (if everything else within `test` has the header)

lostella (changes requested):
@deejlucas thank you! I have some comments on the test, see inline.

The changes `NotImplemented` -> `NotImplementeError` can stay, as a matter of fact I was in the process of doing the same as part of #1223

deejlucas:
@lostella Thank you for the feedback. I've addressed it. Ready for re-review

lostella (approved):
Looks good! Thanks for spotting this and fixing it!

PR #1227, opened 2020-12-17
Add features and lags to Autogluon Tabular model

arianacai1997:
*Issue #, if available:*
Adding the context length as a feature to predict future values. 
*Description of changes:*
There was a PR about using autogluon tabular predictor and estimation. Then there is a new feature which is context length, indicating the length of previous values as the features for prediction.


src/gluonts/nursery/autogluon_tabular.py
```diff
@@ -103,34 +125,42 @@ def to_forecast(
...
 
     def train(self, training_data: Dataset) -> TabularPredictor:
+        # every time there is only one time series passed
+        # list(training_data)[0] is essentially getting the only time series
         dfs = [
-            get_prediction_dataframe(to_pandas(entry))
+            get_prediction_dataframe(series=to_pandas(entry),
+                                     prediction_length=self.prediction_length,
+                                     use_lag=self.use_lag,
+                                     context_length=len(self.lags), train=True)
```
lostella, line 154: So I think the lags are supposed to indicate exactly which previous data points from the time series should be used as features. Here, however, you're just taking the amount of them, and (if I understand correctly), let's call it `n`, use the `n` previous data points as features. Shouldn't you pass `self.lags` here, and use those to construct the dataframe?

src/gluonts/nursery/autogluon_tabular.py
```diff
@@ -59,19 +61,25 @@ def get_prediction_dataframe(series):
...
     df = df.astype(convert_type)
-    return df
+
+    cache = [None] * (context_length + prediction_length) + list(series.values)
+    col = []
+    total = len(df)
+    for i in range(context_length):
+        col.append('lag' + str(i))
+        df['lag' + str(i)] = cache[i:i + total]
+    return df if train else df[-prediction_length:]
```
lostella, line 71: I'm not sure the logic is very clear here: a few lines above you are setting up which columns are in the dataframe, and what values should be in them. Maybe you can do everything there?

src/gluonts/nursery/autogluon_tabular.py
```diff
@@ -103,34 +125,42 @@ def to_forecast(
...
             for entry in training_data
         ]
         df = pd.concat(dfs)
-
-        ag_model = self.task.fit(
-            df, label="target", problem_type="regression", **self.kwargs
-        )
-
-        return TabularPredictor(ag_model, self.freq, self.prediction_length)
+        ag_model = self.task.fit(df, label="target", problem_type="regression", output_directory="Eval2/electricity30",
```
lostella, line 158: Is the output directory supposed to be hard-coded here?

lostella (changes requested):
Thanks! I've left some comment in line.

In general: could you expand tests for `get_prediction_dataframe`? We should really make sure that it is providing what is expected. Even just a single test case, with a simple choice of the lags (say, `lags=[1, 2, 5]`)

src/gluonts/nursery/autogluon_tabular.py
```diff
@@ -59,19 +61,25 @@ def get_prediction_dataframe(series):
...
-        prediction_length: int,
-    ) -> None:
-        self.ag_model = ag_model  # task?
+    def __init__(self, ag_model, freq: str, prediction_length: int, lags: list, use_lag=False) -> None:
+        self.ag_model = ag_model
         self.freq = freq
         self.prediction_length = prediction_length
+        self.use_lag = use_lag
+        self.lags = lags
+        self.auto_regression = False if ((not self.use_lag) or self.lags[-1] < self.prediction_length) else True
```
lostella, line 81:
```suggestion
        self.auto_regression = False if ((not self.use_lag) or self.lags[-1] < self.prediction_length) else True
```

src/gluonts/nursery/autogluon_tabular.py
```diff
@@ -59,19 +61,25 @@ def get_prediction_dataframe(series):
...
 
 class TabularPredictor(Predictor):
-    def __init__(
-        self,
-        ag_model,
-        freq: str,
-        prediction_length: int,
-    ) -> None:
-        self.ag_model = ag_model  # task?
+    def __init__(self, ag_model, freq: str, prediction_length: int, lags: list, use_lag=False) -> None:
```
lostella, line 75: Why do we have the `use_lag` flag here? Maybe the `lags` list can be enough: if empty (which could be the default) then no lags are used

arianacai1997: I think an empty list of `lags` may have two indicates: 1) no lags are used 2) generate the lags by `get_lags_for_frequency`.

src/gluonts/nursery/autogluon_tabular.py
arianacai1997: Sorry, it should be set by users. I'll put it into the kwargs instead.

src/gluonts/nursery/autogluon_tabular.py
arianacai1997: Yeah, it is a mistake. Will correct this accordingly. Thanks!

examples/autogluon_tabular_tutorial.ipynb
```diff
@@ -0,0 +1,192 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [
+    {
+     "ename": "ModuleNotFoundError",
```
jwmueller, line 9: Could you fix this ModuleNotFoundError?

In general, this entire notebook should be run with nice output before being comitted right?

lostella: I removed the notebook and instead expanded the tests: those example notebooks are at risk of becoming outdated, especially when they depend on newer modules

jwmueller: But will there be some sort of tutorial somewhere on how to use this model? 
Otherwise I worry all the various options made available here will be quickly be forgotten.

lostella: Yes, we can add a notebook to the website: those are run as part of the build, so we make sure not to break them or to update them

lostella (approved):

lostella (approved):

PR #1230, opened 2020-12-18
Fixed pytorch's predict to numpy helper

kashif:
currently this does not work when the tensors are on the gpu


lostella:
Woops! Good catch @kashif! I guess this wasn't really tested

lostella (approved):

kashif:
no worries @lostella I have a bit of spare time now so i can test. I have another issue I found which i am investigating so will send over something in the next few days

PR #1231, opened 2020-12-18
Fix type error for using quantile_weights and add a proper Pytest 

youngsuk0723:
*Issue #, if available:*

*Description of changes:*
Quantile loss class has an argument 'quantile_weights' which is set to be 'None' as a default. This argument is going to be used when computing approximate CRPS loss.

1. I found the bug when this argument is passed. The type must be 'List', but it is operated under nd.array.asscalar(), giving error as follows 

`E           AttributeError: 'float' object has no attribute 'asscalar'`
`Python-gluonts/src/gluonts/mx/block/quantile_output.py:101: AttributeError`

2.  Fixed the type/AttributeError and added a proper pytest example under 'test_quantile_loss.py' file. 

### DIFF
--- a/test/model/seq2seq/test_quantile_loss.py
+++ b/test/model/seq2seq/test_quantile_loss.py
@@ -15,7 +15,6 @@ from mxnet import nd
 
 from gluonts.mx.block.quantile_output import QuantileLoss
 
-
 def test_compute_quantile_loss() -> None:
     y_true = nd.ones(shape=(10, 10, 10))
     y_pred = nd.zeros(shape=(10, 10, 10, 2))
@@ -36,3 +35,21 @@ def test_compute_quantile_loss() -> None:
             - correct_qt_loss[idx]
             < 1e-5
         ), f"computing quantile loss at quantile {q} fails!"
+
+def test_compute_weighted_quantile_loss() -> None:
+    y_true = nd.ones(shape=(10, 10, 10))
+    y_pred = nd.zeros(shape=(10, 10, 10, 2))
+
+    quantiles = [0.5, 0.9]
+    quantile_weights = [0.5, 0.5]
+
+    loss = QuantileLoss(quantiles, quantile_weights)
+    # print(nd.mean(loss(y_true, y_pred)))
+    correct_weighted_qt_loss = 1.4
+    assert (
+        nd.mean(
+            loss(y_true, y_pred)
+        )
+        - correct_weighted_qt_loss
+        < 1e+2
+    ), f"computing weighted quantile loss fails!"
--- a/test/model/seq2seq/test_quantile_loss.py
+++ b/test/model/seq2seq/test_quantile_loss.py
@@ -48,18 +48,3 @@ def test_compute_quantile_loss(quantile_weights, correct_qt_loss) -> None:
             nd.mean(loss(y_true, y_pred)) - correct_qt_loss < 1e-5
         ), f"computing weighted quantile loss fails!"
 
-
-#
-# def test_compute_weighted_quantile_loss() -> None:
-#     y_true = nd.ones(shape=(10, 10, 10))
-#     y_pred = nd.zeros(shape=(10, 10, 10, 2))
-#
-#     quantiles = [0.5, 0.9]
-#     quantile_weights = [0.5, 0.5]
-#
-#     loss = QuantileLoss(quantiles, quantile_weights)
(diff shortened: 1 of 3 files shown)


dcmaddix (commented):
Looks great! Thanks for fixing Youngsuk!  You can fix the style check and then I think we should be good to merge.

test/model/seq2seq/test_quantile_loss.py
```diff
@@ -36,3 +36,18 @@ def test_compute_quantile_loss() -> None:
...
+
+def test_compute_weighted_quantile_loss() -> None:
+    y_true = nd.ones(shape=(10, 10, 10))
+    y_pred = nd.zeros(shape=(10, 10, 10, 2))
+
+    quantiles = [0.5, 0.9]
+    quantile_weights = [0.5, 0.5]
+
+    loss = QuantileLoss(quantiles, quantile_weights)
+    # print(nd.mean(loss(y_true, y_pred)))
```
dcmaddix, line 49: let's remove this commented print statement.

test/model/seq2seq/test_quantile_loss.py
```diff
@@ -36,3 +36,18 @@ def test_compute_quantile_loss() -> None:
...
+    correct_weighted_qt_loss = 1.4
+    assert (
+        nd.mean(loss(y_true, y_pred)) - correct_weighted_qt_loss < 1e2
```
dcmaddix, line 52: Should this be `1e-2` for the tolerance?

test/model/seq2seq/test_quantile_loss.py
```diff
@@ -36,3 +36,18 @@ def test_compute_quantile_loss() -> None:
             - correct_qt_loss[idx]
             < 1e-5
         ), f"computing quantile loss at quantile {q} fails!"
+
+
+def test_compute_weighted_quantile_loss() -> None:
```
dcmaddix, line 41: You should be actually able to combine this with the `test_compute_quantile_loss` function for minimal code change by importing `pytest`. 
You can do the following:
`@pytest.mark.parametrize("quantile_weights", [None, [0.5, 0.5]])
def test_compute_quantile_loss(quantile_weights) -> None:`

and it will iterate the test over both values of `quantile_weights`.

test/model/seq2seq/test_quantile_loss.py
```diff
@@ -12,27 +12,39 @@
...
 
 from gluonts.mx.block.quantile_output import QuantileLoss
 
 
-def test_compute_quantile_loss() -> None:
+@pytest.mark.parametrize(
+    "quantile_weights, correct_qt_loss",
+    [
+        (None, [1.0, 1.8]),
+        ([0.5, 0.5], 1.4),
```
dcmaddix, line 24: Great!

test/model/seq2seq/test_quantile_loss.py
```diff
@@ -12,27 +12,39 @@
...
-
-    correct_qt_loss = [1.0, 1.8]
-
-    for idx, q in enumerate(quantiles):
-        assert (
-            nd.mean(
-                loss.compute_quantile_loss(
-                    nd.ndarray, y_true, y_pred[:, :, :, idx], q
+    loss = QuantileLoss(quantiles, quantile_weights)
+    print(quantile_weights, correct_qt_loss)
```
dcmaddix, line 34: remove print statement

test/model/seq2seq/test_quantile_loss.py
```diff
@@ -12,27 +12,39 @@
...
+    if quantile_weights == None:
```
dcmaddix, line 35: Can simplify to `if not quantile_weights:`

test/model/seq2seq/test_quantile_loss.py
```diff
@@ -12,27 +12,39 @@
...
+                    loss.compute_quantile_loss(
+                        nd.ndarray, y_true, y_pred[:, :, :, idx], q
+                    )
                 )
-            )
-            - correct_qt_loss[idx]
-            < 1e-5
-        ), f"computing quantile loss at quantile {q} fails!"
+                - correct_qt_loss[idx]
+                < 1e-5
```
dcmaddix, line 44: Let's define a `tol=1e-5` variable, so don't have to repeat `1e-5` twice.

dcmaddix (approved):

youngsuk0723:
Thanks!

PR #1234, opened 2020-12-21
Enable mypy in subpackages

lostella:

### DIFF
--- a/src/gluonts/mx/block/cnn.py
+++ b/src/gluonts/mx/block/cnn.py
@@ -19,6 +19,13 @@ from mxnet.gluon import nn
 from gluonts.mx import Tensor
 
 
+def _get_singleton(a):
+    if isinstance(a, (list, tuple)):
+        assert len(a) == 1
+        return a[0]
+    return a
+
+
 class CausalConv1D(gluon.HybridBlock):
     """
     1D causal temporal convolution, where the term causal means that output[t]
@@ -60,10 +67,8 @@ class CausalConv1D(gluon.HybridBlock):
     ):
         super(CausalConv1D, self).__init__(**kwargs)
 
-        self.dilation = dilation if isinstance(dilation, int) else dilation[0]
-        self.kernel_size = (
-            kernel_size if isinstance(kernel_size, int) else kernel_size[0]
-        )
+        self.dilation = _get_singleton(dilation)
+        self.kernel_size = _get_singleton(kernel_size)
         self.padding = self.dilation * (self.kernel_size - 1)
         self.conv1d = nn.Conv1D(
             channels=channels,
--- a/src/gluonts/mx/block/cnn.py
+++ b/src/gluonts/mx/block/cnn.py
@@ -19,7 +19,7 @@ from mxnet.gluon import nn
 from gluonts.mx import Tensor
 
 
-def _get_singleton(a):
+def _get_singleton(a: Union[int, List[int], Tuple[int]) -> int:
     if isinstance(a, (list, tuple)):
         assert len(a) == 1
         return a[0]
--- a/src/gluonts/mx/block/cnn.py
+++ b/src/gluonts/mx/block/cnn.py
@@ -53,20 +53,22 @@ class CausalConv1D(gluon.HybridBlock):
     def __init__(
         self,
         channels: int,
-        kernel_size: int,
-        dilation: int = 1,
+        kernel_size: Union[int, Tuple[int], List[int]],
+        dilation: Union[int, Tuple[int], List[int]] = 1,
         activation: Optional[str] = None,
         **kwargs,
     ):
         super(CausalConv1D, self).__init__(**kwargs)
 
-        self.dilation = dilation
-        self.kernel_size = kernel_size
-        self.padding = dilation * (kernel_size - 1)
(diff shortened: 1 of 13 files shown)


src/gluonts/mx/block/cnn.py
```diff
@@ -53,20 +53,22 @@ class CausalConv1D(gluon.HybridBlock):
     def __init__(
         self,
         channels: int,
-        kernel_size: int,
-        dilation: int = 1,
+        kernel_size: Union[int, Tuple[int], List[int]],
```
jaheba, line 56: We could give this type a name.

src/gluonts/mx/block/cnn.py
```diff
@@ -53,20 +53,22 @@ class CausalConv1D(gluon.HybridBlock):
...
+        dilation: Union[int, Tuple[int], List[int]] = 1,
         activation: Optional[str] = None,
         **kwargs,
     ):
         super(CausalConv1D, self).__init__(**kwargs)
 
-        self.dilation = dilation
-        self.kernel_size = kernel_size
-        self.padding = dilation * (kernel_size - 1)
+        self.dilation = dilation if isinstance(dilation, int) else dilation[0]
```
jaheba, line 63: What happens if you provide more than one value?

src/gluonts/mx/block/decoder.py
```diff
@@ -30,9 +30,7 @@ def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
     # noinspection PyMethodOverriding
-    def hybrid_forward(
-        self, F, dynamic_input: Tensor, static_input: Tensor
-    ) -> None:
+    def hybrid_forward(self, F, dynamic_input, static_input):
```
jaheba, line 33: Why remove the type annotations?

src/gluonts/mx/block/regularization.py
```diff
@@ -81,7 +81,7 @@ def hybrid_forward(self, F, *states: List[Tensor]) -> Tensor:
                     state = F.stack(*state, axis=self._time_axis)
                 means.append(
                     self._alpha
-                    * state.__pow__(2).mean(
+                    * F.power(state, 2).mean(
```
jaheba, line 84: Do you need to use `F` here?

src/gluonts/mx/kernels/_kernel_output.py
```diff
@@ -100,7 +100,7 @@ def get_args_proj(self, float_type: DType = np.float32) -> ArgProj:
     # noinspection PyMethodOverriding,PyPep8Naming
     def gp_params_scaling(
         self, F, past_target: Tensor, past_time_feat: Tensor
-    ) -> Tuple[Tensor, Tensor, Tensor]:
+    ) -> Tuple:
```
jaheba, line 103: Blank `Tuple` should be `tuple`.

src/gluonts/testutil/shell.py
```diff
@@ -85,8 +85,8 @@ def invocations(
     def batch_invocations(
         self, data_entries: Iterable[DataEntry]
     ) -> List[dict]:
-        instances = map(serialize_data_entry, data_entries)
-        instances = list(map(json.dumps, instances))
+        instances_pre = map(serialize_data_entry, data_entries)
```
jaheba, line 88: Mypy doesn't like re-assigning?

lostella: looks like so

src/gluonts/mx/block/decoder.py
lostella: Truth is, there appear to be no way of making mypy happy here

src/gluonts/mx/block/regularization.py
lostella: `state` above can change type, I tried asserting that it's a `Tensor` (which supports `__pow__`) but that cannot be done apparently, so using `F.power` fixes it

src/gluonts/mx/block/cnn.py
lostella: added a helper method which does an assetion too

src/gluonts/mx/block/cnn.py
lostella: not sure, because then anyone looking into the docstring will find some custom type without knowing what it is

unless we document that too, but a `Union` is pretty self-explanatory and not too bad I think

src/gluonts/mx/block/decoder.py
lostella: addressed, there was actually some inconsistencies of ours in all this subclassing

src/gluonts/mx/block/cnn.py
```diff
@@ -19,6 +19,13 @@
 from gluonts.mx import Tensor
 
 
+def _get_singleton(a):
```
jaheba, line 22:
```suggestion
def _get_singleton(a: Union[int, List[int], Tuple[int]) -> int:
```

jaheba: Not sure I like the name. Maybe just `_get_int` is better?

jaheba (dismissed):

jaheba (approved):

PR #1235, opened 2020-12-22
Input masking and scaling in NBEATS (following authors' implementation)

prolearner:
*Issue:*
NBEATS implementation differs from [the one of the authors](https://github.com/ElementAI/N-BEATS/blob/master/models/nbeats.py). In particular it lacks proper [masking of the input](https://github.com/ElementAI/N-BEATS/blob/master/models/nbeats.py#L72), which is also described in the [paper](https://arxiv.org/pdf/1905.10437.pdf) at the end of page 26.

*Description of changes:*
Added proper input masking and optional (default to False) input scaling. Input scaling is used in the [meta-learning version of NBEATS](https://arxiv.org/pdf/2002.02887.pdf) to account for time series of different magnitudes. Here, scaling is implemented similarly to DeeAR (using MeanScaler) while in the paper the input/output of the model is scaled/descaled by the maximum value in the input.

Strangely, the authors' implementation also seems to [include part of the input as one of the additive components of the final forecasts](https://github.com/ElementAI/N-BEATS/blob/master/models/nbeats.py#L69). This pull request does not include this change since it is not described in the [original NBEATS paper](https://arxiv.org/pdf/1905.10437.pdf).

### DIFF
--- a/src/gluonts/model/n_beats/_network.py
+++ b/src/gluonts/model/n_beats/_network.py
@@ -15,6 +15,7 @@ from typing import List
 
 import mxnet as mx
 import numpy as np
+from gluonts.block.scaler import MeanScaler, NOPScaler
 
 from gluonts.core.component import validated
 from gluonts.mx import Tensor
@@ -416,6 +417,8 @@ class NBEATSNetwork(mx.gluon.HybridBlock):
         A list of strings of length 1 or 'num_stacks'.
         Default and recommended value for generic mode: ["G"]
         Recommended value for interpretable mode: ["T","S"]
+    scale
+        if True scales the input observations by the mean
     kwargs
         Arguments passed to 'HybridBlock'.
     """
@@ -434,6 +437,7 @@ class NBEATSNetwork(mx.gluon.HybridBlock):
         expansion_coefficient_lengths: List[int],
         sharing: List[bool],
         stack_types: List[str],
+        scale: bool = False,
         **kwargs,
     ) -> None:
         super().__init__(**kwargs)
@@ -447,6 +451,11 @@ class NBEATSNetwork(mx.gluon.HybridBlock):
         self.stack_types = stack_types
         self.prediction_length = prediction_length
         self.context_length = context_length
+
+        if scale:
+            self.scaler = MeanScaler(keepdims=True)
+        else:
+            self.scaler = NOPScaler(keepdims=True)
 
         with self.name_scope():
             self.net_blocks: List[NBEATSBlock] = []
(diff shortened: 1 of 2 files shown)


lostella:
@prolearner looks like `black` is complaining: you need to run [this version](https://github.com/awslabs/gluon-ts/blob/master/requirements/requirements-setup.txt#L1) against `src` and `test`, and commit any re-format it applies

src/gluonts/model/n_beats/_network.py
```diff
@@ -666,7 +694,8 @@ def __init__(self, *args, **kwargs) -> None:
 
     # noinspection PyMethodOverriding,PyPep8Naming
     def hybrid_forward(
-        self, F, past_target: Tensor, future_target: Tensor = None
+            self, F, past_target: Tensor, future_target: Tensor = None,
+            past_observed_values: Tensor = None, future_observed_values: Tensor = None,
```
lostella: Maybe the optional arguments (that can be `None`) should be `Optional[Tensor]`, but I'm not sure whether mypy will like it.

Also it seems like the super classes should be able to accept `None`, since these arguments are passed on to them later on

src/gluonts/model/n_beats/_network.py
```diff
@@ -678,15 +724,23 @@ def hybrid_forward(
             Shape: (batch_size, context_length, target_dim).
         future_target
             Not used.
+        past_observed_values
+            Tensor with past observed values.
+            Shape: (batch_size, context_length, target_dim).
```
lostella, line 729: Same as above

src/gluonts/model/n_beats/_network.py
```diff
@@ -666,7 +707,12 @@ def __init__(self, *args, **kwargs) -> None:
 
     # noinspection PyMethodOverriding,PyPep8Naming
     def hybrid_forward(
-        self, F, past_target: Tensor, future_target: Tensor = None
+        self,
+        F,
+        past_target: Tensor,
+        past_observed_values: Tensor,
+        future_target: Optional[Tensor] = None,
+        future_observed_values: Optional[Tensor] = None,
```
lostella, line 715: I'm wondering if this block could just avoid having these inputs, since they aren't used. I know that `future_target` was there already, but maybe that could be removed too

src/gluonts/model/n_beats/_network.py
```diff
@@ -508,19 +517,37 @@ def __init__(
...
             # connect last block
-            return forecast + self.net_blocks[-1](backcast)
+            forecast = forecast + self.net_blocks[-1](backcast)
+
+        forecast = F.broadcast_mul(forecast, scale)
+
+        return (
+            forecast
+            if future_observed_values is None
+            else forecast * future_observed_values
```
lostella, line 549: Maybe an idea: just return the original (unmasked) `forecast` tensor here, and let the caller mask it where needed (e.g. the training network). Could that work?

In fact: only masking the forecast is probably not enough, and the corresponding target values which are compared in the loss should be masked too. Or otherwise, one can mask the loss values directly. (But I haven't checked in the paper if this is what the authors mean -- I hope so)

With these changes, the `future_` inputs could be removed from this block probably.

lostella (commented):
Thanks @prolearner! I have some comments inline

src/gluonts/model/n_beats/_network.py
prolearner: it seems that the signature of hybrid_forward must match the one of the super class (some tests fail if this is not the case). That's why probably all four inputs are needed for the hybrid_forward of all the 3 classes.

lostella: Ah I see, it's probably `mypy` who's complaining?

prolearner: Can't remember, might be.

src/gluonts/model/n_beats/_network.py
prolearner: I think it's a good idea, I moved the masking inside the loss functions. I followed the [authors' implementation](https://github.com/ElementAI/N-BEATS/blob/04f56c4ca4c144071b94089f7195b1dd606072b0/common/torch/losses.py#L45).
Still did not manage to remove future imports (did not even try since the tests were failing before).

lostella (approved):
Looks good to me! Thanks for fixing this!

PR #1240, opened 2020-12-25
set is_pad_field's type to the data type in spilters

kashif:
Else it ends up being `float64` and needs to be cast down when checking with `past_observed_values` etc.


lostella (approved):
Makes sense! Thanks @kashif! In fact, I think this goes unnoticed when using MXNet models, because MXNet probably does some automatic conversions; but PyTorch would probably complain

PR #1252, opened 2021-01-08
Speed up multivariate evaluator by 10x

kashif:
set super's `num_workers=None` since we loop over each dimension to calculate the metrics and thus multiple workers only slow it down


lostella:
Looks good to me @kashif! Maybe though it's better to re-expose the setting in the constructor? I can imagine multiprocessing to help in case one has a collection of several multi-dimensional series. What do you think? Thought defaulting it to `None` probably makes sense

kashif:
sure I can do that... let me add it thanks!

kashif:
@lostella done. Can you kindly have a look? Thanks!

lostella (approved):
🚀

PR #1253, opened 2021-01-08
Add Settings.

jaheba:
*Issue #, if available:*

*Description of changes:*


jaheba:
Usage:

```python

context = Context(foo=2)

print(context["foo"])

with context(foo=3):
    print(context["foo"])

print(context["foo"])


@context.bind(a=2)
def f(a):
    print(a)


f(1)
with context(a=10):
    f()
    f(a=3)
```

src/gluonts/env.py
```diff
@@ -11,13 +11,22 @@
...
 
-"""
-Maximum number of times a transformation can receive an input without returning an output.
-This parameter is intended to catch infinite loops or inefficiencies, when transformations
-never or rarely return something.
-"""
-GLUONTS_MAX_IDLE_TRANSFORMS = int(
-    os.environ.get("GLUONTS_MAX_IDLE_TRANSFORMS", "100")
-)
+env = Environment()
```
lostella, line 32: This could be `env = Context(max_idle_transforms=..., use_tqdm=...)`

The advantage of subclassing is the type annotation? What does that provide in practice?

src/gluonts/core/context.py
```diff
@@ -0,0 +1,203 @@
...
+from typing import Any
+
+import pydantic
+
+
+class _Config:
+    arbitrary_types_allowed = True
+
+
+class Context:
```
lostella, line 71: I'm wondering if this could create confusion (there's python context managers, mxnet's computational context...)

Would `Namespace` be a better name?

lostella (dismissed):
Looks good to me! We can iterate on this and refine it as needed (the whole thing is a bit verbose maybe? but I guess that's because this has the decorator methods, plus the pydantic stuff, so even using a `ChainMap` would end up being as verbose).

Just have some minor comments in the code.

src/gluonts/env.py
jaheba: A more concise interface and type-checking. So now you get an error if you use a non-supported type.

src/gluonts/core/context.py
jaheba: I see the possible confusion, but I think `Context` is still a better name than `Namespace`. We could name it `Setting` though.

src/gluonts/env.py
lostella: > So now you get an error if you use a non-supported type.

We could have a small test case for that? That can help understanding this use case, and also make sure pydantic is doing what it's supposed to be doing

lostella (approved):

Issue #1158, opened 2020-11-19
Can't use validation_data when training with single time series? (DeepAR)


mariosantosprivate:
Hello, so I have several time series of daily sales spanning 940 days, the goal being to predict the last 40 days, and I've split it into 3 sets:

- ListDataSet train_ds : First 860 days
- ListDataSet val_ds: First 900 days
- ListDataSet test_ds: All records (940)

I've trained a Deep AR model using `validation_data = val_ds` and it worked fine for multiple time series, however when I only train it on a single time series, after a couple of epochs I get the following error:

```python
  0%|          | 0/50 [00:00<?, ?it/s]learning rate from ``lr_scheduler`` has been overwritten by ``learning_rate`` in optimizer.
100%|██████████| 50/50 [02:00<00:00,  2.42s/it, epoch=1/100, avg_epoch_loss=8.9] 
1it [00:00,  2.50it/s, epoch=1/100, validation_avg_epoch_loss=8.6]
100%|██████████| 50/50 [00:11<00:00,  4.27it/s, epoch=2/100, avg_epoch_loss=8.68]
1it [00:00,  2.67it/s, epoch=2/100, validation_avg_epoch_loss=8.62]
100%|██████████| 50/50 [00:11<00:00,  4.38it/s, epoch=3/100, avg_epoch_loss=8.66]
1it [00:00,  2.72it/s, epoch=3/100, validation_avg_epoch_loss=8.59]
100%|██████████| 50/50 [00:11<00:00,  4.35it/s, epoch=4/100, avg_epoch_loss=8.65]
1it [00:00,  2.71it/s, epoch=4/100, validation_avg_epoch_loss=8.42]
100%|██████████| 50/50 [00:11<00:00,  4.41it/s, epoch=5/100, avg_epoch_loss=8.61]
0it [00:00, ?it/s]
---------------------------------------------------------------------------
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-36-8857ef7fe6cd> in <module>()
     21 # Using a validation set training stops as soon as the loss on the validation set (and therefore on unseen data) does not decrease any further.
     22 deepAR_predictor = deepAR_estimator.train(training_data=train_ds, 
---> 23                                           validation_data=val_ds
     24                                           )
     25 filepath = '/content/gdrive/My Drive/Thesis/Models/deepAR_single_ts.pkl'

3 frames
[...]

Issue #997, opened 2020-08-21
UnboundLocalError: local variable 'lv' referenced before assignment


kaijennissen:
## Description
I cannot include a validation dataset during the training of a multivariate model. When training without validation dataset everything works.

## To Reproduce
(Please provide minimal example of code snippet that reproduces the error. For existing examples, please provide link.)

```python
from gluonts.dataset.common import TrainDatasets
from gluonts.dataset.multivariate_grouper import MultivariateGrouper
from gluonts.dataset.repository.datasets import get_dataset
from gluonts.model.gpvar import GPVAREstimator
from gluonts.mx.trainer import Trainer

NUM_OF_SERIES = 8


def load_multivariate_dataset(dataset_name: str):
    ds = get_dataset(dataset_name)
    grouper_train = MultivariateGrouper(max_target_dim=NUM_OF_SERIES)
    grouper_test = MultivariateGrouper(max_target_dim=NUM_OF_SERIES)
    return TrainDatasets(
        metadata=ds.metadata,
        train=grouper_train(ds.train),
        test=grouper_test(ds.test),
    )


dataset = load_multivariate_dataset(
    dataset_name="exchange_rate"
)
metadata = dataset.metadata

estimator = GPVAREstimator(
    prediction_length=metadata.prediction_length,
    target_dim=NUM_OF_SERIES,
    freq=metadata.freq,
    trainer=Trainer(
        epochs=50,
        batch_size=4,
        num_batches_per_epoch=10,
        patience=5,
    )
)

predictor = estimator.train(
    training_data=dataset.train, validation_data=dataset.test)
```

## Error message or code output

```
UnboundLocalError: local variable 'lv' referenced before assignment
```

## Environment
- Operating system: Mac OSX 10.15.5
- Python version: 3.7.6
- GluonTS version: 0.5.0
- MXNet version: 1.6.0

lostella:
The problem occurs also on `master`. The root cause seems appears to be the way `ValidationDataLoader` extracts batches of data: the data transformation pipeline gets applied with `is_train=True` (see [here](...)) so that the "future" target is included in the data; however, this has the consequence that the instance splitter selects a random number of time windows from the validation dataset.

Now, because of the usage of `MultivariateGrouper`, the validation dataset consists of a single 8-dimensional series, out of which 0 or more "validation" windows gets sampled by the instance sampler. When 0 are sampled, no validation loss is computed and `lv` never gets assigned.

[...]

PR #1256, opened 2021-01-12
Rework GluonEstimator interface, splitters, and samplers for more flexibility

lostella:
## Related issues

#997, #1158 

## Description of changes

This proposal addresses a few issues with the current design of `GluonEstimator`: this allows estimator to be more accurate in constructing data loaders for validation purposes in the training loop.

The new interface of `GluonEstimator` is roughly as follows:

```python
class GluonEstimator:
    def create_transformation(self):
        # returns the data pre-processing transformation pipeline
    
    def create_training_data_loader(self, dataset, **kwargs):
        # returns an iterator over batches to be used for training purposes

    def create_validation_data_loader(self, dataset, **kwargs):
        # returns an iterator over batches to be used for validation purposes
    
    def create_training_network(self):
        # returns the network to be used for training the model

    def create_predictor(self, transformation, trained_network):
        # returns a predictor object
```

In addition to those, `GluonEstimator` provides the `train` method that glues everything together:

```python
    def train(
        self, training_data, [validation_data, options...]
    ):
        # 1. creates transformation and applies it to the data, optionally caching it
        # 2. creates data loader(s) given the transformed data
        # 3. creates the network
        # 4. invokes training loop given network and data loaders
        # 5. creates predictor out of the trained network, and return it
```

### Default validation data loaders

Before this PR, the validation data loader would be constructed by applying the transformations with `is_train=True`, so as include the `future_target` field in the batches, and evaluate the loss on it. However, this also means that windows of data from random points in time would be sampled out of the dataset, which usually not what one wants: what if one wants to validate the model performance on some "future data" from the same time series used in training? (A trivial case where this is needed is when one wants to train a model on a single series). Another drawback of random sampling for validation is that #997 can happen.

This PR proposes to default to a deterministic sequence of batches instead, with one instance per time series, taken at the end of it. This way one can easily control what time ranges are used to compute the training vs validation loss.

## Summary of the changes

* Simplifies the interface of `InstanceSampler` types, which now only accept the one argument `ts: np.ndarray` and not the bounds `a` and `b`, which was a bit redundant. Now at construction time, one can specify how many indices should be skipped at the beginning/end of the target for sampling, and `a, b` are computed internally.
* The continuous-time instance samplers change analogously, and accept a single `interval_length: float` argument to `__call__`.
* Make instance splitter transformations independent of the `is_train` flag, making them a little simpler. Now, whether to act in "train" or "test" mode can be controlled through the provided `InstanceSampler` object.
* The above allow to construct a "validation" data loader, that samples similarly to test mode (only pick final windows of the time series), but include future target as in train mode. This wouldn't be possible solely based on the `is_train` flag.
* `GluonEstimator`'s interface was updated to include explicit methods that return the various data loaders to be used. Also the way transformations are applied is changed: we first apply the proper "data" transformation (and an option is there to cache this), and then we pass anything "data-iteration-specific" (such as instance splitter, but this part could also include other transformation steps) to the data loader.

## Related changes

* The `get_hybrid_forward_input_names` utility was updated to work on `HybridBlock` types rather than instances
* The training loop was updated to better display the progress of validation

## To do

- [x] Adapt other estimators to the above changes.
- [x] Make all tests green again.
- [x] Rename `skip_initial` and `skip_final` to `min_past` and `min_future`
- [x] Adjust all affected docstrings.
- [ ] Can we make all of this backward compatible?


src/gluonts/model/simple_feedforward/_estimator.py
```diff
@@ -204,6 +273,17 @@ def create_training_network(self) -> HybridBlock:
     # we now define how the prediction happens given that we are provided a
     # training network.
     def create_predictor(self, transformation, trained_network):
+        prediction_splitter = InstanceSplitter(
+            target_field=FieldName.TARGET,
+            is_pad_field=FieldName.IS_PAD,
+            start_field=FieldName.START,
+            forecast_start_field=FieldName.FORECAST_START,
+            instance_sampler=TestSplitSampler(),
```
mbohlkeschneider, line 281: Not related to this PR, but the naming is actually confusing, since the `TestSplitSampler` does not actually perform any sampling. Maybe we want to rename this in the future.

src/gluonts/model/simple_feedforward/_estimator.py
```diff
@@ -158,34 +170,91 @@ def __init__(
...
+            forecast_start_field=FieldName.FORECAST_START,
+            instance_sampler=self.train_sampler,
+            past_length=self.context_length,
+            future_length=self.prediction_length,
+            time_series_fields=[FieldName.OBSERVED_VALUES],
+        )
+
+        return TrainDataLoader(
+            dataset=training_data,
+            transform=training_splitter + SelectFields(input_names),
```
mbohlkeschneider, line 215: Maybe I missed this, but where did the rest of the transformation go?

src/gluonts/model/simple_feedforward/_estimator.py
```diff
@@ -158,34 +170,91 @@ def __init__(
...
+                skip_final=self.prediction_length
+            ),
+            past_length=self.context_length,
+            future_length=self.prediction_length,
+            time_series_fields=[FieldName.OBSERVED_VALUES],
+        )
+
+        return ValidationDataLoader(
+            dataset=validation_data,
+            transform=validation_splitter + SelectFields(input_names),
```
mbohlkeschneider, line 250: Same here.

src/gluonts/mx/model/estimator.py
```diff
@@ -117,70 +117,70 @@ def create_predictor(
...
-        training_data_loader = TrainDataLoader(
-            dataset=training_data,
-            transform=transformation + SelectFields(input_names),
-            batch_size=self.batch_size,
-            stack_fn=partial(
-                batchify,
-                ctx=self.trainer.ctx,
-                dtype=self.dtype,
-            ),
+        transformed_training_data = TransformedDataset(
```
mbohlkeschneider, line 141: Ahh, now I see. I think I would prefer to have the transformed data chained in the `Estimator` as we had before. This distributes what is happening into too many places, I think.

src/gluonts/transform/convert.py
```diff
@@ -384,21 +378,18 @@ def map_transform(self, data: DataEntry, is_train: bool) -> DataEntry:
...
         return data
 
-    def _preprocess_data(self, data: DataEntry, is_train: bool):
+    def _preprocess_data(self, data: DataEntry):
         """
         Performs several preprocess operations for computing the empirical CDF.
         1) Reshaping the data.
         2) Normalizing the target length.
-        3) Adding noise to avoid zero slopes (training only)
+        3) Adding noise to avoid zero slopes
```
mbohlkeschneider, line 386: Not sure this changes behavior... We need to check.

src/gluonts/transform/feature.py
```diff
@@ -249,18 +247,15 @@ class AddConstFeature(MapTransformation):
...
     Parameters
     ----------
     output_field
         Field name for output.
     target_field
         Field containing the target array. The length of this array will be used.
     pred_length
-        Prediction length (this is necessary since
-        features have to be available in the future)
+        Prediction length (this is necessary since features have to be available
```
mbohlkeschneider, line 257: So to clarify: The change in behavior is that we ALWAYS generate the features for additional `prediction_length` steps, right?

src/gluonts/transform/split.py
```diff
@@ -559,7 +499,9 @@ def flatmap_transform(
                 + data[self.start_field].freq.delta * future_start
             )
 
-            if is_train:  # include the future only if is_train
+            if (
+                self.use_prediction_features
```
mbohlkeschneider, line 503: Why is this needed in this case? This is again the same as `is_train`, no?

mbohlkeschneider (commented):
Great work! Thank you for that! Dropped a few comments.

src/gluonts/model/simple_feedforward/_estimator.py
```diff
@@ -158,34 +170,91 @@ def __init__(
...
+        input_names = get_hybrid_forward_input_names(
+            SimpleFeedForwardTrainingNetwork
+        )
+
+        validation_splitter = InstanceSplitter(
+            target_field=FieldName.TARGET,
+            is_pad_field=FieldName.IS_PAD,
+            start_field=FieldName.START,
+            forecast_start_field=FieldName.FORECAST_START,
+            instance_sampler=ValidationSplitSampler(
```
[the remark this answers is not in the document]
lostella: good point, I think that makes a lot of sense

src/gluonts/mx/model/estimator.py
lostella: The point is, you may want to exploit the difference between completely deterministic transformations (i.e. data pre-processing) from random ones. For example, you may want to cache the former (like it is done here) to speed up training.

I think in general these are two very different kinds of transformations: one part says how all data entries in the datasets should be adjusted to fit the model's needs, the other instructs the data loader on how to split instances that need to be stacked into batches.

But I'll reconsider all of this, I'm thinking of reducing the scope of this PR a bit to make it less breaking

lostella:
@mbohlkeschneider thanks for the comments! I'm thinking of focusing this PR to only fixing the validation story: so, improve how `GluonEstimator` is structured, but leave the `is_train` story there for now.

I'm afraid otherwise there's too much changes going on here that may require a lot of double checking.

src/gluonts/dataset/loader.py
```diff
@@ -243,24 +242,15 @@ def ValidationDataLoader(
...
-    num_prefetch: Optional[int] = None,
-    shuffle_buffer_length: Optional[int] = None,
-):
-    return DataLoader(
-        data_iterable=TransformedDataset(dataset, transform, is_train=False),
-        batch_size=batch_size,
-        stack_fn=stack_fn,
+def ValidationDataLoader(*args, **kwargs):
+    warnings.warn(
+        "gluonts.dataset.loader.ValidationDataLoader is deprecated. Use InferenceDataLoader instead.",
```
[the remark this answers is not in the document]
lostella: The idea is that users who call this function get the warning, stop using it, and then the function is removed from the next release.

But I’ve reverted the change now, so it doesn’t matter :-)

src/gluonts/transform/feature.py
lostella: Reverted

src/gluonts/transform/sampler.py
```diff
@@ -19,28 +21,29 @@
...
-    -------
-    np.ndarray
-        Selected points to sample
+    The sampled indices ``i`` satisfy ``a <= i <= b``, where ``a = skip_initial``
+    and ``b = ts.shape[axis] - skip_final``.
     """
 
-    def __call__(self, ts: np.ndarray, a: int, b: int) -> np.ndarray:
+    def __init__(
+        self, *, axis: int = -1, skip_initial: int = 0, skip_final: int = 0
```
vafl, line 32: This is very flexible and much nicer than the previous bounds approach. But perhaps we can pass additional parameters to `__call__`, such as `context_length`, `past_length`, etc.

The `skip_initial` is an important parameter here that should perhaps be set to `context_length` by default. Otherwise the network may try to predict something from nothing (I know this was the previous behavior...). Also perhaps `min_history` is a better name than `skip_initial`?

vafl (commented):
Looks good overall. Added some comments

src/gluonts/model/simple_feedforward/_estimator.py
```diff
@@ -117,7 +129,8 @@ def __init__(
         batch_normalization: bool = False,
         mean_scaling: bool = True,
         num_parallel_samples: int = 100,
-        train_sampler: InstanceSampler = ExpectedNumInstanceSampler(1.0),
+        train_sampler: Optional[InstanceSampler] = None,
```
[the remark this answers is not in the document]
lostella: The reason is that the default sampler depends on `prediction_length`, I’m not sure there is a way in python to default an argument to the value of another one

src/gluonts/transform/sampler.py
lostella: In my mind, `__call__` should take anything that is entry-dependent, while other fixed settings (that are usually tied to some model hyperparameters) are better set in the constructor: this way we don’t force additional arguments in the API, while each concrete sampler is free to take as many configuration options as it wants.

About the default `skip_initial` value: sure, we can decide to use `context_length` as default, but I think that can be set in each estimator’s default sampler. 

I like `min_history` as name, it tells exactly what requirement the sampled indices will have with respect to the given sequence, while being detached from model-specific concepts (like `context_length`).

What about `skip_final`?

src/gluonts/model/simple_feedforward/_estimator.py
```diff
@@ -158,34 +171,94 @@ def __init__(
...
+    def create_training_data_loader(
+        self,
+        training_data: Dataset,
+        **kwargs,
+    ) -> DataLoader:
+        input_names = get_hybrid_forward_input_names(
+            SimpleFeedForwardTrainingNetwork
+        )
+
+        training_splitter = InstanceSplitter(
```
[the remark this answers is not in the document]
lostella: Yes, each estimator will construct its data loaders using their preferred type of splitter, no restriction there

src/gluonts/transform/sampler.py
vafl: I agree about what should be passed to call in principle, however in this case it makes it difficult to write samplers that have a clear behavior, because they need to know things like `context_length` etc, unless we want to go with writing sampler factories, using functools.partial. or setting some attributes of the sampler later. I think I ran into this with my other comment too, where you cannot instantiate a sampler without knowing `prediction_length`.

lostella: I'm not sure there's a clear advantage in having `context_length` or other quantities passed to `__call__` rather than used at construction time. The potential drawbacks I see are the following:
* models may have multiple "time lengths" as hyperparameters, so I'm afraid that pinning down what exactly should be provided to `__call__` may not always be clear
* moreover, the sampler is used through the splitter, so in order to specify which quantities the splitter will pass on to the sampler, the interfaces of these two objects must be in sync, which may make things too implicit

In general, it's my impression that the instance splitter class should be extremely simple and general. We might not even need to have two separate ones (`InstanceSplitter` and `CanonicalInstanceSplitter`)

vafl: Ok, yeah maybe its more about documentation and good defaults. We can just set the `min_history` to `context_length` when we construct the sampler inside all models, as this should be the default behavior. Then if you provide a sampler yourself you should know what you are doing.

vafl: > What about `skip_final`?

Yeah, or something like `min_future`.

src/gluonts/model/canonical/_estimator.py
```diff
@@ -68,31 +78,75 @@ def __init__(
...
-                    is_pad_field=FieldName.IS_PAD,
-                    start_field=FieldName.START,
-                    forecast_start_field=FieldName.FORECAST_START,
-                    train_sampler=TestSplitSampler(),
-                    time_series_fields=[FieldName.FEAT_TIME],
-                    past_length=self.context_length,
-                    future_length=self.prediction_length,
-                ),
-            ]
+        return (
```
jaheba, line 81: This isn't so much related to the PR, but still:

When I look at this, a few things look odd.

Why is it called `create_transformation? What is transformed? Looking at the body, we do input pre-processing, ensuring that fields have the correct type and shape. Also we seem to add some features.

It looks very verbose and noisy, lots of `FieldName` being mentioned.

If we would introduce schemas, the code would already look much nicer.

Also, we might want to think about how we generate features.

src/gluonts/model/deepar/_estimator.py
```diff
@@ -238,7 +250,16 @@ def __init__(
             else DummyValueImputation(self.distr_output.value_in_support)
         )
 
-        self.train_sampler = train_sampler
+        self.train_sampler = (
+            train_sampler
+            if train_sampler is not None
+            else ExpectedNumInstanceSampler(1.0, min_future=prediction_length)
```
jaheba, line 256: I guess we can't have these as default, since it depends on the `prediction_length`?

src/gluonts/mx/model/estimator.py
```diff
@@ -117,85 +122,98 @@ def create_predictor(
...
-            ),
-            num_workers=num_workers,
-            num_prefetch=num_prefetch,
-            shuffle_buffer_length=shuffle_buffer_length,
-            decode_fn=partial(as_in_context, ctx=self.trainer.ctx),
-            **kwargs,
-        )
+        assert (training_data is None) != (
+            training_data_loader is None
+        ), "You must provide either `training_data` or `training_data_loader`"
```
jaheba, line 148: Can't we just do `training_data: Union[Dataset, DataLoader]`?

I don't like adding more arguments, if it's an either or thing.

src/gluonts/mx/trainer/_base.py
```diff
@@ -272,12 +275,15 @@ def loop(
...
-                    with tqdm(
-                        itertools.islice(
-                            batch_iter, self.num_batches_per_epoch
-                        ),
-                        total=self.num_batches_per_epoch,
-                    ) as it:
+                    batch_iter_length = None
+
+                    if num_batches_to_use:
+                        batch_iter = itertools.islice(
```
jaheba, line 281: Can't this be unconditional, since slicing `None` is the same as not slicing at all?

src/gluonts/transform/sampler.py
```diff
@@ -55,31 +56,64 @@ class UniformSplitSampler(InstanceSampler):
...
-    def __call__(self, ts: np.ndarray, a: int, b: int) -> np.ndarray:
-        return np.array([b])
+    def __init__(
+        self,
+        allow_empty_interval=False,
+        axis: int = -1,
+        min_past: int = 0,
+        min_future: int = 0,
+    ) -> None:
+        super().__init__(axis=axis, min_past=min_past, min_future=min_future)
```
jaheba, line 88: Can't we just convert this into `pydantic` types?

src/gluonts/model/canonical/_estimator.py
lostella: Yes, I agree. This is the same for several other estimators.

src/gluonts/model/deepar/_estimator.py
lostella: That's correct. It's unfortunate...

src/gluonts/mx/model/estimator.py
lostella: I like the idea, but I'm afraid python's typing wouldn't play well with that. How do you distinguish a `Dataset` from a `DataLoader`? They're essentially the same type `Iterable[Dict]`

lostella: The alternative would be to have some other method `train_from_data_loaders` which is called by the current one

lostella: Conclusion is: we can get rid of this. If someone wants to provide custom training/validation data loaders, they probably don't want to be using the "high level" estimator interface, and can just create the network and invoke

```python
trainer.train(net, training_data_loader, validation_data_loader)
```

It's on the single components to be usable as stand-alone, not on the estimator API to accommodate every possible usage.

vafl:
Looks great. Thanks for this and for the detailed explanation.

Since this is a major change, it would be good to run thorough regression tests before the next release.
Maybe even better if we can merge the automatic accuracy tests before this CR or shortly after? @jaheba @arangatang

vafl (dismissed):

src/gluonts/transform/sampler.py
lostella: done, it looks much better

src/gluonts/transform/split.py
```diff
@@ -395,7 +347,7 @@ def flatmap_transform(
                 past_ts = past_ts.transpose() if self.output_NTC else past_ts
                 d[self._past(ts_field)] = past_ts
```
vafl, line 348: BTW, one thing that would be nice here, is to support incomplete prediction windows. Right now we only support full time windows in the future. If we mask the loss using a `future_is_pad` indicator, we can also support incomplete future windows. This would be quite useful when you have short time series (the NBEATS paper does this). It is also more consistent with samplers being able to pick any point from the time series without constraints.

lostella: yes, that could be an option. maybe though we address it separately? we can open a dedicated issue

lostella:
@vafl here are some results for DeepAR, WaveNet, DeepState. The number of runs for each combination is in parentheses in the label of each box plot (10 to 15 for DeepAR, WaveNet, while for DeepState that doesn't matter since data loading is deterministic there). There seem to be no apparent regression, although numbers are a bit noisy (I might run more with DeepAR, WaveNet).

I found something weird with MQ-CNN instead: no regression, but I get some variance in the results while I would expect also that case to be deterministic. I'm looking into that.

image

vafl (dismissed):
Awesome, looks good. Thanks for the thorough comparison @lostella !

benidis (approved):

Issue #1196, opened 2020-12-03
Fewer, shallower imports


lostella:
A typical script or notebook using GluonTS begins as follows:

```
from gluonts.dataset.repository.datasets import get_dataset, dataset_recipes
from gluonts.dataset.common import ListDataset
from gluonts.dataset.field_names import FieldName
from gluonts.evaluation import Evaluator
from gluonts.evaluation.backtest import make_evaluation_predictions
from gluonts.model.simple_feedforward import SimpleFeedForwardEstimator
from gluonts.mx.trainer import Trainer
```

This type of deep import paths makes it hard to remember what-is-where, and forces one to have lots of different import lines to do anything. One could could instead have:

```
from gluonts.dataset import get_dataset, dataset_recipes, ListDataset, FieldName
from gluonts.evaluation import Evaluator, make_evaluation_predictions
from gluonts.mx import SimpleFeedForwardEstimator, Trainer
```

This makes the package more pleasant to use, especially in interactive mode.

One could even shorten it further and import *almost* everything from `gluonts` directly (although it is arguable whether this would be good or not), but *not everything:* since we want to keep some optional dependencies (MXNet, PyTorch, ...), then some components will need to be imported from different sub-packages.

**Note:** in the above I’m moving `SimpleFeedForwardEstimator` to `gluonts.mx` (from `gluonts.model`). This will eventually be necessary to keep dependencies optional, but since all offered models are currently MXNet-based we can start by deprecating the current import paths for models, suggest the new one, but allow both.

## Docs driven change

Since GluonTS code-base is pretty huge and messy by now, it might be hard to do all the changes at once. One way to proceed is to let example and documentation code drive the change: we can use scripts and notebooks there to identify everything that is commonly imported, shorten their paths, and make sure that they work.

## Backward compatible

The above changes could be done in a backward-compatible manner, so no code should be moved. Following the changes, we should also adjust imports in all tests, examples scripts and notebooks, and documentation, so that the “recommended” import paths are always guaranteed to be available.

## Optional: deprecate importing from deeper modules

We could also deprecate importing directly from the modules which are more than two levels down the path: in this way once people stop importing from there we can allow ourselves reorganizing the code as we wish (e.g. renaming modules, merging modules, moving definitions).

However I’m not sure how easy this is, i.e. how easy it is to have the following distinction:

```
from gluonts.dataset import ListDataset  # silently imports
from gluonts.dataset.common import ListDataset  # prints deprecation warning
```

PR #1259, opened 2021-01-13
add shallower import paths

Schmedu:
#1196 *Issue #, if available:*

*Description of changes:*


src/gluonts/dataset/__init__.py
```diff
@@ -14,5 +14,25 @@
...
+    "dataset_recipes",
+    "TrainDataLoader",
+    "StepStrategy",
+    "generate_rolling_dataset",
+]
+
+# fix Sphinx issues, see https://bit.ly/2K2eptM
+for item in __all__:
+    if hasattr(item, "__module__"):
+        setattr(item, "__module__", __name__)
```
jaheba, line 38: Is this needed?

src/gluonts/dataset/__init__.py
```diff
@@ -14,5 +14,25 @@
 # !!! DO NOT MODIFY !!! (pkgutil-style namespace package)
 
 from pkgutil import extend_path
+from .common import ListDataset
```
jaheba, line 17: There should be an empty line here.

src/gluonts/dataset/__init__.py
```diff
@@ -14,5 +14,25 @@
...
+from .field_names import FieldName
+from .repository.datasets import get_dataset, dataset_recipes, load_datasets
```
jaheba, line 19: I think I'm against making these available directly.

In the future, I would like to see `repository` be moved out of `dataset`.

src/gluonts/dataset/__init__.py
```diff
@@ -14,5 +14,25 @@
...
+from .loader import TrainDataLoader
+from .rolling_dataset import StepStrategy, generate_rolling_dataset
```
jaheba, line 21: Also not sure if we want to include these..

src/gluonts/dataset/__init__.py
```diff
@@ -14,5 +14,25 @@
 # !!! DO NOT MODIFY !!! (pkgutil-style namespace package)
 
 from pkgutil import extend_path
+from .common import ListDataset
+from .field_names import FieldName
+from .repository.datasets import get_dataset, dataset_recipes, load_datasets
+from .loader import TrainDataLoader
```
jaheba, line 20: This also looks like something, which should be moved out of `dataset`.

jaheba (commented):
Thanks for the pull request!

Can you adjust the title to be more specific?

PR #1260, opened 2021-01-13
Add streaming interface and RNN-based model

lostella:
cc @benidis 

Tests are at the bare minimum, we could use some more.

### DIFF
--- a/src/gluonts/nursery/streaming_models/forecast_generator.py
+++ b/src/gluonts/nursery/streaming_models/forecast_generator.py
@@ -20,11 +20,17 @@ import numpy as np
 from gluonts.core.component import validated
 from gluonts.dataset.field_names import FieldName
 from gluonts.dataset.loader import DataLoader, InferenceDataLoader
-from gluonts.model.forecast_generator import ForecastGenerator, OutputTransform
+from gluonts.model.forecast_generator import (
+    ForecastGenerator,
+    OutputTransform,
+    recursively_zip_arrays,
+)
 from gluonts.mx.distribution import DistributionOutput
 from gluonts.mx.model.forecast import DistributionForecast
 
 from .predictor import NETWORK_STATE_KEY, PREDICTOR_STATE_KEY
+
+logger = logging.getLogger(__name__)
 
 
 def to_numpy(x: Any) -> Any:
@@ -59,9 +65,9 @@ class StatefulDistributionForecastGenerator(ForecastGenerator):
         **kwargs
     ) -> Iterator[DistributionForecast]:
         if output_transform is not None:
-            logging.info("The `output_transform` argument will be ignored.")
+            logger.info("The `output_transform` argument will be ignored.")
         if num_samples is not None:
-            logging.info(
+            logger.info(
                 "Forecast is not sample based. Ignoring parameter `num_samples`."
             )
         for batch in inference_data_loader:
@@ -70,18 +76,18 @@ class StatefulDistributionForecastGenerator(ForecastGenerator):
                 k: b for k, b in batch.items() if k.startswith("s:")
             }
             outputs = prediction_net(*inputs)
-            distr_args_scale = outputs[:-1]
+            distr_args_scale_loc = outputs[:-1]
             network_state_batch = outputs[-1]
 
-            assert len(distr_args_scale) == 2
+            assert len(distr_args_scale_loc) == 3
 
             distributions = [
-                self.distr_output.distribution(distr_args, scale=scale)
-                for distr_args, scale in _extract_instances(distr_args_scale)
+                self.distr_output.distribution(*u)
+                for u in recursively_zip_arrays(distr_args_scale_loc)
             ]
 
             network_states = [
-                s for s in _extract_instances(network_state_batch)
+                s for s in recursively_zip_arrays(network_state_batch)
             ]
 
             idx = -1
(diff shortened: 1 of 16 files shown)


src/gluonts/nursery/streaming_models/forecast_generator.py
```diff
@@ -0,0 +1,140 @@
...
+        if output_transform is not None:
+            logging.info("The `output_transform` argument will be ignored.")
+        if num_samples is not None:
+            logging.info(
+                "Forecast is not sample based. Ignoring parameter `num_samples`."
+            )
+        for batch in inference_data_loader:
+            inputs = [batch[k] for k in input_names]
+            transformation_states = {
+                k: b for k, b in batch.items() if k.startswith("s:")
```
jaheba, line 70: These one-letter variable names have to stop!

src/gluonts/nursery/streaming_models/forecast_generator.py
```diff
@@ -0,0 +1,140 @@
...
+        else:
+            tmp = [to_numpy(xi) for xi in x]
+            if isinstance(x, tuple):
+                return tuple(tmp)
+            return tmp
+    return x
+
+
+class StatefulDistributionForecastGenerator(ForecastGenerator):
+    @validated()
```
jaheba, line 47: I wonder, can we get around not using `validated`? Does `core.ty.Stateless` work?

src/gluonts/nursery/streaming_models/forecast_generator.py
```diff
@@ -0,0 +1,140 @@
...
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+import logging
```
jaheba, line 14: I think we generally not want to use `logging` directly in the module, but create a logger first instead.

src/gluonts/nursery/streaming_models/forecast_generator.py
```diff
@@ -0,0 +1,140 @@
...
+
+            assert len(distr_args_scale) == 2
+
+            distributions = [
+                self.distr_output.distribution(distr_args, scale=scale)
+                for distr_args, scale in _extract_instances(distr_args_scale)
+            ]
+
+            network_states = [
+                s for s in _extract_instances(network_state_batch)
```
jaheba, line 84: When does it has to stop?

src/gluonts/nursery/streaming_models/forecast_generator.py
```diff
@@ -0,0 +1,140 @@
...
+            ]
+
+            i = -1
```
jaheba, line 87: Now!

jaheba (changes requested):
Single letter variable names.

lostella:
> Single letter variable names.

While I agree in general, I think extremely local scopes like comprehensions may benefit from short names. This said, I can change them :-)

src/gluonts/nursery/streaming_models/rnn/_estimator.py
```diff
@@ -0,0 +1,865 @@
...
+            t = np.concatenate([buffer, target])
+        lags = np.vstack(
+            [t[self.max_lag - l : len(t) - l] for l in self.lag_seq]
+        )
+        data[self.lag_field] = np.nan_to_num(lags)
+        data[self.lag_state_field] = t[-self.max_lag :]
+        return data
+
+
+class AddStreamAggregateLags(SimpleTransformation):
```
benidis, line 166: Maybe we can move all the transformations to `stream_models/transform` so all the transformations are together? I guess there is no reason to split them here.

src/gluonts/nursery/streaming_models/predictor.py
```diff
@@ -0,0 +1,145 @@
...
+    def deserialize(
+        cls,
+        path: Path,
+        ctx: Optional[mx.Context] = None,
+        *args,
+        **kwargs,
+    ) -> "StateAwarePredictorWrapper":
+        ctx = ctx if ctx is not None else get_mxnet_context()
+        with mx.Context(ctx):
+            predictor = Predictor.deserialize(path / "predictor", ctx)  # type: ignore
```
benidis, line 140: This should fail. The `ctx` should be removed (this was for the 0.4.x gluonts). I guess serde is not captured in the included tests that's why there was no issue.

src/gluonts/nursery/streaming_models/masking.py
```diff
@@ -0,0 +1,172 @@
...
+    def mask(
+        self,
+        masking_state: StreamState,
+        forecast: DistributionForecast,
+        values: pd.Series,
+    ) -> Tuple[StreamState, pd.Series, pd.Series]:
+        raise NotImplementedError()
+
+
+class MissingValueMaskingStrategy(MaskingStrategy):
```
benidis, line 39: There are some changes in masking that I think should be included here also since they fix edge cases.

src/gluonts/nursery/streaming_models/rnn/_estimator.py
```diff
@@ -0,0 +1,865 @@
...
+    def __call__(self) -> Any:
+        return pickle.loads(self.pickled_bytes)
+
+    @staticmethod
+    def from_state(state) -> "StateInitializer":
+        s = pickle.dumps(state).decode("latin1")
+        return StateInitializer(s)
+
+
+class StreamingRnnEstimator(Estimator):
```
benidis, line 320: There is the dataset subsampling change that we may or may not want to include now. Just mentioning it here to be aware of what is missing.

src/gluonts/nursery/streaming_models/predictor.py
lostella: Just saw this, I'll fix it

src/gluonts/nursery/streaming_models/rnn/_estimator.py
lostella: Should we keep the subsampling code here? It appears to complicate the estimator significantly, and to some extent limit its applicability to `ListDataset` objects (because `len` needs to be invoked)

PR #1261, opened 2021-01-13
include arrow dataset

vafl:
*Issue #, if available:*

*Description of changes:*

This includes a new dataset based on pyarrow. The dataset is much faster than json for large datasets and in particular for long time series. While for small datasets (e.g. m3) json can be faster, the pyarrow datasets are still fast enough for any training purposes and don't slow down for other datasets.

| dataset            | arrow samples / s | json samples /s | speedup |
|--------------------|------------------:|----------------:|--------:|
| exchange_rate      |           15685.1 |          1699.1 |     9.2 |
| solar-energy       |           29739.7 |          1553.8 |    19.1 |
| electricity        |           33287.7 |           538.7 |    61.7 |
| traffic            |           33130.7 |           682.7 |    48.5 |
| exchange_rate_nips |           18593.4 |          1759.2 |    10.6 |
| electricity_nips   |           32578.2 |          2458.5 |    13.3 |
| traffic_nips       |           34474.2 |          2646.5 |    13.0 |
| solar_nips         |           33199.8 |         1556.46 |    21.3 |
| wiki-rolling_nips  |           35168.1 |         10024.6 |     3.5 |
| taxi_30min         |           34670.0 |         12326.6 |     2.8 |
| m3_monthly         |           33348.8 |         41733.8 |     0.8 |
| m3_quarterly       |           32515.0 |         57618.1 |     0.6 |
| m3_yearly          |           26561.4 |         70285.3 |     0.4 |
| m3_other           |           28783.4 |         51215.7 |     0.6 |
| m4_hourly          |           30906.9 |         10985.2 |     2.8 |
| m4_daily           |           34236.2 |          4400.5 |     7.8 |
| m4_weekly          |           32433.3 |          9363.1 |     3.5 |
| m4_monthly         |           34562.3 |         28598.7 |     1.2 |
| m4_quarterly       |           34958.2 |         47717.1 |     0.7 |
| m4_yearly          |           34416.6 |         69802.5 |     0.5 |
| m5                 |           28913.0 |          1025.6 |    28.2 |


requirements/requirements.txt
```diff
@@ -5,3 +5,4 @@ pandas~=1.1
 pydantic~=1.1,<1.7
 tqdm~=4.23
 toolz~=0.10
+pyarrow==2.0.0
```
jaheba, line 8: I wonder, can we make this optional?

src/gluonts/dataset/artificial/generate_synthetic.py
```diff
@@ -24,6 +24,7 @@
     ComplexSeasonalTimeSeries,
     ConstantDataset,
 )
+from gluonts.dataset.common import ArrowWriter
```
jaheba, line 27: Can we do this in a second PR?

src/gluonts/dataset/common.py
```diff
@@ -207,13 +213,144 @@ def files(self) -> List[Path]:
...
-        # TODO: given that we only support json, should we also filter json
-        # TODO: in the extension?
-        return not (path.name.startswith(".") or path.name == "_SUCCESS")
+        valid_endings = [
+            ".json",
+            ".jsonl",
+            ".json.gz",
+            ".jsonl.gz",
+        ]
+        if not any(path.name.endswith(ending) for ending in valid_endings):
```
jaheba, line 226:
```suggestion
        if not path.suffix in valid_endings:
```

src/gluonts/dataset/util.py
```diff
@@ -73,9 +74,7 @@ def get_bounds_for_mp_data_loading(dataset_len: int) -> DataLoadingBounds:
     return DataLoadingBounds(lower=lower, upper=upper)
 
 
-def _split(
-    it: Iterator[T], fn: Callable[[T], bool]
-) -> Tuple[List[T], List[T]]:
+def _split(it: List[T], fn: Callable[[T], bool]) -> Tuple[List[T], List[T]]:
```
jaheba, line 77: What's the reason for this change?

vafl: I was using lists for some operations and it was simpler to just use them everywhere instead of passing `iter(mylist)` to please mypy. Also materializing the list of file names is never a problem.

src/gluonts/dataset/common.py
vafl: Would not work for `.json.gz`.

src/gluonts/dataset/artificial/generate_synthetic.py
vafl: I'd prefer not to invest time in splitting the PR if not necessary.

jaheba:
These are some great numbers! Have you tried caching the dataset as an alternative benchmark?

src/gluonts/dataset/common.py
jaheba: Oh, of course! There is `.suffixes` which would work, but not sure it reads much nicer than this.

src/gluonts/dataset/util.py
```diff
@@ -87,20 +86,33 @@ def _split(
...
 
 
-def _list_files(directory: Path) -> Iterator[Path]:
-    for dirname, _, filenames in os.walk(directory):
-        for filename in filenames:
-            yield Path(dirname, filename)
+def _list_files(paths: List[Path]) -> List[Path]:
+    files = []
+    for p in paths:
+        if p.exists() and p.is_file():
```
jaheba, line 92: Shouldn't just `is_file()` work here, since if it doesn't exist, it's also a file.

src/gluonts/dataset/common.py
```diff
@@ -207,13 +213,144 @@ def files(self) -> List[Path]:
...
+        cls,
+        paths: Union[str, Path, List[str], List[Path]],
+        freq: str,
+        chunk_size: int = 100,
+    ):
+        paths = util.resolve_paths(paths) if paths is not None else None
+        files = util.find_files(paths, lambda p: str(p).endswith(".arrow"))
+        assert (
+            len(files) > 0
+        ), f"Could not find any arrow files in paths: {paths}"
```
jaheba, line 266:
```suggestion
        assert files, f"Could not find any arrow files in paths: {paths}"
```

src/gluonts/dataset/common.py
```diff
@@ -475,9 +612,22 @@ def load_datasets(
         An object collecting metadata, training data, test data.
     """
     meta = MetaData.parse_file(Path(metadata) / "metadata.json")
-    train_ds = FileDataset(path=train, freq=meta.freq)
-    test_ds = FileDataset(path=test, freq=meta.freq) if test else None
-
+    files = util._list_files(util.resolve_paths(train))
+    has_arrow_files = any(str(fname).endswith(".arrow") for fname in files)
+    if use_arrow and has_arrow_files:
+        print(f"loading arrow files from {train}, {test}")
```
jaheba, line 618: Let's not use `print`.

src/gluonts/dataset/common.py
```diff
@@ -207,13 +213,144 @@ def files(self) -> List[Path]:
...
+            ".jsonl",
+            ".json.gz",
+            ".jsonl.gz",
+        ]
+        if not any(path.name.endswith(ending) for ending in valid_endings):
+            return False
+        return not path.name.startswith(".")
+
+
+class ArrowDataset(Dataset):
```
jaheba, line 231: I think I don't want this to be in `common`, since I want to see `common` to be removed in the future.

Can't we just define it all in `dataset/arrow.py`, and then import the `ArrowDataset` in `dataset/__init__.py`, so that we can then do:

```py
from gluonts.dataset import ArrowDataset
```

src/gluonts/dataset/common.py
```diff
@@ -207,13 +213,144 @@ def files(self) -> List[Path]:
...
+        it: Iterable[DataEntry],
+        file_path: Union[str, Path],
+        chunk_size=100,
+    ) -> None:
+        """
+        Write time series records as an arrow dataset to the file_path.
+        """
+        file_path = Path(file_path)
+        msg = f"Converting to arrow dataset: {file_path}"
+        l: Optional[int] = None
```
jaheba, line 303: I really dislike one-letter variable names. Looking just at this line, I can only guess what it should be.

src/gluonts/dataset/common.py
```diff
@@ -207,13 +213,144 @@ def files(self) -> List[Path]:
...
+        try:
+            l = len(it)  # type: ignore
+        except:
+            l = None
+        with ArrowWriter(file_path, chunk_size=chunk_size) as writer:
+            for rec in tqdm(it, total=l, desc=msg):
+                rec = rec.copy()
+                float_np_fields = [
```
jaheba, line 311: This looks somewhat problematic to me, since it appears we are only supporting a few fixed fields.

I'm fine with leaving it as it is, since we have it elsewhere too.

src/gluonts/dataset/common.py
```diff
@@ -207,13 +213,144 @@ def files(self) -> List[Path]:
...
+                if "start" in rec:
+                    rec["start"] = pd.Timestamp(rec["start"])
+                if "source" in rec:
+                    del rec["source"]
+                writer.write_record(rec)
+
+
+class MemmapArrowDataset(ArrowDataset):
+    def __init__(self, files: List[Path], freq: str, chunk_size: int):
+        """
```
jaheba, line 332: Docstring should go on the class.

src/gluonts/dataset/common.py
```diff
@@ -207,13 +213,144 @@ def files(self) -> List[Path]:
...
+            row_col = pa.array(np.arange(len(t)))
+            t = t.append_column(self._ROW_COL_NAME, row_col)
+            tables.append(t)
+        if len(tables) > 1:
+            table = pa.concat_tables(tables)
+        else:
+            table = tables[0]
+        super().__init__(table, freq=freq, chunk_size=chunk_size)
+
+    def __del__(self):
```
jaheba, line 351: I don't like relying on `__del__`.

But can we at least have a `close()` method which is then called in `__del__`?

jaheba (changes requested):

src/gluonts/dataset/util.py
jaheba: Also, I think this can be more easily done with `Path.rglob`.

Will look into this later or tomorrow.

src/gluonts/dataset/common.py
vafl: Yes, will do. I only added this cleanup because something there is an issue on windows, where you cannot delete a directory with a memmaped file.

src/gluonts/dataset/common.py
vafl: yes, makes sense

src/gluonts/dataset/common.py
vafl: Yes, the `ArrowWriter` is very general and can write numpy arrays, lists etc. But this part is specific for our stuff. It just makes sure that the data we write is in the right format. If we write the target as a list for example, it slows things down again when reading, since you have to convert it to an array.

src/gluonts/dataset/common.py
vafl: will do

src/gluonts/dataset/arrow.py
```diff
@@ -0,0 +1,171 @@
...
+        Strings of the form 'yyyy-mm-dd ...' are converted to timestamps if possible.
+
+        Note:
+        Numpy arrays with dimension > 1 are stored in two columns one with a 1d flat data
+        an additional column called "<array_name>._np_shape" that stores the array shape.
+        One could also do this via pyarrow/pandas extension types, but it seemed overly
+        complicated for this use case.
+        """
+        self.path: Path = Path(path)
+        if not str(path).endswith(".arrow"):
```
jaheba, line 45: This time `.suffix` looks like the right thing :D

```suggestion
        if not self.path.suffix == ".arrow":
```

src/gluonts/dataset/common.py
```diff
@@ -207,13 +213,144 @@ def files(self) -> List[Path]:
...
+    ):
+        paths = util.resolve_paths(paths) if paths is not None else None
+        files = util.find_files(paths, lambda p: str(p).endswith(".arrow"))
+        assert (
+            len(files) > 0
+        ), f"Could not find any arrow files in paths: {paths}"
+        return MemmapArrowDataset(files, freq, chunk_size)
+
+    def __iter__(self) -> Iterator[DataEntry]:
+        row_number = 0
```
jaheba, line 270: Why not use `enumerate` for this?

src/gluonts/dataset/common.py
```diff
@@ -207,13 +213,144 @@ def files(self) -> List[Path]:
...
+        self.reader = ArrowReader(self.table, chunk_size=self.chunk_size)
+
+    @classmethod
+    def load_files(
+        cls,
+        paths: Union[str, Path, List[str], List[Path]],
+        freq: str,
+        chunk_size: int = 100,
+    ):
+        paths = util.resolve_paths(paths) if paths is not None else None
```
jaheba, line 262: According to the type, `paths` should never be `None`.

src/gluonts/dataset/repository/_util.py
```diff
@@ -47,13 +50,14 @@ def serialize(x):
...
 def save_to_file(path: Path, data: List[Dict]):
-    print(f"saving time-series into {path}")
     path_dir = os.path.dirname(path)
     os.makedirs(path_dir, exist_ok=True)
     with open(path, "wb") as fp:
-        for d in data:
+        for d in tqdm(data, desc=f"saving time-series into {path}"):
             fp.write(json.dumps(d).encode("utf-8"))
             fp.write("\n".encode("utf-8"))
+    arrow_path = Path(str(path).replace(".json", ".arrow"))
```
jaheba, line 59:
```suggestion
    arrow_path = path.with_suffix(".arrow")
```

src/gluonts/dataset/common.py
```diff
@@ -158,18 +164,18 @@ class FileDataset(Dataset):
 
     def __init__(
         self,
-        path: Path,
+        path: Union[str, Path, List[Path], List[str]],
```
jaheba, line 167: What's the use case for supporting multiple paths?

requirements/requirements.txt
jaheba: Also, we try not to fix depend on specific versions.

src/gluonts/dataset/common.py
```diff
@@ -10,7 +10,7 @@
 # on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
-
+import re
```
jaheba, line 13:
```suggestion
```

src/gluonts/dataset/common.py
```diff
@@ -475,9 +612,22 @@ def load_datasets(
         An object collecting metadata, training data, test data.
     """
     meta = MetaData.parse_file(Path(metadata) / "metadata.json")
-    train_ds = FileDataset(path=train, freq=meta.freq)
-    test_ds = FileDataset(path=test, freq=meta.freq) if test else None
-
+    files = util._list_files(util.resolve_paths(train))
+    has_arrow_files = any(str(fname).endswith(".arrow") for fname in files)
+    if use_arrow and has_arrow_files:
```
jaheba, line 617: Should we handle the case where both `.json` and `.arrow` are provided?

src/gluonts/dataset/arrow.py
```diff
@@ -0,0 +1,171 @@
...
+class ArrowReader:
+    def __init__(self, table: pa.Table, chunk_size: int = 100):
+        self.table = table
+        self.chunk_size = chunk_size
+        self._np_shape_cols: Optional[List[Tuple[str, str]]] = None
+
+    def iter_slice(
+        self, start: Optional[int] = None, length: Optional[int] = None
+    ):
+        start = start if start is not None else 0
```
jaheba, line 146:
```suggestion
        self, start: int = 0, length: Optional[int] = None
    ):
```

src/gluonts/dataset/common.py
vafl: > These are some great numbers! Have you tried caching the dataset as an alternative benchmark?

Yes, if data is in memory samples/s is much higher (around 1.0E7). The main purpose of the ArrowDataset is for large datasets that you cannot cache in memory.

src/gluonts/dataset/common.py
```diff
@@ -455,7 +592,7 @@ def __call__(self, data: DataEntry) -> DataEntry:
 
 
 def load_datasets(
-    metadata: Path, train: Path, test: Optional[Path]
+    metadata: Path, train: Path, test: Optional[Path], use_arrow=True
```
chm90, line 595: One thing I like about gluonts in general is how few options and settings there are. What do you think about optionally specifying the file format in MetaData? Inferring could also be one option.

src/gluonts/dataset/common.py
chm90: @vafl Multiple files are already supported. No need to supply multiple paths?

Schmedu:
Hey @vafl, @jaheba asked me to add some of his suggested changes. Please have both a look and let me know if it's alright so far. Thanks

src/gluonts/dataset/util.py
```diff
@@ -87,15 +88,7 @@ def _split(it: List[T], fn: Callable[[T], bool]) -> Tuple[List[T], List[T]]:
...
-    files = []
-    for p in paths:
-        if p.exists() and p.is_file():
-            files.append(p)
-            continue
-        for dirname, _, filenames in os.walk(p):
-            for filename in filenames:
-                files.append(Path(dirname, filename))
-    return files
+    return list(concat(p.rglob("*") for p in paths))
```
jaheba, line 91: No single letter variables, por favor!

src/gluonts/dataset/common.py
```diff
@@ -300,13 +299,13 @@ def write_table_from_records(
...
         file_path = Path(file_path)
         msg = f"Converting to arrow dataset: {file_path}"
-        l: Optional[int] = None
+        length: Optional[int] = None
         try:
-            l = len(it)  # type: ignore
+            length = len(it)  # type: ignore
         except:
-            l = None
+            length = None
```
jaheba: This looks like passing `length` does nothing.

Schmedu: It's used two lines later in: `for rec in tqdm(it, total=length, desc=msg):`

src/gluonts/dataset/util.py
Schmedu: removed

src/gluonts/dataset/common.py
jaheba: Ah, sorry, I thought length was passed as an argument, which it is not.

We should not do the blank except however.

requirements/requirements.txt
```diff
@@ -5,4 +5,4 @@ pandas~=1.1
 pydantic~=1.1,<1.7
 tqdm~=4.23
 toolz~=0.10
-pyarrow==2.0.0
+pyarrow~=2.0
```
vafl, line 8: Version 3 has been released in the mean time.

src/gluonts/dataset/common.py
```diff
@@ -259,17 +259,16 @@ def load_files(
...
+        files = util.find_files(paths, lambda p: p.suffix == ".arrow")
+        assert files, f"Could not find any arrow files in paths: {paths}"
         return MemmapArrowDataset(files, freq, chunk_size)
 
     def __iter__(self) -> Iterator[DataEntry]:
-        row_number = 0
         bounds = get_bounds_for_mp_data_loading(len(self))
-        for rec in self.reader.iter_slice(bounds.lower, bounds.upper):
+        for row_number, rec in enumerate(
+            self.reader.iter_slice(bounds.lower, bounds.upper)
```
vafl, line 270: Should maybe be `enumerate(..., start=bounds.lower)` ?

vafl (commented):
Thanks a lot for bringing this over the line.

src/gluonts/dataset/common.py
Schmedu: @vafl did something change in the logic? Previously you started the `now_number` at 0. Why do you want to start it now at `bounds.lower`? I mean, if you say that's the correct way to implement it, I'll do so, I just want to point out that it seems to be inconsistent with your previous solution.

vafl: Yes, I think my previous version was wrong. Its not super important, but the purpose of the row number is to track where the data came from and I think it makes more sense if its the row number in the full table.

Schmedu: Ok, pushed it, thanks for the explanation!

requirements/requirements.txt
```diff
@@ -1,7 +1,8 @@
 holidays>=0.9
 matplotlib~=3.0
-numpy~=1.16
+numpy~=1.16,<1.20.0
```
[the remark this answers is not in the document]
Schmedu: There is an issue on PyArrows side: https://github.com/Azure/MachineLearningNotebooks/issues/1314
Here's a test run without the numpy restriction with PyArrow 3.0: https://github.com/Schmedu/gluon-ts/runs/2222295991
It also fails with PyArrow 2.0 and without the restriction: https://github.com/Schmedu/gluon-ts/runs/2222623650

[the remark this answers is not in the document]
jaheba: Thanks for the info!

Thinking about this, I don't think we want to have `pyarrow` as a direct dependency, but instead have it optional as an extra. So you would do `pip install "gluonts[arrow]"` to install the dependencies.

kashif:
BTW, so I managed to add a time series dataset ("electricity") to huggingface's datasets repo and it seems to work nicely via the `ListDataset(...)` modulo the meta-data. I would appreciate it if you have some feedback on this... 

https://github.com/huggingface/datasets/pull/3722

I plan to add the rest of the datasets...

requirements/requirements.txt
lostella: Looking into this PR ~1 year later: version 7 is now there :-)

kashif:
@lostella in the meantime I have added the Monash time series repository to datasets as well: https://github.com/huggingface/datasets/pull/3743

lostella:
Closing in favor of #2000

PR #1263, opened 2021-01-13
Add support for forward fields.

### LINKED ISSUE #806: Add forward-fields support to shell for batch-transform
When doing batch transform, one might want to store additional fields such as `item_id` alongside the prediction to be able to simply match input and output values.

Currently, we just ignore all additional values.

jaheba:
Fixes #806.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/dataset/common.py
+++ b/src/gluonts/dataset/common.py
@@ -505,6 +505,8 @@ def serialize_data_entry(data):
             field = field.astype(np.object_)
             field[nan_ix] = "NaN"
             return field.tolist()
+        if isinstance(field, (int, float)):
+            return field
         return str(field)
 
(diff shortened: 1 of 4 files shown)


mbohlkeschneider (dismissed):
Thank you!

src/gluonts/dataset/common.py
```diff
@@ -505,6 +505,8 @@ def serialize_field(field):
             field = field.astype(np.object_)
             field[nan_ix] = "NaN"
             return field.tolist()
+        if isinstance(field, (int, float)):
```
jaheba, line 508: This needs further investigation. I'm not sure it's doing what it's supposed to do.

mbohlkeschneider (approved):

PR #1265, opened 2021-01-14
added the kwargs for in autogluon-tabular

arianacai1997:
*Issue #, if available:*
In autogluon, the code `task.fit` contains some arguments such as presets. It is fixed that in the time-series case, the prediction is a matter of regression instead of others. Therefore, at least the `problem_type` is fixed.

*Description of changes:*
Added a fixed `fit` argument `problem_type` and implement others as the kwargs.


src/gluonts/nursery/autogluon_tabular.py
```diff
@@ -117,7 +113,46 @@ def train(self, training_data: Dataset) -> TabularPredictor:
...
+                "optimize_for_deployment",
+            ]
+        else:
+            presets = self.kwargs["presets"]
+
+        if "eval_metric" not in self.kwargs:
+            eval_metric = "mean_absolute_error"
+        else:
+            eval_metric = self.kwargs["eval_metric"]
+
```
lostella, line 144: I think you can do these in a more manageable way as follows: within `__init__` do

```python
default_kwargs = {
    "excluded_model_types": ["KNN", "XT", "RF"],
    "hyperparameters": "light",
    ...
}
self.kwargs = {**default_kwargs, **kwargs}
```

this way, for any key which is in both `kwargs` and `default_kwargs`, the value in `kwargs` will take precedence

lostella:
@arianacai1997 looks like you have added a lot of undesired files (probably autogluon models) to the PR, could you remove them?

src/gluonts/nursery/autogluon_tabular.py
```diff
@@ -103,23 +103,36 @@ def to_forecast(
...
+        ag_model = self.task.fit(
+            df, label="target", problem_type="regression", **self.kwargs
+        )
+
         return TabularPredictor(ag_model, self.freq, self.prediction_length)
 
 
 def LocalTabularPredictor(*args, **kwargs) -> Localizer:
-    return Localizer(TabularEstimator(*args, **kwargs))
+    return Localizer(TabularEstimator(*args, **kwargs))
```
lostella, line 138:
```suggestion
    return Localizer(TabularEstimator(*args, **kwargs))

```

src/gluonts/nursery/autogluon_tabular.py
```diff
@@ -103,21 +103,34 @@ def to_forecast(
...
-    def __init__(self, freq: str, prediction_length: int) -> None:
+    def __init__(self, freq: str, prediction_length: int, **kwargs) -> None:
         super().__init__()
         self.task = task
         self.freq = freq
         self.prediction_length = prediction_length
+        default_kwargs = {
+            "excluded_model_types": ["KNN", "XT", "RF"],
+            "hyperparameters": "light",
+            "num_bagging_sets": 1,
```
jwmueller, line 114: I would run a small set of benchmarks to see whether this default_kwarg is better (in terms of forecast-accuracy, disk space used) than an alternative version of default_kwarg where you instead delete: `hyperparameters`, `num_bagging_sets`, 
and replace 
`presets: [ "high_quality_fast_inference_only_refit", "optimize_for_deployment",]`

arianacai1997: Working on it. Should `eval_metric="mean_absolute_error"` also be the hard-set argument instead of being in the kwargs?

jwmueller: No that should be left as default kwarg, as some users may want to change its value

lostella (approved):
Looks good to me!

jwmueller (approved):
looks good!

PR #1266, opened 2021-01-14
Add distribution-based forecast types for torch, output layers, tests

lostella:
cc @kashif 


kashif:
awesome! Sorry I couldnt find the time this last week.

lostella:
> awesome! Sorry I couldnt find the time this last week.

@kashif no worries at all! let me know if you can spot anything, I did some minor changes (like extending `DistributionOutput` instead of `IndependentOutput`)

kashif:
Right... the `IndependentOutput` was something on our side which helped in code reuse when specifying independent multivariate distributions... So all looks good!

lostella:
Makes sense. I think it's nice to have something like that, but maybe one could wrap a `DistributionOutput` class (rather than having an additional type) in a similar fashion to [how pytorch distributions handle turning event dimensions into batch ones](https://pytorch.org/docs/stable/distributions.html#torch.distributions.independent.Independent)

kashif:
BTW the tests use the script in the workflows to install the torch stuff... perhaps you can change that to use the requirements?

kashif:
@lostella  ah i remember the issue was that the default pytorch pip will pull in the gpu version on Linux etc. which is ok when using the library but not ok when running the tests CI/CD machines...

kashif (dismissed):

src/gluonts/torch/model/forecast_generator.py
```diff
@@ -0,0 +1,89 @@
...
+def _extract_instances(x: Any) -> Any:
+    """
+    Helper function to extract individual instances from batched
+    mxnet results.
+    For a tensor `a`
+      _extract_instances(a) -> [a[0], a[1], ...]
+    For (nested) tuples of tensors `(a, (b, c))`
+      _extract_instances((a, (b, c)) -> [(a[0], (b[0], c[0])), (a[1], (b[1], c[1])), ...]
+    """
+    if isinstance(x, (np.ndarray, torch.Tensor)):
```
mbohlkeschneider, line 36: This could be just factored out in one function which operates both on mxnet and torch

src/gluonts/torch/model/forecast_generator.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def _extract_instances(x: Any) -> Any:
+    """
+    Helper function to extract individual instances from batched
+    mxnet results.
+    For a tensor `a`
+      _extract_instances(a) -> [a[0], a[1], ...]
+    For (nested) tuples of tensors `(a, (b, c))`
+      _extract_instances((a, (b, c)) -> [(a[0], (b[0], c[0])), (a[1], (b[1], c[1])), ...]
```
mbohlkeschneider, line 34: Not sure here: If the `b`, `c` tensors have smaller dimension in the respective axis, than the returned batches will be limited to this. Is this intended?

src/gluonts/torch/model/forecast_generator.py
```diff
@@ -0,0 +1,89 @@
...
+        freq: str,
+        output_transform: Optional[OutputTransform],
+        num_samples: Optional[int],
+        **kwargs
+    ) -> Iterator[DistributionForecast]:
+        for batch in inference_data_loader:
+            inputs = [batch[k] for k in input_names]
+            outputs = prediction_net(*inputs)
+            if output_transform is not None:
+                outputs = output_transform(batch, outputs)
```
mbohlkeschneider, line 71: That seems to be the only difference in code to the mxnet version, right?

kashif: also `outputs` need to be converted to numpy here

src/gluonts/torch/model/forecast.py
```diff
@@ -0,0 +1,106 @@
...
+        distribution: Distribution,
+        start_date: pd.Timestamp,
+        freq: str,
+        item_id: Optional[str] = None,
+        info: Optional[Dict] = None,
+    ) -> None:
+        self.distribution = distribution
+        self.shape = (
+            self.distribution.batch_shape + self.distribution.event_shape
+        )
```
jaheba, line 62: I guess we can omit the `self` to get this on one line

src/gluonts/torch/model/forecast.py
```diff
@@ -0,0 +1,106 @@
...
+        self.prediction_length = self.shape[0]
+        self.item_id = item_id
+        self.info = info
+
+        assert isinstance(
```
jaheba, line 67: I think we could use `ty.checked` for this :)

src/gluonts/torch/model/forecast.py
```diff
@@ -0,0 +1,106 @@
...
+            start_date, pd.Timestamp
+        ), "start_date should be a pandas Timestamp object"
+        self.start_date = start_date
+
+        assert isinstance(freq, str), "freq should be a string"
+        self.freq = freq
+        self._mean = None
```
jaheba, line 74: Why can't we compute it directly? Is this too expensive?

src/gluonts/torch/model/forecast.py
```diff
@@ -0,0 +1,106 @@
...
+    def mean_ts(self) -> pd.Series:
+        """
+        Forecast mean, as a pandas.Series object.
+        """
+        return pd.Series(data=self.mean, index=self.index)
+
+    def quantile(self, level: Union[float, str]) -> np.ndarray:
+        level = Quantile.parse(level).value
+        q = self.distribution.icdf(torch.tensor([level])).cpu().numpy()
+        return q
```
jaheba, line 97:
```suggestion
        return self.distribution.icdf(torch.tensor([level])).cpu().numpy()
```

src/gluonts/torch/model/forecast.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def to_sample_forecast(self, num_samples: int = 200) -> SampleForecast:
```
jaheba, line 99: Where do the `200` come from?

lostella: it is the same as in the corresponding method of `gluonts.mx.model.forecast.DistributionForecast`

src/gluonts/torch/model/forecast_generator.py
lostella: now that I look at it, I'm not sure `.numpy()` would work here @kashif? since the network might output a tuple of tensors (and it usually does)

I think there is a lot to be unified here between `gluonts.mx` and `gluonts.torch`, I'll try to leverage `@singledispatch` a little more and see if we can avoid rewriting this logic at all 💪

src/gluonts/torch/model/forecast_generator.py
lostella: yes, this is not detected and raised as an issue, but hopefully we have tests in place that check that you get N forecasts out of a dataset of size N

src/gluonts/torch/model/forecast.py
lostella: It might be, I guess it depends a bit on the type of distribution

lostella:
@mbohlkeschneider @kashif I've consolidated things quite a bit, now the `DistributionForecastGenerator` is implemented once, and the specific type of `DistributionForecast` which it generates (from `gluonts.mx` or `gluonts.torch`) depends on the type of `Distribution` it's given (either from `gluonts` or `torch`).

Plus I added some tests which I forgot to include earlier today.

@kashif note my comment on the output transformation: I think it doesn't really make sense to apply it, please confirm. I believe the `__call__` method of `ForecastGenerator` types is getting way too much information, and maybe the class could be refactored so that many of those details (e.g. the network) are constructor arguments. But this suggests that a bigger redesign of how `Predictor`/`ForecastGenerator` interact is required, so that can go to another time.

src/gluonts/model/forecast_generator.py
```diff
@@ -145,3 +183,46 @@ def __call__(
...
+        self.distr_output = distr_output
+
+    def __call__(
+        self,
+        inference_data_loader: DataLoader,
+        prediction_net,
+        input_names: List[str],
+        freq: str,
+        output_transform: Optional[OutputTransform],
+        num_samples: Optional[int],
```
lostella, line 200: Among these arguments, it seems to me like only `inference_data_loader` is really required "at run time" -- everything else could maybe be fixed at construction time. But then the `Predictor` classes that use this (see eg [here](https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/mx/model/predictor.py#L150-L174)) are really left with very little purpose.

src/gluonts/model/forecast_generator.py
```diff
@@ -20,17 +20,16 @@
...
-@singledispatch
-def predict_to_numpy(prediction_net, tensor) -> np.ndarray:
-    raise NotImplementedError
+OUTPUT_TRANSFORM_NOT_SUPPORTED_MSG = (
+    "The `output_transform` argument is not supported and will be ignored."
+)
+NOT_SAMPLE_BASED_MSG = "Forecast is not sample based. Ignoring parameter `num_samples` from predict method."
 
 
 def log_once(msg):
```
lostella, line 36: This should not use the `logging` module directly

src/gluonts/model/forecast_generator.py
```diff
@@ -40,14 +39,55 @@ def log_once(msg):
...
+# different deep learning frameworks generate predictions and the tensor to numpy conversion differently,
+# use a dispatching function to prevent needing a ForecastGenerators for each framework
+@singledispatch
+def predict_to_numpy(prediction_net, tensor) -> np.ndarray:
+    raise NotImplementedError
+
+
+@singledispatch
+def recursively_zip_arrays(x) -> Iterator:
+    raise NotImplementedError
```
lostella, line 51: Docstring missing here

src/gluonts/model/forecast_generator.py
```diff
@@ -20,24 +20,81 @@
...
     raise NotImplementedError
 
 
-def log_once(msg):
-    global LOG_CACHE
-    if msg not in LOG_CACHE:
-        logging.info(msg)
-        LOG_CACHE.add(msg)
+@singledispatch
+def recursively_zip_arrays(x) -> Iterator:
```
mbohlkeschneider, line 52: Nice!

mbohlkeschneider (approved):
Really nice. Lots of stuff here. Maybe another pair of eyes would help :-). Otherwise, I'm all in for merging and iterating...

kashif:
awesome! next we can do a jax backend too 😃

PR #1267, opened 2021-01-14
Add to_quantile_forecast method to Forecast classes

yx1215:
…tion forecast to quantile forecast

*Issue #, if available:*

method that allows conversion from sample forecast to quantile forecast and distribution forecast to quantile forecast might be helpful in developing AutoGluon for GluonTS as quantile forecasts are easier to be shown as a data frame.

*Description of changes:*
Add two methods: 
SampleForecast.to_quantile_forecast
DistributionForecast.to_quantile_forecast

### DIFF
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -455,14 +455,16 @@ class SampleForecast(Forecast):
             ]
         )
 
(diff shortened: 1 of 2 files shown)


lostella (changes requested):
@yx1215 thanks, looks good! you just need to apply `black`:

```
black src
```

make sure you have `black==20.8b1` installed

lostella (approved):

PR #1272, opened 2021-01-18
Add distribution output classes for PyTorch, add tests

lostella:
cc @kashif 

### DIFF
--- a/test/torch/modules/test_torch_distribution_inference.py
+++ b/test/torch/modules/test_torch_distribution_inference.py
@@ -268,14 +268,12 @@ def test_poisson(rate: float) -> None:
     ), f"rate did not match: rate = {rate}, rate_hat = {rate_hat}"
 
 
-@pytest.mark.parametrize("total_count_logit", [(2.5, 0.7)])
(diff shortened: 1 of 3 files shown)


kashif:
Looks fine

test/torch/modules/test_torch_distribution_inference.py
```diff
@@ -165,3 +240,64 @@ def test_studentT_likelihood(df: float, loc: float, scale: float):
...
+        learning_rate=0.05,
+    )
+
+    assert (
+        np.abs(rate_hat - rate) < TOL * rate
+    ), f"rate did not match: rate = {rate}, rate_hat = {rate_hat}"
+
+
+@pytest.mark.parametrize("total_count_logit", [(2.5, 0.7)])
+def test_neg_binomial(total_count_logit: Tuple[float, float]) -> None:
```
lostella, line 272: Looks like these test parameters are the same as for the mxnet case, but here the parametrization is different

canerturkmen (approved):

PR #1281, opened 2021-01-21
Added thetaf method to the R predictor

prolearner:
*Description of changes:*
Added the [thetaf method](https://rdrr.io/cran/forecast/man/thetaf.html) to the R predictor. The method does not require additional R libraries (it's in the forecast library) but the implementation works differently than the other methods: it returns point forecasts and prediction intervals. Due to this difference, when `ouput_types="samples"` the same forecast is copied `num_samples` times.

I tested the method on the M3 dataset (only with `ouput_types="samples"`) and was able to get an sMAPE of 13.06, which is very similar to the one in [the paper discussing the M3 results](https://www.forecastingprinciples.com/files/pdf/Makridakia-The%20M3%20Competition.pdf) (13.01).


vafl (approved):
Great! Thanks a lot!

PR #1299, opened 2021-02-01
Fix serde for np.dtype.

jaheba:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1300, opened 2021-02-01
Merges for v0.6.5

lostella:
* Fix serde for np.dtype. (#1299)

### DIFF
--- a/src/gluonts/core/serde/np.py
+++ b/src/gluonts/core/serde/np.py
@@ -27,7 +27,7 @@ def encode_np_dtype(v: np.dtype) -> Any:
     """
     return {
         "__kind__": Kind.Instance,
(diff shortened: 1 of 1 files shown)


Schmedu (approved):

kashif:
I would appreciate all but the last couple of breaking changes in a release... then I can make a new cut of pytorch-ts which depends on gluon-ts. Else I would need to also go over and update my estimators. Any idea what you plan to include here?

lostella:
@kashif we only include bug fixes in bugfix releases (no new features).

I assume you are referring to #1256? Do you have any concern about that?

lostella:
(This particular backport is rather urgent)

kashif:
No concern... just didn’t get around to integrating it.

Issue #1313, opened 2021-02-05
dropout not applied during training


geoalgo:
## Description
I suspect that the training mode is set correctly when training our models. Normally, the training mode needs to be specified as some layers behave differently (typically dropout).

For instance:

```python
with autograd.train_mode():
  # dropout is applied
  net(x)

with autograd.predict_mode():
  # dropout is not applied (inference)
  net(x)

```

By default, mxnet assume that it is in predict_mode which means that dropout is not applied if the model is not told specifically that it is in training mode. Consequently, my understanding is that dropout is not applied currently in our model.

Reference: https://gluon.mxnet.io/chapter03_deep-neural-networks/mlp-dropout-gluon.html

## To Reproduce

I checked the result of the forward pass in DeepAR and zone-out cell in trainer/_base.py. 
The result of the forward loop is always the same but by doing `with autograd.train_mode()`, it is varying as expected.

lostella:
@geoalgo thanks! Should be easy to fix in the Trainer class I guess?

geoalgo:
Yes, I can send a PR that modifies `trainer/_base.py` as follow:

```python
if is_training:
    with autograd.train_mode():
        output = net(*batch.values())
else:
    with autograd.predict_mode():
        output = net(*batch.values())

```

lostella:
Looks like this applies also to the v0.6.x branch (see [here](https://github.com/awslabs/gluon-ts/blob/b3f1a64264448189865520dd7cd7fe91261652f7/src/gluonts/mx/trainer/_base.py#L266)) so we should backport the fix

geoalgo:
Yes, I did another test to be sure where I did a quick benchmark of electricity CRPS:

* DeepAR default dropout, current code: 0.06330868858104806
* DeepAR dropout=0.99, current code: 0.10863092213276618 -> no effect
* DeepAR dropout=0.99, fixed code: 0.35484217043263466  -> worse prediction as it should

lostella:
@geoalgo the difference here is just noise then?

>     * DeepAR default dropout, current code: 0.06330868858104806
>     * DeepAR dropout=0.99, current code: 0.10863092213276618 -> no effect

geoalgo:
Yes it is training with little epochs.

PR #1315, opened 2021-02-05
fixing bug where dropout was not used, also remove unused halt option

geoalgo:
Issue #1313 :

Description of changes:

* Fixing bug that caused dropout to not be called
* Remove unused and dead code halt

The PR is the same than https://github.com/awslabs/gluon-ts/pull/1314 except that I used recent version (I accidentally used one that was from one year ago and the merge was too painful).


lostella (approved):
Looks good! Thanks again!

PR #1319, opened 2021-02-09
Enable test workflow on Python 3.8 and 3.9

lostella:
This is to track test results on newer Python versions


.github/workflows/tests.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       max-parallel: 4
       fail-fast: false
       matrix:
-        python-version: [3.6, 3.7]
+        python-version: [3.6, 3.7, 3.8, 3.9]
```
jaheba, line 12:
```suggestion
        python-version: [3.6, 3.7, 3.8]
```

jaheba (commented):
Should we start enabling `3.8` since that one appears to work?

jaheba (approved):

PR #1323, opened 2021-02-11
Set default num_samples, shorten some import paths

lostella:
* Set a default `num_samples` in `make_evaluation_predictions`, which is annoying to set every time even when it is not needed.
* Fix docstrings
* Shorten import paths across the `gluonts.evaluation` package (backward compatible, old paths still available, but references to them are removed here)

### DIFF
--- a/src/gluonts/evaluation/backtest.py
+++ b/src/gluonts/evaluation/backtest.py
@@ -34,13 +34,16 @@ from gluonts.transform import TransformedDataset
 
 
 def make_evaluation_predictions(
-    dataset: Dataset, predictor: Predictor, num_samples: int
+    dataset: Dataset,
+    predictor: Predictor,
+    num_samples: Optional[int] = 100,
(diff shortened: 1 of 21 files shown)


src/gluonts/evaluation/backtest.py
```diff
@@ -34,13 +34,16 @@
 
 
 def make_evaluation_predictions(
-    dataset: Dataset, predictor: Predictor, num_samples: int
+    dataset: Dataset,
+    predictor: Predictor,
+    num_samples: Optional[int] = 100,
```
jaheba, line 39: Is this really `Optional`?

lostella: Some predictor may not need it. But maybe that's in the sense of "not have it passed at all", so this could just be an `int`.

jaheba (approved):

jaheba (approved):

jaheba (approved):

Issue #750, opened 2020-04-13
Prediction index is wrong with weekly frequency


mbataillou:
## Description
Gluon-ts uses pandas' date_range function to create an index this will change the start day and shift the following ones.

## To Reproduce

```python
import numpy as np

from gluonts.dataset.common import ListDataset
from gluonts.distribution import MultivariateGaussianOutput
from gluonts.model.deepar import DeepAREstimator
from gluonts.trainer import Trainer
from gluonts.evaluation.backtest import make_evaluation_predictions


train_dataset = ListDataset(
    data_iter=[
        {
            "start": "2019-01-01 00:00:00",
            "target": np.ones(shape=(4, 4)),
        },
    ],
    freq="W",
    one_dim_target=False,
)

test_dataset = ListDataset(
    data_iter=[
        {
            "start": "2019-01-01 00:00:00",
            "target": np.ones(shape=(4, 5)),
        },
    ],
    freq="W",
    one_dim_target=False,
)

estimator = DeepAREstimator(
    'W', prediction_length=1, trainer=Trainer(epochs=3, hybridize=False),
    distr_output=MultivariateGaussianOutput(dim=4),
)

predictor = estimator.train(train_dataset)

forecast_it, ts_it = make_evaluation_predictions(dataset=test_dataset, predictor=predictor, num_samples=10)

forecast_list = list(forecast_it)
ts_list = list(ts_it)
```

## Error message or code output
Where ts_list starts on the first sunday it found: '2019-01-06' instead of the provided start date: '2019-01-01'

```
DatetimeIndex(['2019-01-06', '2019-01-13', '2019-01-20', '2019-01-27',
               '2019-02-03'],
              dtype='datetime64[ns]', freq='W-SUN')
```

## Environment
- Operating system: Amazon Linux
- Python version: 3.6
- GluonTS version: a5539612 internal

vafl:
Gluonts aligns time stamps to the pandas frequency. Since "W" is short for "W-SUN", all time stamps in the dataset are aligned to sundays.  We can relatively easily include support for [other pandas frequencies](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases) such as "W-TUE", which would be weekly time series with time stamps falling on tuesdays.

What we cannot easily support are a dataset with weekly time series where each series can start on a different day, because the feature processing relies on global time stamps that all time series align with.

Would the inclusion of additional frequencies like "W-TUE" solve your problem?

[...]

PR #1326, opened 2021-02-11
Normalized frequency-name.

kashif:
can now specify any freq="W-..." string for issue #750

*Description of changes:*

just look at the first part of the week freq string

### DIFF
--- a/src/gluonts/time_feature/lag.py
+++ b/src/gluonts/time_feature/lag.py
@@ -16,6 +16,7 @@ from typing import List, Optional
 import numpy as np
 from pandas.tseries.frequencies import to_offset
 
+from gluonts.time_feature import norm_freq_str
 
 def _make_lags(middle: int, delta: int) -> np.ndarray:
     """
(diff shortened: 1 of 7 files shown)


jaheba:
Thanks!

I like the idea. Maybe it's safer to split the name on `-` and then use the first part:

```python
In [14]: "W".split("-")[0]
Out[14]: 'W'

In [15]: "W-SUN".split("-")[0]
Out[15]: 'W'
```

There appear to be some frequencies with names longer than one character (e.g. `SM` for semi-month).

So we could have something like `norm_freq()` to get that value.

src/gluonts/time_feature/lag.py
```diff
@@ -83,16 +85,22 @@ def _make_lags_for_month(multiple, num_cycles=3):
     # multiple, granularity = get_granularity(freq_str)
     offset = to_offset(freq_str)
 
-    if offset.name == "A-DEC":  # normalized string for YearEnd aka Y
+    if (
```
jaheba, line 88: Can you put these comments on the line before? Black does some weird formatting otherwise :|

jaheba (commented):
Looks good. Just one minor thing :)

jaheba (approved):

jaheba:
@kashif Thanks a lot!

PR #1328, opened 2021-02-12
Fix serialization of lead_time

lostella:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/mx/model/predictor.py
+++ b/src/gluonts/mx/model/predictor.py
@@ -178,6 +178,13 @@ class GluonPredictor(Predictor):
         if type(self) != type(that):
             return False
 
+        if not equals(self.freq, that.freq):
+            return False
+        if not equals(self.prediction_length, that.prediction_length):
+            return False
+        if not equals(self.lead_time, that.lead_time):
+            return False
+
         # TODO: also consider equality of the pipelines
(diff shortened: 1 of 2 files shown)


src/gluonts/mx/model/predictor.py
```diff
@@ -178,6 +178,13 @@ def __eq__(self, that):
         if type(self) != type(that):
             return False
 
+        if not equals(self.freq, that.freq):
+            return False
+        if not equals(self.prediction_length, that.prediction_length):
+            return False
+        if not equals(self.lead_time, that.lead_time):
+            return False
```
lostella, line 186: this is to catch the issue

test/model/lstnet/test_lstnet.py
```diff
@@ -84,6 +87,12 @@ def test_lstnet(
     )
 
     predictor = estimator.train(dataset.train)
+
+    with tempfile.TemporaryDirectory() as directory:
+        predictor.serialize(Path(directory))
+        predictor_copy = Predictor.deserialize(Path(directory))
+        assert predictor == predictor_copy
```
lostella, line 94: this would catch the issue because of the difference in `lead_time`

jaheba (approved):

PR #1345, opened 2021-02-26
fix edge case in iteration based model averaging

benidis:
*Description of changes:*

Edge case for the `NTA` iteration-based model averaging. If the averaging is not triggered the model that is returned is the one of the last epoch which is not even the best regardless of averaging.

With this fix there is a fallback clause so the last alpha epochs will be averaged eitherway (as in `Alpha_Suffix`) if the averaging is not triggered earlier.


test/trainer/test_model_iteration_averaging.py
```diff
@@ -73,8 +73,13 @@ def initialize_model() -> nn.HybridBlock:
 def test_NTA_V1(n: int, last_n_trigger: bool):
     model = initialize_model()
     params = model.collect_params()
-    avg_strategy = NTA(n=n, last_n_trigger=last_n_trigger)
     loss_list = [5, 4, 3, 2, 3, 3, 3, 3, 3, 3, 3]
+    avg_strategy = NTA(
```
mbohlkeschneider, line 77: Does this test actually reflect that edge case? If not, can we add it?

mbohlkeschneider (approved):
LGTM. Minor comment to add the edge case to the test (if this hasn't been done already).

benidis:
The tests do not reflect the edge case. Will try to fix them asap.

PR #1348, opened 2021-02-27
Typo in import

miguelgfierro:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ We also specify some minimal training options.
 
 ```python
 from gluonts.model.deepar import DeepAREstimator
-from gluonts.mx.trainer import Trainer
+from gluonts.trainer import Trainer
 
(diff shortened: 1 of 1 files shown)


lostella:
Hi @miguelgfierro! I think the import path is actually correct, since most mxnet-specific modules (among which `trainer`) were moved under `gluonts.mx` since `v0.6.x`. In fact, in the `master` branch there is https://github.com/awslabs/gluon-ts/tree/master/src/gluonts/mx/trainer

So the content of README.md probably won’t work on older releases: you can find its content for the `v0.5.x` branch here https://github.com/awslabs/gluon-ts/tree/v0.5.x

miguelgfierro:
got it, no problem, I found this error when trying the tutorial with 0.5.0

PR #1365, opened 2021-03-16
Added lightgbm support for QRX/Rotbaum

zoolhasson:
*Issue #, if available:*

*Description of changes:* Added lightgbm support for QRX/Rotbaum. Instantiate a an lgb_wrapper object, and then feed into the model parameter in QRX/Rotbaum.

### DIFF
--- a/src/gluonts/model/rotbaum/_model.py
+++ b/src/gluonts/model/rotbaum/_model.py
@@ -13,6 +13,7 @@
 
 from typing import Dict, List, Optional
 
+import copy
 import numpy as np
 import pandas as pd
 import xgboost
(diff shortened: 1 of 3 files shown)


src/gluonts/model/rotbaum/_model.py
```diff
@@ -99,7 +100,7 @@ def __init__(
             true values associated with each prediction.
         """
         if model:
-            self.model = model
+            self.model = copy.deepcopy(model)
```
dcmaddix, line 103: Do we need a `deepcopy` here?

zoolhasson: This is copying an untrained model, so it's not a drain on resources. We do need it do deepcopy, because Rotbaum will call multiple QRX models (one for each timestep in the future), and we want each of them to be a different model trained on different data.

dcmaddix (approved):

PR #1372, opened 2021-03-19
Made dealing with pandas dataframes native to QRX

zoolhasson:
*Issue #, if available:*

*Description of changes:* This allows removing the lightgbm wrapper. The alternative was that QRX turns whatever it gets (including pandas dataframes) into numpy arrays and then the lightgbm wrapper turns the numpy array back to pandas dataframes. Other then being too convoluted, this causes issues with the dtypes changing.


src/gluonts/model/rotbaum/_model.py
```diff
@@ -124,29 +124,49 @@ def _create_xgboost_model(model_params: Optional[dict] = None):
...
             If not None, choose without replacement
             min(max_sample_size, len(x_train)) many datapoints to train on.
         seed: int
             seed for sampling purposes
+        x_train_is_dataframe: bool
+            This should be False for XGBoost, but True if one uses lightgbm
         """
         self.quantile_dicts = {}
-        x_train, y_train = np.array(x_train), np.array(y_train)  # xgboost
+        if x_train_is_dataframe == False:
```
jaheba, line 157:
```suggestion
        if not x_train_is_dataframe:
```

src/gluonts/model/rotbaum/_model.py
```diff
@@ -158,8 +178,13 @@ def fit(self, x_train, y_train, max_sample_size=None, seed=1, **kwargs):
         self.model.fit(x_train, y_train, **kwargs)
         y_train_pred = self.model.predict(x_train)
         self.df = pd.DataFrame(
-            {"x": list(x_train), "y_true": y_train, "y_pred": y_train_pred}
-        )
+            {
+                "x": list(np.array(x_train)),  # this uniformly deals with
```
jaheba, line 182:
```suggestion
                "x": np.array(x_train).tolist(),  # this uniformly deals with
```

AFAIK, you generally want to use `.tolist()`, since it also works for multi-dimensional arrays.

I think using `list(...)` is a bit of a hack, since it just iterates over the array and then collects the elements.

```py
In [5]: np.arange(9).reshape((3, 3)).tolist()
Out[5]: [[0, 1, 2], [3, 4, 5], [6, 7, 8]]

In [6]: list(np.arange(9).reshape((3, 3)))
Out[6]: [array([0, 1, 2]), array([3, 4, 5]), array([6, 7, 8])]
```

src/gluonts/model/rotbaum/_model.py
zoolhasson: Done.

src/gluonts/model/rotbaum/_model.py
zoolhasson: Actually, it is not necessary to keep track of x, and it just makes the models more bulky. I dropped this altogether.

requirements/requirements-mxnet.txt
```diff
@@ -1 +1 @@
-mxnet~=1.7
+mxnet==1.7.*
```
mbohlkeschneider, line 1: I think we just changed the requirements to mxnet in another PR and this should be aligned.

src/gluonts/model/rotbaum/_model.py
```diff
@@ -124,29 +125,50 @@ def _create_xgboost_model(model_params: Optional[dict] = None):
             }
         return xgboost.sklearn.XGBModel(**model_params)
 
-    def fit(self, x_train, y_train, max_sample_size=None, seed=1, **kwargs):
+    def fit(
```
mbohlkeschneider, line 128: Can we add type annotations?

src/gluonts/model/rotbaum/_model.py
```diff
@@ -124,29 +125,50 @@ def _create_xgboost_model(model_params: Optional[dict] = None):
...
         Fits self.model and partitions R^n into cells. More accurately,
         it creates a dictionary taking predictions on train to lists of
         associated true values, and puts it in self.cell_values.
 
         Parameters
         ----------
-        x_train: list
-            list of lists
-        y_train: list
+        x_train: pd.DataFrame (if x_train_is_dataframe) or list (else)
```
mbohlkeschneider, line 144: Maybe move type descriptions as type annotations into the arguments?

mbohlkeschneider (changes requested):

requirements/requirements-mxnet.txt
jaheba: This looks a bit strange, since this should be the same as what is on master, and thus doesn't show in the diff.

src/gluonts/model/rotbaum/_model.py
zoolhasson: Done.

src/gluonts/model/rotbaum/_model.py
zoolhasson: Done.

requirements/requirements-mxnet.txt
zoolhasson: Yes, I'm not sure what that was about. But with the new commits this longer appears as a change.

zoolhasson:
It looks like Docs is failing for reasons that are not relevant to the PR.

jaheba (dismissed):

dcmaddix (approved):

src/gluonts/model/rotbaum/_model.py
mbohlkeschneider: Sorry, maybe I was not clear. I think type annotations in the function definition are better than in the doc string and as far as I understand, this is what we converged to in GluonTS.

mbohlkeschneider:
Something is weird here. There seem to be changes that are already merged that show up in this PR. Can you maybe resolve this?

mbohlkeschneider (changes requested):
Can we clean up the attached text files? There are two readme files, for example. Are they all intended? Also, there are several blank files we should remove from this PR.

mbohlkeschneider (approved):

Issue #1054, opened 2020-09-25
#898 causes cuda initialization error


szhengac:
When I submitted a sagemaker job with the latest gluonts master, I received cuda initialization error in https://github.com/awslabs/gluon-ts/blob/3ff95ebcf70e39f6424d76b9b58bd861b36f7363/src/gluonts/mx/trainer/_base.py#L242. After I remove the commit #898, everything goes well.

lostella:
@szhengac do you think you could isolate the problem into a MWE?

lostella:
Can reproduce with the following:

```python
import mxnet as mx

from gluonts.dataset.repository.datasets import get_dataset
from gluonts.model.simple_feedforward import SimpleFeedForwardEstimator
from gluonts.trainer import Trainer

dataset = get_dataset("electricity", regenerate=False)

estimator = SimpleFeedForwardEstimator(freq="H", prediction_length=24, trainer=Trainer(epochs=3, ctx=mx.gpu()))

predictor = estimator.train(dataset.train, num_workers=2)
```

which fails on master but succeeds in c3150eb7b3f58181ddda385a2ef273349003c782 (i.e. prior to #898)

PascalIversen:
The source of the error seems to be placing any data on the gpu at the `with mx.gpu()` statement.
This reproduces the error:

```python
import mxnet as mx
from gluonts.mx.batchify import batchify
from gluonts.dataset.loader import TrainDataLoader
from gluonts.dataset.repository.datasets import get_dataset
from functools import partial
from gluonts.transform import (
    InstanceSplitter,
    ExpectedNumInstanceSampler
)
from gluonts.dataset.field_names import FieldName
from gluonts.mx.batchify import batchify 
from gluonts.dataset.loader import TrainDataLoader


with mx.gpu():
    a =mx.nd.array([42])
    
dataset = get_dataset("electricity", regenerate=False)



transformation = InstanceSplitter(
                target_field=FieldName.TARGET,
                is_pad_field=FieldName.IS_PAD,
                start_field=FieldName.START,
                forecast_start_field=FieldName.FORECAST_START,
                train_sampler=ExpectedNumInstanceSampler(num_instances=2),
                past_length=24,
                future_length=12)



batch_size = 32
num_batches_per_epoch = 300

data_loader = TrainDataLoader(dataset.train, batch_size=batch_size, stack_fn=partial(batchify, ctx=mx.gpu()),
                              transform=transformation,
[...]

PR #1376, opened 2021-03-23
Fix dockerfiles

arangatang:
*Description of changes:*
Some of the [dockerfiles](examples/dockerfiles) cannot be built/does not execute properly, this PR fixes this. 

These are the problems which this PR addresses:

1. The GPU image crashes on SageMaker due to MXNET_USE_FUSION being enabled. 
2. The GPU image crashes when passing `num_workers` as a hyperparameter. 
3. The mxnet version in the requirements file is unavailable when building the [R image](examples/dockerfiles/Dockerfile.R) so the R image fails to build.

*Problem 1* 

This occurs when the gpu image is run. 
For example running the SimpleFeedForwardEstimator on the electricity dataset with these hyperparameters:
```
epochs	1
forecaster_name	gluonts.model.simple_feedforward.SimpleFeedForwardEstimator
freq	D
prediction_length	7
```

Results in this error (removed redundant parts)

```
AlgorithmError: Traceback (most recent call last): 
File "src/operator/fusion/fused_op.cu", line 647 MXNetError: 
Check failed: compileResult == NVRTC_SUCCESS (5 vs. 0) : NVRTC Compilation failed. 
Please set environment variable MXNET_USE_FUSION to 0.
. . .
```
Following the advice in the logs and setting `MXNET_USE_FUSION=0` fixes the issue.

**Problem 2** (requires problem 1 to be fixed beforehand)
Executing an algorithm with num_workers set to a value causes a CUDA initializtion error. 
A fix for a similar issue was proposed by @lostella for #1054 applying this fix when `num_workers` has been passed fixes problem 2. [link](https://github.com/awslabs/gluon-ts/issues/1054#issuecomment-706161305)

**problem 3**
The latest available mxnet version which can be installed in the R image is mxnet 1.6.0 but the current [mxnet requirement](requirements/requirements-mxnet.txt) is 1.7.0 which breaks the build. Changing the installation requirement to `>=1.6.0` fixes the issue.

I tested the changes on the SimpleFeedForwardEstimator as well as the DeepFactorEstimator on a `ml.g4dn.2xlarge` instance on SageMaker. 

### DIFF
--- a/src/gluonts/shell/train.py
+++ b/src/gluonts/shell/train.py
@@ -13,6 +13,7 @@
 
 import json
 import logging
+import multiprocessing
 from typing import Any, Optional, Type, Union
 
 import gluonts
@@ -90,6 +91,8 @@ def run_train(
         if "num_workers" in hyperparameters.keys()
         else None
     )
+    if num_workers:
+        multiprocessing.set_start_method("spawn", force=True)
     shuffle_buffer_length = (
         int(hyperparameters["shuffle_buffer_length"])
         if "shuffle_buffer_length" in hyperparameters.keys()
--- a/src/gluonts/shell/train.py
+++ b/src/gluonts/shell/train.py
@@ -32,6 +32,7 @@ from gluonts.transform import FilterTransformation, TransformedDataset
 
 from .env import TrainEnv
 
+multiprocessing.set_start_method("spawn", force=True)
 logger = logging.getLogger(__name__)
 
 
@@ -91,8 +92,6 @@ def run_train(
         if "num_workers" in hyperparameters.keys()
         else None
     )
-    if num_workers:
-        multiprocessing.set_start_method("spawn", force=True)
     shuffle_buffer_length = (
         int(hyperparameters["shuffle_buffer_length"])
         if "shuffle_buffer_length" in hyperparameters.keys()
--- a/examples/dockerfiles/Dockerfile.gpu
+++ b/examples/dockerfiles/Dockerfile.gpu
@@ -16,5 +16,6 @@ ADD . /gluonts
 
 RUN pip3.7 install mxnet-cu92>=1.6.0
 RUN pip3.7 install /gluonts[shell]
+ENV MXNET_USE_FUSION=0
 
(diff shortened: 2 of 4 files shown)


src/gluonts/shell/train.py
```diff
@@ -90,6 +91,8 @@ def run_train(
         if "num_workers" in hyperparameters.keys()
         else None
     )
+    if num_workers:
+        multiprocessing.set_start_method("spawn", force=True)
```
lostella, line 95: Maybe minor, but still: this `set_start_method` could be invoked at the beginning of the module, right after the imports, regardless of the number of workers. (If no workers are used, then no subprocess will be launched, so it doesn't matter what start method is being used)

arangatang: True, I moved it!

arangatang:
It seems as if the setting the multiprocessing method to spawn introduces some additional issues. Maybe this fix only applies to GPUs, ill look into it further.

mbohlkeschneider (commented):
Are here any concerns @jaheba, @lostella? Makes sense to me.

lostella:
Looks good to me, let's wait for @arangatang to report on the additional issues he found

arangatang:
```
multiprocessing.set_start_method("spawn", force=True)
``` 
Caused the shell test to fail when a process is forked in the testutils/shell package on [line 139](https://github.com/awslabs/gluon-ts/blob/97ea08cb394f1c0d411234c1b33beff37ae30cd0/src/gluonts/testutil/shell.py#L139). 

Explicitly starting that process as a fork resolves the issue while still allowing gpu acceleration in the shell.

**Tested on**
mxnet:
mxnet-cu101 1.7.0
mxnet 1.7.0

GPU:
Titan RTX

NVCC:
version 10.1

arangatang:
I recon the changes are good to go! 
What do you think?

@lostella @mbohlkeschneider

arangatang:
I believe the failing CodeQL test is unrelated to my changes, should I merge it?

jaheba:
Yes, let's merge it.

PR #1380, opened 2021-03-26
Deepar imputation model

mbohlkeschneider:
*Description of changes:*

This introduces the "imputation mode" for DeepAR. This means when unrolling the LSTM, DeepAR will draw samples from the current model parameters to generate a sample mean at time `i` that will be fed to the later time steps `i+n` through the lags. Effectively, we allow DeepAR impute values with the current model parameters, therefore using DeepAR itself as the imputation model. Note this will only impute values in `context` and `prediction` range. Other target values (that might be drawn as lags) will still be imputed by the set `ImputationMethod` in the estimator. 

The vanilla behavior is not changed and the imputation model can be invoked by setting `impute_missing_values=True` in the `DeepAREstimator`

Other changes: 
#### Default scale for MeanScaler
Allows for setting a `default_scale` in case the entire context range is unobserved. This might make training more stable when lots of empty context ranges are drawn because the scale will be fixed and does not vary every batch (using the mean scale of the batch is default behavior. 

#### Include zeros in denominator for weighted average 
Currently, we do NOT include the zeros in the denominator when computing the weighted average. For sparse time series, that can lead to spiky losses when only a single observation is present in the context + prediction range during training and that single observation is not modeled well. Including the zeros in the denominator leads to smoother training, effectively giving time series with few observations a smaller contribution in the total loss. 

#### LaplaceFixedVarianceOutput
Added `LaplaceFixedVarianceOutput` to have a deterministic L1 distribution. 


src/gluonts/model/deepar/_estimator.py
```diff
@@ -140,6 +140,19 @@ class DeepAREstimator(GluonEstimator):
...
+        The minimum scale that is returned by the MeanScaler
+    default_scale
+        Default scale that is applied if the context length window is
+        completely unobserved. If not set, the scale in this case will be
+        the mean scale in the batch.
+    impute_missing_values
+        Whether to impute the missing values during training by using the
+        current model parameters. Recommended if the dataset contains many
+        missing values. However, this is a lot slower than the default mode.
+    imputation_samples
```
melopeo, line 153: Maybe we should change this name.
Currently by *imputation_samples* I understand that this variable contains samples.
What about *num_imputation_samples* or something alike?

mbohlkeschneider: > Maybe we should change this name.
> Currently by _imputation_samples_ I understand that this variable contains samples.
> What about _num_imputation_samples_ or something alike?

Done

src/gluonts/model/deepar/_network.py
```diff
@@ -39,6 +40,44 @@ def prod(xs):
     return p
 
 
+def mx_switch(F, *args, **kwargs):
+    """
+    A switch statement for mxnet
+
```
shubhamkapoor, line 46: Shouldn't we move mx_switch to mx/utils.py? It will be clean in a case other models use it.

mbohlkeschneider: Done.

shubhamkapoor (dismissed):

shubhamkapoor (approved):

melopeo (approved):

PR #1384, opened 2021-03-31
Update forecast_methods.R

shruti32:
I have added set.seed(123) to the existing models to remove randomness from the generated results.

*Issue #, if available:*

*Description of changes:*


mbohlkeschneider (changes requested):
I think hard-coding the seed is not what we want. Having that as an optional parameter is fine but hard-coding it increases the risk of overfitting a model to a particular seed.

PR #1385, opened 2021-04-01
Allow serialization and deserialization for auto tabular predictor

yx1215:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/nursery/autogluon_tabular/estimator.py
+++ b/src/gluonts/nursery/autogluon_tabular/estimator.py
@@ -46,9 +46,6 @@ class TabularEstimator(Estimator):
         Frequency of the data to handle
     prediction_length
         Prediction length
-    ag_path_prefix
-        path prefix set to where the ag_model will be saved. If None, ag_model will be saved to
-        default directory 'ag_tabular_models/'
     lag_indices
         List of indices of the lagged observations to use as features. If
         None, this will be set automatically based on the frequency.
--- a/src/gluonts/nursery/autogluon_tabular/estimator.py
+++ b/src/gluonts/nursery/autogluon_tabular/estimator.py
@@ -12,10 +12,11 @@
 # permissions and limitations under the License.
 
 from typing import Callable, Optional, List, Tuple
-
+from pathlib import Path
 import pandas as pd
-from autogluon import TabularPrediction as task
+from autogluon.tabular import TabularPredictor
 
+from gluonts.core.component import validated
 from gluonts.dataset.common import Dataset
 from gluonts.dataset.util import to_pandas
 from gluonts.model.estimator import Estimator
@@ -26,7 +27,7 @@ from gluonts.time_feature import (
 )
 
 from .predictor import (
-    TabularPredictor,
+    GluonTSTabularPredictor,
     mean_abs_scaling,
     get_features_dataframe,
 )
@@ -65,10 +66,14 @@ class TabularEstimator(Estimator):
         This will make predictions more efficient, but may impact their accuracy.
     """
 
+    ag_path = Path('ag_tabular_models')
+
+    @validated()
     def __init__(
         self,
         freq: str,
         prediction_length: int,
+        path: Optional[Path] = None,
         lag_indices: Optional[List[int]] = None,
         time_features: Optional[List[TimeFeature]] = None,
         scaling: Callable[
@@ -80,9 +85,10 @@ class TabularEstimator(Estimator):
     ) -> None:
         super().__init__()
 
-        self.task = task
+        self.task = TabularPredictor
         self.freq = freq
         self.prediction_length = prediction_length
+        self.path = path
         self.lag_indices = (
             lag_indices
             if lag_indices is not None
@@ -103,7 +109,7 @@ class TabularEstimator(Estimator):
             ]
 
         default_kwargs = {
-            "eval_metric": "mean_absolute_error",
+            "time_limit": 10,
             "excluded_model_types": ["KNN", "XT", "RF"],
             "presets": [
                 "high_quality_fast_inference_only_refit",
@@ -112,7 +118,7 @@ class TabularEstimator(Estimator):
         }
         self.kwargs = {**default_kwargs, **kwargs}
 
-    def train(self, training_data: Dataset) -> TabularPredictor:
+    def train(self, training_data: Dataset) -> GluonTSTabularPredictor:
         dfs = [
             get_features_dataframe(
                 series=self.scaling(to_pandas(entry))[0],
@@ -122,10 +128,15 @@ class TabularEstimator(Estimator):
             for entry in training_data
         ]
         df = pd.concat(dfs)
-        ag_model = self.task.fit(
-            df, label="target", problem_type="regression", **self.kwargs
+        if self.path is not None:
+            self.path = self.path / self.ag_path
+        else:
+            self.path = self.ag_path
+
+        ag_model = self.task(path=self.path, label="target", problem_type="regression", eval_metric="mean_absolute_error").fit(
+            df, **self.kwargs
         )
-        return TabularPredictor(
+        return GluonTSTabularPredictor(
             ag_model=ag_model,
             freq=self.freq,
             prediction_length=self.prediction_length,
@@ -133,4 +144,6 @@ class TabularEstimator(Estimator):
             lag_indices=self.lag_indices,
             scaling=self.scaling,
             batch_size=self.batch_size,
+            ag_path=self.path
         )
+
--- a/src/gluonts/nursery/autogluon_tabular/estimator.py
+++ b/src/gluonts/nursery/autogluon_tabular/estimator.py
@@ -14,7 +14,7 @@
 from typing import Callable, Optional, List, Tuple
 from pathlib import Path
 import pandas as pd
-from autogluon.tabular import TabularPredictor
+from autogluon import tabular
 
 from gluonts.core.component import validated
 from gluonts.dataset.common import Dataset
@@ -27,7 +27,7 @@ from gluonts.time_feature import (
 )
 
 from .predictor import (
-    GluonTSTabularPredictor,
+    TabularPredictor,
     mean_abs_scaling,
     get_features_dataframe,
 )
@@ -47,6 +47,9 @@ class TabularEstimator(Estimator):
         Frequency of the data to handle
     prediction_length
         Prediction length
+    ag_path_prefix
+        path prefix set to where the ag_model will be saved. If None, ag_model will be saved to
+        default directory 'ag_tabular_models/'
     lag_indices
         List of indices of the lagged observations to use as features. If
         None, this will be set automatically based on the frequency.
@@ -73,7 +76,7 @@ class TabularEstimator(Estimator):
         self,
         freq: str,
         prediction_length: int,
-        path: Optional[Path] = None,
+        ag_path_prefix: Optional[Path] = None,
         lag_indices: Optional[List[int]] = None,
         time_features: Optional[List[TimeFeature]] = None,
         scaling: Callable[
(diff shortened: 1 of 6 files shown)


lostella (changes requested):
Thanks! I've left some comments inline.

Tests are failing, see https://github.com/awslabs/gluon-ts/runs/2245357197#step:5:17 there seem to be a problem with some autogluon import, maybe the version we require needs to be bumped?

Also, it would be nice to update the tests we have to include serialization/deserialization, see test/nursery/test_autogluon_tabular.py

src/gluonts/nursery/autogluon_tabular/estimator.py
```diff
@@ -103,7 +109,7 @@ def __init__(
             ]
 
         default_kwargs = {
-            "eval_metric": "mean_absolute_error",
+            "time_limit": 10,
```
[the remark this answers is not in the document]
yx1215: This makes it easier for testing, I will change the default back when everything else is done.

src/gluonts/nursery/autogluon_tabular/predictor.py
```diff
@@ -333,3 +338,58 @@ def predict(
...
+
+    @classmethod
+    def deserialize(
+        cls,
+        path: Path,
+        scaling: Callable[
+            [pd.Series], Tuple[pd.Series, float]
+        ] = mean_abs_scaling,
+        # ctx: Optional[mx.Context] = None
+        ag_path: Optional[Path] = None,
```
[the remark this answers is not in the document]
yx1215: Right, we don't need this as this can always be got by deserializing the parameters. I'm going to delete it.

yx1215:
> Thanks! I've left some comments inline.
> 
> Tests are failing, see https://github.com/awslabs/gluon-ts/runs/2245357197#step:5:17 there seem to be a problem with some autogluon import, maybe the version we require needs to be bumped?
> Yes, we need to bump the dependency for autogluon.tabular to 0.1.0 as the API has changed. We no longer has a task in autogluon.tabular, TabularPredictor is used directly instead.

> Also, it would be nice to update the tests we have to include serialization/deserialization, see test/nursery/test_autogluon_tabular.py

src/gluonts/nursery/autogluon_tabular/predictor.py
```diff
@@ -333,3 +338,48 @@ def predict(
...
+                lag_indices=self.lag_indices,
+                ag_path=self.ag_path,
+            )
+            print(dump_json(parameters), file=fp)
+
+    @classmethod
+    def deserialize(
+        cls,
+        path: Path,
+        scaling: Callable[
```
lostella, line 368:
```suggestion
        # TODO this is temporary, we should make the callable object serializable in the first place
        scaling: Callable[
```

src/gluonts/nursery/autogluon_tabular/predictor.py
```diff
@@ -333,3 +338,48 @@ def predict(
...
+            [pd.Series], Tuple[pd.Series, float]
+        ] = mean_abs_scaling,
+        # ctx: Optional[mx.Context] = None
+        **kwargs,
+    ) -> "Predictor":
+        # with mx.Context(ctx):
```
lostella, line 374:
```suggestion
```

test/nursery/test_autogluon_tabular.py
```diff
@@ -173,23 +175,31 @@ def test_get_features_dataframe(
...
         freq=freq,
         prediction_length=prediction_length,
+        ag_path_prefix=ag_path_prefix,
         lag_indices=lag_indices,
-        time_limits=10,
+        time_limit=10,
         disable_auto_regression=disable_auto_regression,
     )
 
+    os.makedirs(path, exist_ok=True)
```
lostella, line 198: If you go the tempdir way (see comment above) there should be no way to create any directory explicitly

test/nursery/test_autogluon_tabular.py
```diff
@@ -173,23 +175,31 @@ def test_get_features_dataframe(
 )
 @pytest.mark.parametrize("lag_indices", [[], [1, 2, 5]])
 @pytest.mark.parametrize("disable_auto_regression", [False, True])
+@pytest.mark.parametrize("path", [Path("test_autogluon_dir")])
+@pytest.mark.parametrize("ag_path_prefix", [None, Path("test_autogluon_dir")])
```
lostella, line 179: There is no need to parametrize the test with these paths, I think: It would be better to use a temporary directory in tests, see https://docs.python.org/3.7/library/tempfile.html

src/gluonts/nursery/autogluon_tabular/estimator.py
```diff
@@ -80,9 +88,13 @@ def __init__(
...
         super().__init__()
 
-        self.task = task
+        self.task = tabular.TabularPredictor
         self.freq = freq
         self.prediction_length = prediction_length
+        if ag_path_prefix is not None:
+            self.path = ag_path_prefix / self.ag_path
+        else:
+            self.path = self.ag_path
```
lostella, line 97: I’m wondering: why not take `ag_path` directly as input? With default value `”ag_tabular_models”`. That way these four lines can be replaced by `self.ag_path = ag_path`

src/gluonts/nursery/autogluon_tabular/predictor.py
```diff
@@ -333,3 +338,48 @@ def predict(
...
             )
+
+    def serialize(self, path: Path) -> None:
+        # call Predictor.serialize() in order to serialize the class name
+
+        super().serialize(path)
+
+        # serialize self.ag_model
+        # auto gluon predictor should be saved automatically?
+        # self.ag_model.save()
```
lostella, line 349: The autogluon model is saved already somewhere: but maybe we want to save it again under the path specified here? I understand the current mechanism: the estimator tells the prediction (when creating it) where the autogluon model is stored. However there is a point in making the “product” of the current `serialize` method here self-contained: if you serialize to path A, you can zip it and store it and it will have everything.

What do you think?

src/gluonts/nursery/autogluon_tabular/estimator.py
```diff
@@ -80,9 +88,13 @@ def __init__(
     ) -> None:
         super().__init__()
 
-        self.task = task
+        self.task = tabular.TabularPredictor
```
jwmueller, line 91: Can you just get rid of this line and use `tabular.TabularPredictor` in place of `self.task` below?
I don't think `self.task` is used anywhere else, and this seems like a confusing attribute to have.

src/gluonts/nursery/autogluon_tabular/estimator.py
```diff
@@ -122,9 +138,13 @@ def train(self, training_data: Dataset) -> TabularPredictor:
             for entry in training_data
         ]
         df = pd.concat(dfs)
-        ag_model = self.task.fit(
-            df, label="target", problem_type="regression", **self.kwargs
-        )
+
+        ag_model = self.task(
```
jwmueller, line 142: just use `tabular.TabularPredictor` here instead of `self.task`

src/gluonts/nursery/autogluon_tabular/estimator.py
```diff
@@ -12,10 +12,11 @@
 # permissions and limitations under the License.
 
 from typing import Callable, Optional, List, Tuple
-
+from pathlib import Path
 import pandas as pd
-from autogluon import TabularPrediction as task
+from autogluon import tabular
```
jwmueller, line 17: change to: `from autogluon.tabular import TabularPredictor`

src/gluonts/nursery/autogluon_tabular/predictor.py
jwmueller: Inside serialize, we could just move all the autogluon files to this new location. And then do:
```
ag_model = None
ag_model = TabularPredictor.load(NEW_LOCATION)
...
```

src/gluonts/nursery/autogluon_tabular/estimator.py
yx1215: We have a TabularPredictor in gluon-ts also. Doing that might result in name conflict. See [here](https://github.com/awslabs/gluon-ts/blob/bee033d964951eb632414153d8ba32f0f2f3718c/src/gluonts/nursery/autogluon_tabular/predictor.py#L99)

src/gluonts/nursery/autogluon_tabular/predictor.py
lostella: Yes!

src/gluonts/nursery/autogluon_tabular/estimator.py
lostella: If we really need to make a distinction in the names, the GluonTS class could be called AutogluonTabularPredictor, indicating that it is a predictor based on Autogluon. The previously proposed name (GluonTSTabularPredictor) was redundant in my opinion

jwmueller: Yes I vote for:
`from autogluon.tabular import TabularPredictor as AutogluonTabularPredictor`

test/nursery/test_autogluon_tabular.py
yx1215: I'm confused about what this means..

src/gluonts/nursery/autogluon_tabular/estimator.py
yx1215: I no longer ask the user to enter a path when estimator is created. See the comments below.

src/gluonts/nursery/autogluon_tabular/predictor.py
yx1215: I can do it in this way. But there is a problem, if I move the ag_model to NEW_LOCATION, then all the files for ag_model are in NEW_LOCATION but the path in ag_model still refer to the OLD_LOCATION where it is first created. I think this might lead to some errors.

yx1215: Currently I remove the input of path when training the estimator. And all the files for tabular will be automatically moved to the place specified when doing the serialization.

test/nursery/test_autogluon_tabular.py
```diff
@@ -16,14 +16,17 @@
 import pytest
 import numpy as np
 import pandas as pd
+import os
```
lostella, line 19: I think this is not needed anymore, right?

src/gluonts/nursery/autogluon_tabular/estimator.py
```diff
@@ -80,7 +85,7 @@ def __init__(
     ) -> None:
         super().__init__()
 
-        self.task = task
+        self.ag_predictor = AutogluonTabularPredictor
```
lostella, line 88: Why storing this as attribute?

src/gluonts/nursery/autogluon_tabular/estimator.py
```diff
@@ -122,9 +131,12 @@ def train(self, training_data: Dataset) -> TabularPredictor:
             for entry in training_data
         ]
         df = pd.concat(dfs)
-        ag_model = self.task.fit(
-            df, label="target", problem_type="regression", **self.kwargs
-        )
+
+        ag_model = self.ag_predictor(
```
lostella, line 135: You could use `AutogluonTabularPredictor(...).fit(...)` here

lostella:
The test failures on Linux are unrelated to the changes in this PR: there is (at least) one flaky test that should be (for now) marked as such: https://github.com/awslabs/gluon-ts/blob/6ac4cc6b2beef045453b58ab10876e96fbd9f502/test/torch/modules/test_torch_distribution_inference.py#L273

You can mark it as "flaky" by decorating it with e.g.

```
@pytest.mark.flaky(max_runs=5, min_passes=1)
```

However, it seems that some of the related tests are also failing on Windows

lostella (approved):
Looks good! Thank you!

PR #1400, opened 2021-04-13
add time-varying example

yifeim:
*Issue #, if available:* n/a

*Description of changes:*
* Add example for time-varying exponential distribution. the inferred parameter shows correct rank with respect to the true lambda parameter with a 20% relative error in values.
* Modify infer_lambda function to include xmin as well as xmax. Both of these configurations are keyword-only to avoid ambiguities when this function is overloaded in the future.


yifeim:
Time-varying exponential:
* Bugfix such that lamb>0 to ensure `integrate(lamb, 0, infty)=infty <=> cdf(infty)=1`
* Add thinning sampler and Hawkes sampler

GMM:
* Add `log_prior` estimation in GMM and rename some variables
* Explain the extra bias in slope estimation due to `lamb'(t)`

dcmaddix (approved):

PR #1401, opened 2021-04-15
fix bug about attention mask

Xiaoxiong-Liu:
score attention mask should be -inf, so that after softmax, score in masked will be zero.

*Issue #, if available:*

*Description of changes:*
Mask with 1e-9 make no sense, it should be -1e9


mbohlkeschneider (approved):
Thanks for catching this! 

One note: 1e-9 is probably fine most of the time as well (unless you have very low scores),

Xiaoxiong-Liu:
> Thanks for catching this!
> 
> One note: 1e-9 is probably fine most of the time as well (unless you have very low scores),
Thank you for your work to provide open source code of TFT model and prompt reply.
There is no guarantee that the values in `score` matrix greater than 0,  
So after softmax, if masked with 0 , this code may not work as expected .

Issue #1225, opened 2020-12-17
MAPE metric calculation in GLUONTS


ammarwm:
## Description
The calculation of the MAPE metric looks strange. You are removing zero actual values but still you calculate the mean including the zeros. Any explanation for that?

```python
    @staticmethod
    def mape(target, forecast):
        r"""
        .. math::

            mape = mean(|Y - Y_hat| / |Y|))
        """

        denominator = np.abs(target)
        flag = denominator == 0

        mape = np.mean(
            (np.abs(target - forecast) * (1 - flag)) / (denominator + flag)
        )
        return mape
```

PR #1402, opened 2021-04-15
Move methods for metric calculations from Evaluator class

arangatang:
*Issue #, if available:*
#1135 & #1225 

*Description of changes:*
This PR aims to cleanup the Evaluator class and changing the way that metrics are being handled and calculated. 
This will be done in several steps as detailed below.

- [x] Move methods for calculating metrics outside of the Evaluator class
- [ ] Remove masking of denominators close to zero for MAPE, MASE, sMAPE, MSIS
- [ ] Write unit tests for all metrics (In progress)
- [ ] Write docstrings for metrics


src/gluonts/evaluation/_base.py
```diff
@@ -31,9 +31,22 @@
 import numpy as np
 import pandas as pd
 
+from gluonts.evaluation.metrics import (
```
jaheba, line 34: I think relative imports would make more sense here. So `from .metrics import ...`

jaheba (commented):
Looks good so far. I think we can merge this in a first step and then do the other steps later.

lostella (commented):
Looks good, I agree with proceeding with multiple PRs. @arangatang could you make sure the PR title & commit message reflects the change before merging? So it's easier when we do the next release

src/gluonts/evaluation/_base.py
arangatang: Sure thing, changed it!

lostella (approved):

Issue #1374, opened 2021-03-22
Tracking: MXNet 1.8 compatibility


jaheba:
Our tests currently break because of the new mxnet release.

lostella:
Running tests locally (macOS, py37, mxnet==1.8.0.post0) I only get the following failures:

```
test/activation/test_activations.py
test/distribution/test_issue_287.py
```

In total, 2 test cases failed. Was anything else observed on different platforms? If not, this may be not too bad.

lostella:
Windows wheels for mxnet 1.8 are not available yet, see discussion here https://github.com/apache/incubator-mxnet/issues/18800#issuecomment-813465016

PR #1403, opened 2021-04-15
Fixes for MXNet 1.8

lostella:
*Issue #, if available:* #1374

Tests against mxnet 1.8 cannot be enabled because the pypi package for windows is not available. However this PR passes tests on macos and linux when using mxnet 1.8.

### DIFF
--- a/test/activation/test_activations.py
+++ b/test/activation/test_activations.py
@@ -36,24 +36,19 @@ from gluonts.mx.activation import get_activation, get_activation_deriv
     ],
 )
 def test_activation_deriv(activation, kwargs):
-    def get_deriv_autograd(input, act):
-        input.attach_grad()
+    def get_deriv_autograd(x, act):
+        x.attach_grad()
         with autograd.record():
-            output = act(input)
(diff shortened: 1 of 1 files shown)


mbohlkeschneider (commented):
There were issues when doing inference on SageMaker with mxnet 1.7. Does this problem still persist with mxnet 1.8?

lostella:
> There were issues when doing inference on SageMaker with mxnet 1.7. Does this problem still persist with mxnet 1.8?

I'm not sure, we would have to try: this fix only concerns failing tests. What kind of issues are there on SM with 1.7?

mbohlkeschneider:
There wasn't a particular error message. Inference jobs where not running, basically (i.e. running forever).

lostella:
I see, that doesn't sound right. I think the best thing to do is to try out an mxnet 1.8-based container to see whether it was mxnet 1.7 causing the issue. From the tests I ran on mxnet 1.8 for the present PR, everything was green except the one I went fixing

jaheba (approved):

PR #1404, opened 2021-04-16
Custom scm version handling.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/requirements/requirements-setup.txt
+++ b/requirements/requirements-setup.txt
@@ -1,4 +1,3 @@
 black==20.8b1
 mypy==0.761
 setuptools>=40.1.0
-setuptools_scm>=3.3.3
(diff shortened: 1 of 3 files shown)


jaheba:
Example:

```
$ python -c "import gluonts; print(gluonts.__version__)"                                                                
dev.custom-scm-version+ga8d6cdd
```

```
$ git tag v0.7.0 
$ python -c "import gluonts; print(gluonts.__version__)"
v0.7.0
```

lostella:
Does this work is the running environment doesn't have `git`? Say, one pip-installs gluon-ts from PyPI, but is not developing it

jaheba:
In that case you should get `"0.0.0-unknown"`.

lostella:
> In that case you should get `"0.0.0-unknown"`.

but you probably want the correct version instead, right?

PR #1409, opened 2021-04-21
Refactor metric calculations and the Evaluator class.

arangatang:
*Issue #, if available:*

*Description of changes:*
The PR refactors the metric functions so that they "follow the book". Previous alterations made has either been made optional through estimator parameters or removed completely. 

**Overview of changes:**

- [x] Moved method call to `nan_if_masked` from metric definitions to evaluator class 
- [x] `nan_if_masked` is now run if `nan_if_masked_timeseries=True` is passed when instantiating the evaluator
- [x] Made masking of invalid values optional via the `mask_invalid_timeseries` parameter.
- [x] Removed handling of denominators close to 0 from metrics methods.
- [x] Updated tests for Evaluator and metrics

**Questions:**

1. Does it make sense to expose both `nan_if_masked_timeseries` and `mask_invalid_timeseries`?
2. Shouldn't the forecast also be masked for invalid values and not only the `past_data` and the `pred_target`
3. Should we leave the default values of the evaluator with both `nan_if_masked_timeseries=True` and  `mask_invalid_timeseries=True` or should these be disabled by default? 

Any feedback would be highly appreciated.


src/gluonts/evaluation/metrics.py
```diff
@@ -86,44 +77,31 @@ def mase(
...
-    if exclude_zero_denominator and np.isclose(seasonal_error, 0.0):
-        return np.nan
-
-    return nan_if_masked(np.mean(np.abs(target - forecast)) / seasonal_error)
+    return np.mean(np.abs(target - forecast)) / seasonal_error
 
 
 def mape(
-    target: np.ndarray, forecast: np.ndarray, exclude_zero_denominator=True
+    target: np.ndarray, forecast: np.ndarray, exclude_zero_denominator=False
```
lostella, line 84: `exclude_zero_denominator` could actually be removed?

arangatang: Thanks, it should go indeed.

src/gluonts/evaluation/_base.py
```diff
@@ -30,6 +30,7 @@
 
 import numpy as np
 import pandas as pd
+import toolz
```
jaheba, line 33: Personally, I like to import things directly from toolz: `from toolz import valmap, ...`

Makes it easier to read at places where it is used.

src/gluonts/evaluation/_base.py
```diff
@@ -114,6 +120,8 @@ def __init__(
         custom_eval_fn: Optional[Dict] = None,
         num_workers: Optional[int] = multiprocessing.cpu_count(),
         chunk_size: int = 32,
+        mask_invalid_timeseries: Optional[bool] = True,
```
jaheba, line 123: I think this shouldn't be `Optional`. `Optional` means that a value also can be `None`, not that the argument is optional.

src/gluonts/evaluation/_base.py
```diff
@@ -101,6 +102,11 @@ class Evaluator:
     chunk_size
         Controls the approximate chunk size each workers handles at a time.
         Default is 32.
+    mask_invalid_timeseries
```
jaheba, line 105: Wouldn't `mask_invalid_values` be more accurate name, since we don't mask the entire ts, but just some values?

src/gluonts/evaluation/_base.py
```diff
@@ -259,15 +269,15 @@ def get_metrics_per_ts(
...
-        # required for seasonal_error and owa calculation
         past_data = np.array(self.extract_past_data(time_series, forecast))
-        past_data = np.ma.masked_invalid(past_data)
+
+        if self.mask_invalid_timeseries:
+            past_data = np.ma.masked_invalid(past_data)
+            pred_target = np.ma.masked_invalid(pred_target)
 
         try:
             mean_fcst = forecast.mean
```
jaheba, line 279: I think this can just be `mean_fcst = getattr(forecast, "mean", None)`, so we would get around the `try-except` entirely.

src/gluonts/evaluation/metrics.py
```diff
@@ -50,34 +46,29 @@ def calculate_seasonal_error(
...
 def quantile_loss(target: np.ndarray, forecast: np.ndarray, q: float) -> float:
-    return nan_if_masked(
-        2 * np.sum(np.abs((forecast - target) * ((target <= forecast) - q)))
-    )
+    return 2 * np.sum(np.abs((forecast - target) * ((target <= forecast) - q)))
 
 
 def coverage(target: np.ndarray, forecast: np.ndarray) -> float:
-    return nan_if_masked(np.mean((target < forecast)))
+    return np.mean((target < forecast))
```
jaheba, line 65:
```suggestion
    return np.mean(target < forecast)
```

jaheba (commented):
Thanks a lot!

I have some minor comments, especially on naming.

src/gluonts/evaluation/_base.py
arangatang: Alrighty, changed it

src/gluonts/evaluation/_base.py
arangatang: Thanks!

src/gluonts/evaluation/_base.py
arangatang: I think that makes sense yes, ill change it.

arangatang: I also change the `nan_if_masked_timeseries` to `nan_if_masked_metric` as this also is more specific.

src/gluonts/evaluation/_base.py
arangatang: Nice!

src/gluonts/evaluation/metrics.py
arangatang: Thanks.

src/gluonts/evaluation/_base.py
```diff
@@ -101,6 +102,11 @@ class Evaluator:
     chunk_size
         Controls the approximate chunk size each workers handles at a time.
         Default is 32.
+    mask_invalid_timeseries
+        Ignore `NaN` and `inf` values in the timeseries when calculating metrics.
+    nan_if_masked_timeseries
```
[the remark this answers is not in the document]
arangatang: True that.

src/gluonts/evaluation/_base.py
arangatang: This breaks the tests, as some forecasts have the attribute but raises a `NotImplementedError`.

src/gluonts/evaluation/_base.py
```diff
@@ -259,21 +270,22 @@ def get_metrics_per_ts(
...
+
+        if self.mask_invalid_values:
+            past_data = np.ma.masked_invalid(past_data)
+            pred_target = np.ma.masked_invalid(pred_target)
 
         try:
-            mean_fcst = forecast.mean
-        except:
+            mean_fcst = getattr(forecast, "mean", None)
+        except NotImplementedError:
```
jaheba, line 281: Oh, I see. Now the `getattr` doesn't look so sensible anymore. Maybe we should just not raise an exception for `forecast.mean`.

src/gluonts/evaluation/_base.py
```diff
@@ -16,6 +16,7 @@
 import sys
 from functools import partial
 from itertools import chain, tee
+from toolz import valmap
```
jaheba, line 19: `toolz` is 3rd party, so it should be in the next block

src/gluonts/evaluation/_base.py
```diff
@@ -101,6 +103,11 @@ class Evaluator:
     chunk_size
         Controls the approximate chunk size each workers handles at a time.
         Default is 32.
+    mask_invalid_values
```
jaheba, line 106: Now I wonder if `ignore_invalid_values` is an even better name.

src/gluonts/evaluation/_base.py
```diff
@@ -345,6 +357,9 @@ def get_metrics_per_ts(
                 pred_target, forecast_quantile
             )
 
+        if self.nan_if_masked_metric:
+            metrics = valmap(nan_if_masked, metrics)
```
jaheba, line 361: I wonder, why wouldn't you do it always? What's the benefit of getting back `masked` as a metric?

src/gluonts/evaluation/_base.py
arangatang: Sure, im fine with that.

src/gluonts/evaluation/_base.py
arangatang: It allows users to differentiate between if the metric calculation resulted in a `nan` or not when looking at the item metrics.
For the purpose of aggregating the metrics it should not matter.

src/gluonts/evaluation/_base.py
arangatang: True, although i think that changing the `forecast` goes outside the scope of this pr?

src/gluonts/evaluation/_base.py
arangatang: Perhaps we should remove `nan_if_masked_metric` instead.

src/gluonts/evaluation/_base.py
jaheba: Agreed.

src/gluonts/evaluation/_base.py
jaheba: Not sure we want to make a differentiation there. I'm for just removing the option and always calling `nan_if_masked`.

@lostella what do you think?

lostella (approved):

PR #1410, opened 2021-04-21
spliced binned pareto

elenaehrlich:
*Issue #, if available:* N/A

*Description of changes:* Pytorch code of the Spliced Binned-Pareto (SBP) distribution and an example notebook its fit to non-stationary, heavy-tailed time series using a DistributionalTCN.


src/gluonts/nursery/spliced_binned_pareto/spliced_binned_pareto.py
```diff
@@ -0,0 +1,498 @@
...
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+
+import torch
+from torch.nn import functional as F
+from torch import optim
+import torch.nn.functional as F
```
lostella, line 18: This is imported twice

src/gluonts/nursery/spliced_binned_pareto/spliced_binned_pareto.py
```diff
@@ -0,0 +1,498 @@
...
+import torch.nn as nn
+
+from typing import List, Union, Optional
+
+import numpy as np
+
+try:
+    from .genpareto import GenPareto
+except:
+    from genpareto import GenPareto
```
lostella, line 29: Shouldn't the relative import work?

src/gluonts/nursery/spliced_binned_pareto/training_functions.py
```diff
@@ -0,0 +1,274 @@
...
+import pandas as pd
+import numpy as np
+import os
+from scipy import stats
+from tqdm import tqdm
+
+import torch
+from torch.nn import functional as F
+from torch import optim
+import torch.nn.functional as F
```
lostella, line 28: Imported twice

src/gluonts/nursery/spliced_binned_pareto/training_functions.py
```diff
@@ -0,0 +1,274 @@
...
+from torch.autograd import Variable
+import torch.nn as nn
+
+try:
+    from .distr_tcn import DistributionalTCN
+except:
+    from distr_tcn import DistributionalTCN
```
lostella, line 35: Same as above, I think the relative import should work

src/gluonts/nursery/spliced_binned_pareto/distr_tcn.py
```diff
@@ -0,0 +1,162 @@
...
+# SOFTWARE.
+
+import pandas as pd
+import numpy as np
+from tqdm import tqdm
+
+import torch
+from torch.nn import functional as F
+from torch import optim
+import torch.nn.functional as F
```
lostella, line 69: Imported twice

src/gluonts/nursery/spliced_binned_pareto/distr_tcn.py
```diff
@@ -0,0 +1,162 @@
...
+from torch.autograd import Variable
+import torch.nn as nn
+
+try:
+    from .tcn import TCNBlock
+except:
+    from tcn import TCNBlock
```
lostella, line 76: Relative import should suffice

src/gluonts/nursery/spliced_binned_pareto/gaussian_model.py
```diff
@@ -0,0 +1,83 @@
...
+import torch
+import torch.nn.functional as F
+import torch.nn as nn
+
+from torch.distributions import constraints
+from torch.distributions.distribution import Distribution
+from torch.distributions.normal import Normal
+
+
+class GaussianModel(nn.Module):
```
lostella, line 24: Is this class used anywhere? I don't get what its purpose is, it seems like a suspiciously thin wrapper around pytorch's `Normal`

lostella (changes requested):
If relative imports don't work (see my comments in the code about the try-catch around imports) maybe it is because the `__init__.py` file is missing in the folder.

Also, is the html file required? It's called the same as the notebook, so it looks like it's exported from it: maybe keeping the notebook is enough?

src/gluonts/nursery/spliced_binned_pareto/gaussian_model.py
elenaehrlich: This class is used in the example notebook, to demonstrate a comparison for the user how to use DistributionalTCN on a familiar distribution and not just our Spliced Binned-Pareto. 

Aside from the function to switch device id, it demonstrates a forward with mini-batching updates, and the forward does not require this way of registering the parameters: https://stackoverflow.com/questions/63047762/correct-way-to-register-a-parameter-for-model-in-pytorch.

lostella (approved):

PR #1411, opened 2021-04-21
made QRX faster by not computing quantiles for all bins

zoolhasson:
*Issue #, if available:*

*Description of changes:* Made QRX faster by not computing quantiles for all bins ahead of time, but rather in an ad hoc fashion and caching the results.


dcmaddix (approved):
Do you tests showing that it works?

PR #1416, opened 2021-04-23
Add strategies for aggregating metrics to Evaluator class.

arangatang:
*Issue #, if available:*

*Description of changes:*
This pr adds logic which allows users to choose how per-timeseries metrics should be aggregated when calling `Evaluator.get_aggregate_metrics`.

### DIFF
--- a/src/gluonts/evaluation/_base.py
+++ b/src/gluonts/evaluation/_base.py
@@ -54,6 +54,51 @@ def worker_function(evaluator: "Evaluator", inp: tuple):
     return evaluator.get_metrics_per_ts(ts, forecast)
 
 
+def aggregate_all(
+    metric_per_ts: pd.DataFrame, agg_funs: Dict[str, str]
+) -> Dict[str, float]:
+    """
+    No filtering applied.
+
+    Both `nan` and `inf` possible in aggregate metrics.
+    """
+    return {
+        key: metric_per_ts[key].agg(agg, skipna=False)
+        for key, agg in agg_funs.items()
+    }
+
+
+def aggregate_no_nan(
+    metric_per_ts: pd.DataFrame, agg_funs: Dict[str, str]
+) -> Dict[str, float]:
+    """
+    Filter all `nan` but keep `inf`.
+
+    `nan` is only possible in the aggregate metric if all timeseries
+    for a metric resulted in `nan`.
+    """
+    return {
+        key: metric_per_ts[key].agg(agg, skipna=True)
+        for key, agg in agg_funs.items()
+    }
+
+
+def aggregate_valid(
+    metric_per_ts: pd.DataFrame, agg_funs: Dict[str, str]
+) -> Dict[str, Union[float, np.ma.core.MaskedConstant]]:
+    """
+    Filter all `nan` & `inf` values from `metric_per_ts`.
+
+    If all metrics in a column of `metric_per_ts` are `nan` or `inf` the
+    result will be `np.ma.masked` for that column.
+    """
+    metric_per_ts = metric_per_ts.apply(np.ma.masked_invalid)
+    return {
+        key: metric_per_ts[key].agg(agg, skipna=True)
+        for key, agg in agg_funs.items()
+    }
+
+
 class Evaluator:
     """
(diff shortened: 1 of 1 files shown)


src/gluonts/evaluation/_base.py
```diff
@@ -99,6 +99,18 @@ class Evaluator:
         Default is 32.
     ignore_invalid_values
         Ignore `NaN` and `inf` values in the timeseries when calculating metrics.
+    aggregation_strategy:
+        Selects how invalid values in per-timeseries metrics should be filtered
+        when calculating the aggregate metric. Available options are:
+        `"all" | "valid" | "inf_only"`.
```
lostella, line 105: Maybe mention here what the default is

src/gluonts/evaluation/_base.py
```diff
@@ -112,6 +124,7 @@ def __init__(
         custom_eval_fn: Optional[Dict] = None,
         num_workers: Optional[int] = multiprocessing.cpu_count(),
         chunk_size: int = 32,
+        aggregation_strategy: Optional[str] = "inf_only",
```
lostella, line 127: This should not be `Optional` right?

src/gluonts/evaluation/_base.py
```diff
@@ -364,20 +379,29 @@ def get_aggregate_metrics(
...
             set(metric_per_ts.columns) >= agg_funs.keys()
         ), "Some of the requested item metrics are missing."
 
+        agg_kwargs = {}
+        if self.aggregation_strategy == "inf_only":
+            pass
+        elif self.aggregation_strategy == "all":
+            agg_kwargs = {"skipna": False}
+        elif self.aggregation_strategy == "valid":
+            metric_per_ts = metric_per_ts.apply(np.ma.masked_invalid)
```
lostella, line 400: Do we need an `else` here? Or otherwise you can validate the possible options at construction time.

lostella (changes requested):

kashif:
@arangatang do you think it would be possible to add helpers to calculate multivariate metrics? I have 2 in mind:

```py
def energy_score(forecast, target):
        obs_dist = np.mean(np.linalg.norm((forecast - target), axis=-1))
        pair_dist = np.mean(
            np.linalg.norm(forecast[:, np.newaxis, ...] - forecast, axis=-1)
        )
        return obs_dist - pair_dist * 0.5
```

and 

```py
def variogram_score(forecast, target, p=0.5):
        _, T, D = forecast.shape
        score = np.zeros(T)

        for i in range(D):
            for j in range(D):
                vdat = np.mean(np.power(np.abs(forecast[..., i] - forecast[..., j]), p), 0)
                vy = np.power(np.abs(target[..., i] - target[..., j]), p)
                score += np.power(vy - vdat, 2)

        return np.mean(score)
```

src/gluonts/evaluation/_base.py
jaheba: If this was a true strategy-pattern, the implementation would be in a different object. This implementation therefore also breaks the open-closed principle.

lostella:
> @arangatang do you think it would be possible to add helpers to calculate multivariate metrics? I have 2 in mind:

@kashif we can definitely add those, but I would do it in a separate PR

src/gluonts/evaluation/_base.py
lostella: There are cases in which a pre-defined set of options makes sense: this might be one?

src/gluonts/evaluation/_base.py
arangatang: We could validate them there that is true, however the `metric_per_ts.apply(...)` would still need to be run at this point.

src/gluonts/evaluation/_base.py
arangatang: True that, updated it.

src/gluonts/evaluation/_base.py
arangatang: Moved them to seperate methods instead.

lostella (approved):
Looks good to me! Thanks a lot!

jaheba (approved):
Looks good!

PR #1421, opened 2021-04-26
Add transform objects for dealing with sparse time series.

canerturkmen:
*Issue #, if available:*

Related to creating dedicated neural intermittent demand models in GluonTS.

*Description of changes:*

This PR adds two simple transformation objects that are meant to specifically deal with intermittent time series. Namely, `AddTrailingZeros` simply adds the number of zero valued time steps that are trailing in each of the target time series, allowing models to explicitly condition on the number of steps since the last 'demand' point when making a prediction. `SparseToDense` transformation takes a univariate time series and converts it to a size-interval format as often used in intermittent demand forecasting (e.g., in Croston's method).

### DIFF
--- a/src/gluonts/transform/convert.py
+++ b/src/gluonts/transform/convert.py
@@ -761,7 +761,7 @@ def cdf_to_gaussian_forward_transform(
     return outputs
 
 
-class SparseToDense(FlatMapTransformation):
+class ToIntervalSizeFormat(FlatMapTransformation):
     """
     Convert a sparse univariate time series to the `interval-size` format,
     i.e., a two dimensional time series where the first dimension corresponds
@@ -802,15 +802,17 @@ class SparseToDense(FlatMapTransformation):
 
     def _process_sparse_time_sample(self, a: List) -> Tuple[List, List]:
         a = np.array(a)
-        nz = np.nonzero(a)[0]
+        (non_zero_index,) = np.nonzero(a)
 
-        if len(nz) == 0:
+        if len(non_zero_index) == 0:
             return [], []
 
-        times = np.diff(np.r_[-1, np.nonzero(a)[0]]).astype(np.float).tolist()
-        sizes = a[np.nonzero(a)].tolist()
+        times = np.diff(non_zero_index, prepend=-1.0).tolist()
+        sizes = a[non_zero_index].tolist()
 
-        return (times[1:], sizes[1:]) if self.discard_first else (times, sizes)
+        if self.discard_first:
+            return times[1:], sizes[1:]
+        return times, sizes
 
     def flatmap_transform(
         self, data: DataEntry, is_train: bool
@@ -820,6 +822,5 @@ class SparseToDense(FlatMapTransformation):
         times, sizes = self._process_sparse_time_sample(target)
 
         if len(times) > 0 or not self.drop_empty:
-            new_data = data.copy()
-            new_data[self.target_field] = [times, sizes]
-            yield new_data
+            data[self.target_field] = [times, sizes]
+            yield data
--- a/src/gluonts/transform/convert.py
+++ b/src/gluonts/transform/convert.py
@@ -759,3 +759,67 @@ def cdf_to_gaussian_forward_transform(
             tensor_to_numpy(input_batch["intercepts"]),
         )
     return outputs
+
+
+class SparseToDense(FlatMapTransformation):
+    """
+    Convert a sparse univariate time series to the `interval-size` format,
+    i.e., a two dimensional time series where the first dimension corresponds
+    to the time since last positive value (1-indexed), and the second dimension
+    corresponds to the size of the demand. This format is used often in the
+    intermittent demand literature, where predictions are performed on this
+    "dense" time series, e.g., as in Croston's method.
+
+    As an example, the time series `[0, 0, 1, 0, 3, 2, 0, 4]` is converted into
+    the 2-dimensional time series `[[3, 2, 1, 2], [1, 3, 2, 4]]`, with a
+    shape (2, M) where M denotes the number of non-zero items in the time series.
+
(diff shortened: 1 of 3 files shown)


src/gluonts/transform/convert.py
```diff
@@ -759,3 +759,67 @@ def _empirical_cdf_inverse_transform(
...
+        discard_first: bool = False,
+    ) -> None:
+
+        self.target_field = target_field
+        self.drop_empty = drop_empty
+        self.discard_first = discard_first
+
+    def _process_sparse_time_sample(self, a: List) -> Tuple[List, List]:
+        a = np.array(a)
+        nz = np.nonzero(a)[0]
```
jaheba, line 805: Not sure it's better, but I like to use:

```python
(nz,) = np.nonzero(a)
```

jaheba: However, couldn't you just do:

```py
sizes = a[a != 0]
if len(sizes) == 0:

    return [], []

...
```

src/gluonts/transform/convert.py
```diff
@@ -759,3 +759,67 @@ def _empirical_cdf_inverse_transform(
...
+
+        if len(nz) == 0:
+            return [], []
+
+        times = np.diff(np.r_[-1, np.nonzero(a)[0]]).astype(np.float).tolist()
```
jaheba, line 810:
```suggestion
        (non_zero_idx,) = np.nonzero(a)
        times = np.diff(non_zero_idx, prepend=-1.0).tolist()
```

src/gluonts/transform/convert.py
```diff
@@ -759,3 +759,67 @@ def _empirical_cdf_inverse_transform(
...
+        sizes = a[np.nonzero(a)].tolist()
+
+        return (times[1:], sizes[1:]) if self.discard_first else (times, sizes)
```
jaheba, line 813: I think this is easier written as proper statements

```suggestion
        if self.discard_first:
            return times[1:], sizes[1:]

        return times, sizes
```

src/gluonts/transform/convert.py
```diff
@@ -759,3 +759,67 @@ def _empirical_cdf_inverse_transform(
...
+
+    def flatmap_transform(
+        self, data: DataEntry, is_train: bool
+    ) -> Iterator[DataEntry]:
+        target = data[self.target_field]
+
+        times, sizes = self._process_sparse_time_sample(target)
+
+        if len(times) > 0 or not self.drop_empty:
+            new_data = data.copy()
```
jaheba, line 823: `data` is already copied in `FlatMapTransform`:

```python
for result in self.flatmap_transform(
                    data_entry.copy(), is_train
                ):
```

src/gluonts/transform/feature.py
```diff
@@ -555,3 +555,39 @@ def map_transform(self, data: DataEntry, is_train: bool) -> DataEntry:
...
+        self,
+        new_field: str = "trailing_zeros",
+        target_field: str = "target",
+    ) -> None:
+        self.target_field = target_field
+        self.new_field = new_field
+
+    def map_transform(self, data: DataEntry, is_train: bool) -> DataEntry:
+        return self.transform(data)
+
```
jaheba, line 586:
```suggestion
```

This is defined in `SimpleTransformation`.

src/gluonts/transform/feature.py
```diff
@@ -555,3 +555,39 @@ def map_transform(self, data: DataEntry, is_train: bool) -> DataEntry:
...
+    def transform(self, data: DataEntry) -> DataEntry:
+        target = data[self.target_field]
+        if len(target) == 0:
+            return data
+        nz = np.nonzero(target[::-1])[0]
+        data[self.new_field] = min(nz) if len(nz) > 0 else len(target)
```
jaheba, line 592: How about

```python
data[self.new_field] = len(target) - len(np.trim_zeros(target, "b"))

File: src/gluonts/transform/convert.py [hunk 7]
```diff
@@ -759,3 +759,67 @@ def _empirical_cdf_inverse_transform(
             tensor_to_numpy(input_batch["intercepts"]),
         )
     return outputs
+
+
+class SparseToDense(FlatMapTransformation):
```
lostella, line 764: The naming here may be misleading I think. The given array is dense, in the sense that it is represented as a dense array; however it is "sparse" in that it contains lots of zeros, so here we turn it into a compressed format.

I would personally like `DenseToSparse` better, but at this point it's clear that this creates confusion. Maybe `ToIntervalSizeFormat` is more to the point?

src/gluonts/transform/feature.py
```diff
@@ -555,3 +555,39 @@ def map_transform(self, data: DataEntry, is_train: bool) -> DataEntry:
         )
 
         return data
+
+
+class AddTrailingZeros(SimpleTransformation):
```
lostella, line 560: How about `CountTrailingZeros`? "Add" makes it sound like some right-padding is done on the target

canerturkmen: Yes, thank you!

src/gluonts/transform/convert.py
canerturkmen: `ToIntervalSizeFormat` it is. Thanks.

src/gluonts/transform/feature.py
canerturkmen: Yes, thanks.

src/gluonts/transform/convert.py
canerturkmen: I think I would need the nonzero index anyway below.

jaheba (approved):
LGTM

Let's maybe also wait for @lostella to approve.

lostella (approved):

PR #1433, opened 2021-05-03
Remove travis file.

jaheba:
*Issue #, if available:*

*Description of changes:*


lostella (approved):
Are we setting up an alternative workflow for this? (I guess based on gh actions?)

jaheba:
We should.

However, I think we want to break the automatism of `tag pushed -> release` and instead have more manual control.

PR #1434, opened 2021-05-03
Use miniver for version resolution.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.devtools/license
+++ b/.devtools/license
@@ -42,6 +42,11 @@ license = dedent(
 ).strip()
 
 
+miniver = (
+    "# This file is part of 'miniver': https://github.com/jbweston/miniver"
+)
+
+
 def check_file(path: Path) -> bool:
     """Check whether `path` is complicit.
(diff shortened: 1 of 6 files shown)


jaheba:
This uses https://github.com/jbweston/miniver to resolve dependencies based on tags.

jaheba:
Hmm, how should we handle the license headers? Should we still copy our header in the miniver files?

lostella:
> Hmm, how should we handle the license headers? Should we still copy our header in the miniver files?

One would need to check what the license for miniver says about that

jaheba:
https://github.com/jbweston/miniver/blob/master/LICENSE

Looks like we can do whatever we want :)

jaheba:
This seems to work now. If I run `python setup.py build`, the contents of `gluonts/_static_version.py` is replaced.

lostella (approved):

Issue #1335, opened 2021-02-18
Keep pandas 1.0 in version 0.7


alexcombessie:
## Description
My application has a hard dependency on pandas 1.0. I am thus a happy user of gluon-ts v0.6 😄 

I have seen on the master branch that pandas has been updated to 1.1. Would it be possible to relax that requirement to keep pandas 1.0 along with pandas 1.1, i.e. use `pandas>=1.0`

Pandas 1.0 is barely one-year-old, IMO it's reasonable to wait a bit before forcing a switch to the 6-month old 1.1 release.

## References
https://github.com/awslabs/gluon-ts/blob/master/requirements/requirements.txt

jaheba:
Hello Alex,

the requirement was changed in #1233. I'm not 100% why, maybe @lostella can shed some light here :)

Generally, gluon-ts should work with `pandas==1.0` though.

lostella:
@jaheba @alexcombessie looking at it again, I suppose that requirement bump was not really necessary.

@jaheba did you mean compatible with `pandas~=1.0`?

jaheba:
> @jaheba did you mean compatible with `pandas~=1.0`?

What I meant is that `gluon-ts` should work if you install `pandas==1.0`.

We should change the dependency to `pandas~=1.0` still :)

jaheba:
@alexcombessie Do you want to do the change?

alexcombessie:
Hi @jaheba,

Thanks, that's kind of you to offer 🙏  Here is a short PR: #1344 

Cheers,

Alex

jaheba:
PR is merged

lostella:
Coming back to this: the restriction was actually required because `DatetimeIndex.isocalendar()` was introduced in `pandas==1.1` cc @Schmedu

PR #1437, opened 2021-05-03
fix compatibility for pandas < 1.1 in `time_feature/_base.py`

Schmedu:
*Issue #, if available:*
```
     def __call__(self, index: pd.DatetimeIndex) -> np.ndarray:
 >       return (index.isocalendar().week - 1).map(float)
 E       AttributeError: 'DatetimeIndex' object has no attribute 'isocalendar'
```
According to https://github.com/awslabs/gluon-ts/issues/1335 we support pandas >= 1.0, but when using pandas < 1.1 we get the displayed error for `class WeekOfYearIndex` and `class WeekOfYear`.

*Description of changes:*
Added try block around `isocalendar()` function call and replaced it for the given error with the respective implementation.

### DIFF
--- a/src/gluonts/time_feature/_base.py
+++ b/src/gluonts/time_feature/_base.py
@@ -125,22 +125,30 @@ class WeekOfYear(TimeFeature):
     """Week of year encoded as value between [-0.5, 0.5]"""
 
     def __call__(self, index: pd.DatetimeIndex) -> np.ndarray:
+        # TODO:
+        # * pandas >= 1.1 does not support `.week`
+        # * pandas == 1.0 does not support `.isocalendar()`
+        # as soon as we drop support for `pandas == 1.0`, we should remove this
         try:
-            return (index.isocalendar().week - 1) / 52.0 - 0.5
+            week = index.isocalendar().week
         except AttributeError:
-            # compatibility for pandas < 1.1
-            return (index.week - 1) / 52.0 - 0.5
+            week = index.week
+        return (week - 1) / 52.0 - 0.5
 
 
 class WeekOfYearIndex(TimeFeature):
     """Week of year encoded as zero-based index, between 0 and 52"""
 
     def __call__(self, index: pd.DatetimeIndex) -> np.ndarray:
+        # TODO:
+        # * pandas >= 1.1 does not support `.week`
+        # * pandas == 1.0 does not support `.isocalendar()`
+        # as soon as we drop support for `pandas == 1.0`, we should remove this
         try:
-            return (index.isocalendar().week - 1).map(float)
+            week = index.isocalendar().week
         except AttributeError:
-            # compatibility for pandas < 1.1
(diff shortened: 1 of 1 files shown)


lostella:
@Schmedu on the other hand https://pandas.pydata.org/docs/reference/api/pandas.DatetimeIndex.week.html#pandas.DatetimeIndex.week

Schmedu:
> @Schmedu on the other hand https://pandas.pydata.org/docs/reference/api/pandas.DatetimeIndex.week.html#pandas.DatetimeIndex.week

Did not see that, my bad!

Schmedu:
It's really strange that `isocalendar()` is not listed as a function in [DatetimeIndex](https://pandas.pydata.org/docs/reference/api/pandas.DatetimeIndex.html)

But now it should be compatibly with pandas < 1.1 and >= 1.1.

jaheba:
I wonder, are there strong arguments against requiring `pandas >= 1.1`?

lostella:
> I wonder, are there strong arguments against requiring `pandas >= 1.1`?

There was one request for this [here](https://github.com/awslabs/gluon-ts/issues/1335)

jaheba:
> > I wonder, are there strong arguments against requiring `pandas >= 1.1`?
> 
> There was one request for this in #1335

He is free to keep using `gluonts == 0.6` :)

But I think this raises the more general question of how long we want to support things like this. So when does it start to be ok to have these kind of breaking changes.

For now I'm fine with having this hack, but I think we should add a `TODO` comment indicating that we should remove it in the future. /cc @Schmedu

src/gluonts/time_feature/_base.py
```diff
@@ -125,14 +125,22 @@ class WeekOfYear(TimeFeature):
     """Week of year encoded as value between [-0.5, 0.5]"""
 
     def __call__(self, index: pd.DatetimeIndex) -> np.ndarray:
-        return (index.isocalendar().week - 1) / 52.0 - 0.5
+        try:
+            return (index.isocalendar().week - 1) / 52.0 - 0.5
+        except AttributeError:
+            # compatibility for pandas < 1.1
+            return (index.week - 1) / 52.0 - 0.5
```
jaheba, line 132:
```suggestion
        # TODO:
        # * pandas >= 1.1 does not support `.week`
        # * pandas == 1.0 does not support `.isocalendar()`
        # as soon as we drop support for `pandas == 1.0`, we should remove this
        try:
            week = index.isocalendar().week
        except AttributeError:
            return week = index.week

        return (week - 1) / 52.0 - 0.5
```

jaheba:
@Schmedu Can you add a simple test?

src/gluonts/time_feature/_base.py
Schmedu: added it

jaheba (approved):

PR #1442, opened 2021-05-04
adjustments to variable length functionality in batchify

canerturkmen:
*Issue #, if available:*

*Description of changes:*

batchify functions for handling variable length arrays handled mxnet arrays, which were marked as todos. this PR removes this, restricting the accepted array types to numpy arrays. moreover, it exposes the `is_right_pad` option, already available in `pad_to_size`, also in `batchify` and `stack`. This allows users to specify whether the arrays should be padded on the right (as is the mxnet convention, in functions like `SequenceMask`) or left.

### DIFF
--- a/src/gluonts/mx/batchify.py
+++ b/src/gluonts/mx/batchify.py
@@ -22,41 +22,30 @@ from gluonts.dataset.common import DataBatch
 from gluonts.support.util import pad_to_size
 
 
-# TODO: should the following contempate mxnet arrays or just numpy arrays?
 def _is_stackable(arrays: List, axis: int = 0) -> bool:
     """
     Check if elements are scalars, have too few dimensions, or their
     target axes have equal length; i.e. they are directly `stack` able.
     """
-    if isinstance(arrays[0], (mx.nd.NDArray, np.ndarray)):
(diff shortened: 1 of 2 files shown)


lostella (approved):
Looks good to me.

Following our offline discussion (but related to this PR since you removed the TODOs): we may want to actually be able to batchify directly fields which are MXNet arrays already on the right device (or Torch ones, or whatever, but not just NumPy like it is now), so that one can cache the (pre-processed, feature-complete) dataset say on GPU, and then sample and batchify instances directly there, for efficiency reasons. However the setup prior to this PR, with `Union` arguments, and `if isinstance` everywhere, is probably not the right way to achieve that.

Moreover, the `InstanceSplitter`-type transformations (see [this](https://github.com/awslabs/gluon-ts/blob/7fdfa4b74f36bf7db9729e1fe6ca37ad889b04c3/src/gluonts/transform/split.py#L62-L204) or the similar classes that follow) are currently only numpy-to-numpy. So it seems to me that, before we can think of directly sampling & slicing & stacking on GPU, something needs to be reworked a bit (read: simplified), especially if we want to use the same exact transformations on mxnet & torch ;-).

In summary, this is stuff we can address separately from this PR!

PR #1443, opened 2021-05-05
add label requirement information to pull request template

Schmedu:
*Issue #, if available:*
A release note is created for every major release that lists all commits since the last release. These commits are divided into the categories:
- breaking
- new feature
- bug fix
- other change or improvement

We could simplify this process by automatically generating a release note using at least one required label for each pull request (one of the four categories). This would cost only a little effort from every contributor or reviewer but would speed up the process for each release.

*Description of changes:*
Added a new GitHub workflow using a forked workflow action from the GitHub marketplace (see https://github.com/jesusvasquez333/verify-pr-label-action).


lostella:
The only concern I have is that this doesn't make it apparent in the PR what should be done to make it merge-able. Anyone who's not aware/doesn't remember should click on the failing job, figure it out, add the label, restart the job. One way to overcome this would be to have a group of maintainers taking care of merges.

Maybe it would be nice to also allow another "dev setup" label, so that PRs related to dev tooling, workflows and so on (like this one) don't end up in the other categories (arguably, release notes can skip mentioning these).

Schmedu:
> The only concern I have is that this doesn't make it apparent in the PR what should be done to make it merge-able. Anyone who's not aware/doesn't remember should click on the failing job, figure it out, add the label, restart the job. One way to overcome this would be to have a group of maintainers taking care of merges.

We could also solve this with two other approaches:
- we could add a [template for pull requests only](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository), where we refer to use at least one of the labels
- let a bot create a comment when the pr is opened and no label is set

This is of course not exclusive to your suggested approach.

I have also modified the action to run whenever a PR is labeled or unlabeled, so the job does not need to be rerun manually.

> Maybe it would be nice to also allow another "dev setup" label, so that PRs related to dev tooling, workflows and so on (like this one) don't end up in the other categories (arguably, release notes can skip mentioning these).

Sounds good!

lostella:
> I have also modified the action to run whenever a PR is labeled or unlabeled, so the job does not need to be rerun manually.

Oh, nice, this makes it much easier!

lostella:
> let the workflow action create a comment when the pr is opened and no valid label is set

If this is possible, it would be ideal

lostella (dismissed):

lostella (approved):

PR #1460, opened 2021-05-12
PyTorch implementation of DeepAR

lostella:
Work in progress, open for comments.

This ports the PyTorch implementation of DeepAR from PyTorchTS (cc @kashif), with some changes:
* The estimator class was slightly refactored, and in particular the way data loaders are set up is more in line with other estimators (but I want to try out a few things here, this is giving me some thoughts)
* No specific "trainer" class was implemented, and instead the estimator relies on PyTorch Lightning for this.
* The network is now down to a single class implementing both loss computation and sample paths prediction, following torch's `.training` convention
* A thin extension to the network provides the interface used by Lightning

A few surrounding, related changes are also included.

Some open questions:
1. should the `dtype` and `device` be specified at constructor time for the estimator? Or is it something we want to pass to the `train` method?
2. the base estimator class is really PyTorch Lightning oriented: should it be called PyTorchLightningEstimator?
3. we would now have `gluonts.model` containing existing models (mxnet based) and `gluonts.torch.model` containing this one; should the mxnet ones moved to `gluonts.mx.model` for the sake of symmetry?

TODOs (partial list, probably):
- [x] cover also validation data in tests
- [x] remove the `input_size` parameter from the estimator (this should probably be inferred from the other ones)
- [x] re-include the option to pseudo-shuffle batches at training time
- [x] improve tests (also serde and so on)
- [x] open issue on the left-over features of the model, and make it release-blocking
- [x] run experiments to check the model accuracy


lostella:
[This](https://github.com/awslabs/gluon-ts/pull/1460/checks?check_run_id=2572703445#step:7:220) is weird, I wasn’t getting this locally

kashif:
Looking good... I'll have a look at this branch hopefully today too 👁️  👁️

lostella:
> Looking good... I'll have a look at this branch hopefully today too 👁️ 👃🏽 👁️

Thanks! I still want to at least resolve the TODOs above, so I'm expecting this to take a couple days more

mbohlkeschneider (commented):
Nice! 

This seems to be "only" vanilla DeepAR. We have added quite a few features to the Mxnet DeepAR (like model missing value imputation and activation regularization). Thus, I'm wondering whether we should have it here as well?

lostella:
> Nice!
> 
> This seems to be "only" vanilla DeepAR. We have added quite a few features to the Mxnet DeepAR (like model missing value imputation and activation regularization). Thus, I'm wondering whether we should have it here as well?

@mbohlkeschneider I think so, but we can maybe track that in separate PRs? So that this PR can bring in some of the necessary “scaffolding”, as well as on example of model, for anyone working on other PyTorch-based models to take inspiration from.

I’ll add an issue for that and make it release-blocking for v0.8

src/gluonts/torch/model/deepar/deepar_network.py
```diff
@@ -0,0 +1,430 @@
...
+            + self.num_feat_dynamic_real
+            + self.num_feat_static_real
+        )
+
+    @property
+    def _past_length(self) -> int:
+        return self.context_length + max(self.lags_seq)
+
+    @staticmethod
+    def get_lagged_subsequences(
```
mbohlkeschneider, line 116: Not for this PR, but maybe we can pull this out and use the `@singledispatch` mechanism for having this in torch/mxnet?

src/gluonts/torch/model/deepar/deepar_network.py
```diff
@@ -0,0 +1,430 @@
...
+            length of the subsequences to be extracted.
+
+        Returns
+        --------
+        lagged : Tensor
+            a tensor of shape (N, S, C, I), where S = subsequences_length and
+            I = len(indices), containing lagged subsequences. Specifically,
+            lagged[i, j, :, k] = sequence[i, -indices[k]-S+j, :].
+        """
+        sequence_length = sequence.shape[1]
```
mbohlkeschneider, line 141: <3

src/gluonts/torch/model/deepar/deepar_network.py
```diff
@@ -0,0 +1,430 @@
...
+        lags = self.get_lagged_subsequences(
+            sequence=sequence,
+            indices=self.lags_seq,
+            subsequences_length=subsequences_length,
+        )
+
+        # scale is computed on the context length last units of the past target
+        # scale shape is (batch_size, 1, *target_shape)
+        _, scale = self.scaler(
+            past_target[:, self.context_length :, ...],
```
mbohlkeschneider, line 192: Shouldn't this be `past_target[:, -self.context_length :, ...]`? Otherwise it starts with the `context_length` index until the end of that axis? Same for observed values below.

src/gluonts/torch/model/deepar/deepar_network.py
```diff
@@ -0,0 +1,430 @@
...
+            past_observed_values[:, self.context_length :, ...],
+        )
+        embedded_cat = self.embedder(feat_static_cat)
+        static_feat = torch.cat(
+            (
+                embedded_cat,
+                feat_static_real,
+                scale.log()
+                if len(self.target_shape) == 0
```
mbohlkeschneider, line 201: Not sure I get this.

test/torch/model/test_deepar.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mbohlkeschneider, line 1: Maybe add a simple accuracy test as well (on constant or so)?

mbohlkeschneider (dismissed):
Nice! I think I found a bug in the network that needs fixing. Otherwise, mostly stylistic comments.

src/gluonts/torch/model/deepar/deepar_estimator.py
```diff
@@ -0,0 +1,325 @@
...
+            embedding_dimension
+            if embedding_dimension is not None
+            else [min(50, (cat + 1) // 2) for cat in self.cardinality]
+        )
+        self.scaling = scaling
+        self.lags_seq = (
+            lags_seq
+            if lags_seq is not None
+            else get_lags_for_frequency(
+                freq_str=freq, lag_ub=self.context_length
```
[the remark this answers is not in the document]
kashif: yes thats my fault... sorry! we can revert it to the mxnet DeepAR

src/gluonts/torch/model/deepar/deepar_network.py
kashif: yes also my fault! good catch!

src/gluonts/torch/model/deepar/__init__.py
```diff
@@ -11,8 +11,12 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-# !!! DO NOT MODIFY !!! (pkgutil-style namespace package)
+from .deepar_estimator import DeepAREstimator
```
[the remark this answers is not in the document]
kashif: again this is my style... feel free to change it back 😄

src/gluonts/torch/model/deepar/deepar_estimator.py
```diff
@@ -0,0 +1,325 @@
...
+        self,
+        freq: str,
+        prediction_length: int,
+        trainer: pl.Trainer = pl.Trainer(),
+        context_length: Optional[int] = None,
+        num_layers: int = 2,
+        num_cells: int = 40,
+        cell_type: str = "LSTM",
+        dropout_rate: float = 0.1,
+        num_feat_dynamic_real: int = 0,
```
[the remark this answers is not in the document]
lostella: Yes, I agree. This doesn’t have much to do with Lightning, and is only needed in constructing the network because PyTorch needs to know the input size for the RNN construction (while in PyTorchTS that value is asked directly, I think that asking the size of all different inputs to the `forward` method—and then summing them up—makes more sense from the user perspective).

The “auto mode” you propose makes perfect sense in the estimator. We could just peek into the first (transformed) dataset entry and fill these value up.

src/gluonts/torch/model/deepar/deepar_network.py
```diff
@@ -0,0 +1,430 @@
...
+        self.prediction_length = prediction_length
+        self.dropout_rate = dropout_rate
+        self.num_feat_dynamic_real = num_feat_dynamic_real
+        self.num_feat_static_cat = num_feat_static_cat
+        self.num_feat_static_real = num_feat_static_real
+        self.cardinality = cardinality
+        self.embedding_dimension = (
+            embedding_dimension
+            if embedding_dimension is not None or self.cardinality is None
+            else [min(50, (cat + 1) // 2) for cat in self.cardinality]
```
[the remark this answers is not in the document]
lostella: I think we can maybe keep this in the network only. What I’m trying to achieve here is to have as much defaults as possible in the network directly, so that one can use it as a standalone object (with custom data loading, training procedure, whatever)

src/gluonts/torch/model/deepar/deepar_estimator.py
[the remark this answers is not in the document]
lostella: Ah, thanks, hadn’t noticed this! Sure, I’ll fix it

test/torch/model/test_deepar.py
lostella: Yes, that I’ll add. I also wanted to run a few training/evaluation jobs to compare with the MXNet implementation on some dataset, just to get a feeling for any apparent discrepancy: I’ll post results here as soon as I do

src/gluonts/torch/model/deepar/deepar_estimator.py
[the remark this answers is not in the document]
mbohlkeschneider: Could be a follow-up PR, for sure. Just wanted to point it out.

src/gluonts/torch/model/deepar/deepar_network.py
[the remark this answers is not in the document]
mbohlkeschneider: Good point.

src/gluonts/torch/model/deepar/deepar_estimator.py
```diff
@@ -0,0 +1,325 @@
...
+                FieldName.FEAT_TIME,
+                FieldName.OBSERVED_VALUES,
+            ],
+            dummy_value=self.distr_output.value_in_support,
+        )
+
+    def create_training_data_loader(
+        self,
+        data: Dataset,
+        shuffle_buffer_length: Optional[int] = None,
```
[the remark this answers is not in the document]
lostella: Yes, I agree. I think enabling shuffling by default would raise the question of "how much to shuffle": since this doesn't really enable uniform random picking of time series from which to select training windows (that would require indexable datasets, allowing random access), but only "pseudo-random" iteration, then one should come up with a reasonable buffer length here. In some cases, putting 10 here might mean that the shuffling buffer will contain the whole dataset twice, in other cases that might be a tiny fraction of the overall dataset.

I think we should look into using indexable datasets: those will be enough for several use cases, and many things like sampling windows completely at random would become almost a no-brainer in those cases.

src/gluonts/torch/model/deepar/deepar_network.py
```diff
@@ -0,0 +1,430 @@
...
+        )
+        self.target_shape = distr_output.event_shape
+        self.proj_distr_args = distr_output.get_args_proj(num_cells)
+        if self.cardinality is not None:
+            self.embedder = FeatureEmbedder(
+                cardinalities=self.cardinality,
+                embedding_dims=self.embedding_dimension,
+            )
+        else:
+            self.embedder = None
```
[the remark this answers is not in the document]
lostella: I think let's make `cardinality` compulsory in the network for now. I think this is a leftover from when I was trying to make inputs to `forward` optional: eventually that would be nice to have, so that if one sets `num_feat_xxx_yyy = 0` the coresponding input can be missing.

src/gluonts/torch/model/deepar/deepar_network.py
lostella: Yeah I think the scaling part can be simplified

src/gluonts/torch/model/deepar/_network.py
```diff
@@ -0,0 +1,450 @@
...
+    def __init__(
+        self,
+        *args,
+        loss: DistributionLoss = NegativeLogLikelihood(),
+        optimizer: Optional[torch.optim.Optimizer] = None,
+        **kwargs,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.loss = loss
+        self.optimizer = optimizer or torch.optim.Adam(
```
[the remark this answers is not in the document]
borchero: You might get around this by copying the `state_dict` of the optimizer, but I wouldn't do that ... you can either let the user specify `optimizer_class` and `optimizer_kwargs` or let him provide a lambda function.

src/gluonts/torch/model/deepar/_network.py
```diff
@@ -0,0 +1,450 @@
...
+        return self.sampling_decoder(
+            past_target=past_target,
+            time_feat=future_time_feat,
+            static_feat=static_feat,
+            scale=scale,
+            begin_states=state,
+        )
+
+
+class DeepARLightningNetwork(DeepARNetwork, pl.LightningModule):
```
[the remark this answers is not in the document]
borchero: Also, I would use composition over inheritance, i.e. pass the initialized `DeepARNetwork` to the initializer.

src/gluonts/torch/model/deepar/_network.py
```diff
@@ -0,0 +1,450 @@
...
+        optimizer: Optional[torch.optim.Optimizer] = None,
+        **kwargs,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.loss = loss
+        self.optimizer = optimizer or torch.optim.Adam(
+            self.parameters(), lr=1e-3, weight_decay=1e-8
+        )
+
+    def _compute_loss(
```
borchero, line 397: Would move it to a standalone module `class DeepARLoss(nn.Module)`.

src/gluonts/torch/model/estimator.py
```diff
@@ -0,0 +1,181 @@
...
+    `create_training_data_loader`, and `create_validation_data_loader`.
+    """
+
+    @validated()
+    def __init__(
+        self,
+        trainer: pl.Trainer,
+        lead_time: int = 0,
+        device: torch.device = torch.device("cpu"),
+        dtype: torch.dtype = torch.float32,
```
borchero, line 50: Device and dtype are already handled by `pl.Trainer`

src/gluonts/torch/model/estimator.py
```diff
@@ -0,0 +1,181 @@
...
+        cache_data: bool = False,
+        **kwargs,
+    ) -> TrainOutput:
+        transformation = self.create_transformation()
+
+        transformed_training_data = TransformedDataset(
+            training_data, transformation, is_train=True
+        )
+
+        training_network = self.create_network().to(self.device, self.dtype)
```
borchero, line 122: Don't move, the trainer will do that

src/gluonts/torch/model/deepar/_network.py
```diff
@@ -0,0 +1,450 @@
...
+        )
+
+        if len(self.target_shape) == 0:
+            loss_weights = observed_values
+        else:
+            loss_weights = observed_values.min(dim=-1, keepdim=False)
+
+        return weighted_average(loss, weights=loss_weights)
+
+    def _loss_step(self, batch):
```
borchero, line 421: Missing types?

src/gluonts/torch/model/deepar/_network.py
```diff
@@ -0,0 +1,450 @@
...
+        )
+        return self._compute_loss(
+            distr,
+            batch["past_target"],
+            batch["future_target"],
+            batch["past_observed_values"],
+            batch["future_observed_values"],
+        )
+
+    def training_step(self, batch, *args, **kwargs):
```
borchero, line 440: Missing types? Apart from batch, only `batch_index: int` is passed

src/gluonts/torch/modules/loss.py
```diff
@@ -0,0 +1,52 @@
...
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+import torch
+
+
+class DistributionLoss(torch.nn.Module):
```
borchero, line 17: Why do we have this base class?

src/gluonts/torch/util.py
```diff
@@ -0,0 +1,94 @@
...
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+from typing import Optional, Type
+import inspect
+
+import torch
+
+
+def copy_parameters(
```
borchero, line 20: Guess, this should be obsolete?

src/gluonts/torch/util.py
```diff
@@ -0,0 +1,94 @@
...
+    strict:
+        whether to strictly enforce that the keys
+        in :attr:`state_dict` match the keys returned by this module's
+        :meth:`~torch.nn.Module.state_dict` function. Default: ``True``
+    """
+
+    net_dest.load_state_dict(net_source.state_dict(), strict=strict)
+
+
+def get_forward_input_names(module: Type[torch.nn.Module]):
```
borchero, line 43: Also think, that we don't need this anymore, the LightningModule does it in a more explicit fashion.

src/gluonts/torch/util.py
```diff
@@ -0,0 +1,94 @@
...
+            weights.sum(dim=dim) if dim else weights.sum(), min=1.0
+        )
+        return (
+            weighted_tensor.sum(dim=dim) if dim else weighted_tensor.sum()
+        ) / sum_weights
+    else:
+        return x.mean(dim=dim)
+
+
+class IterableDataset(torch.utils.data.IterableDataset):
```
borchero, line 89: Is this really necessary? `torch`'s `IterableDataset` is really just `__iter__` + some `NotImplementedError`s raised when accessing `__len__` or `__getitem__`

src/gluonts/torch/model/deepar/_network.py
```diff
@@ -0,0 +1,450 @@
...
+        )
+        outputs, state = self.rnn(inputs)
+
+        # outputs: (batch_size, seq_len, num_cells)
+        # state: list of (num_layers, batch_size, num_cells) tensors
+        # scale: (batch_size, 1, *target_shape)
+        # static_feat: (batch_size, num_features + prod(target_shape))
+        return outputs, state, scale, static_feat
+
+    def sampling_decoder(
```
borchero, line 229: Would introduce a new class `DeepARDecoder` to make this class a little smaller

src/gluonts/torch/model/deepar/_network.py
```diff
@@ -0,0 +1,450 @@
...
+        self,
+        feat_static_cat: torch.Tensor,
+        feat_static_real: torch.Tensor,
+        past_time_feat: torch.Tensor,
+        past_target: torch.Tensor,
+        past_observed_values: torch.Tensor,
+        future_time_feat: torch.Tensor,
+        future_target: Optional[torch.Tensor] = None,
+        sampling: Optional[bool] = None,
+    ) -> Union[torch.Tensor, torch.distributions.Distribution]:
```
borchero, line 359: (1) `Union` types don't work with TorchScript
(2) `torch.distributions.Distribution` cannot be moved between devices (they are missing a `to` method). Not sure if training on GPUs is possible when doing it like this.

borchero: I would rather always return a `torch.Tensor` and construct the distribution object in the LightningModule

src/gluonts/torch/model/deepar/_estimator.py
```diff
@@ -0,0 +1,326 @@
...
+            lags_seq=self.lags_seq,
+            scaling=self.scaling,
+            num_parallel_samples=self.num_parallel_samples,
+        )
+
+    def create_predictor(
+        self,
+        transformation: Transformation,
+        network: DeepARNetwork,
+        device: torch.device,
```
borchero, line 314: Why do we need the device here?

src/gluonts/torch/model/deepar/_network.py
```diff
@@ -0,0 +1,450 @@
...
+        self,
+        feat_static_cat: torch.Tensor,
+        feat_static_real: torch.Tensor,
+        past_time_feat: torch.Tensor,
+        past_target: torch.Tensor,
+        past_observed_values: torch.Tensor,
+        future_time_feat: Optional[torch.Tensor] = None,
+        future_target: Optional[torch.Tensor] = None,
+    ) -> Tuple[
+        torch.Tensor, Union[torch.Tensor, List], torch.Tensor, torch.Tensor
```
borchero, line 165: (1) Again, `Union` doesn't work with TorchScript
(2) `List` is missing generic type parameter

src/gluonts/torch/model/deepar/_network.py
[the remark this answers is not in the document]
lostella: Yeah I'm not sure about that, I was thinking about it and couldn't spot any clear advantage to that (but maybe I forgot from our previous discussions?)

src/gluonts/torch/model/deepar/_network.py
```diff
@@ -0,0 +1,450 @@
...
+            scale=scale,
+            begin_states=state,
+        )
+
+
+class DeepARLightningNetwork(DeepARNetwork, pl.LightningModule):
+    def __init__(
+        self,
+        *args,
+        loss: DistributionLoss = NegativeLogLikelihood(),
```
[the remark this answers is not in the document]
lostella: And not expose this as an option, you mean? There are other things one may want to optimize, for example CRPS

src/gluonts/torch/model/deepar/_network.py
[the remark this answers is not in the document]
lostella: OK well, the other option could be not to expose this and if someone wants to customize it, can extend the class and re-define `configure_optimizers`

src/gluonts/torch/model/estimator.py
lostella: Nice! I'll happily get rid of them

src/gluonts/torch/model/deepar/_network.py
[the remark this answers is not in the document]
borchero: Agree. You should still be able to pass very basic config options for Adam (i.e. the learning rate), I would say.

src/gluonts/torch/model/deepar/_network.py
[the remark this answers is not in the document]
borchero: Ah ok, didn't think about that. Never mind then

src/gluonts/torch/modules/loss.py
lostella: To make it explicit, when an object of type `DistributionLoss` is asked, that it should handle a `Distribution` as first argument to `forward`. Not having it would make this requirement a bit more implicit... but it's also not a big deal

src/gluonts/torch/util.py
lostella: Yeah we will eventually get rid of obsolete code :-)

src/gluonts/torch/util.py
lostella: Same as above, we will do clean up

src/gluonts/torch/model/deepar/_network.py
lostella: Good points. It is hard however to make this type-stable if we want to use it for both training and prediction: to return a distribution you would need to return a tuple of tensors.

src/gluonts/torch/util.py
lostella: Exactly, and also `__iter__` is not defined there. Its docstring says:
```
All datasets that represent an iterable of data samples should subclass it.
```
so I did it 😄

lostella:
Some results I got with the current code from the PR

```
                            count      mean       std       min       25%       50%       75%       max
dataset     model
electricity DeepAR_pytorch    5.0  0.068094  0.010365  0.052888  0.061976  0.073358  0.074637  0.077613
m4_daily    DeepAR_pytorch    5.0  0.028392  0.006089  0.023068  0.025569  0.027012  0.027466  0.038843
m4_hourly   DeepAR_pytorch    5.0  0.064823  0.026216  0.037337  0.043694  0.060085  0.084379  0.098621
traffic     DeepAR_pytorch    5.0  0.133552  0.001598  0.132023  0.132279  0.132940  0.135086  0.135431
```

src/gluonts/torch/model/deepar/__init__.py
```diff
@@ -11,8 +11,23 @@
...
 # permissions and limitations under the License.
 
-# !!! DO NOT MODIFY !!! (pkgutil-style namespace package)
+from ._module import (
+    LaggedLSTM,
+    DeepARModel,
+)
 
-from pkgutil import extend_path
+from ._lightning_module import DeepARLightningModule
```
borchero, line 19: Why do we use hidden package names?

src/gluonts/torch/model/deepar/__init__.py
```diff
@@ -11,8 +11,23 @@
...
 
-__path__ = extend_path(__path__, __name__)  # type: ignore
+from ._estimator import DeepAREstimator
+
+__all__ = [
+    "LaggedLSTM",
```
borchero, line 24: Wouldn't expose it from the "top-level" DeepAR package

src/gluonts/torch/model/deepar/_lightning_module.py
```diff
@@ -0,0 +1,88 @@
...
+            batch["past_target"],
+            batch["past_observed_values"],
+            batch["future_time_feat"],
+            batch["future_target"],
+        )
+        distr = self.model.output_distribution(params, scale)
+        past_target = batch["past_target"]
+        future_target = batch["future_target"]
+        past_observed_values = batch["past_observed_values"]
+        future_observed_values = batch["future_observed_values"]
```
borchero, line 51: Put these assignments at the top? They can already be used in `unroll_lagged_rnn`

src/gluonts/torch/model/deepar/_module.py
```diff
@@ -0,0 +1,323 @@
...
+        num_feat_static_cat: int,
+        cardinality: List[int],
+        embedding_dimension: Optional[List[int]] = None,
+        num_layers: int = 2,
+        hidden_size: int = 40,
+        dropout_rate: float = 0.1,
+        distr_output: DistributionOutput = StudentTOutput(),
+        lags_seq: Optional[List[int]] = None,
+        scaling: bool = True,
+        num_parallel_samples: int = 100,
```
borchero, line 140: I would not pass this as an initialization option here, but rather pass this to the `forward` method directly. The number of samples doesn't have an impact on the model architecture.

src/gluonts/torch/model/deepar/_module.py
```diff
@@ -0,0 +1,323 @@
...
+        )
+
+        features = torch.cat((expanded_static_feat, time_feat), dim=-1)
+
+        output, new_state = self.lagged_rnn(prior_input, input, features)
+
+        params = self.param_proj(output)
+        return params, scale, static_feat, new_state
+
+    def output_distribution(
```
borchero, line 249: I'm not sure if this function is fully JIT supported for all `torch.Distribution` objects. Mark it as `@torch.jit.ignore` otherwise.

borchero (changes requested):

src/gluonts/torch/model/deepar/_module.py
lostella: True, on the other hand, the predictor that wraps only provides data to the model. A compromise is to leave it here, and allow an optional override in the `forward` signature

src/gluonts/torch/model/deepar/_estimator.py
```diff
@@ -0,0 +1,323 @@
...
+]
+
+
+class DeepAREstimator(PyTorchLightningEstimator):
+    @validated()
+    def __init__(
+        self,
+        freq: str,
+        prediction_length: int,
+        trainer: pl.Trainer = pl.Trainer(max_epochs=100),
```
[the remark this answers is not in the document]
lostella: oh ok, I didn't know this was there actually, nice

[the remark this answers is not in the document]
lostella: this is however enforced below in the class, by islice-ing the data loader

lostella:
I ran a bit more comparisons: the minimum scores are very close between the MXNet and PyTorch implementations, with the PyTorch one sometimes better. In some cases, higher variance is observed with the PyTorch estimator, resulting in worst worst-case runs, especially in `electricity` and `exchange_rate`. I think this is a problem with some different training options, like gradient clipping and "best model selection" during training (both absent in the PyTorch training loop I ran).

I will do some more runs on `electricity` and `exchange_rate` with the training options better aligned to check if variance can be reduced. But the minimum scores suggest that the networks are essentially equivalent: another way to check this would be to initialize the networks with the same weights, and check their output on the same input data.

```
mean_wQuantileLoss
                            count      mean       std       min       25%       50%       75%       max
dataset       model
electricity   DeepAR_mxnet   10.0  0.058250  0.003542  0.054607  0.055732  0.057031  0.060223  0.065524
              DeepAR_torch   10.0  0.070389  0.019396  0.050715  0.057407  0.067428  0.075375  0.114468
exchange_rate DeepAR_mxnet   10.0  0.011979  0.003958  0.007581  0.009066  0.010209  0.014335  0.019022
              DeepAR_torch   10.0  0.015999  0.009422  0.007932  0.009165  0.012377  0.022294  0.035100
solar-energy  DeepAR_mxnet   10.0  0.395402  0.006397  0.386605  0.390082  0.396136  0.397910  0.408057
              DeepAR_torch   10.0  0.397074  0.006896  0.386119  0.392976  0.395508  0.402427  0.408164
traffic       DeepAR_mxnet   10.0  0.130629  0.003020  0.126329  0.129284  0.129797  0.133450  0.134651
              DeepAR_torch   10.0  0.131482  0.006321  0.126205  0.126695  0.130697  0.132835  0.147017
```

kashif:
@lostella may I ask what the `fixed bugs` commit does (dba30e9)? thanks!

lostella:
> @lostella may I ask what the `fixed bugs` commit does ([dba30e9](https://github.com/awslabs/gluon-ts/commit/dba30e940f798dd96f41610d8e9e3dd2e1ee0f21))? thanks!

@kashif I refactored the network a bit, and had introduced bugs in doing that :-) (in the sampling procedure in particular) 

I got the dimensions wrong in the first sample, plus I was handling scale incorrectly: I was getting samples on the original scale, and feeding them directly into the RNN. Now the loop produces scaled samples (I don't give the `scale` argument when constructing the distribution), and samples are scaled back only at the end

lostella:
Adding model checkpointing provide much more stable results on `electricity` and `exchange_rate`:

```
mean_wQuantileLoss
                            count      mean       std       min       25%       50%       75%       max
dataset       model
electricity   DeepAR_torch   10.0  0.054265  0.004268  0.050959  0.051395  0.053759  0.054148  0.065283
exchange_rate DeepAR_torch   10.0  0.015017  0.005611  0.007968  0.009751  0.015622  0.018340  0.024659
```

src/gluonts/torch/model/estimator.py
```diff
@@ -147,16 +146,7 @@ def train_model(
...
-            monitor=monitor,
-            mode="min",
-            verbose=True,
-            patience=10,
-            check_on_train_epoch_end=validation_data is None,
-        )
-        trainer = pl.Trainer(
-            callbacks=[checkpoint, early_stopping], **self.trainer_kwargs
-        )
+        trainer = pl.Trainer(callbacks=[checkpoint], **self.trainer_kwargs)
```
borchero, line 149: Checkpointing is still overridden by passed callbacks, but not too important for this PR.

lostella: well it actually is, because who knows what model will be loaded at the end... so I'd say let's keep the default list of callbacks maybe

borchero (approved):

mdancho84:
Is there any reason that the torch.deepar module wasn't included in the `gluonts` 0.8.0 release? I'm not seeing it in there, and I was looking forward to testing it out.

lostella:
@mdancho84 the module should be there: https://github.com/awslabs/gluon-ts/tree/v0.8.x/src/gluonts/torch/model/deepar

lostella:
@mdancho84 I confirm after installing `gluonts==0.8.0` in a fresh environment, that the following runs fine (note that you need to have `pytorch_lightning` installed as well to use the `DeepAREstimator` (besides `torch` obviously)

```
```

mdancho84:
@lostella Thanks for this! The problem was not having pytorch lightning installed. 

Also, for what it's worth, I might as well introduce myself. 
 - I am the creator of `modeltime`, an R forecasting ecosystem. https://github.com/business-science/modeltime
 - I've integrated GluonTS into the system here: https://github.com/business-science/modeltime.gluonts

I've been a big fan of GluonTS for several years and even teach it in my [Time Series Course](https://university.business-science.io/p/ds4b-203-r-high-performance-time-series-forecasting/). 

I really appreciate all of the work you are doing.

mdancho84:
@lostella Just to Follow Up: I've successfully got the Pytorch DeepAR implementation working. Initial results are promising:

- Speedup of 1.6X 
- Similar accuracy and forecast results of GluonTS vs Torch Implementations

Test:
- 10 Epochs on Walmart Time Series
- 1.12 min to train 10 epochs vs 42 sec

image

image

lostella:
> Also, for what it's worth, I might as well introduce myself.

Oh, thanks for doing all this, and for the kind words!

The results look indeed promising, I had also observed some speedup with the PyTorch version. Is not 100% feature complete compared to the MXNet one: some new imputation strategies for handling missing values during the RNN unrolling were added recently, which would take some time to replicate in PyTorch. We will see about these, but it's important that the biggest part of the model is there.

I think we will generally look into improving PyTorch support, tooling, model components, and try to add new models in PyTorch going forward. So please let me know if you find anything off in your experiments or in the code, we're very interested.

kashif:
@lostella I forgot if I talked about this with you but the mxnet deepAR's imputation scheme is not quite correct as it uses the RNN to impute the values if they are missing, but it's doing so from the past... while imputation would need to consider the past and future values from the point of view of the missing time-point...

mdancho84:
@lostella Thanks for your help. The kind words are well deserved. 

If the API has stabilized for the Torch `DeepAREstimator`, then the next steps in my opinion are more algorithms. I've been getting good results with Deep State and GP Forecaster. 

The speed-up and pylightning logging gives a nice compliment to the GluonTS versions. The GluonTS framework then makes it easy to forecast and compare. Win win.

lostella:
@kashif true but you would need some other architecture for that, like bi-directional RNN or so, right? With the architecture at hand, the implemented imputation is one thing one can do, not necessarily the best of course.

@mdancho84 the API should be stable enough, but also what we tried enforcing in the torch-based DeepAR implementations is to have as much independent network/estimator classes as possible, so that potentially one can start implementing the “core” of the model while the higher level wrapper interface stabilizes. I think the ideal situation is one where the network is 100% usable in its own, in the sense that one can construct it easily (so: options are documented in the network, and have reasonable defaults) and train it with custom data loading logic and training loop.

mdancho84:
@lostella Sounds great. I can always enforce `gluonts==0.8.0` so my students have a working Torch DeepAR. Then upgrade when more models are released (if anything changes). I just pushed the latest version of Modeltime GluonTS. Looks good: https://github.com/business-science/modeltime.gluonts/issues/30

PR #1491, opened 2021-05-26
Remove loader-specific args from train.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/mx/model/estimator.py
+++ b/src/gluonts/mx/model/estimator.py
@@ -100,7 +100,7 @@ class GluonEstimator(Estimator):
         self.batch_size = batch_size
         self.trainer = trainer
         self.dtype = dtype
-        self.loader_args = loader
+        self.loader = loader
 
     @classmethod
     def from_hyperparameters(cls, **hyperparameters) -> "GluonEstimator":
--- a/src/gluonts/mx/model/estimator.py
+++ b/src/gluonts/mx/model/estimator.py
@@ -16,7 +16,7 @@ from functools import partial
 
 import numpy as np
 from mxnet.gluon import HybridBlock
-from pydantic import ValidationError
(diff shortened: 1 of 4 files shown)


src/gluonts/mx/model/estimator.py
```diff
@@ -151,11 +156,6 @@ def train_model(
 
         training_data_loader = self.create_training_data_loader(
             transformed_training_data
-            if not cache_data
-            else Cached(transformed_training_data),
-            num_workers=num_workers,
-            num_prefetch=num_prefetch,
-            shuffle_buffer_length=shuffle_buffer_length,
```
lostella, line 158: These can actually be updated to use `loader_args` right?

jaheba: Just need to find where they are used.

src/gluonts/mx/model/estimator.py
```diff
@@ -35,6 +35,39 @@
...
+        return TrainDataLoader(
+            dataset=dataset,
+            transform=transform,
+            batch_size=batch_size,
+            stack_fn=stack_fn,
+            decode_fn=decode_fn,
+            num_workers=self.num_workers,
+            num_prefetch=self.num_prefetch,
+            shuffle_buffer_length=self.shuffle_buffer_length,
+        )
```
lostella, line 57: Would it make sense to directly do here what happens in `TrainDataLoader`?

Also, this overloads the "data loader" terminology a bit, would it make sense to rename this class to something else, like `DataLoaderConfig` or something similar?

lostella: For the first point: that would mean getting rid of these functions https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/dataset/loader.py#L226-L377 but we can also keep them around

PR #1519, opened 2021-05-31
Serde: Allow encoding of functions and methods.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/core/_base.py
+++ b/src/gluonts/core/_base.py
@@ -12,18 +12,25 @@
 # permissions and limitations under the License.
 
 
-def fqname_for(cls: type) -> str:
+def fqname_for(obj: type) -> str:
     """
-    Returns the fully qualified name of ``cls``.
+    Returns the fully qualified name of ``obj``.
 
     Parameters
     ----------
-    cls
+    obj
         The class we are interested in.
 
     Returns
     -------
     str
-        The fully qualified name of ``cls``.
+        The fully qualified name of ``obj``.
     """
-    return f"{cls.__module__}.{cls.__qualname__}"
+
+    if "<locals>" in obj.__qualname__:
+        raise RuntimeError(
+            "Can't get fully qualified name of locally defined object. "
+            f"{obj.__qualname__}"
+        )
+
+    return f"{obj.__module__}.{obj.__qualname__}"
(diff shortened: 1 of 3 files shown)


jaheba:
This should allow us also to define transformations more easily in terms of functions.

test/core/test_serde.py
```diff
@@ -188,3 +189,13 @@ def test_timestamp_encode_decode() -> None:
 )
 def test_string_escape(serialize_fn) -> None:
     assert serialize_fn(r"a\b") == r"a\b"
+
+
+def test_serde_fq():
+    serde.encode(test_serde_fq)
```
lostella, line 195: Is this testing something? I guess we should have a test running `decode(encode(f))` for some function `f` to check that it reconstructs it, right?

test/core/test_serde.py
```diff
@@ -21,6 +21,7 @@
 from pydantic import BaseModel
 
 from gluonts.core import serde
+from gluonts.mx.prelude import *
```
lostella, line 24: Not sure why this import is needed: I’m guessing to bring names into scope so that deserialization works, but I don’t see it used in the test changed by this PR

jaheba: I couldn't run the test by itself, since it's relying on being able to serialize numpy and mxnet arrays.

src/gluonts/core/_base.py
```diff
@@ -12,18 +12,25 @@
 # permissions and limitations under the License.
 
 
-def fqname_for(cls: type) -> str:
+def fqname_for(obj: type) -> str:
```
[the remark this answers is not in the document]
jaheba: Not anymore I think.

jaheba:
One thing I'm also unsure about is how to handle bound methods:

```py
class X:
    def m(self):
        pass
```

`X.m` and `X().m` resolve to the same name, whilst the bound method has `.__self__` which points to the object and `.__func__` which points to `X.m`.

We could translate this to `partial(obj.__func__, obj.__self__)` and then serialise it.

However if you serialise `[x, x.m]` then the two `x` become two separate instances after initialisation (the same is true for `[x, x]` though.

lostella:
thanks, windows

lostella (approved):

PR #1524, opened 2021-06-01
Mark `test_rotbaum_smoke` as flaky.

Schmedu:
*Issue #, if available:*
Test is flaky as we saw in https://github.com/awslabs/gluon-ts/pull/1523
*Description of changes:*


jaheba:
Do we know why this is flaky?

Schmedu:
> Do we know why this is flaky?

No, but looking at some other prs (like https://github.com/awslabs/gluon-ts/pull/440, https://github.com/awslabs/gluon-ts/pull/380, https://github.com/awslabs/gluon-ts/pull/353) there is no elaboration on what the source of the flakyness is, but sure, I agree with you that we should investigate it. Though, I think solving it for now with a hotfix is more important than to block this pr and let it bother other people. Could we for now just add a `TODO` and investigate it later as this is probably a Mandelbug and therefore hard to replicate, so this could take some time to solve it. Maybe we should do this also for the other prs?

Schmedu:
No longer necessary

PR #1525, opened 2021-06-01
Fix test workflows

lostella:
* Disable tests on Windows (too flaky, they slow down development; we run nightly jobs on the latest release and that's OK for now)
* Give relevant names to workflows jobs (instead of just "build")
* Add missing Autogluon dependency in nightly workflows (they have been failing for ~3 weeks due to this)


jaheba (approved):

lostella:
Interesting, changing the job names messes up with the approval rules, I'll disable-and-re-enable

PR #1527, opened 2021-06-02
Simplify transformation chains construction

lostella:
*Description of changes:* This is an attempt at simplify the way estimators set up the data transformation chains they need. Goal is to make the code more readable (less verbose, more linear).

Most important changes:
* Using `+` and `+=` extensively, in place of constructing lists to be passed to `Chain`
* Replacing sequences of `SetField` + `AsNumpyArray` with the new `SetArrayField`
* Replacing `AddAgeFeature` transform in favor of `AddTimeFeature` with dedicated `Age` and `LogAge` features


src/gluonts/model/deepar/_estimator.py
```diff
@@ -306,71 +305,65 @@ def create_transformation(self) -> Transformation:
         if not self.use_feat_dynamic_real:
             remove_field_names.append(FieldName.FEAT_DYNAMIC_REAL)
 
-        return Chain(
-            [RemoveFields(field_names=remove_field_names)]
-            + (
-                [SetField(output_field=FieldName.FEAT_STATIC_CAT, value=[0.0])]
-                if not self.use_feat_static_cat
-                else []
+        transform = RemoveFields(field_names=remove_field_names)
```
borchero, line 308: Maybe call it "transforms" or explicitly state that it's a `Chain` here? Using `+=` on a single transform seems weird at first sight.

borchero:
Already looks much more clear than before!! Although `create_transformation` is still a bit of a beast :D

src/gluonts/model/tft/_estimator.py
```diff
@@ -155,161 +155,106 @@ def __init__(
...
+            transform += AsNumpyArray(field=name, expected_ndim=1)
+
+        for name in chain(
+            self.static_feature_dims.keys(),
+            self.dynamic_cardinalities.keys(),
+        ):
+            transform += AsNumpyArray(field=name, expected_ndim=1)
+
+        for name in self.dynamic_feature_dims.keys():
+            transform += AsNumpyArray(field=name, expected_ndim=2)
```
borchero, line 170: Why is `expected_ndim` set to `2` here? Could we maybe group transforms with `expected_ndim = 1` and `expected_ndim = 2` and give it a more descriptive name?

src/gluonts/model/tft/_estimator.py
```diff
@@ -155,161 +155,106 @@ def __init__(
...
+
+        transform += AddObservedValuesIndicator(
+            target_field=FieldName.TARGET,
+            output_field=FieldName.OBSERVED_VALUES,
```
borchero, line 174: Could we use defaults here? Especially for this transform, it seems like `target_field` and `output_field` almost never change.

src/gluonts/model/tft/_estimator.py
```diff
@@ -155,161 +155,106 @@ def __init__(
...
         )
 
         if self.static_cardinalities:
-            transforms.append(
-                VstackFeatures(
-                    output_field=FieldName.FEAT_STATIC_CAT,
-                    input_fields=list(self.static_cardinalities.keys()),
-                    h_stack=True,
-                )
+            transform += HstackFeatures(
```
borchero, line 186: Ha! So much better than `Vstack(hstack=True)` :D

src/gluonts/model/tft/_estimator.py
```diff
@@ -155,161 +155,106 @@ def __init__(
...
-                        output_field=FieldName.FEAT_STATIC_CAT,
-                        value=[0.0],
-                    ),
-                    AsNumpyArray(
-                        field=FieldName.FEAT_STATIC_CAT, expected_ndim=1
-                    ),
-                ]
+            transform += SetArrayField(
+                output_field=FieldName.FEAT_STATIC_CAT,
+                value=[0.0],
```
borchero, line 193: Again, would a default of `[0.0]` make sense here? I haven't seen another value yet.

src/gluonts/transform/convert.py
```diff
@@ -87,7 +88,7 @@ def transform(self, data: DataEntry) -> DataEntry:
         return data
 
 
-class VstackFeatures(SimpleTransformation):
+class StackFeatures(SimpleTransformation):
```
jaheba, line 91: Why don't we just call it `Stack`?

src/gluonts/model/tft/_estimator.py
lostella: Yes, I forgot to update this: `AddObservedTargetIndicator` fixes those arguments

src/gluonts/model/tft/_estimator.py
lostella: Because that's the expected number of dimensions of the array 😉 we expect lists-of-lists or 2-dimensional arrays whenever there's an additional axis other than time. In this case, you expect potentially multiple time-varying features

src/gluonts/model/tft/_estimator.py
lostella: `[[0.0]]` is used at times, in general I think it's hard to have a default here because of shape

PR #1538, opened 2021-06-04
Extend FileDataset's Parameters to load_datasets

unibeck:
*Description of changes:*
Added one_dim_target to load_datasets. This allows a developer to serialize their multivariate dataset as FileDataset to then load it in later without error. Also added FileDataset's cache parameter to load_datasets to enable full pass through of FileDataset's capabilities. 


unibeck:
Hello @jaheba and @lostella. I opened this PR almost a week ago and I am looking for an update so I can set expectations. Please let me know if I am missing something, you guys would like me to change something, or if there are other higher priority tasks at hand. While creating the PR it mentioned that I should add a label for faster triaging, though I don't seem to have the ability/permissions to change my PR.

P.S. if there is anyone else relevant to this matter that I may have left out, please include them.

src/gluonts/dataset/common.py
```diff
@@ -455,7 +455,11 @@ def __call__(self, data: DataEntry) -> DataEntry:
 
 
 def load_datasets(
-    metadata: Path, train: Path, test: Optional[Path]
+        metadata: Path,
```
lostella, line 458: The indentation here could be reduced, I think

lostella (changes requested):
@unibeck thanks for this! The change makes perfect sense to me. I only have a comment in-line about formatting, after which I think this could be merged

src/gluonts/dataset/common.py
unibeck: Must have been my IDE's default settings. This should be fixed now.

unibeck: Ha okay, now it should be all good. I ran `black src` as suggested by the pipeline.

lostella (approved):

PR #1544, opened 2021-06-07
QRX fixes and added functionality

zoolhasson:
*Issue #, if available:*

*Description of changes:*

1. _model.py:
1.A. Fixed the issue of the last cell being too small by making the last cell be itself merged with the penultimate cell. This improves performance significantly.
1.B. Added model_is_already_trained to fit function. If True, don't fit the underlying model. This allows one to apply QRX to pre-trained models.
2. _predictor.py: Added train_QRX_only_using_timestep. If it is -1, nothing changes. Otherwise, it should be an integer between 0 and forecast horizon - 1. In that case, instead of creating a different partition for each timestep in the forecast horizon (and therefore train a different underlying model for each timestep in the forecast horizon), it will only train one underlying model (and therefore have the same partition for all timesteps) based on the the target from the train_QRX_only_using_timestep^{th} timestep. This is an experimental option, which cuts down on training time significantly, while degrading performance in general. On the other hand, it allows you to use more context windows to create the partition.
3. Changed the name clump_size to min_bin_size, which is more descriptive. 
4. Added seed support.


lostella:
marked this as breaking to prevent it from ending up in a previous release branch

src/gluonts/model/rotbaum/_model.py
```diff
@@ -170,7 +172,8 @@ def fit(
             )
             x_train = x_train[idx]
             y_train = y_train[idx]
-        self.model.fit(x_train, y_train, **kwargs)
+        if model_is_already_trained is False:
```
dcmaddix, line 175: can simplify to:
`if not model_is_already_trained:`

src/gluonts/model/rotbaum/_model.py
```diff
@@ -196,6 +201,15 @@ def fit(
         del cell_values_dict
         gc.collect()
 
+        df = pd.DataFrame({"preds": self.sorted_train_preds})
+        df["bins"] = df["preds"].apply(lambda x: self.preds_to_id[x])
+        final_id = df["bins"].drop_duplicates().values[-1]
```
dcmaddix, line 206: `bins = df["bins"].drop_duplicates().values()`, so you only have to compute once and then take the last and second to last element

src/gluonts/model/rotbaum/_model.py
```diff
@@ -196,6 +201,15 @@ def fit(
...
+        penultimate_id = df["bins"].drop_duplicates().values[-2]
+        if len(self.id_to_bins[final_id]) < self.min_bin_size:
+            self.id_to_bins[final_id] = (
```
dcmaddix, line 209: `self.id_to_bins[final_id] += self.id_to_bins[penultimate_id]`

dcmaddix (changes requested):
sounds good to me - just minor changes we can merge after checking with @lostella on the breaking change.

dcmaddix (approved):

Issue #1513, opened 2021-05-31
Training on M4 Daily fails for multiple models


borchero:
## Description

Training on the M4 Daily fails on multiple models provided by GluonTS, namely:

- DeepAR
- NBEATS
- Simple Feedforward
- Temporal Fusion Transformer

Funnily, training always fails after 70 epochs when using a batch size of 32 and 2472 batches per epoch. The issue seems to be related to data transformations.

Note that training works using MQRNN/MQCNN.

## To Reproduce

```python
from gluonts.dataset.repository.datasets import get_dataset
from gluonts.model.deepar import DeepAREstimator
from gluonts.mx.trainer import Trainer
from gluonts.mx.trainer.learning_rate_scheduler import LearningRateReduction

dataset = get_dataset("m4_daily")
callbacks = [
    LearningRateReduction(
        objective="min", patience=9, base_lr=1e-3, decay_factor=0.5
    )
]
estimator = DeepAREstimator(
    freq=dataset.metadata.freq,
    prediction_length=dataset.metadata.prediction_length,
    trainer=Trainer(epochs=81, num_batches_per_epoch=2472, callbacks=callbacks),
    context_length=4 * dataset.metadata.prediction_length,
)
estimator.train(dataset.train)
```

And wait until epoch 70...

## Error message or code output

Stacktrace for training using any of the models above:

```
File "/usr/local/lib/python3.8/site-packages/gluonts/mx/model/estimator.py", line 201, in train
    return self.train_model(
  File "/usr/local/lib/python3.8/site-packages/gluonts/mx/model/estimator.py", line 176, in train_model
    self.trainer(
  File "/usr/local/lib/python3.8/site-packages/gluonts/mx/trainer/_base.py", line 436, in __call__
    epoch_loss = loop(
  File "/usr/local/lib/python3.8/site-packages/gluonts/mx/trainer/_base.py", line 325, in loop
    for batch_no, batch in enumerate(it, start=1):
  File "/usr/local/lib/python3.8/site-packages/tqdm/std.py", line 1178, in __iter__
    for obj in iterable:
  File "/usr/local/lib/python3.8/site-packages/gluonts/itertools.py", line 51, in get_batch
    return list(itertools.islice(it, batch_size))
  File "/usr/local/lib/python3.8/site-packages/gluonts/transform/dataset.py", line 53, in __iter__
    yield from self.transformation(
[...]

PR #1546, opened 2021-06-09
Set max_idle_transforms to the length of the dataset

borchero:
Solves #1513 

*Description of changes:*

The `max_idle_transforms` in the `FlatMapTransformation` is set dynamically based on the number of time series in the dataset. The value is passed to the `InstanceSplitter` when the estimator constructs the training and validation transforms.

It also fixes a couple of issues indicated by pylint and removes dead code in some of the files touched.


lostella:
Maybe it would make sense to split this into two PRs? One to fix the linting issues, the other one addressing #1513 what do you think?

jaheba:
Something we could do is to set the `max_idle_transformations` directly on the env. Then we don't need to pass that argument along.

borchero:
> Something we could do is to set the `max_idle_transformations` directly on the env. Then we don't need to pass that argument along.

Although this might work well in practice, it seems like a hack. If you (theoretically) were to run multiple training runs at the same time, you could run into issues.

src/gluonts/dataset/common.py
```diff
@@ -26,6 +26,7 @@
     Union,
     cast,
 )
+from typing_extensions import Protocol
```
jaheba, line 29: This seems to be introduced in `3.8`.

There are some other nice things in `3.8`, which I would like to use. Maybe we can have a discussion if we want to require `3.8` at some point.

borchero:
> Maybe it would make sense to split this into two PRs? One to fix the linting issues, the other one addressing #1513 what do you think?

I think that I only modified files (and mostly only classes) where I changed something. Since it also fixes a few bugs, I think it should be included here.

src/gluonts/dataset/common.py
borchero: Yes I thought the same. Also, the introduction of `Literal` is nice for multiple use cases.

jaheba:
> > Something we could do is to set the `max_idle_transformations` directly on the env. Then we don't need to pass that argument along.
> 
> Although this might work well in practice, it seems like a hack. If you (theoretically) were to run multiple training runs at the same time, you could run into issues.

Yes, it's a limitation, but what we could do is to bind it to the function and then it's fixed once it's used.

The settings module is not thread-safe at the moment, maybe that's something we need to consider in general.

borchero:
@jaheba what about some kind of context mechanism?

jaheba:
> @jaheba what about some kind of context mechanism?

What do you have in mind?

borchero:
Something like the following:

```
with env(max_idle_transforms=500):
    splitter = InstanceSplitter()
```

where the `InstanceSplitter` can access a global object and receives the correct value. I don't yet know how to implement it but I've seen it^^

jaheba:
There is env._push for that :)

borchero:
Ah, haven't seen that :D

jaheba:
Actually, there is `env._let` which you can use with a `with` statement. The `Settings` class is basically just a stack of key-value pairs, on which you can push new settings.

It is not thread-safe, because another thread can just push/pop things.

src/gluonts/dataset/common.py
```diff
@@ -216,10 +221,36 @@ def is_valid(cls, path: Path) -> bool:
         return not (path.name.startswith(".") or path.name == "_SUCCESS")
 
 
+class InMemoryDataset(Dataset):
```
jaheba, line 224: I'm a bit torn on this. Ideally, we should just get rid of the `ListDataset` and move the `ProcessDataEntry` code downstream to the models.

Then you can just use `dataset = [...]` as your dataset, and don't need to wrap it into `ListDataset`.

src/gluonts/env.py
```diff
@@ -21,8 +21,8 @@ class Environment(Settings):
     # returning an output. This parameter is intended to catch infinite loops
     # or inefficiencies, when transformations never or rarely return
     # something.
-    max_idle_transforms: int = os.environ.get(
-        "GLUONTS_MAX_IDLE_TRANSFORMS", "100"
+    max_idle_transforms: int = int(
```
jaheba, line 24: I think this should be handled in `settings.py` by setting `validate_all` in the `_Config`.

borchero: Mhh ok, a little confusing tbh but I'll change it^^

src/gluonts/dataset/common.py
borchero: I'll just remove it for now, should be part of another PR, I guess.

src/gluonts/env.py
jaheba: What exactly is confusing?

borchero: It's not really obvious how the types are converted (at least I didn't find out at a glance)

src/gluonts/dataset/common.py
```diff
@@ -31,6 +31,7 @@
 import pandas as pd
 import pydantic
 from pandas.tseries.offsets import Tick
+from typing_extensions import Protocol
```
jaheba, line 34: I wonder if we need a `requirement` for this.

This seems to be pulle in via either `black` or `mypy` which are not actual requirements of gluonts.

src/gluonts/dataset/common.py
```diff
@@ -109,6 +114,7 @@ def save(self, path_str: str, overwrite=True) -> None:
             Whether to delete previous version in this folder.
         """
         import shutil
```
jaheba, line 116: Why are the `imports` here at all?

src/gluonts/dataset/loader.py
```diff
@@ -203,22 +203,23 @@ def TrainDataLoader(
...
         )
-
-    batches = iter(batches)
+        iterator = iter(batches)
+    else:
+        iterator = iter(transformed_dataset)
 
     if num_batches_per_epoch is None:
-        return batches
+        return iterator
```
jaheba, line 220: I prefer that this keeps the name `batches`.

src/gluonts/dataset/loader.py
```diff
@@ -203,22 +203,23 @@ def TrainDataLoader(
...
         batches = MultiProcessLoader(
-            batches,
+            transformed_dataset,
             decode_fn=decode_fn,
             num_workers=num_workers,
             max_queue_size=num_prefetch,
         )
-
-    batches = iter(batches)
+        iterator = iter(batches)
```
jaheba, line 215: You don't like renaming?

src/gluonts/dataset/repository/datasets.py
```diff
@@ -152,11 +147,11 @@ def materialize_dataset(
     dataset_recipe = dataset_recipes[dataset_name]
 
     if not dataset_path.exists() or regenerate:
-        logging.info(f"downloading and processing {dataset_name}")
+        logging.info("downloading and processing %s", dataset_name)
```
jaheba, line 150: Any benefit of this over the existing usage?

src/gluonts/env.py
```diff
@@ -21,7 +22,7 @@ class Environment(Settings):
     # returning an output. This parameter is intended to catch infinite loops
     # or inefficiencies, when transformations never or rarely return
     # something.
-    max_idle_transforms: int = os.environ.get(
+    max_idle_transforms: int = os.environ.get(  # type: ignore
```
jaheba, line 25: Pylint complains?

src/gluonts/dataset/common.py
borchero: Mh good point, didn't notice...

src/gluonts/dataset/common.py
borchero: Moved it to the top

src/gluonts/dataset/loader.py
borchero: Most tools discourage shadowing ... for mypy you need to explicitly set that you are allowed to change the type (which GluonTS probably does)

src/gluonts/dataset/repository/datasets.py
borchero: Pylint complains^^ this is the idiomatic usage for the logging package

src/gluonts/env.py
borchero: Almost, it's mypy :D

jaheba: Of course!

src/gluonts/dataset/repository/datasets.py
jaheba: I still like the previous version more, since there is one less step I have to do mentally.

To hell with pylint!

src/gluonts/dataset/loader.py
jaheba: Right, because it makes more complicated for them. Personally, I think it's still fine.

src/gluonts/dataset/common.py
jaheba: Thanks

src/gluonts/env.py
jaheba: All fields are converted using `pydantic`. Actually, this should just work as it is.

src/gluonts/itertools.py
```diff
@@ -113,3 +122,6 @@ def __init__(self, iterable: Iterable, length: Optional[int]) -> None:
 
     def __iter__(self):
         return itertools.islice(self.iterable, self.length)
+
+    def __len__(self) -> int:
+        return len(self.iterable)
```
jaheba, line 127: Is this needed? And what happens if `self.iterable` doesn't support `len`?

src/gluonts/transform/split.py
```diff
@@ -269,16 +267,18 @@ def __init__(
...
         use_prediction_features: bool = False,
         prediction_length: Optional[int] = None,
     ) -> None:
+        super().__init__()
+
         self.instance_sampler = instance_sampler
         self.instance_length = instance_length
         self.output_NTC = output_NTC
-        self.dynamic_feature_fields = time_series_fields
+        self.dynamic_feature_fields = time_series_fields or []
```
jaheba, line 281: If this is using `validated`, this might not be required, since `pydantic` always returns a `copy`.

src/gluonts/itertools.py
borchero: It's required for setting the max number of idle transformations. If it doesn't support `len`, it'll throw (probably an `AttributeError`). The base datasets always should implement it, right?

src/gluonts/transform/split.py
borchero: Mhh ok, pydantic is super weird :D

src/gluonts/itertools.py
jaheba: Yes, but this class doesn't operate on `Dataset` but on `Iterable`.

borchero: Sure, but the current behavior won't change if `self.iterable` does not provide a `__len__`. If we don't put this here, the entire PR is useless since datasets are most often being transformed :D I just noticed though that we need to catch whether `__len__` throws when setting `max_idle_transforms` ... I will fix that

src/gluonts/model/wavenet/_estimator.py
```diff
@@ -308,7 +311,7 @@ def _create_instance_splitter(self, mode: str):
             instance_sampler=instance_sampler,
             past_length=self.context_length,
             future_length=self.prediction_length
-            if mode is "test"
+            if mode == "test"
```
jaheba, line 314: Let's not do this in this PR.

jaheba: Maybe, let's just still do it.

jaheba (approved):

src/gluonts/model/deepar/_estimator.py
```diff
@@ -401,7 +403,8 @@ def create_training_data_loader(
         **kwargs,
     ) -> DataLoader:
         input_names = get_hybrid_forward_input_names(DeepARTrainingNetwork)
-        instance_splitter = self._create_instance_splitter("training")
+        with env._let(max_idle_transforms=maybe_len(data) or 0):
```
lostella, line 406: I’m wondering whether a less invasive way of doing this would be to execute training within this context, instead of the splitter construction: the former could be done in the superclass `GluonEstimator` if I’m not wrong (or by the estimator user when  calling `train`)

PR #1557, opened 2021-06-11
Fix computation of OWA

### LINKED ISSUE #1556: M4 competition metric "OWA" is computed incorrectly
## Description

It seems like GluonTS computes the M4 competition metric "OWA" incorrectly. GluonTS computes the OWA per time-series and then averages it. However, the [competition guide](https://mk0mcompetitiont8ake.kinstacdn.com/wp-content/uploads/2018/09/M4-Competitors-Guide.pdf) states that the OWA is only computed on the final sMAPE and MASE values:

> Note that MASE and sMAPE are first estimated per series by averaging the error computed per forecasting horizon and then averaged again across the 3003 time series to compute their value for the whole dataset. On the other hand, OWA is computed only once at the end for the whole sample.

borchero:
Fixes #1556 

*Description of changes:*

- For each time series, only compute the sMAPE and MASE of Naive2
- Compute the OWA only for the final stats

**TBD:** new test

### DIFF
--- a/src/gluonts/evaluation/_base.py
+++ b/src/gluonts/evaluation/_base.py
@@ -14,7 +14,6 @@
 import logging
 import multiprocessing
 import sys
-import warnings
 from functools import partial
 from itertools import chain, tee
 from typing import (
@@ -470,7 +469,7 @@ class Evaluator:
         # Compute OWA if required
         if self.calculate_owa:
             if totals["sMAPE_naive2"] == 0 or totals["MASE_naive2"] == 0:
-                warnings.warn(
+                logging.warning(
                     "OWA cannot be computed as Naive2 yields an sMAPE or MASE of 0."
                 )
                 totals["OWA"] = np.nan
--- a/src/gluonts/evaluation/_base.py
+++ b/src/gluonts/evaluation/_base.py
@@ -362,7 +362,8 @@ class Evaluator:
                             target_fcst,
                         )
                     }
-                except:
+                except Exception:
+                    logging.warning(f"Error occured when evaluating {k}.")
                     val = {k: np.nan}
 
                 metrics.update(val)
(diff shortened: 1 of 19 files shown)


src/gluonts/evaluation/_base.py
```diff
@@ -31,6 +31,10 @@
 import numpy as np
 import pandas as pd
 
+from gluonts.gluonts_tqdm import tqdm
+from gluonts.model.forecast import Forecast, Quantile
+from gluonts.model.naive_2 import naive_2
```
jaheba, line 36: This requires `statsmodules` which is not installed by default

borchero: Changed where it's imported

vafl (approved):
Thanks @borchero !

It would be nice to add a small test. Maybe based on the example in the m4 [guide lines](https://mk0mcompetitiont8ake.kinstacdn.com/wp-content/uploads/2018/09/M4-Competitors-Guide.pdf), but not urgent. I can also put it in the backlog.

src/gluonts/evaluation/_base.py
```diff
@@ -461,6 +466,25 @@ def get_aggregate_metrics(
                 for q in self.quantiles
             ]
         )
+
+        # Compute OWA if required
+        if self.calculate_owa:
+            if totals["sMAPE_naive2"] == 0 or totals["MASE_naive2"] == 0:
+                warnings.warn(
+                    "OWA cannot be computed as Naive2 yields an sMAPE or MASE of 0."
+                )
```
lostella, line 475: Elsewhere we use the `logging` module for warnings, would it make sense to also use it here?

lostella (approved):
Looks good! Thanks for spotting this!

kashif:
awesome find @borchero 😎

PR #1562, opened 2021-06-14
Allow validation data for TabularEstimator.

yx1215:
*Issue #, if available:*

*Description of changes:*
Allow validation data for TabularEstimator

User will have three options regarding to the validation data.

- Users can only one dataset and an int provide last_k_for_val so that the estimator will split the last k data points of each time series out to be the validation dataset(Auto stacking will be turned off in this case).
- Users can provide both train and validation data by themselves(Auto stacking will be turned off in this case).
- Users can provide only one dataset and let TabularPredictor in AutoGluon to the splitting(This might break the time order).


src/gluonts/nursery/autogluon_tabular/estimator.py
```diff
@@ -127,13 +130,49 @@ def train(
             )
             for entry in training_data
         ]
-        df = pd.concat(dfs)
+        if validation_data is not None or last_k_for_val is not None:
+            self.kwargs["auto_stack"] = False
+            logging.warning(
```
lostella, line 135: It is recommended to create a specific logger object for this module, instead of using the default one from the `logging` module. Just add this on top after the imports

```
```

and use `logger.log(…)` throughout the code in this module

lostella (changes requested):
I see you updated the example, could you maybe also update the tests on this estimator class to cover all three settings?

src/gluonts/nursery/autogluon_tabular/estimator.py
```diff
@@ -111,12 +113,15 @@ def __init__(
...
+            "auto_stack": True,
         }
         self.kwargs = {**default_kwargs, **kwargs}
 
     def train(
         self,
         training_data: Dataset,
+        validation_data: Optional[Dataset] = None,
+        last_k_for_val: Optional[int] = None,
         eval_metric="mean_absolute_error",
```
lostella, line 125: I think these two settings are better suited as constructor arguments, so that the signature of `train` is minimal: `train` it only gets data, and everything else is a hyperparameter and is given to the constructor. What do you think? Could you also update the class docstring to include these when you move them?

src/gluonts/nursery/autogluon_tabular/estimator.py
```diff
@@ -127,13 +132,49 @@ def train(
...
+            validation_dfs = [
+                get_features_dataframe(
+                    series=self.scaling(to_pandas(entry))[0],
+                    time_features=self.time_features,
+                    lag_indices=self.lag_indices,
+                )
+                for entry in validation_data
+            ]
+            train_df = pd.concat(dfs)
+            val_df = pd.concat(validation_dfs)
```
lostella, line 163: This way, `validation_data` is ignored if `last_k_for_val` is set. Maybe the other way around would be better: if `validation_data` is provided, then use it, otherwise check whether `last_k_for_val` is set: I think this is less of a "surprise" if one explicitly gives validation data.

lostella (approved):
Looks good! Thanks a lot for this!

PR #1563, opened 2021-06-15
Merges for v0.7.6

Schmedu:
Merge commits:
- Fix serde issue with some distribution output types, add test (#1543)


lostella (approved):
🚢

PR #1564, opened 2021-06-15
R fixes for methods that produce point forecasts or prediction intervals directly

rshyamsundar:
*Issue #, if available:  [255](https://github.com/awslabs/gluon-ts/issues/255)*

*Description of changes:*
Several fixes:

* For `thetaf`, instead of repeating mean forecasts, the fix now returns quantile forecasts correctly by first obtaining prediction intervals
* `simulate` is not defined on tbats model; returns quantile forecasts by first obtaining prediction intervals
* Fixes the non-functional `Croston` and `MLP` methods; need to pass prediction length to obtain forecasts directly.
* On the python-side, allows obtaining quantile forecasts from R instead of samples-based forecast.

This worked on the following versions:
* R version 3.6.3 (2020-02-29) and `forecast` version 8.15
* R version 4.1.0 (2021-05-18) and `forecast` version 8.14

### DIFF
--- a/src/gluonts/model/r_forecast/R/forecast_methods.R
+++ b/src/gluonts/model/r_forecast/R/forecast_methods.R
@@ -69,7 +69,7 @@ croston <- function(ts, params) {
 tbats <- function(ts, params) {
     model <- forecast::tbats(ts)
 
-    # R doesn't allow `simulate` on the tbats model. We generate prediction intervals directly.
+    # R doesn't allow `simulate` on tbats model. We obtain prediction intervals directly.
     forecasts <- forecast::forecast(model, h=params$prediction_length, level=unlist(params$intervals))
     handleQuantileForecast(forecasts, params)
 }
@@ -83,6 +83,7 @@ mlp <- function(ts, params) {
 }
 
 thetaf <- function(ts, params) {
+    # For thetaf, we obtain prediction intervals directly.
     forecasts <- forecast::thetaf(y=ts, h=params$prediction_length, level=unlist(params$intervals))
     handleQuantileForecast(forecasts, params)
 }
--- a/src/gluonts/model/r_forecast/R/forecast_methods.R
+++ b/src/gluonts/model/r_forecast/R/forecast_methods.R
@@ -39,10 +39,10 @@ handlePointForecast <- function(forecasts, params) {
     outputs = list()
     output_types = params$output_types
     if ("samples" %in% output_types) {
-        print("This forecasting method only produces point forecasts!")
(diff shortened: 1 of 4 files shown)


src/gluonts/model/r_forecast/R/forecast_methods.R
```diff
@@ -16,6 +16,40 @@ handleForecast <- function(model, params) {
...
+        outputs$mean <- forecasts$mean
+    }
+    outputs
+}
+
+handlePointForecast <- function(forecasts, params) {
+    outputs = list()
+    output_types = params$output_types
+    if ("samples" %in% output_types) {
+        print("This forecasting method only produces point forecasts!")
```
lostella, line 42: Maybe in this error message we can suggest to use “mean” as (only) output type

src/gluonts/model/r_forecast/_predictor.py
```diff
@@ -111,12 +111,30 @@ def __init__(
...
+        point_forecast_methods = ["croston", "mlp"]
+        quantile_forecast_methods = ["tbats", "thetaf"]
+
+        if self.method_name in point_forecast_methods:
+            self.point_forecasts = True
+        else:
+            self.point_forecasts = False
+
+        if self.method_name in quantile_forecast_methods:
+            self.quantlie_forecasts = True
```
lostella, line 130: Typo: `quantile` (Also other occurrences)

src/gluonts/model/r_forecast/_predictor.py
```diff
@@ -178,38 +229,70 @@ def predict(
...
         num_samples: int = 100,
+        intervals: Optional[List] = None,
         save_info: bool = False,
         **kwargs,
-    ) -> Iterator[SampleForecast]:
-        for entry in dataset:
-            if isinstance(entry, dict):
-                data = entry
-            else:
-                data = entry.data
```
lostella, line 188: I’m still unsure how this was supposed to work…

lostella (changes requested):
See comments inline. Maybe since it is hard to get tests running on GitHub (one would need to install R and so on I guess) could you note in the PR message what version of R and Forecast package you tested this on?

We could also include a very simple test actually (just a sanity check on constant data or whatever) to make sure that the various options run fine, and mark it as `@pytest.mark.skip` so that one can run it locally if they want to check

src/gluonts/model/r_forecast/_predictor.py
rshyamsundar: :)

Yea, it is always dict and `data` = `entry`.

lostella (approved):

PR #1567, opened 2021-06-16
Improve tests for R methods

rshyamsundar:
*Issue #, if available:*

*Description of changes:*
Use the `constant` dataset for tests. Also assert that the forecast error is small.


lostella (approved):
Makes sense! Thanks for the scrutiny in setting up these tests :-)

PR #1592, opened 2021-06-28
shallow import for `gluonts.mx` module

Schmedu:
*Issue #, if available:*
https://github.com/awslabs/gluon-ts/issues/1196
*Description of changes:*

### DIFF
--- a/docs/tutorials/forecasting/extended_tutorial.md
+++ b/docs/tutorials/forecasting/extended_tutorial.md
@@ -574,7 +574,7 @@ Finally, each estimator is configured by a `Trainer`, which defines how the mode
 
 ```python
 from gluonts.model.simple_feedforward import SimpleFeedForwardEstimator
-from gluonts.mx.trainer import Trainer
+from gluonts.mx import Trainer
 ```
 
 
@@ -845,11 +845,11 @@ from mxnet.gluon import HybridBlock
 from gluonts.core.component import validated
 from gluonts.dataset.loader import TrainDataLoader
 from gluonts.model.predictor import Predictor
-from gluonts.mx.batchify import as_in_context, batchify
-from gluonts.mx.model.estimator import GluonEstimator
-from gluonts.mx.model.predictor import RepresentableBlockPredictor
-from gluonts.mx.trainer import Trainer
-from gluonts.mx.util import copy_parameters, get_hybrid_forward_input_names
+from gluonts.mx import as_in_context, batchify
+from gluonts.mx import GluonEstimator
+from gluonts.mx import RepresentableBlockPredictor
+from gluonts.mx import Trainer
+from gluonts.mx import copy_parameters, get_hybrid_forward_input_names
 from gluonts.transform import (
     ExpectedNumInstanceSampler,
     Transformation,
@@ -1016,7 +1016,7 @@ Note that in all the tensors that we handle there is an initial dimension that r
 
 
 ```python
-from gluonts.mx.distribution import DistributionOutput, GaussianOutput
+from gluonts.mx import DistributionOutput, GaussianOutput
 ```
 
 
@@ -1236,7 +1236,7 @@ An important issue that a practitioner needs to deal with often is the different
 
 
 ```python
-from gluonts.mx.block.scaler import MeanScaler, NOPScaler
+from gluonts.mx import MeanScaler, NOPScaler
 ```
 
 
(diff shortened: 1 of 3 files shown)


lostella:
@Schmedu are these the only import paths that can be reasonably shortened? One way to check this, is to go through the tutorial notebooks (which are generated from the markdown files [here](https://github.com/awslabs/gluon-ts/tree/master/docs/tutorials) and see what imports are being used. We should also shorten the imports in the tutorials themselves.

lostella:
For example in the extended forecasting tutorial there is 

```python
from gluonts.mx.batchify import as_in_context, batchify
from gluonts.mx.model.estimator import GluonEstimator
from gluonts.mx.model.predictor import RepresentableBlockPredictor
from gluonts.mx.util import copy_parameters, get_hybrid_forward_input_names
```

so maybe more stuff can be exposed

Schmedu:
> @Schmedu are these the only import paths that can be reasonably shortened? One way to check this, is to go through the tutorial notebooks (which are generated from the markdown files [here](https://github.com/awslabs/gluon-ts/tree/master/docs/tutorials) and see what imports are being used. 

I did not update my last analyses about the import paths, but will do so now.

> We should also shorten the imports in the tutorials themselves.

That's also a very good point, thanks!

docs/tutorials/forecasting/extended_tutorial.md
```diff
@@ -845,11 +845,11 @@ from mxnet.gluon import HybridBlock
...
-from gluonts.mx.batchify import as_in_context, batchify
-from gluonts.mx.model.estimator import GluonEstimator
-from gluonts.mx.model.predictor import RepresentableBlockPredictor
-from gluonts.mx.trainer import Trainer
-from gluonts.mx.util import copy_parameters, get_hybrid_forward_input_names
+from gluonts.mx import as_in_context, batchify
+from gluonts.mx import GluonEstimator
+from gluonts.mx import RepresentableBlockPredictor
+from gluonts.mx import Trainer
+from gluonts.mx import copy_parameters, get_hybrid_forward_input_names
```
lostella, line 852: I guess these could be all in one? As in

```
from gluonts.mx import (
   ...
   ...
)
```

src/gluonts/mx/__init__.py
```diff
@@ -24,3 +24,27 @@
...
 
 from . import prelude as _  # noqa
+from .trainer import Trainer  # noqa
+from .util import get_hybrid_forward_input_names, copy_parameters  # noqa
+from .kernels import RBFKernel  # noqa
+from .batchify import as_in_context, batchify  # noqa
+from .model.estimator import GluonEstimator  # noqa
+from .model.predictor import RepresentableBlockPredictor  # noqa
+from .distribution import DistributionOutput, GaussianOutput  # noqa
+from .block.scaler import MeanScaler, NOPScaler  # noqa
```
lostella, line 34: We could sort the imports here (`isort` to the rescue!)

src/gluonts/mx/__init__.py
```diff
@@ -24,3 +24,27 @@
...
+    "GaussianOutput",
+    "RepresentableBlockPredictor",
+    "GluonEstimator",
+    "as_in_context",
+    "batchify",
+    "copy_parameters",
+    "RBFKernel",
+    "get_hybrid_forward_input_names",
+    "Trainer",
+]
```
lostella, line 50: I think this is needed only if we want to *exclude* stuff from `__all__`. Is that the case? For example, this way `Tensor` (defined above) is not visible in `gluonts.mx`

lostella (changes requested):

src/gluonts/mx/__init__.py
Schmedu: Removed ✔️

src/gluonts/mx/__init__.py
Schmedu: Sorted ✔️

docs/tutorials/forecasting/extended_tutorial.md
Schmedu: Concatenated ✔️

lostella (approved):
Looks good! Thanks!

src/gluonts/mx/__init__.py
```diff
@@ -24,3 +24,11 @@
...
 
 from . import prelude as _  # noqa
+from .batchify import as_in_context, batchify  # noqa
+from .block.scaler import MeanScaler, NOPScaler  # noqa
+from .distribution import DistributionOutput, GaussianOutput  # noqa
+from .kernels import RBFKernel  # noqa
+from .model.estimator import GluonEstimator  # noqa
+from .model.predictor import RepresentableBlockPredictor  # noqa
+from .trainer import Trainer  # noqa
+from .util import copy_parameters, get_hybrid_forward_input_names  # noqa
```
jaheba, line 34: Something we could do is to define an `__all__`, which would allow us to remove the noqa-comments.

PR #1604, opened 2021-07-06
loosen RTOL in `test/distribution/test_flows.py` to make `test_flow_invertibility` pass

Schmedu:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/test/distribution/test_flows.py
+++ b/test/distribution/test_flows.py
@@ -73,7 +73,7 @@ def test_flow_invertibility(
     ), f"y and y_hat did not match: y = {inp}, y_hat = {y_hat}"
     assert allclose(
         inp, x_hat, rtol=1.0e-4
-    ), f"y and y_hat did not match: x = {inp}, x_hat = {x_hat}"
+    ), f"x and x_hat did not match: x = {inp}, x_hat = {x_hat}"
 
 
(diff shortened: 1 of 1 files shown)


test/distribution/test_flows.py
```diff
@@ -69,10 +69,10 @@ def test_flow_invertibility(
     y_hat = bijection.f(bijection.f_inv(inp))
     x_hat = bijection.f_inv(bijection.f(inp))
     assert allclose(
-        inp, y_hat
+        inp, y_hat, rtol=1.0e-4
     ), f"y and y_hat did not match: y = {inp}, y_hat = {y_hat}"
     assert allclose(
-        inp, x_hat
+        inp, x_hat, rtol=1.0e-4
     ), f"y and y_hat did not match: x = {inp}, x_hat = {x_hat}"
```
lostella, line 76: Just realized there’s a typo here in the error string, should be x and x_hat

lostella (changes requested):

test/distribution/test_flows.py
Schmedu: Good catch! Is fixed

lostella (approved):
🚀

PR #1611, opened 2021-07-12
Add quantitle regression mode to AutoGluon-based TabularEstimator

jindingars:
*Issue #, if available:*
Autogluon has quantile method 

*Description of changes:*
1. add quantile parameters to TabularEstimator to allow user to decide use quantile prediction model or not and assign the quantile value to training process
2. update TabularPredictor to fit quantile prediction output
3. add quantile_example.py because the forecast visualization is different from non-quantile prediction.
4. changed model exclude because RF is required for quantile prediction


src/gluonts/nursery/autogluon_tabular/estimator.py
```diff
@@ -112,7 +125,7 @@ def __init__(
 
         default_kwargs = {
             "time_limit": 60,
-            "excluded_model_types": ["KNN", "XT", "RF"],
+            # "excluded_model_types": ["KNN", "XT", "RF"],
```
lostella, line 128: Is this intended?

src/gluonts/nursery/autogluon_tabular/estimator.py
```diff
@@ -80,6 +87,8 @@ def __init__(
         batch_size: Optional[int] = 32,
         disable_auto_regression: bool = False,
         last_k_for_val: Optional[int] = None,
+        quantile=False,
+        quantiles_topredict=[0.1, 0.5, 0.9],
```
lostella, line 91: Would it make sense to control the behaviour with a single argument here? For example, `quantiles_to_predict=None` would give the previous behaviour, while setting it to a list of quantiles would enable the new behaviour

src/gluonts/nursery/autogluon_tabular/predictor.py
```diff
@@ -130,6 +130,10 @@ def _to_forecast(
         start_timestamp: pd.Timestamp,
         item_id=None,
     ) -> Iterator[SampleForecast]:
+        if ag_output.size > self.prediction_length:
+            samples = ag_output.transpose()
+        else:
+            samples = ag_output.reshape((1, self.prediction_length))
         samples = ag_output.reshape((1, self.prediction_length))
         sample = SampleForecast(
```
lostella, line 138: This is using `SampleForecast` to encode the prediction, which is OK for point prediction. But if the model explicitly outputs quantiles, it makes sense to use `QuantileForecast` instead. To do that, the predictor object will have to be configured with the quantile levels.

src/gluonts/nursery/autogluon_tabular/predictor.py
```diff
@@ -130,6 +130,10 @@ def _to_forecast(
         start_timestamp: pd.Timestamp,
         item_id=None,
     ) -> Iterator[SampleForecast]:
+        if ag_output.size > self.prediction_length:
+            samples = ag_output.transpose()
+        else:
+            samples = ag_output.reshape((1, self.prediction_length))
```
lostella, line 136: This line is happening in any case just below here, is this correct?

lostella (commented):
Thanks @jindingars! I have left some comments, mainly on how the model is configured and the type of forecasts it produces

src/gluonts/nursery/autogluon_tabular/estimator.py
jindingars: yeah this is intended. If we exclude "RF", then no quantile model could be used. I'm curious why we need to exclude some models at initial. That also why I didn't remove this line of code

src/gluonts/nursery/autogluon_tabular/estimator.py
jindingars: make sense to me! Thanks. I'll update in the next version.

src/gluonts/nursery/autogluon_tabular/predictor.py
jindingars: yes. Let me explore `QuantileForecast`. Thanks for your suggestion!

src/gluonts/nursery/autogluon_tabular/predictor.py
jindingars: Sorry I'm not very clear about this comment. What line are you talking about? 
sample output:
`samples=numpy.array([[135.67691040039062, 171.1056671142578, 183.63880920410156, 197.9020233154297, 162.87466430664062, 147.75, 138.01918029785156, 136.3420867919922, 139.5883331298828, 130.777099609375, 100.32427978515625, 106.0, 108.5605239868164, 93.05574798583984, 103.6478042602539, 119.0, 147.0, 118.28972625732422, 117.5522689819336, 121.49999237060547, 113.92756652832031, 131.35231018066406, 144.86109924316406, 154.62594604492188], [175.91529846191406, 208.87501525878906, 217.50601196289062, 224.64553833007812, 190.80322265625, 175.60281372070312, 165.0, 155.90975952148438, 162.31201171875, 159.0, 126.79875946044922, 126.79875946044922, 130.4644317626953, 117.0, 130.80892944335938, 150.4036407470703, 181.0, 143.8137969970703, 137.3653106689453, 149.0178680419922, 141.72166442871094, 170.82974243164062, 190.0, 210.4620819091797], [227.33328247070312, 243.37208557128906, 255.0, 257.380126953125, 215.36550903320312, 204.04937744140625, 198.0, 181.0, 178.6450653076172, 185.9631805419922, 150.58973693847656, 156.0, 160.5, 151.05474853515625, 160.5927734375, 174.84234619140625, 221.3333282470703, 169.0, 164.63075256347656, 184.5719451904297, 185.25, 207.8796844482422, 231.82571411132812, 249.3555908203125]]`

lostella: What I mean is: line 137, immediately following 136, is doing the same reshaping, so this seems odd

jindingars: I see. Thanks! I'll change it to `QuantileForecast`

src/gluonts/nursery/autogluon_tabular/estimator.py
yx1215: I believe these models are removed because usually they do not give a better results than DL models while training them is going to consume a lot of space. Do you mean that if we do not enable RF then we will not be able to use the quantiles?

jindingars: yes. If we don't enable RF then there's no quantile model available.

src/gluonts/nursery/autogluon_tabular/predictor.py
```diff
@@ -129,14 +131,24 @@ def _to_forecast(
...
-        sample = SampleForecast(
-            freq=self.freq,
-            start_date=pd.Timestamp(start_timestamp, freq=self.freq),
-            item_id=item_id,
-            samples=samples,
-        )
+    ) -> Iterator[Forecast]:
+        if ag_output.size > self.prediction_length:
+            forecasts = ag_output.transpose()
+            sample = QuantileForecast(
```
lostella, line 137:
```suggestion
            return QuantileForecast(
```

src/gluonts/nursery/autogluon_tabular/predictor.py
```diff
@@ -129,14 +131,24 @@ def _to_forecast(
...
+                freq=self.freq,
+                start_date=pd.Timestamp(start_timestamp, freq=self.freq),
+                item_id=item_id,
+                forecast_arrays=forecasts,
+                forecast_keys=self.quantiles_to_predict,
+            )
+        else:
+            samples = ag_output.reshape((1, self.prediction_length))
```
lostella, line 145:
```suggestion
            return ag_output.reshape((1, self.prediction_length))
```

src/gluonts/nursery/autogluon_tabular/predictor.py
```diff
@@ -129,14 +131,24 @@ def _to_forecast(
...
+            sample = SampleForecast(
+                freq=self.freq,
+                start_date=pd.Timestamp(start_timestamp, freq=self.freq),
+                item_id=item_id,
+                samples=samples,
+            )
         return sample
```
lostella, line 152:
```suggestion
```

lostella (changes requested):
There is a test for this class here: https://github.com/awslabs/gluon-ts/blob/663268032cfde371a88dd5fd177911f2a3949852/test/nursery/test_autogluon_tabular.py#L216

Maybe you could expand it to cover the behavior introduced in this PR?

src/gluonts/nursery/autogluon_tabular/predictor.py
```diff
@@ -365,6 +377,7 @@ def serialize(self, path: Path) -> None:
     def deserialize(
         cls,
         path: Path,
+        quantiles_to_predict=None,
```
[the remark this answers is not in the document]
jindingars: agree. will address it

src/gluonts/nursery/autogluon_tabular/predictor.py
```diff
@@ -129,14 +131,24 @@ def _to_forecast(
...
-    ) -> Iterator[SampleForecast]:
-        samples = ag_output.reshape((1, self.prediction_length))
-        sample = SampleForecast(
-            freq=self.freq,
-            start_date=pd.Timestamp(start_timestamp, freq=self.freq),
-            item_id=item_id,
-            samples=samples,
-        )
+    ) -> Iterator[Forecast]:
+        if ag_output.size > self.prediction_length:
```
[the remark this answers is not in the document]
jindingars: yup but no matter SampleForecast or QuantileForecast, they are all extended from Forecast. That's why I use it here.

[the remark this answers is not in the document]
lostella: Sure, what I mean is: the condition in this `if` is not very clear, shouldn’t this class behavior depend on something more explicit? Like the value of `quantiles_to_predict`

jindingars:
> There is a test for this class here:
> 
> https://github.com/awslabs/gluon-ts/blob/663268032cfde371a88dd5fd177911f2a3949852/test/nursery/test_autogluon_tabular.py#L216
> 
> Maybe you could expand it to cover the behavior introduced in this PR?

I tried to quantile test case with existing test data. autogluon quantile method can't train a model with the existing test data. The quantile method failed with existing data but it succeed with electricity dataset. I'll add the test case if I find out a suitable sample data for quantile method later.

src/gluonts/nursery/autogluon_tabular/predictor.py
```diff
@@ -129,23 +131,33 @@ def _to_forecast(
...
-    ) -> Iterator[SampleForecast]:
-        samples = ag_output.reshape((1, self.prediction_length))
-        sample = SampleForecast(
-            freq=self.freq,
-            start_date=pd.Timestamp(start_timestamp, freq=self.freq),
-            item_id=item_id,
-            samples=samples,
-        )
-        return sample
+    ) -> Iterator[Forecast]:
```
lostella, line 134:
```suggestion
    ) -> Forecast:
```

src/gluonts/nursery/autogluon_tabular/predictor.py
```diff
@@ -129,23 +131,33 @@ def _to_forecast(
...
+            )
+        else:
+            samples = ag_output.reshape((1, self.prediction_length))
+            sample = SampleForecast(
+                freq=self.freq,
+                start_date=pd.Timestamp(start_timestamp, freq=self.freq),
+                item_id=item_id,
+                samples=samples,
+            )
+            return sample
```
lostella, line 152:
```suggestion
            return SampleForecast(
                freq=self.freq,
                start_date=pd.Timestamp(start_timestamp, freq=self.freq),
                item_id=item_id,
                samples=samples,
            )
```

src/gluonts/nursery/autogluon_tabular/predictor.py
```diff
@@ -104,6 +104,7 @@ def __init__(
         lag_indices: List[int],
         scaling: Callable[[pd.Series], Tuple[pd.Series, float]],
         batch_size: Optional[int] = 32,
+        quantiles_to_predict: List[str] = None,
```
lostella, line 107:
```suggestion
        quantiles_to_predict: Optional[list] = None,
```

src/gluonts/nursery/autogluon_tabular/estimator.py
```diff
@@ -80,6 +84,7 @@ def __init__(
         batch_size: Optional[int] = 32,
         disable_auto_regression: bool = False,
         last_k_for_val: Optional[int] = None,
+        quantiles_to_predict=None,
```
lostella, line 87:
```suggestion
        quantiles_to_predict: Optional[list]=None,
```

src/gluonts/nursery/autogluon_tabular/estimator.py
```diff
@@ -80,6 +84,7 @@ def __init__(
         batch_size: Optional[int] = 32,
         disable_auto_regression: bool = False,
         last_k_for_val: Optional[int] = None,
+        quantiles_to_predict: Optional[list]=None,
```
lostella, line 87:
```suggestion
        quantiles_to_predict: Optional[list] = None,
```

lostella:
Hey @jindingars I pushed some final changes and made sure that the example runs. I think this can be merged now, and then we can figure out how to add proper tests: the existing one would need to run for much longer in order to work for quantile regression as well, so I didn't add it yet. But the example runs fine! Thanks for this!

lostella (approved):

PR #1612, opened 2021-07-13
Move callbacks example to docs

lostella:
This way the example is run as part of the documentation build.


src/gluonts/mx/__init__.py
```diff
@@ -13,6 +13,8 @@
 
 # !!! DO NOT MODIFY !!! (pkgutil-style namespace package)
 
+# flake8: noqa
```
lostella, line 16: @jaheba hope you don't mind this? I would avoid over-engineering these init modules if possible

jaheba: So what we want is to re-expose the classes and functions without sphinx noticing?

lostella: I'm not sure what is needed for sphinx here, I just put this annotation so that we don't need to have `# noqa` at every line (which btw complicates things when multi-line imports are there, which are sometimes enforced by `black`). I just want to avoid all this style checking to not be in the way :-)

docs/tutorials/mxnet_models/trainer_callbacks.md
```diff
@@ -0,0 +1,239 @@
...
+```
+
+```python
+print(history.loss_history) # The training loss history of all 3+3 epochs we trained the model for
+```
+
+## Default callbacks
+
+In addition to the callbacks you specify, the `Trainer` class uses the two default callbacks `ModelAveraging` and `LearningRateReduction`.
+You can turn them off by setting add_default_callbacks=False when initializing the Trainer.
```
borchero, line 65: Missing code block for `add_default_callbacks=False`

docs/tutorials/mxnet_models/trainer_callbacks.md
```diff
@@ -0,0 +1,239 @@
...
+```
+
+## Custom callbacks
+
+To implement your own callback you can write a class which inherits from `gluonts.mx.trainer.Callback`, and overwrite one or more of the hooks.
+Have a look at the abstract `Callback` class, the hooks take different arguments which you can use. 
+Hook methods with boolean return value stop the training if False is returned.
+
+Here is an example implementation of a Metric value based early stopping custom callback implementation.
+It only implements the hook method "on_epoch_end()" which gets called after all batches of one epoch have been processed.
```
borchero, line 80: Code block instead of quotes?

docs/tutorials/mxnet_models/trainer_callbacks.md
```diff
@@ -0,0 +1,239 @@
...
+trainer=Trainer(epochs=20, add_default_callbacks=False) # use no callback at all
+```
+
+## Custom callbacks
+
+To implement your own callback you can write a class which inherits from `gluonts.mx.trainer.Callback`, and overwrite one or more of the hooks.
+Have a look at the abstract `Callback` class, the hooks take different arguments which you can use. 
+Hook methods with boolean return value stop the training if False is returned.
+
+Here is an example implementation of a Metric value based early stopping custom callback implementation.
```
borchero, line 79: "Here is an example for a custom callback implementation which terminates training early based on the value of some Metric (such as the RMSE)"?

src/gluonts/mx/__init__.py
jaheba: The other way to suppress the warning would be to add these things to `__all__`. However, that would have the effect that sphinx autogen would include these entities in the modules documentation -- which is probably not what we want.

So I think this is the way to go.

rshyamsundar (approved):

PR #1617, opened 2021-07-14
Adding NCAD (Neural Contextual Anomaly Detection for Time Series) to nursery

Francois-Aubet:
*Description of changes:*
Adding the Neural Contextual Anomaly Detection for Time Series to nursery.

The whole code source is added here.

**Please tag this pr with at least one of these labels to make our release process faster:** other change


mbohlkeschneider (changes requested):
Thank you! 

Some high-level questions: 

1. This adds a lot of files (many of which are json). Is this really needed? 
2. Is there a test implemented somewhere? I did not find one and at least a smoke test should be the bare minimum.

PR #1625, opened 2021-07-16
Remove setting __module on __all__ items.

jaheba:
*Issue #, if available:*

*Description of changes:*


lostella:
At this point I don’t remember why these were needed: are these not needed anymore?

jaheba:
I can only speculate, but I think the idea is that things like `trainer._base.Trainer` are documented in sphinx as `trainer.Trainer` by re-setting the `__module__` attribute on these entities.

However, this is normally done by listing the entities in `__all__` of the re-exporting module, and sphinx will do the abbreviation itself, so there should be no need to write to `__module__` just for sphinx. (Whether we should do it for serde is different question).

However, and more importantly, the to be removed code doesn't work, since it tries to set `__module__` on the strings which represent the names and not the actual objects. In that sense, it didn't do anything.

lostella:
> However, and more importantly, the to be removed code doesn't work, since it tries to set __module__ on the strings which represent the names and not the actual objects. In that sense, it didn't do anything.

Awesome! So maybe we should just make sure Sphinx is generating the API docs as we want it, before the next release: I’m not sure it’s doing the right thing in case it relies on the `__module__` property

lostella (approved):

PR #1631, opened 2021-07-19
Add `TimeLimitCallback` to `mx/trainer` callbacks.

yx1215:
*Issue #, if available:*

*Description of changes:*
Add TimelimitCallback so that user can set a time limit to the training process.


src/gluonts/mx/trainer/callback.py
```diff
@@ -338,3 +339,31 @@ def on_network_initializing_end(
         self, training_network: nn.HybridBlock
     ) -> None:
         copy_parameters(self.predictor.prediction_net, training_network)
+
+
+class TimeLimitCallback(Callback):
```
jaheba, line 344: Can we have a short doc-string describing the class and it's parameters?

src/gluonts/mx/trainer/callback.py
```diff
@@ -338,3 +339,31 @@ def on_network_initializing_end(
...
+    @validated()
+    def __init__(self, time_limit=None):
```
jaheba, line 346: `validated` only makes sense if you have type annotations really.

src/gluonts/mx/trainer/callback.py
```diff
@@ -338,3 +339,31 @@ def on_network_initializing_end(
...
+        self.start_time = None
+        self.time_limit = time_limit
```
jaheba, line 348: Can we reverse these? Parameters which are passed should be handled first.

src/gluonts/mx/trainer/callback.py
```diff
@@ -338,3 +339,31 @@ def on_network_initializing_end(
...
+        epoch_loss: float,
+        training_network: nn.HybridBlock,
+        trainer: gluon.Trainer,
+        best_epoch_info: Dict[str, Any],
+        ctx: mx.Context,
+    ) -> bool:
+        if self.time_limit is not None:
+            cur_time = time.time()
+            if cur_time - self.start_time > self.time_limit:
+                logging.warning(
```
jaheba, line 365: We don't want to use `logging` directly`, but use a logger instance instead.

src/gluonts/mx/trainer/callback.py
jaheba: Let's give it a more descriptive name.

yx1215: How about `TrainingTimeLimitCallback`?

src/gluonts/mx/trainer/callback.py
```diff
@@ -338,3 +341,39 @@ def on_network_initializing_end(
...
+        epoch_no: int,
+        epoch_loss: float,
+        training_network: nn.HybridBlock,
+        trainer: gluon.Trainer,
+        best_epoch_info: Dict[str, Any],
+        ctx: mx.Context,
+    ) -> bool:
+        if self.time_limit is not None:
+            cur_time = time.time()
+            if cur_time - self.start_time > self.time_limit:
```
jaheba, line 374:
```suggestion
            elapsed = time.time() - self.start_time
            if elapsed > self.time_limit:
```

src/gluonts/mx/trainer/callback.py
```diff
@@ -338,3 +341,39 @@ def on_network_initializing_end(
         self, training_network: nn.HybridBlock
     ) -> None:
         copy_parameters(self.predictor.prediction_net, training_network)
+
+
+class TrainingTimeLimitCallback(Callback):
```
jaheba, line 346: I think we omitted the `callback` name from other callbacks.

```suggestion
class TrainingTimeLimit(Callback):
```

src/gluonts/mx/trainer/callback.py
jaheba:
```suggestion
    def __init__(self, time_limit: Optional[int] = None) -> None:
```

src/gluonts/mx/trainer/callback.py
```diff
@@ -338,3 +341,39 @@ def on_network_initializing_end(
...
+    def __init__(self, time_limit=None):
+        """
+        Used when you want to set a time limit to the training process.
+        Once passed into model.train(), the training process will end roughly after 'time_limit' seconds.
+
+        Attributes
+        ----------
+        time_limit: int
+                    time in seconds, after which your training process will end.
```
jaheba, line 355:
```suggestion
        time_limit: int
            time in seconds, after which training ends
```

src/gluonts/mx/trainer/callback.py
```diff
@@ -338,3 +341,39 @@ def on_network_initializing_end(
         self, training_network: nn.HybridBlock
     ) -> None:
         copy_parameters(self.predictor.prediction_net, training_network)
+
+
+class TrainingTimeLimitCallback(Callback):
+    def __init__(self, time_limit=None):
+        """
+        Used when you want to set a time limit to the training process.
+        Once passed into model.train(), the training process will end roughly after 'time_limit' seconds.
```
jaheba, line 350: Can we a bit more precise? Why is this "roughly"?

src/gluonts/mx/trainer/callback.py
```diff
@@ -338,3 +341,39 @@ def on_network_initializing_end(
...
+        training_network: nn.HybridBlock,
+        trainer: gluon.Trainer,
+        best_epoch_info: Dict[str, Any],
+        ctx: mx.Context,
+    ) -> bool:
+        if self.time_limit is not None:
+            cur_time = time.time()
+            if cur_time - self.start_time > self.time_limit:
+                logger.warning(
+                    "Time limit exceed during training, stop training."
```
jaheba, line 376:
```suggestion
                    "Time limit exceeded during training, stopping training."
```

jaheba (commented):
Before I forget: Thanks for the PR!

src/gluonts/mx/trainer/callback.py
yx1215: The actual training time might be a little bit over `time_limit`. For example, if `time_limit` is 60s, the first 10 epochs take 55s, and the 11st epoch takes 10s, the training process will stop after epoch 11, which takes 65s in total.

jaheba:
@borchero Can you also take a look?

src/gluonts/mx/trainer/callback.py
borchero: Could we add this to the documentation? Something like "the training process will end after the first batch exceeding `time_limit` seconds"?

borchero: Ah it's actually epoch, not batch.

src/gluonts/mx/trainer/callback.py
```diff
@@ -338,3 +341,39 @@ def on_network_initializing_end(
...
+        epoch_loss: float,
+        training_network: nn.HybridBlock,
+        trainer: gluon.Trainer,
+        best_epoch_info: Dict[str, Any],
+        ctx: mx.Context,
+    ) -> bool:
+        if self.time_limit is not None:
+            elapsed = time.time() - self.start_time
+            if elapsed > self.time_limit:
+                logger.warning(
```
borchero, line 375: `loger.info`?

borchero (changes requested):
This is very nice in general, I was implementing something similar very recently :D however, I would consider a couple more points:

- Should we time validation? We could "stop" timing `on_validation_epoch_start` and resume on `on_train_epoch_start`.
- Should we extend the `Callback` base class to return a value from `on_train_batch_end`? This way, we could stop training after the first batch exceeding the time limit instead of the first epoch (which is more useful in my opinion).

jaheba:
> * Should we time validation? We could "stop" timing `on_validation_epoch_start` and resume on `on_train_epoch_start`.

What is the intuition here? Why would I want to stop training after a certain amount of time?

> * Should we extend the `Callback` base class to return a value from `on_train_batch_end`? This way, we could stop training after the first batch exceeding the time limit instead of the first epoch (which is more useful in my opinion).

Would there be other places where we also would like to be able to stop training?

yx1215:
> > * Should we time validation? We could "stop" timing `on_validation_epoch_start` and resume on `on_train_epoch_start`.
> 
> What is the intuition here? Why would I want to stop training after a certain amount of time?
> 
> Sometimes user might only have limit time resource. They won't know with that limit time, how much epoch they can run. so they can set the epoch to 9999 and give a time limit to make it stop once the time is used up. 
> > * Should we extend the `Callback` base class to return a value from `on_train_batch_end`? This way, we could stop training after the first batch exceeding the time limit instead of the first epoch (which is more useful in my opinion).
> 
> Would there be other places where we also would like to be able to stop training?
>
> Maybe we can do what @borchero says, recording only the training process, not the validation process. And if we want to be more precise with the time limit, we can check after every batch

borchero:
> What is the intuition here? Why would I want to stop training after a certain amount of time?

One very common use case is hyperparameter optimization. In the successive halving algorithm, you train many configurations for a "budget" N, then take the best 50% of configurations, train for a total budget of 2N and so on ... While you can look at the budget as the number of epochs, it is often useful to actually use time as the budget (as it also often determines the money spent) -- especially if you want your budget to be independent from model size and/or dataset size.

> Would there be other places where we also would like to be able to stop training?

I think, we could consider allowing to do that after every hook which is called at the end of some iteration (i.e. end of training/validation batch, end of training/validation epoch, end of epoch).

borchero:
> Maybe we can do what @borchero says, recording only the training process, not the validation process. And if we want to be more precise with the time limit, we can check after every batch.

I would say that we should have a flag in the `__init__` of the callback to determine whether validation should be recorded.

jaheba:
I think I'm against treating treating validation differently and to have a net-train mode. It makes the code more complicated and is less intuitive (if I have a budged, I don't care too much how it is spend). If we realise there is still a need for this, we can add it later.

jaheba:
We can have a flag, which controls whether we check after each batch or after each full epoch.

What happens if we stop after a batch? Is that entire epoch invalidated?

borchero:
> I think I'm against treating treating validation differently and to have a net-train mode. It makes the code more complicated and is less intuitive (if I have a budged, I don't care too much how it is spend). If we realise there is still a need for this, we can add it later.

I actually needed this exactly (i.e. only track training time and not callbacks/validation since callbacks were potentially very time-consuming) few weeks ago and ended up rewriting the `Trainer` class (for some other reasons as well). Would be nice to have this option out-of-the-box and it's not too much work imo.

borchero:
> What happens if we stop after a batch? Is that entire epoch invalidated?

I would just stop the epoch prematurely and treat it like the epoch has been completed.

jaheba:
> > I think I'm against treating treating validation differently and to have a net-train mode. It makes the code more complicated and is less intuitive (if I have a budged, I don't care too much how it is spend). If we realise there is still a need for this, we can add it later.
> 
> I actually needed this exactly (i.e. only track training time and not callbacks/validation since callbacks were potentially very time-consuming) few weeks ago and ended up rewriting the `Trainer` class (for some other reasons as well). Would be nice to have this option out-of-the-box and it's not too much work imo.

Fair enough.

jaheba:
> > What happens if we stop after a batch? Is that entire epoch invalidated?
> 
> I would just stop the epoch prematurely and treat it like the epoch has been completed.

Then let's do this.

yx1215:
I'm writing the conclusion here to make sure we are on the same page.
We will have to add the following:
- a flag that controls whether validation epoch should be recorded in the time limit
- a flag that controls whether we stop at the end of each epoch or each batch, and if we stop after a batch, we will treat it as the whole epoch ends.

Did I miss anything?

src/gluonts/mx/trainer/callback.py
```diff
@@ -355,25 +362,96 @@ def __init__(self, time_limit: Optional[int] = None) -> None:
...
+        self.checkpoint = -1.0
 
     def on_train_start(self, max_epochs: int) -> None:
-        self.start_time = time.time()
+        self.checkpoint = time.time()
+
+    def on_train_batch_end(self, training_network: nn.HybridBlock) -> bool:
+        print(
+            "on_train_batch_end", self.time_spent
+        )  # for debugging purpose, will be deleted before merging
```
borchero, line 378: Just to keep track of the comment^^

src/gluonts/mx/trainer/callback.py
```diff
@@ -355,25 +362,96 @@ def __init__(self, time_limit: Optional[int] = None) -> None:
...
-        ctx: mx.Context,
     ) -> bool:
-        if self.time_limit is not None:
-            elapsed = time.time() - self.start_time
-            if elapsed > self.time_limit:
+        self.time_spent += time.time() - self.checkpoint
+        self.checkpoint = time.time()
+        print(
+            "on_train_epoch_end", self.time_spent
+        )  # for debugging purpose, will be deleted before merging
```
borchero, line 397: Only for tracking.

src/gluonts/mx/trainer/callback.py
```diff
@@ -355,25 +362,96 @@ def __init__(self, time_limit: Optional[int] = None) -> None:
...
+        self.checkpoint = time.time()
+
+    def on_train_batch_end(self, training_network: nn.HybridBlock) -> bool:
+        print(
+            "on_train_batch_end", self.time_spent
+        )  # for debugging purpose, will be deleted before merging
+        self.time_spent += time.time() - self.checkpoint
+        self.checkpoint = time.time()
+        if self.stop_during_epoch:
+            if self.time_spent > self.time_limit:
```
borchero, line 382: Log for consistency?

src/gluonts/mx/trainer/callback.py
```diff
@@ -355,25 +362,96 @@ def __init__(self, time_limit: Optional[int] = None) -> None:
...
+        )  # for debugging purpose, will be deleted before merging
+        tmp_time_spent = time.time() - self.checkpoint
+        self.checkpoint = time.time()
+
+        if self.include_validation_in_time_limit:
+            self.time_spent += tmp_time_spent
+
+        if self.stop_during_epoch:
+            if self.time_spent > self.time_limit:
+                return False
```
borchero, line 417: Logging for consistency

borchero (changes requested):
Nice, thanks!!

src/gluonts/mx/trainer/callback.py
yx1215: I think we don't need log here, otherwise, the log will be printed twice(one after the batch, and one after the epoch). Because regardless of whether we are to stop after one batch, time limit will always be check after one epoch.

src/gluonts/mx/trainer/callback.py
yx1215: Same as above.

jaheba:
Looks like there is some code duplication. Can we make the checking more reusable?

yx1215:
> Looks like there is some code duplication. Can we make the checking more reusable?

I've just reduced the redundancy of the code.

src/gluonts/mx/trainer/_base.py
```diff
@@ -404,6 +408,8 @@ def loop(  # todo call run epoch
                         logger.info(
                             f"Number of parameters in {net_name}: {num_model_param}"
                         )
+                    if not should_continue:
+                        break
```
jaheba, line 412: This doesn't set the outer `should_continue` and thus we will call `loop` again and again.

src/gluonts/mx/trainer/callback.py
```diff
@@ -338,3 +353,108 @@ def on_network_initializing_end(
         self, training_network: nn.HybridBlock
     ) -> None:
         copy_parameters(self.predictor.prediction_net, training_network)
+
+
+class TrainingTimeLimit(Callback):
+    def __init__(
+        self,
+        time_limit: int,
+        track_validation_duration: bool = True,
```
jaheba, line 362: I think something like `use_net_training` is better, since time can be spent in a lot of places, including validation.

src/gluonts/mx/trainer/callback.py
```diff
@@ -338,3 +353,108 @@ def on_network_initializing_end(
...
+        self.track_validation_duration = track_validation_duration
+        self.stop_during_epoch = stop_during_epoch
+        self.time_spent = 0.0
+        self.checkpoint = -1.0
+        self.report_exceed_timelimit = True
+
+    def on_train_start(self, max_epochs: int) -> None:
+        self.checkpoint = time.time()
+
+    def should_continue_by_timelimit(self, record_time=True, should_stop=True):
```
jaheba, line 388: I think this should be two separate methods, which do not have default arguments. It's not really clear to me at a quick glance what this is supposed to be doing.

src/gluonts/mx/trainer/callback.py
yx1215: I'm not sure what this actually means.... What would we use ```use_net_training``` to record?

src/gluonts/mx/trainer/_base.py
yx1215: I fixed this by setting ```self.halt=True``` before we break from the loop.

lostella:
Marked as breaking because of the change in the output signature for the `on_train_batch_end` and `on_validation_batch_end` methods

jaheba:
@lostella Should we have tests, at least for `serde`?

lostella:
> @lostella Should we have tests, at least for `serde`?

Yes that would be good. For example the `TerminateOnNan` callback doesn't seem to use any of the serialization mechanisms other classes rely on.

jaheba:
However not sure we can really test it functionally. Would also take some time, which we don't have.

lostella (approved):

PR #1632, opened 2021-07-19
Add additional datasets from the Monash Time Series Forecasting Repository

borchero:
*Description of changes:*

This PR adds 11 more datasets from the [Monash Time Series Forecasting Repository](https://forecastingdata.org/). It uses a default prediction length whenever none is available, depending on the frequency of the data. Additionally, this PR includes the following changes:

- Users are able to specify a custom `prediction_length` in the `get_dataset` function. This will override any defaults and takes effect whenever `regenerate=True` or the dataset is generated for the first time.
- The PR fixes a bug in the frequency calculation for the Monash datasets where (1) a minutely frequency was parsed as monthly and (2) `half_<frequency>` couldn't be parsed.

### DIFF
--- a/src/gluonts/dataset/repository/_tsf_datasets.py
+++ b/src/gluonts/dataset/repository/_tsf_datasets.py
@@ -13,16 +13,17 @@
 
 from pathlib import Path
 from tempfile import TemporaryDirectory
-from typing import List, Dict, NamedTuple
+from typing import Dict, List, NamedTuple, Optional
 from urllib import request
 from zipfile import ZipFile
 
 from gluonts import json
 from gluonts.dataset import jsonl
+from gluonts.dataset.field_names import FieldName
 from gluonts.gluonts_tqdm import tqdm
 
-from ._tsf_reader import frequency_converter, TSFReader
-from ._util import metadata, to_dict, request_retrieve_hook
+from ._tsf_reader import TSFReader, frequency_converter
+from ._util import metadata, request_retrieve_hook, to_dict
 
 
 class Dataset(NamedTuple):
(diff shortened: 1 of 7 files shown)


src/gluonts/dataset/repository/_tsf_datasets.py
```diff
@@ -150,13 +199,39 @@ def generate_forecasting_dataset(dataset_path: Path, dataset_name: str):
...
-    save_metadata(
-        dataset_path,
-        len(data),
-        frequency_converter(meta.frequency),
-        prediction_length,
-    )
 
-    save_datasets(dataset_path, data, prediction_length)
+def default_prediction_length_from_frequency(freq: str) -> int:
+    freq = freq.strip("0123456789").upper()
```
jaheba, line 222: I think you can use`pandas.tseries.frequencies.to_offset`for this:

```py
In [3]: to_offset("60min").name
Out[3]: 'T'
```

src/gluonts/dataset/repository/_tsf_datasets.py
```diff
@@ -150,13 +199,39 @@ def generate_forecasting_dataset(dataset_path: Path, dataset_name: str):
...
+    if freq == "H":
+        return 48
+    if freq == "D":
+        return 30
+    if freq == "W":
+        return 8
+    if freq == "M":
+        return 12
+    if freq == "Y":
+        return 4
```
jaheba, line 234: Looks like this could be done with a dict instead?

jaheba (commented):
Generally lgtm, thanks!

jaheba:
Looks like it passes `None` for `prediction_length` in tests:

```py
ValidationError: 1 validation error for SimpleFeedForwardEstimatorModel
prediction_length
  none is not an allowed value (type=type_error.none.not_allowed)
```

Looks like pydantic is good for something :)

jaheba (approved):

PR #1652, opened 2021-08-09
Fix frequency string conversion from ts format, add test

### LINKED ISSUE #1649: _tsf_reader.convert_base doesn't handle "10 minutes" frequency correctly
## Description
For Monash datasets with the "10 minutes" frequency, the frequency converter will convert it to a frequency 10 MonthEnd (10M), instead of the expect 10 Minutes (10T) frequency.

There is already code to properly handle the "minutely" frequency, but it checks for that string explicitly, so it doesn't catch the "10 minutes" case.

## To Reproduce
One dataset which has this frequency is the 10 minutes observation Solar dataset: https://zenodo.org/record/4656144
filename: `"solar_10_minutes_dataset.zip"`
record: `"4656132"`

lostella:
*Issue #, if available:* Fixes #1649

*Description of changes:*

### DIFF
--- a/src/gluonts/dataset/repository/_tsf_reader.py
+++ b/src/gluonts/dataset/repository/_tsf_reader.py
@@ -48,7 +48,7 @@ def frequency_converter(freq: str):
 
 
 def convert_base(text: str) -> str:
-    if text.lower() == "minutely":
+    if text.lower() in ["minutely", "minutes"]:
         return "T"
     return text[0].upper()
 
--- a/src/gluonts/dataset/repository/_tsf_reader.py
+++ b/src/gluonts/dataset/repository/_tsf_reader.py
@@ -56,6 +56,7 @@ BASE_FREQ_TO_PANDAS_OFFSET: Dict[str, str] = {
     "hourly": "H",
     "hours": "H",
     "daily": "D",
+    "days": "D",
     "weekly": "W",
     "weeks": "W",
     "monthly": "M",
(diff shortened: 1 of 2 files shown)


lostella:
@borchero could you check whether the problematic cases you've found are covered by the test?

jaheba:
I wonder if we should have a more explicit lookup, ie a mapping from names to pandas frequencies. And then throw an error if we can't find the given input name.

lostella:
> I wonder if we should have a more explicit lookup, ie a mapping from names to pandas frequencies. And then throw an error if we can't find the given input name.

Yeah makes sense, we could do that

src/gluonts/dataset/repository/_tsf_reader.py
```diff
@@ -47,10 +49,32 @@ def frequency_converter(freq: str):
...
-    if text.lower() == "minutely":
-        return "T"
-    return text[0].upper()
+    try:
+        offset_str = BASE_FREQ_TO_PANDAS_OFFSET[text]
+    except KeyError:
+        raise GluonTSDataError(
+            f'"{text}" is not recognized as a frequency string'
+        )
+    return offset_str
```
jaheba, line 77:
```suggestion
    try:
        return BASE_FREQ_TO_PANDAS_OFFSET[text]
    except KeyError:
        raise GluonTSDataError(
            f'"{text}" is not recognized as a frequency string'
        )
```

src/gluonts/dataset/repository/_tsf_reader.py
```diff
@@ -47,10 +49,31 @@ def frequency_converter(freq: str):
...
+    "minutes": "T",
+    "hourly": "H",
+    "hours": "H",
+    "daily": "D",
+    "days": "D",
+    "weekly": "W",
+    "weeks": "W",
+    "monthly": "M",
+    "months": "M",
+    "quarterly": "Q",
```
borchero, line 64: "quarters" for completeness?

borchero (approved):
LGTM

PR #1654, opened 2021-08-12
Backports for v0.7.7

lostella:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/test/torch/modules/test_torch_distribution_inference.py
+++ b/test/torch/modules/test_torch_distribution_inference.py
@@ -268,8 +268,13 @@ def test_poisson(rate: float) -> None:
     ), f"rate did not match: rate = {rate}, rate_hat = {rate_hat}"
 
 
-# The following parameters match the Gluon-based test, adjusted for the different parametrization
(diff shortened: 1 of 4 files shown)


mbohlkeschneider (approved):

test/torch/modules/test_torch_distribution_inference.py
```diff
@@ -268,8 +268,13 @@ def test_poisson(rate: float) -> None:
     ), f"rate did not match: rate = {rate}, rate_hat = {rate_hat}"
 
 
-# The following parameters match the Gluon-based test, adjusted for the different parametrization
-@pytest.mark.parametrize("total_count, logit", [(1 / 0.7, np.log(2.5 * 0.7))])
```
mbohlkeschneider, line 272: Any particular reason these have been changed?

lostella: Mostly readability. There was a bug in the `NegativeBinomialOutput` which went unnoticed because the logit was passed were a probability was required: it just happens that `np.log(2.5 * 0.7)) ≈ 0.56` so it's both valid as a logit and a probability. So I just replaced this test case parameters with explicit numbers, and added another one where the logit is clearly *not* a probability

PR #1662, opened 2021-08-23
Replace miniver with custom versioning code.

jaheba:
*Issue #, if available:*

*Description of changes:*


src/gluonts/_version.py
```diff
@@ -1,214 +1,244 @@
...
+When using git, `setup.py` needs to be located at the root of the repository.
 
+Thus, a typical setup looks like this::
 
-def get_static_version_info(version_file=STATIC_VERSION_FILE):
-    version_info = {}
-    with open(os.path.join(package_root, version_file), "rb") as f:
-        exec(f.read(), {}, version_info)
-    return version_info
+    my_package/
```
lostella, line 35: my_repo maybe?

src/gluonts/_version.py
```diff
@@ -1,214 +1,244 @@
...
+        .git
+        my_package/
+            __init__.py
+            _version.py
```
lostella, line 39: In this case setup.py is just one level up, unlike in gluonts

lostella (commented):
Just high level comments, I'll try to dive deeper later

src/gluonts/_version.py
jaheba: Sounds clearer, thanks. Although, the name of the two folders is often the same.

src/gluonts/_version.py
jaheba: Right, but that's what the `SEARCH_PACKAGE_LEVELS` thing is there for :)

lostella (commented):
What happens if you check out for development, and do `pip install -e .`? Doesn't `setup.py` run in this case, and set a static version file?

jaheba:
No, that only happens when you run it `sdist` or `build_py`.

It works as intended with `-e`.

PR #1665, opened 2021-08-24
Hierarchical Forecaster: End-to-End model based on DeepVAR

rshyamsundar:
*Issue #, if available:*

*Description of changes:*

Implementation of the end-to-end hierarchical forecasting method described in the ICML 2021 paper: [End-to-End Learning of Coherent Probabilistic Forecasts for Hierarchical Time Series](http://proceedings.mlr.press/v139/rangapuram21a.html)

* The code is based on top of `DeepVAR` method, so the estimator/network of the new method only contain things required for enforcing hierarchical constraints and computing the appropriate loss on samples. This also meant minor changes in the existing code especially `LowrankMultivariateGaussian` (making it work for the case `r=0`). @mbohlkeschneider can you have a look at those changes?
* I added a separate module for empirical distribution to compute CRPS loss on samples. Note that `crps` method of existing parametric distributions cannot be used because the loss is computed on the *modified/reconciled* samples for which there is no underlying parametric distribution.
* Added simple end-to-end test that depends on synthetic hierarchical data. Since `gluonts` does not support hierarchical time series dataset yet, I kept the generation of synthetic data to the `test` package currently. We need to agree on the right format for hierarchical datasets.


lostella:
@rshyamsundar you committed some file from `.git` by mistake I think

src/gluonts/model/deepvar_hierarchical/_estimator.py
```diff
@@ -0,0 +1,308 @@
...
+
+    m, m_K = S.shape
+    m_agg = m - m_K
+
+    # The top `m_agg` rows of the matrix `S` give the aggregation constraint matrix.
+    S_agg = S[:m_agg, :]
+    A = np.hstack((np.eye(m_agg), -S_agg))
+
+    M = np.eye(m) - A.T @ np.linalg.pinv(A @ A.T) @ A
+
```
melopeo, line 71: Minor comment. Do we need the pseudo inverse?
Since the rows of A are independent, then AA^T is invertible.

rshyamsundar: I think I added this for avoiding some numerical issues because it was getting to close to being singular.

src/gluonts/mx/distribution/lowrank_multivariate_gaussian.py
```diff
@@ -259,30 +277,33 @@ def sample_rep(self, num_samples: int = None, dtype=np.float32) -> Tensor:
             tensor with shape (num_samples, ..., dim)
         """
 
-        def s(mu: Tensor, D: Tensor, W: Tensor) -> Tensor:
+        def s(mu: Tensor, D: Tensor, W: Optional[Tensor] = None) -> Tensor:
```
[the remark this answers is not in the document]
rshyamsundar: Thanks, I will add a test..

src/gluonts/model/deepvar/_network.py
```diff
@@ -621,6 +621,8 @@ def repeat(tensor):
 
             # (batch_size, 1, target_dim)
             new_samples = distr.sample()
+            if self._post_process_samples:
```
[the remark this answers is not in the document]
rshyamsundar: very good point.. I updated it and will be part of the next revision.

src/gluonts/model/deepvar_hierarchical/_network.py
```diff
@@ -0,0 +1,439 @@
...
+            Tensor of coherent samples.
+
+        """
+        coherent_samples = self.reconcile_samples(samples=samples)
+
+        assert_shape(coherent_samples, samples.shape)
+
+        # assert that A*X_proj ~ 0
+        if self.assert_reconciliation:
+            assert self.reconciliation_error(samples=coherent_samples) < 1e-2
```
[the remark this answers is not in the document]
rshyamsundar: This is only done during prediction as a sanity check. By default this is disabled so the user would not see this. Maybe  a better option is to just report the final coherence error and not terminate if it is above the given tolerance.

src/gluonts/model/deepvar_hierarchical/_network.py
```diff
@@ -0,0 +1,439 @@
...
+    @validated()
+    def __init__(
+        self,
+        num_parallel_samples: int,
+        assert_reconciliation: bool,
+        coherent_pred_samples: bool,
+        **kwargs,
+    ) -> None:
+        super().__init__(num_parallel_samples=num_parallel_samples, **kwargs)
+        self._post_process_samples = coherent_pred_samples
```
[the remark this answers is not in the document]
rshyamsundar: again good point.. refactored it.

src/gluonts/model/gpvar/_network.py
```diff
@@ -271,6 +271,7 @@ class GPVARPredictionNetwork(GPVARNetwork):
     def __init__(self, num_parallel_samples: int, **kwargs) -> None:
         super().__init__(**kwargs)
         self.num_parallel_samples = num_parallel_samples
+        self._post_process_samples = False
```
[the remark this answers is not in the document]
rshyamsundar: yes, refactored!

src/gluonts/mx/distribution/empirical_distribution.py
```diff
@@ -0,0 +1,153 @@
...
+    Multivariate target is also supported.
+
+    Parameters
+    ----------
+    samples
+        Tensor containing samples, of shape `(num_samples, *batch_shape, *event_shape)`.
+    """
+
+    @validated()
+    def __init__(self, samples: Tensor) -> None:
```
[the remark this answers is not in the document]
rshyamsundar: right.. this could work for any event dimension; updated it.

rshyamsundar:
@lostella I addressed your comments and also refactored the hierarchical network further: it does not even have `train_hybrid_forward` now; everything is used from `deepvar` network; it has bare minimum to do the hierarchical specific stuff.

src/gluonts/model/deepvar_hierarchical/_estimator.py
```diff
@@ -0,0 +1,308 @@
...
+
+# Relative imports
+from ._network import (
+    DeepVARHierarchicalPredictionNetwork,
+    DeepVARHierarchicalTrainingNetwork,
+)
+
+
+def projection_mat(S, return_constraint_mat: bool = False):
```
mbohlkeschneider, line 40: Type of S? Could you also add return types to the function declaration rather than in the docstring? I think this is overall the more consistent approach and also helps because it flares up in an IDE if violated.

src/gluonts/model/deepvar_hierarchical/_estimator.py
```diff
@@ -0,0 +1,308 @@
...
+    m, m_K = S.shape
+    m_agg = m - m_K
+
+    # The top `m_agg` rows of the matrix `S` give the aggregation constraint matrix.
+    S_agg = S[:m_agg, :]
+    A = np.hstack((np.eye(m_agg), -S_agg))
+
+    M = np.eye(m) - A.T @ np.linalg.pinv(A @ A.T) @ A
+
+    if return_constraint_mat:
```
mbohlkeschneider, line 72: I'm not sure we should make the number of return objects dependent on a flag. This would cause that the downstream code does/does not work because it needs to take the flag into account. Could we just always return the `constraint_mat` and ignore it downstream if not used? Seems there is no option to set this in the estimator anyway.

src/gluonts/model/deepvar_hierarchical/_estimator.py
```diff
@@ -0,0 +1,308 @@
...
+    embedding_dimension
+        Dimension of the embeddings for categorical features
+        (default: 5])
+    distr_output
+        Distribution to use to evaluate observations and sample predictions
+        (default: LowrankMultivariateGaussianOutput with dim=target_dim and
+        rank=0, i.e., covariance matrix is assumed to be diagonal).
+        Note that target dim of the DistributionOutput and the estimator constructor
+        call need to match.
+    rank
```
mbohlkeschneider, line 142: Should we print a warning to inform the user of unused parameters? I think this is an issue that we have a lot in GluonTS and the docstring can be easily missed.

src/gluonts/model/deepvar_hierarchical/_estimator.py
```diff
@@ -0,0 +1,308 @@
...
+        scaling: bool = True,
+        pick_incomplete: bool = False,
+        lags_seq: Optional[List[int]] = None,
+        time_features: Optional[List[TimeFeature]] = None,
+        conditioning_length: int = 200,
+        use_marginal_transformation: bool = False,
+        batch_size: int = 32,
+        warmstart_epoch_frac: float = 0.0,
+        sample_LH: bool = False,
+        seq_axis: List[int] = None,
```
mbohlkeschneider, line 198: This seems to be missing in docstring? It is not clear what it is either.

src/gluonts/model/deepvar_hierarchical/_network.py
```diff
@@ -0,0 +1,341 @@
...
+            scaling=scaling,
+            **kwargs,
+        )
+
+        self.M = M
+        self.M_broadcast = None
+        self.A = A
+        self.seq_axis = seq_axis
+
+    def reconcile_samples(self, samples):
```
mbohlkeschneider, line 81: Can we add types here?

src/gluonts/model/deepvar_hierarchical/_network.py
```diff
@@ -0,0 +1,341 @@
...
+            out = mx.nd.moveaxis(
+                out, list(range(len(self.seq_axis))), self.seq_axis
+            )
+            return out
+        else:
+            return mx.nd.linalg.gemm2(
+                self.M_broadcast, samples.expand_dims(-1)
+            ).squeeze(axis=-1)
+
+    def reconciliation_error(self, samples):
```
mbohlkeschneider, line 158: type?

src/gluonts/model/deepvar_hierarchical/_network.py
```diff
@@ -0,0 +1,341 @@
...
+        A_broadcast = mx.nd.broadcast_to(
+            A_expanded, shape=samples.shape[0:1] + constraint_mat_shape
+        )
+        return mx.nd.max(
+            mx.nd.abs(
+                mx.nd.linalg_gemm2(A_broadcast, samples, transpose_b=True)
+            )
+        ).asnumpy()[0]
+
+    def loss(self, target, distr):
```
mbohlkeschneider, line 186: Personally, I think it might be cleaner to split the different losses into sub-functions and just put the flag logic here. It makes it hard to read otherwise.

test/model/deepvar_hierarchical/test_deepvar_hierarchical.py
```diff
@@ -0,0 +1,100 @@
...
+prediction_length = 10
+train_datasets = sine7(
+    seq_length=seq_length, prediction_length=prediction_length
+)
+
+
+@pytest.mark.parametrize(
+    "likelihood_weight, CRPS_weight, coherent_train_samples, coherent_pred_samples, warmstart_epoch_frac",
+    [
+        # Hier-E2E
```
mbohlkeschneider, line 32: Just curious: Does `hybridize` work here?

mbohlkeschneider (changes requested):
Thank you for adding this. Very exciting!

Two overall comments:
1. Should that `.typesafe` file be here?
2. Some functions miss proper type annotations and return types. Maybe we can add this?

test/model/deepvar_hierarchical/test_deepvar_hierarchical.py
rshyamsundar: No, it does not and it is not possible I think given the problem specific not-learned tensors like projection matrix are used in the loss computation.

src/gluonts/model/gpvar/_estimator.py
```diff
@@ -379,7 +379,6 @@ def create_training_network(self) -> GPVARTrainingNetwork:
             lags_seq=self.lags_seq,
             scaling=self.scaling,
             distr_output=self.distr_output,
-            conditioning_length=self.conditioning_length,
```
rshyamsundar, line 382: Removing `conditioning_length` from network since it is not used there. @mbohlkeschneider please confirm this.

src/gluonts/model/deepvar_hierarchical/_estimator.py
rshyamsundar: Yes, forgot to add..

src/gluonts/model/deepvar_hierarchical/_estimator.py
rshyamsundar: yes, anyway I need both, so returning both of them.

src/gluonts/model/deepvar_hierarchical/_estimator.py
rshyamsundar: I removed unused parameters from the constructor arguments and explicitly set them at the beginning of constructor. If the method is extended, if at all, then they can be moved into arguments.

src/gluonts/model/deepvar_hierarchical/_estimator.py
rshyamsundar: Yes, updated.

src/gluonts/model/deepvar_hierarchical/_network.py
rshyamsundar: done!

src/gluonts/model/deepvar_hierarchical/_network.py
rshyamsundar: yes!

src/gluonts/model/deepvar_hierarchical/_network.py
rshyamsundar: yes, you are right.. This allows various combinations of computing loss (for ablation studies) and it is a bit involved. Refactored it as per your suggestion; check if it still needs to be improved.

rshyamsundar:
> Thank you for adding this. Very exciting!
> 
> Two overall comments:
> 
>     1. Should that `.typesafe` file be here?
> 
>     2. Some functions miss proper type annotations and return types. Maybe we can add this?

Thanks for the detailed look. I added missing annotations. The `.typesafe` file is at the right place I think, suggesting the type check can be done in this folder.

src/gluonts/model/deepvar/_network.py
```diff
@@ -60,7 +60,6 @@ def __init__(
         dropout_rate: float,
         lags_seq: List[int],
         target_dim: int,
-        conditioning_length: int,
```
mbohlkeschneider, line 63: Why is this removed?

src/gluonts/model/gpvar/_estimator.py
mbohlkeschneider: It is not used in the network but rather in the transformation. So it can be removed here but maybe should still be in `DeepVAR` to be able to use the cdf transform.

src/gluonts/model/deepvar/_network.py
rshyamsundar: This is needed in the input transformation part before the transformed inputs are fed into the network. The network does not need it afterwards and is also not used anywhere in `_network.py` module. 

The actual `predictor`has two parts: `input_transform` (where this is needed and is passed correctly) plus the prediction network (which directly gets the transformed features with suffix `_cdf`) right?

src/gluonts/model/gpvar/_network.py
```diff
@@ -348,3 +348,19 @@ def hybrid_forward(
...
+        ----------
+        samples
+            Tensor of shape (num_parallel_samples*batch_size, 1, target_dim)
+
+        Returns
+        -------
+            Tensor of samples with the same shape.
+
+        """
+        return samples
```
lostella, line 366: Isn’t this already defined as the identity in the superclass?

src/gluonts/model/deepvar/_network.py
```diff
@@ -860,3 +878,19 @@ def hybrid_forward(
             past_is_pad=past_is_pad,
             future_time_feat=future_time_feat,
         )
+
+    def post_process_samples(self, samples: Tensor) -> Tensor:
+        """
+        Identity map.
```
lostella, line 884: Maybe you can describe what this method does in general, ie how it is used and how it affects the model, so that it’s clear what one can achieve by overriding it.

src/gluonts/mx/distribution/empirical_distribution.py
```diff
@@ -0,0 +1,258 @@
...
+from gluonts.mx import Tensor
+
+from .distribution import Distribution, getF
+from .distribution_output import DistributionOutput, ArgProj
+
+
+class EmpiricalDistribution(Distribution):
+    r"""
+    A class representing empirical distribution.
+    By default, the target is assumed to be vector-valued, i.e., `event_shape` is 1. However, note that each dimension
```
lostella, line 29: There should be an empty line after the short description (before starting a longer paragraph)

src/gluonts/mx/distribution/empirical_distribution.py
```diff
@@ -0,0 +1,258 @@
...
+    ----------
+    samples
+        Tensor containing samples, of shape `(num_samples, *batch_shape, *event_shape)`.
+    event_dim
+        Number of event dimensions, i.e., length of the `event_shape` tuple.
+        This is `0` for distributions over scalars, `1` over vectors, `2` over matrices, and so on.
+    """
+
+    @validated()
+    def __init__(self, samples: Tensor, event_dim: int = 1) -> None:
```
lostella, line 45: I’m not sure I understand this: why is 1 the default here? Should this not be a mandatory argument, without any default?

src/gluonts/mx/distribution/empirical_distribution.py
```diff
@@ -0,0 +1,258 @@
...
+    def cdf(self, x: Tensor):
+        # Note: computes CDF on each dimension of the target independently.
+        self.CDF_sorted = self.F.linspace(
+            start=1 / self.samples.shape[0],
+            stop=1,
+            endpoint=True,
+            num=self.samples.shape[0],
+        )
+
+        # Replace this when mxnet is updated to a version where `searchsorted` is available!
```
lostella, line 95: :-( wouldn’t it be better to rely on numpy for this one?

test/distribution/test_distribution_output_serde.py
```diff
@@ -53,6 +55,9 @@
         DeterministicOutput(value=42.0),
         DirichletMultinomialOutput(dim=3, n_trials=5),
         DirichletOutput(dim=4),
+        EmpiricalDistribution(
```
lostella, line 58: Wait: how does this object fit this test? All others are DistributionOutput

test/model/deepvar_hierarchical/generate_hierarchical_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+"""
+Loads the datasets used in Salinas et al. 2019 (https://tinyurl.com/woyhhqy).
+This wrapper downloads and unpacks them so they don'thave to be attached as
```
lostella, line 16: I cannot find where anything is being downloaded, maybe this is an outdated docstring?

test/model/deepvar_hierarchical/generate_hierarchical_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+    freq: str
+
+
+class HierarchicalTrainDatasets(NamedTuple):
+    train: ListDataset
+    test: ListDataset
+    metadata: HierarchicalMetaData
+
+
+def sine7(seq_length: int, prediction_length: int):
```
lostella, line 38: ~There’s no test function in this script, is this script needed?~
Disregard this, my browser doesn’t allow me to delete this comment for some reason

src/gluonts/mx/distribution/empirical_distribution.py
rshyamsundar: I wanted to use numpy first, but looks like it only supports 1D arrays.

mbohlkeschneider (commented):
What is the status here? Shall we consolidate the open points and make another iteration? I would like to not have this linger too much in "PR phase".

test/model/deepvar_hierarchical/generate_hierarchical_dataset.py
rshyamsundar: ya, removed.

test/distribution/test_distribution_output_serde.py
rshyamsundar: updated it..

src/gluonts/mx/distribution/empirical_distribution.py
rshyamsundar: ya, removed.

src/gluonts/mx/distribution/empirical_distribution.py
rshyamsundar: Done.

src/gluonts/model/deepvar/_network.py
rshyamsundar: Done.

src/gluonts/model/gpvar/_network.py
rshyamsundar: It was not earlier; moved the method to `DeepVARNetwork` from `DeepVarPredictionNetwork`. Now this is not longer required. Thanks!

mbohlkeschneider (approved):
N I C E! I'm good now.

lostella (approved):

PR #1667, opened 2021-08-25
Adding SplicedBinnedPareto to the torch distributions.

Francois-Aubet:
*Description of changes:*

This PR is adding the BinnedUniforms, GeneralizedPareto, SplicedBinnedPareto to the torch distributions as well as an example file to show how they can be used.
This is also creating a torch/distribution folder similar to the one there is in the mx/distribution folder.

**Please tag this pr with at least one of these labels to make our release process faster:** new feature


lostella:
Two early observations:
- the `gluonts.torch.distributions` (with an `s`) is there (since yesterday: so close!) so you could use that
- I would avoid adding scripts to the `examples` folder, because these get broken, and add tests instead. If some example is needed for documentation purposes, you could add it in the form of a notebook (see markdown files [here](https://github.com/awslabs/gluon-ts/tree/master/docs/tutorials))

PR #1685, opened 2021-09-15
R: Add Python wrapper for calling R's hierarchical methods

rshyamsundar:
*Issue #, if available:*

*Description of changes:*

Adds a python wrapper for calling R's hierarchical forecasting methods. The processing of the input (multivariate) target and the forecasts are slightly different from what we have in the current wrapper for uni-variate methods. So, factored out the common stuff by adding a new base class `_predictor.py` and keeping the specifics in `_univariate_predictor.py` and `_hierarchical_predictor.py`.

Also added implementation of the naive bottom up reconciliation method. This is tested on a synthetic hierarchical dataset using both `ETS` and `ARIMA` as base forecasting methods.


src/gluonts/support/pandas.py
```diff
@@ -11,13 +11,31 @@
...
+    entry
+        Dictionary containing `start` and `target` time series.
+
+    Returns
+    -------
+        Timestamp
+            Forecast start date.
+
+    """
+    return frequency_add(entry["start"], np.shape(entry["target"])[-1])
```
lostella, line 41: Minor: how about making the time axis configurable? With default `-1`

src/gluonts/model/r_forecast/_predictor.py
```diff
@@ -49,57 +50,46 @@
...
     <http://pkg.robjhyndman.com/forecast/>`_.
 
-    The `RForecastPredictor` is a thin wrapper for calling the R forecast
-    package.  In order to use it you need to install R and run::
+    The `RForecastPredictor` is a thin wrapper for calling R `forecast` and `hts` packages.
+    In order to use it you need to install R and run::
 
         pip install 'rpy2>=2.9.*,<3.*'
         R -e 'install.packages(c("forecast", "nnfor"), repos="https://cloud.r-project.org")'
+        R -e 'install.packages(c("hts"), repos="https://cloud.r-project.org")'
```
lostella, line 64: Maybe we can avoid mentioning what is required here, since further changes will require updating the docstring again. How about mentioning that “specific R packages are required, depending on which wrapper one needs to run”?

src/gluonts/model/r_forecast/_predictor.py
```diff
@@ -49,57 +50,46 @@
...
-SUPPORTED_METHODS = (
-    SAMPLE_FORECAST_METHODS
-    + QUANTILE_FORECAST_METHODS
-    + POINT_FORECAST_METHODS
-)
-
 
 class RForecastPredictor(RepresentablePredictor):
     """
     Wrapper for calling the `R forecast package
```
lostella, line 56: This is not accurate anymore, right? This is now to invoke any of the R wrappers we provide, given that the necessary dependencies are installed

lostella (commented):
This is breaking the `RForecastPredictor` class; do you think this could be avoided? I think it would be nice to avoid breaking changes if possible.

For example, keeping the `RForecastPredictor` compatible with the existing one, adding `RHtsPredictor` (requiring the `hts` package), both inheriting from a new `RBasePredictor` class containing common functionality.

src/gluonts/support/pandas.py
rshyamsundar: I addressed all comments including this one.. this is hidden maybe because of resolving conflicts manually as the old `support` package is removed. It's [here](https://github.com/awslabs/gluon-ts/pull/1685/files#diff-07bbcc5e4c37496a9b8ef2a3bf07961fc70ff34599d14052434b657f4db15177R53).

test/model/r_forecast/test_r_hierarchical_predictor.py
```diff
@@ -0,0 +1,115 @@
...
+    SUPPORTED_HIERARCHICAL_METHODS,
+)
+
+
+# conditionally skip these tests if `R` and `rpy2` are not installed
+if not R_IS_INSTALLED or not RPY2_IS_INSTALLED:
+    skip_message = "Skipping test because `R` and `rpy2` are not installed!"
+    pytest.skip(msg=skip_message, allow_module_level=True)
+
+from test.model.deepvar_hierarchical.generate_hierarchical_dataset import sine7
```
melopeo: Running this locally I get the following:
`ModuleNotFoundError: No module named 'test.model'`

We need to relocate the dataset `sine7` so that it can be used in this test and the test in `test/model/deepvar_hierarchical/test_deepvar_hierarchical.py`.

A possibility for this would be to relocate `sine7` as a fixture in a `conftest.py` so that `sine7` can be passed to both tests.

melopeo (changes requested):
Current test for `test/model/r_forecast/test_r_hierarchical_predictor.py` is failing. It can be fixed by relocating the required `sine7` to a suitable `conftest.py`

src/gluonts/dataset/util.py
```diff
@@ -32,8 +34,23 @@ def frequency_add(ts: pd.Timestamp, amount: int) -> pd.Timestamp:
...
+    entry
+        Dictionary containing `start` and `target` time series.
+    time_axis
+        The time axis; default: -1.
+    Returns
+    -------
+        Timestamp
+            Forecast start date.
+    """
+    return frequency_add(entry["start"], np.shape(entry["target"])[time_axis])
```
melopeo: Recent modifications in gluon-ts have removed `frequency_add`, leading to a new formulation of `forecast_start`, e.g., https://github.com/awslabs/gluon-ts/blob/dev/src/gluonts/dataset/util.py#L20

It might be that we need to do a quick PR on gluon-ts so that we can translate the `time_axis` parameter to the new version of `forecast_start `.

The suggested PR would "maybe" go in the following direction
```
def forecast_start(entry, time_axis: int = -1):
    return entry[FieldName.START] + np.shape(entry[FieldName.TARGET])[time_axis]
```

melopeo: This task is now addressed in PR https://github.com/awslabs/gluon-ts/pull/2057

src/gluonts/model/r_forecast/_univariate_predictor.py
```diff
@@ -196,7 +196,7 @@ def _forecast_dict_to_obj(
                 ),
                 forecast_keys=list(quantile_forecasts_dict.keys()),
                 start_date=forecast_start_date,
-                freq=self.freq,
+                # freq=self.freq,
```
rshyamsundar: We don't need this anymore? Then remove it instead of commenting it.

src/gluonts/dataset/util.py
```diff
@@ -25,14 +25,12 @@ def to_pandas(instance: dict, freq: Optional[str] = None) -> pd.Series:
...
     Transform a dictionary into a pandas.Series object, using its "start" and
     "target" fields.
-
     Parameters
     ----------
     instance
         Dictionary containing the time series data.
     freq
         Frequency to use in the pandas.Series index.
-
```
lostella: Are these changes intended? Maybe we can leave them out

lostella (commented):
Seems like GitHub is complaining about conflicts, but I don't understand it. Maybe try to rebase on top of `dev` and force-push?

src/gluonts/dataset/util.py
rshyamsundar: unintended, removed.

rshyamsundar:
> Seems like GitHub is complaining about conflicts, but I don't understand it. Maybe try to rebase on top of `dev` and force-push?

This is due to the [recent bug fix](https://github.com/awslabs/gluon-ts/commit/b47a602dbb2dce8326a27348e0ae8cd2d8ee46ee) on mainline. Resolved the conflict manually and this change now appears in both  [`_univariate_predictor`](https://github.com/awslabs/gluon-ts/pull/1685/files#diff-a6e5c35ffddfb718c320a1aa77eda274ee8fc8742b33fa66aeaa5de2912b2674R154-R155) and [`_hierarchical_predictor`](https://github.com/awslabs/gluon-ts/pull/1685/files#diff-b2d1f6a096c727776da94749dcb2ddef24807379758bce91d518a15318d096a3R214-R215).

lostella (approved):

PR #1693, opened 2021-09-16
Adding IQF to remove quantile crossing and required retraining for ne…

dcmaddix:
…w quantiles for MQ-CNN

*Issue #, if available:*
- Remove the quantile crossing issue associated with MQ-CNN

*Description of changes:*
- Use the Incremental Quantile Function (IQF) method to remove the quantile crossing associated with MQ-CNN
- Add linear interpolation between knots and exponential extrapolation at the tails to remove the retraining for MQ-CNN with new query quantiles.
-  Add `is_iqf` hyperparameters to MQ-CNN to toggle between the original QF and new IQF

### DIFF
--- a/src/gluonts/model/seq2seq/_transform.py
+++ b/src/gluonts/model/seq2seq/_transform.py
@@ -11,15 +11,12 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-# Standard library imports
 from collections import Counter
 from typing import Iterator, List, Optional
 
-# Third-party imports
 import numpy as np
 from numpy.lib.stride_tricks import as_strided
 
-# First-party imports
 from gluonts.core.component import validated
 from gluonts.dataset.common import DataEntry
 from gluonts.dataset.field_names import FieldName
--- a/src/gluonts/model/seq2seq/_transform.py
+++ b/src/gluonts/model/seq2seq/_transform.py
@@ -42,6 +42,7 @@ class ForkingSequenceSplitter(FlatMapTransformation):
         is_pad_out: str = "is_pad",
         start_input_field: str = "start",
     ) -> None:
+        super().__init__()
 
         assert enc_len > 0, "The value of `enc_len` should be > 0"
         assert dec_len > 0, "The value of `dec_len` should be > 0"
--- a/src/gluonts/model/seq2seq/_transform.py
+++ b/src/gluonts/model/seq2seq/_transform.py
@@ -107,9 +107,7 @@ class ForkingSequenceSplitter(FlatMapTransformation):
             if len(target) < self.dec_len:
                 return
 
-            sampling_indices = self.instance_sampler(
-                target, 0, len(target) - self.dec_len
-            )
+            sampling_indices = self.instance_sampler(target)
         else:
             sampling_indices = [len(target)]
 
--- a/src/gluonts/model/seq2seq/_transform.py
+++ b/src/gluonts/model/seq2seq/_transform.py
@@ -11,11 +11,15 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
+# Standard library imports
+from collections import Counter
 from typing import Iterator, List, Optional
 
+# Third-party imports
 import numpy as np
 from numpy.lib.stride_tricks import as_strided
 
+# First-party imports
 from gluonts.core.component import validated
 from gluonts.dataset.common import DataEntry
 from gluonts.dataset.field_names import FieldName
@@ -28,7 +32,7 @@ class ForkingSequenceSplitter(FlatMapTransformation):
     @validated()
     def __init__(
         self,
-        instance_sampler,
+        train_sampler,
         enc_len: int,
         dec_len: int,
         num_forking: Optional[int] = None,
@@ -37,15 +41,15 @@ class ForkingSequenceSplitter(FlatMapTransformation):
         decoder_series_fields: Optional[List[str]] = None,
         encoder_disabled_fields: Optional[List[str]] = None,
         decoder_disabled_fields: Optional[List[str]] = None,
+        prediction_time_decoder_exclude: Optional[List[str]] = None,
         is_pad_out: str = "is_pad",
         start_input_field: str = "start",
     ) -> None:
-        super().__init__()
 
         assert enc_len > 0, "The value of `enc_len` should be > 0"
         assert dec_len > 0, "The value of `dec_len` should be > 0"
 
-        self.instance_sampler = instance_sampler
+        self.train_sampler = train_sampler
         self.enc_len = enc_len
         self.dec_len = dec_len
         self.num_forking = (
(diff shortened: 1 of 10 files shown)


youngsuk0723:
Looks good. Could you check the failure Flake8 / flake8 (pull_request)?

dcmaddix:
> Looks good. Could you check the failure Flake8 / flake8 (pull_request)?

Thanks - yes, it is failing due to using the math notation in the comments for the CRPS in the `QuantileOutput` class. It doesn't like the characters '\s' and '\e'.

youngsuk0723 (approved):
approved!

src/gluonts/mx/block/quantile_output.py
```diff
@@ -35,26 +36,31 @@ def __init__(
 
         Parameters
         ----------
-        quantiles
+        quantiles:
```
lostella, line 39: The colon is actually not needed there

src/gluonts/model/seq2seq/_transform.py
```diff
@@ -112,22 +133,32 @@ def flatmap_transform(
             pad_length_enc = max(0, -enc_len_diff)
             pad_length_dec = max(0, -dec_len_diff)
 
-            for ts_field in ts_fields:
+            for ts_field in list(ts_fields_counter.keys()):
```
lostella, line 136: No need to turn `.keys()` into a list

src/gluonts/mx/block/quantile_output.py
```diff
@@ -142,6 +148,64 @@ def compute_quantile_loss(
...
+        Reordering w.r.t. e_j for j=1, ..., n, gives
+        .. math::
+            :nowrap:
+
+                CRPS = 0.5 * (q_2 - q_1) z_1
+                + 0.5 * (q_3 - q_1)  z_2
+                + ....
+                + 0.5 * (q_n - q_{n-2}) z_{n-1}
+                + 0.5 * (q_n - q_{n-1}) z_n
+        , where each coefficient of z_j is quantile weight w_j.
```
lostella, line 173: Maybe the initial comma can be removed from this line?

Also, I’m not sure whether there should be an empty line preceding this?

src/gluonts/mx/block/quantile_output.py
```diff
@@ -142,6 +148,64 @@ def compute_quantile_loss(
...
+        Compute weight of each quantile
+
+        Math:
+        Let quantiles = [q_1, ..., q_n] with quantile estimates [z_1, ..., z_n].
+        Then, approximated CRPS with the linear interpolation is
+
+        .. math::
+            :nowrap:
+
+                    CRPS = sum_{i=1}^{n-1} 0.5 * (q_{i+1} - q_{i}) * (z_{i+1} + z_{i}).
```
lostella, line 162: There appears to be one level of indent too much here

src/gluonts/mx/block/quantile_output.py
```diff
@@ -150,16 +214,31 @@ class ProjectParams(nn.HybridBlock):
 
     Parameters
     ----------
-    num_quantiles
+    num_quantiles:
```
lostella, line 217: Colon not needed

src/gluonts/support/util.py
```diff
@@ -175,3 +175,241 @@ def erfinv(x: np.array) -> np.array:
...
+    def __init__(
+        self, quantile_predictions: dict, interpolation_type: str = "linear"
+    ):
+        self.quantile_predictions = quantile_predictions
+        self.quantiles = sorted(map(float, self.quantile_predictions))
+        self.num_quantiles = len(self.quantile_predictions)
+        self.interpolation_type = interpolation_type
+
+    def __call__(self, inference_quantile):
+        if self.interpolation_type == "linear":
```
lostella, line 201: This does not play well with extension: if you want to add new interpolation strategies, then you need to modify this class. It is better in my opinion to have a `LinearInterpolation` class with `__call__` method, and then one can add more classes with whatever interpolation

src/gluonts/support/util.py
```diff
@@ -175,3 +175,241 @@ def erfinv(x: np.array) -> np.array:
...
+    Parameters
+    ----------
+    quantile_predictions
+        Dict with key = str(quantile) and value is np.array with size prediction_length
+    interpolation_type
+
+    """
+
+    def __init__(
+        self, quantile_predictions: dict, interpolation_type: str = "linear"
```
lostella, line 193: Is it necessary to tie this class to the “quantile” naming? Can’t this be used for general interpolation?

src/gluonts/support/util.py
```diff
@@ -175,3 +175,241 @@ def erfinv(x: np.array) -> np.array:
...
+
+    def __call__(self, inference_quantile):
+        if self.interpolation_type == "linear":
+            return self.linear_interpolation(inference_quantile)
+        else:
+            raise NotImplementedError(
+                f"unknown interpolation type {self.interpolation_type}"
+            )
+
+    def linear_interpolation(self, inference_quantile: float):
```
lostella, line 208: Could numpy.interp be used to compute this? https://numpy.org/doc/stable/reference/generated/numpy.interp.html#numpy-interp

lostella (commented):
Hey @dcmaddix it seems like the docs build was failing here: could someone look into fixing that? Otherwise every PR will fail on that

dcmaddix:
> Hey @dcmaddix it seems like the docs build was failing here: could someone look into fixing that? Otherwise every PR will fail on that

Sure will fix the docs build and thanks for the helpful comments!

src/gluonts/support/util.py
dcmaddix: @youngsuk0723 @KelvinKan can you update this part when you merge in your changes Kelvin? Thanks!

src/gluonts/mx/block/quantile_output.py
dcmaddix: Updated and removed the other colons from the comments here too!

src/gluonts/mx/block/quantile_output.py
dcmaddix: removed comma and added empty line.

src/gluonts/mx/block/quantile_output.py
dcmaddix: Fixed! Do you know how to allow flake8 to pass on '\s' characters so that we can use '\sum' here? Also I couldn't get flake8 to pass with `\begin{aligned}` and `\end{aligned}` blocks, similar to what we have in the GP code.

src/gluonts/mx/block/quantile_output.py
dcmaddix: updated

src/gluonts/support/util.py
dcmaddix: Sure, will update the name to `LinearInterpolation`. @KelvinKan will you be adding any other additional interpolation methods here?

src/gluonts/support/util.py
dcmaddix: Yes, I think @youngsuk0723's implementation here was doing interpolation on the quantile predictions, where the knots are the quantiles but this can be generalized.

src/gluonts/model/seq2seq/_transform.py
dcmaddix: will update, thanks!

src/gluonts/support/util.py
dcmaddix: It is very close, but I'm not sure that we can directly call it here since our xp is a  list of quantiles of length num_quantiles and our fp is a list of lists, where the outer dimension is num_quantiles and the inner dimension is length prediction_length for each of the predictions. I think this function only works on scalar lists.

PR #1699, opened 2021-09-18
Fixing doc build from mqcnn-iqf commit

dcmaddix:
*Issue #, if available:*

*Description of changes:*
- Fixing doc build and comments from MQCNN-IQF PR: https://github.com/awslabs/gluon-ts/pull/1693
-  Fix `mean` method for `QuantileForecast`
- Generalize `LinearInterpolation` method for arbitrary points rather than just `quantile_predictions`

### DIFF
--- a/src/gluonts/support/util.py
+++ b/src/gluonts/support/util.py
@@ -15,7 +15,7 @@ import os
 import signal
 import time
 from pathlib import Path
-from typing import Any, Callable, Dict, Optional
+from typing import Any, Callable, Dict, List, Optional, Tuple
 
 import numpy as np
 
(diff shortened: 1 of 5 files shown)


lostella (commented):
The long docstring is still giving problems. Is it necessary to have it? Or can it be explained with words? It seems like it’s just computing the weights for some quadrature formula, so maybe it’s fine to just say that. (Is it trapezoidal quadrature btw? Didn’t check thoroughly). The function name should also reflect that probably

dcmaddix:
> The long docstring is still giving problems. Is it necessary to have it? Or can it be explained with words? It seems like it’s just computing the weights for some quadrature formula, so maybe it’s fine to just say that. (Is it trapezoidal quadrature btw? Didn’t check thoroughly). The function name should also reflect that probably

yes it looks like it is still failing due to an extra indent in that function and I can't find where. @youngsuk0723 can we remove the equations here?

dcmaddix:
@lostella I got the doc build fixed and generalized the linear interpolation method for arbitrary x and y, rather than just for quantile predictions.

src/gluonts/support/util.py
```diff
@@ -177,239 +177,245 @@ def erfinv(x: np.array) -> np.array:
...
-            approximation_type if self.num_quantiles > 1 else None
+            approximation_type if len(self.x_coord) > 1 else None
         )
         self.tol = tol
         if not self.approximation_type:
             pass
         elif self.approximation_type == "exponential":
-            self.init_exponential_tail_weights()
             self.tail_function_left = self.exponential_tail_left
             self.tail_function_right = self.exponential_tail_right
```
lostella, line 282: Similarly to the interpolation class: I think we could avoid putting too much logic and options in this class. Other tail approximations can be added as new classes; they can share the interface, and if some code is also to be shared among them, we will figure out how to better factor things.

So I think this class could be called `ExponentialTailApproximation`, right?

src/gluonts/support/util.py
```diff
@@ -177,239 +177,245 @@ def erfinv(x: np.array) -> np.array:
...
         Returns
         -------
-        same type and shape as any value of quantile_predictions dict,
-        i.e., type=np.array, shape = (prediction_length, )
+        np.ndarray
+            Interpolated values same shape as self.y_coord
         """
         if not self.approximation_type:
-            return self.quantile_predictions[str(self.quantiles[0])]
+            return self.y_coord[0]
```
lostella, line 305: If I understand correctly, in this case we are saying that the function is constant everywhere to the left of the leftmost data point. This could be its own `ConstantTailApproximation` class implementing this

lostella (changes requested):
@dcmaddix thanks! Looks better now! I only have a few more comments, especially around the `TailApproximation` class, which I think could be simplified similar to the interpolation one

src/gluonts/support/util.py
dcmaddix: yes will clean it up and update - thanks!

src/gluonts/support/util.py
dcmaddix: yes, I think we can simplify the logic more and just remove this case since this would only happen in the case that there is only 1 point, which isn't supported.

dcmaddix:
> @dcmaddix thanks! Looks better now! I only have a few more comments, especially around the `TailApproximation` class, which I think could be simplified similar to the interpolation one

Sure thanks @lostella ! I made the updates to the `TailApproximation` class to just be `ExponentialTailApproximation`.

lostella (approved):

PR #1702, opened 2021-09-21
Fix scaling for pytorch negative binomial output

kashif:
scale neg bin parameters as it should return integers.

*Issue #, if available:*

*Description of changes:*


lostella (approved):
Thanks for catching this @kashif! I must have overlooked it. As soon as #1699 is merged this one should be all green

PR #1704, opened 2021-09-24
Settings._let will pop element it added instead of just the last one.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/core/settings.py
+++ b/src/gluonts/core/settings.py
@@ -79,6 +79,64 @@ import pydantic
 from pydantic.utils import deep_update
 
 
+class StackElement:
+    def __init__(self, stack, val, prv=None, nxt=None):
+        self.stack = stack
+        self.val = val
+        self.prv = prv
+        self.nxt = nxt
+
+    def pop(self):
+        if self.prv is not None:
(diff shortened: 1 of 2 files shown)


src/gluonts/core/settings.py
```diff
@@ -79,6 +79,64 @@ def fn(debug):
...
+    def pop(self):
+        if self.prv is not None:
+            self.prv.nxt = self.nxt
+
+        if self.nxt is not None:
+            self.nxt.prv = self.prv
+
+        if self.stack.end is self:
+            self.stack.end = self.prv
+            self.stack.len -= 1
```
lostella: Should the stack length be decreased unconditionally?

lostella (commented):
Do we need our own custom `LinkedStack`? Wouldn't a regular Python `list` do the job?

jaheba:
In short no, let me add some more docs to the PR, to argue why the the approach is needed.

jaheba:
I wonder if we want to remove support for item-assignment at all. If you want to alter the `setting`, you should only use `push` and `let`.

lostella (approved):

PR #1721, opened 2021-10-06
Fix gluonts.json; added bdump/bdumps.

jaheba:
Fixes: 1708.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/dataset/jsonl.py
+++ b/src/gluonts/dataset/jsonl.py
@@ -22,14 +22,12 @@ from gluonts.exceptions import GluonTSDataError
 
 
 def load(file_obj):
-    for line in file_obj:
-        yield json.loads(line)
+    return map(json.loads, file_obj)
 
 
 def dump(objects, file_obj):
     for object_ in objects:
-        file_obj.write(json.dumps(object_))
-        file_obj.write("\n")
+        json.dump(object_, file_obj, nl=True)
 
 
 class Span(NamedTuple):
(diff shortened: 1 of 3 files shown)


src/gluonts/dataset/jsonl.py
```diff
@@ -22,14 +22,12 @@
 
 
 def load(file_obj):
-    for line in file_obj:
-        yield json.loads(line)
+    return map(json.loads, file_obj)
```
mbohlkeschneider, line 25: Doesn't this change the behavior when we change the generator to this? Is `map` evaluated lazy?

mbohlkeschneider (approved):

src/gluonts/dataset/jsonl.py
jaheba: Yes, this is lazy. `map` and `filter` don't return lists anymore in Python 3 (they used to in Python 2).

src/gluonts/json.py
```diff
@@ -48,17 +85,31 @@ def _json():
...
 
 for fn in _orjson, _ujson, _json:
     try:
-        variant, _methods = fn()
-
-        load = _methods["load"]
-        loads = _methods["loads"]
-        dump = _methods["dump"]
-        dumps = _methods["dumps"]
+        globals().update(**fn())
```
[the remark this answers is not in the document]
jaheba: I've added a test.

`globals()` returns a dictionary which is the modules global namespace. So this is a handy way to set all of these.

mbohlkeschneider (approved):

PR #1722, opened 2021-10-06
Hierarchical: Compute relative reconciliation error and add tests

rshyamsundar:
*Issue #, if available:*

*Description of changes:*

A couple of enhancements/refactoring on the reconciliation part:

* Compute the reconciliation error on relative scale. Move it out of the network as it is independent.
* Move `reconcile_samples` out of the network as we no longer need to cache big projection matrix.
* Tests for checking projection/reconciliation works correctly and the reconciliation error is computed as expected.

### DIFF
--- a/src/gluonts/model/deepvar_hierarchical/_network.py
+++ b/src/gluonts/model/deepvar_hierarchical/_network.py
@@ -295,7 +295,7 @@ class DeepVARHierarchicalNetwork(DeepVARNetwork):
             if self.assert_reconciliation:
                 assert (
                     reconciliation_error(self.A, samples=coherent_samples)
-                    < 1e-2
+                    < self.reconciliation_tol
                 )
 
             return coherent_samples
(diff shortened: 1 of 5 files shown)


src/gluonts/model/deepvar_hierarchical/_network.py
```diff
@@ -32,6 +33,106 @@
 )
 
 
+def reconcile_samples(
```
rshyamsundar, line 36: This function is just moved out of the network and `if` and `else` blocks are swapped for better readability.

rshyamsundar:
> Thank you very much!
> 
> Two minor comments in the review. Another question is whether the reconciliation code should still live into the model or whether it should be moved into another explicit module that just deals with it? We typically avoid having to import code from other models into another model. Just want to hear your opinion on this.
> 
> Otherwise, no concerns.

Yeah, that is one option. But it was suggested in another review to keep all the stuff related to one model inside the model itself. We could move this out if there is one more hierarchical model.

PR #1746, opened 2021-10-26
Adding ISQF distribution class

KelvinKan:
*Issue #, if available:*

*Description of changes:*


KelvinKan:
I incorporated your suggestions except for cumsum and test file. The code now passes all the checks. Thanks for reviewing the code again, @dcmaddix . When you have time, please also check the FIXMEs.

For the mxnet built-in cumsum, weirdly it seems that it doesn't support the option of axis=-1.

Adding the test scripts will be my next thing to do.

dcmaddix:
> I incorporated your suggestions except for cumsum and test file. The code now passes all the checks. Thanks for reviewing the code again, @dcmaddix . When you have time, please also check the FIXMEs.
> 
> For the mxnet built-in cumsum, weirdly it seems that it doesn't support the option of axis=-1.
> 
> Adding the test scripts will be my next thing to do.

Yes, it actually doesn't support `axis=-1`, so we hard coded `axis=3`.  You should actually know the shape since the tensor should have a fixed shape.

src/gluonts/mx/distribution/ISQF.py
```diff
@@ -0,0 +1,864 @@
...
+        alpha_plus = F.expand_dims(alpha_plus, axis=-1)
+
+        d = F.broadcast_add(F.broadcast_mul(x, (alpha_plus - alpha)), alpha)
+        delta_d = F.broadcast_mul(delta_x, (alpha_plus - alpha))
+
+        return d, delta_d
+
+    @staticmethod
+    def parametrize_tail(
+        F, beta: Tensor, alpha: Tensor, q: Tensor
```
dcmaddix, line 182: Add comment description on parameters, and same for below methods.

src/gluonts/mx/distribution/ISQF.py
```diff
@@ -0,0 +1,864 @@
...
+
+        # axis=-2 - passed at training when we evaluate quantiles at spline knots in order to compute alpha_tilde
+        # This is only for the quantile_spline function
+        # shape of input_alpha = (*batch_shape, num_qk-1, num_pieces)
+        # it will be expanded to (*batch_shape, num_qk-1, num_pieces, 1) to perform operation
+        # The shapes of spline and quantile knots parameters should be
+        # (*batch_shape, num_qk-1, 1, num_pieces) and (*batch_shape, num_qk-1, 1), respectively
+        # We expand axis=-2 and axis=-1 for spline and quantile knots parameters, respectively
+
+        if axis is not None:
```
dcmaddix, line 241: Can simplify to:` if not axis`

src/gluonts/mx/distribution/__init__.py
```diff
@@ -65,6 +65,10 @@
     PiecewiseLinearOutput,
     TransformedPiecewiseLinear,
 )
+from .ISQF import (
+    ISQF,
```
dcmaddix, line 69: Do you need to import this? It looks like only `ISQFOutput` is used.

dcmaddix (commented):
Looks good Kelvin! Thanks for incorporating the changes. I think the only additional is the unit tests for the interpolation/extrapolation methods similar to what we have for IQF https://github.com/awslabs/gluon-ts/pull/1699/files#diff-32fb84dbe82dcfe75aaeab4a49f1da2a93b853189454bf7c1a356fbbe2b94c84R246.

src/gluonts/mx/distribution/ISQF.py
```diff
@@ -0,0 +1,864 @@
...
+        beta_l = distr_args[3]
+        # can be deleted after reshape issue has been resolved
+        q = distr_args[2]
+        F = getF(beta_l)
+
+        # FIXME number 1
+        # Convert alpha from list of len=num_qk to Tensor of shape (*batch_shape, num_qk)
+        # For example, if alpha = [0.1, 0.5, 0.9],
+        # then alpha_reshape will be a Tensor of shape (*batch_shape, 3) with the last dimension being [0.1, 0.5, 0.9]
+        # In PyTorch, it would be torch.tensor(alpha).repeat(*batch_shape,1)
```
dcmaddix, line 814: I think you should be able to use `tile` here: https://mxnet.apache.org/versions/1.6/api/r/docs/api/mx.symbol.tile.html.

src/gluonts/mx/distribution/ISQF.py
```diff
@@ -0,0 +1,864 @@
...
+            dim=-1,
+        )
+
+        # FIXME number 2
+        # knots and heights have shape (*batch_shape, (num_qk-1)*num_pieces)
+        # I want to convert the shape to (*batch_shape, (num_qk-1), num_pieces)
+        # Here I make a shape_holder with the target shape, and use reshape_like
+
+        # create a shape holder of shape (*batch_shape, num_qk-1, num_pieces)
+        shape_holder = F.slice_axis(q, axis=-1, begin=0, end=-1)
```
dcmaddix, line 829: Do you have another array of the desired shape, like `heights`? Then you should be able to multiply the values of `q` by F.ones_like(height)`.

src/gluonts/mx/distribution/ISQF.py
KelvinKan: I put their descriptions in init. I will move them to the functions.

src/gluonts/mx/distribution/ISQF.py
```diff
@@ -0,0 +1,864 @@
...
+        delta_d = F.broadcast_mul(delta_x, (alpha_plus - alpha))
+
+        return d, delta_d
+
+    @staticmethod
+    def parametrize_tail(
+        F, beta: Tensor, alpha: Tensor, q: Tensor
+    ) -> Tuple[Tensor, Tensor]:
+        # Refer to the descriptions in init for details of parameters
+        a = 1 / beta
```
dcmaddix, line 185: more descriptive variable names

src/gluonts/mx/distribution/ISQF.py
dcmaddix: Combine the following 3 lines into 1

src/gluonts/mx/distribution/ISQF.py
```diff
@@ -0,0 +1,864 @@
...
+    alpha
+        the x-positions of quantile knots
+    tol
+        tolerance for numerical safeguarding
+    """
+    distr_cls: type = ISQF
+
+    @validated()
+    def __init__(
+        self, num_pieces: int, alpha: List[float], tol: float = 1e-4
```
dcmaddix, line 702: Update to type `Tensor`

KelvinKan:
I had to rename the script to lower-case letters. But it seems that Github has lost track of the rename, so all the comments became outdated, and I cannot respond to them. Therefore I will respond in this comment.

Update on the code:

1. Used more descriptive variable names: x-/y-positions for quantile knots: qk_x/qk_y, x-/y-positions for spline knots: sk_x/sk_y. Coefficient a/b for the tail: tail_a/tail_b
2. Changed function names as suggested
3. Combined the three lines of shape_holder
4. It seems that we cannot change the input type of __init__ from List[float] to Tensor, because mxnet does not allow applying symbol operation to an mxnet.nd.array when Hybridize=True
5. Added a pytest script to extensively test ISQF as a Distribution class

dcmaddix (approved):

src/gluonts/mx/distribution/isqf.py
```diff
@@ -0,0 +1,966 @@
...
+    def args(self) -> List:
+        return [
+            self.spline_knots,
+            self.spline_heights,
+            self.beta_l,
+            self.beta_r,
+            self.qk_y_all,
+        ]
+
+    @staticmethod
```
jaheba, line 133: I'm not a fan of these staticmethods. Couldn't we just declare these as functions?

src/gluonts/mx/distribution/isqf.py
```diff
@@ -0,0 +1,966 @@
...
+        Returns
+        -------
+        Tensor
+            Quantiles tensor, of the same shape as alpha
+        """
+
+        F = self.F
+        qk_x, qk_x_l, qk_x_plus = self.qk_x, self.qk_x_l, self.qk_x_plus
+
+        # The following describes the parameters reshaping in quantile_internal, quantile_spline and quantile_tail
```
jaheba, line 279: Can you ensure to also have these lines at a max width of 79 chars?

src/gluonts/mx/distribution/isqf.py
```diff
@@ -0,0 +1,966 @@
...
+
+        quantile = F.where(
+            alpha < qk_x_l,
+            self.quantile_tail(alpha, axis=axis, left_tail=True),
+            self.quantile_tail(alpha, axis=axis, left_tail=False),
+        )
+
+        spline_val = self.quantile_spline(alpha, axis=axis)
+
+        for i in range(self.num_qk - 1):
```
jaheba, line 318: There certainly is a better name than `i` for this.

src/gluonts/mx/distribution/isqf.py
```diff
@@ -0,0 +1,966 @@
...
+        assert (
+            isinstance(num_pieces, int) and num_pieces > 0
+        ), "num_pieces should be an integer and greater than 0"
+
+        self.num_pieces = num_pieces
+        self.qk_x = sorted(qk_x)
+        self.num_qk = len(qk_x)
+        self.tol = tol
+        self.args_dim = cast(
+            Dict[str, int],
```
jaheba, line 811: Not sure cast is necessary here. You could probably also just do `self.args_dim: Dict[str, int] = ...`

jaheba (approved):
Thanks!

Issue #1657, opened 2021-08-13
Support unknown future features


fernandocamargoai:
## Description

I've been using gluon-ts for quite some time now and got used to somehow trying to pass all the future features during inference. However, I'm dealing with a problem where I have a very important feature (% of the time the product was available for selling during the day) that I don't know the future value. At first, I was simply setting it to 100% for all future days. Then, I tested training another model to forecast this feature and use it as input and the results improved a lot. However, having two models increases my work by two. And if I had others that I needed to forecast, it would be even worse.

For that reason, I started looking around and found out that [PyTorch Forecasting](https://pytorch-forecasting.readthedocs.io/en/latest/) actually allows us to tell the features that are known in the future and the ones that are unknown (`time_varying_known_reals` vs `time_varying_unknown_reals`). I was curious to understand how it works and read their DeepAR code. It seems that all the features are passed to the encoder, but only the known ones are passed to the decoder. I guess this way the model learns to forecast without those unknown future values, probably increasing its uncertainty.

I'm planning to give that library a try and see if the results are better. But I hope we have this feature here as well since I have a lot of code using gluon-ts.

mbohlkeschneider:
Hi @fernandocamargoai, 

thank you for raising this issue. You are correct - this is not implemented in our DeepAR implementation. However, you could try MQCNN/MQRNN, which does implement this feature: https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/model/seq2seq/_mq_dnn_estimator.py#L122. 

Maybe that works for you, we would appreciate your feedback on this.

fernandocamargoai:
Hey, @mbohlkeschneider. Thanks for your answer. It's great to know that there's a model that supports that. However, from what I understood by looking at the code, it uses the real dynamic features in the encoder, and doesn't it in the decoder, right? What about features that I actually know their future values (holidays, for example)? [PyTorch Forecasting](https://pytorch-forecasting.readthedocs.io/en/latest/) allows me to point out which features are known and which are unknown in the future.

mbohlkeschneider:
Actually, MQCNN/RNN does support both `use_feat_dynamic_real` (known values in the future) and `use_past_feat_dynamic_real` (values are not known in the future). So both should be possible using this method.

PR #1757, opened 2021-11-05
Future features

mbohlkeschneider:
*Issue #, if available:*

Addresses  issue #1657. 

*Description of changes:*

Modified `DeepAR` to take optional `past_feat_dynamic_real` features that are only known in training range. This required some changes in the `InstanceSplitter` to accommodate that feature as a separate field. Additionally, some changes in the `ForecastGenerator` and `SelectFields` transformation where required to support optional fields. 

### DIFF
--- a/src/gluonts/model/deepar/_network.py
+++ b/src/gluonts/model/deepar/_network.py
@@ -944,7 +944,7 @@ class DeepARTrainingNetwork(DeepARNetwork):
         past_time_feat: Tensor,
         past_target: Tensor,
         past_observed_values: Tensor,
-        past_is_pad: Optional,
+        past_is_pad: Tensor,
         future_time_feat: Tensor,
         future_target: Tensor,
         future_observed_values: Tensor,
--- a/src/gluonts/model/deepar/_network.py
+++ b/src/gluonts/model/deepar/_network.py
@@ -55,6 +55,7 @@ class DeepARNetwork(mx.gluon.HybridBlock):
         num_imputation_samples: int = 1,
         minimum_scale: float = 1e-10,
         impute_missing_values: bool = False,
+        use_past_feat_dynamic_real: bool = False,
         default_scale: Optional[float] = None,
         **kwargs,
     ) -> None:
@@ -72,6 +73,7 @@ class DeepARNetwork(mx.gluon.HybridBlock):
         self.num_cat = len(cardinality)
         self.scaling = scaling
         self.dtype = dtype
+        self.impute_missing_values = impute_missing_values
 
         assert len(cardinality) == len(
             embedding_dimension
@@ -85,9 +87,10 @@ class DeepARNetwork(mx.gluon.HybridBlock):
         self.lags_seq = lags_seq
 
         self.distr_output = distr_output
-        RnnCell = {"lstm": mx.gluon.rnn.LSTMCell, "gru": mx.gluon.rnn.GRUCell}[
-            self.cell_type
-        ]
+        self.RnnCell = {
+            "lstm": mx.gluon.rnn.LSTMCell,
+            "gru": mx.gluon.rnn.GRUCell,
+        }[self.cell_type]
 
         self.target_shape = distr_output.event_shape
 
@@ -96,7 +99,7 @@ class DeepARNetwork(mx.gluon.HybridBlock):
             len(self.target_shape) <= 1
         ), "Argument `target_shape` should be a tuple with 1 element at most"
 
-        Dropout = {
+        self.Dropout = {
             "ZoneoutCell": ZoneoutCell,
             "RNNZoneoutCell": RNNZoneoutCell,
             "VariationalDropoutCell": VariationalDropoutCell,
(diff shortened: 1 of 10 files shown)


dcmaddix:
Nice, thanks @mbohlkeschneider! @lovvge Michael has extended DeepAR to also support past dynamic features, similar to MQ-CNN.

src/gluonts/model/deepar/_network.py
```diff
@@ -96,34 +99,22 @@ def __init__(
...
-                    )
-                elif "Dropout" in self.dropoutcell_type:
-                    cell = (
-                        Dropout(cell, drop_states=dropout_rate)
-                        if dropout_rate > 0.0
-                        else cell
-                    )
-                self.rnn.add(cell)
-            self.rnn.cast(dtype=dtype)
+            self.rnn = self.rnn_block(
```
rshyamsundar, line 110: Maybe rename this as `self.encoder` for better readability? Then `self.decoder` is either a reference to or an entirely different block depending on past only features.

src/gluonts/model/deepar/_network.py
```diff
@@ -877,10 +944,11 @@ def hybrid_forward(
         past_time_feat: Tensor,
         past_target: Tensor,
         past_observed_values: Tensor,
-        past_is_pad: Optional[Tensor],
+        past_is_pad: Tensor,
         future_time_feat: Tensor,
         future_target: Tensor,
         future_observed_values: Tensor,
+        past_past_feat_dynamic_real: Optional[Tensor] = None,
```
rshyamsundar, line 951: past two times? You mean `past_only_feat_dynamic_real` or `past_only_time_feat`? Also docstring needs to be updated for this new argument.

src/gluonts/model/deepar/_network.py
```diff
@@ -1140,6 +1209,8 @@ def hybrid_forward(
         Tensor
             Predicted samples
         """
+        if past_past_feat_dynamic_real is not None:
+            past_past_feat_dynamic_real.mean()
```
rshyamsundar, line 1213: Why is this required?

src/gluonts/model/forecast_generator.py
```diff
@@ -170,7 +170,7 @@ def __call__(
         **kwargs
     ) -> Iterator[Forecast]:
         for batch in inference_data_loader:
-            inputs = [batch[k] for k in input_names]
+            inputs = [batch[k] for k in input_names if k in batch.keys()]
```
rshyamsundar, line 173: `input_names` is completely ignored then? Is there a way to update it directly by modifying `get_hybrid_forward_input_names`? I think it is hard to determine this difference (with/out past only features) on-the-fly during runtime?

src/gluonts/transform/field.py
```diff
@@ -127,4 +127,4 @@ def __init__(self, input_fields: List[str]) -> None:
         self.input_fields = input_fields
 
     def map_transform(self, data: DataEntry, is_train: bool) -> DataEntry:
-        return {f: data[f] for f in self.input_fields}
+        return {f: data[f] for f in self.input_fields if f in data.keys()}
```
rshyamsundar, line 130: maybe this results in silent errors in other cases (ignoring required fields simply because they are not in the data)?

src/gluonts/model/deepar/_network.py
mbohlkeschneider: Thank you. Yes, I struggled with the naming a bit. In the `canonical` DeepAR, `encoder` would sound weird (although I think this would still be fine).

src/gluonts/model/deepar/_network.py
mbohlkeschneider: Yes this is weird as well. My goal was to use the build-in feature naming from `FieldName` for the `hybrid_forward` call: https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/dataset/field_names.py#L30, which would be `past_feat_dynamic_real`. This then gets turned to `past_past_feat_dynamic_real` in the `InstanceSplitter`https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/transform/split.py#L179. 

One way to circumvent is to have a different channel in the `InstanceSplitter` that does not do the field name modification to that channel (past only time series channel)? 

Do you have other suggestions?

src/gluonts/model/deepar/_network.py
mbohlkeschneider: Thanks for spotting this. Will remove it, this was to test whether the data gets actually passed.

src/gluonts/transform/field.py
mbohlkeschneider: Good point. I don't think the error would be silent in this case (at least mxnet models will complain for sure!). This change is required to have optional fields in the network forward passes, though. I think the fundamental question that we have to ask: 

Do we want to have optional inputs for the networks or not?

DushyantSahoo:
Is there a timeline for this feature?

PR #1767, opened 2021-11-12
Use of independent random seed.

jaheba:
*Issue #, if available:*

*Description of changes:*


conftest.py
```diff
@@ -85,7 +85,7 @@ def pytest_configure():
 
     module_seed_str = os.getenv("MXNET_MODULE_SEED")
     if module_seed_str is None:
-        seed = np.random.randint(0, np.iinfo(np.int32).max)
+        seed = int.from_bytes(os.urandom(4), "big")
```
melopeo, line 88: If the goal is to have an independent random number generator, another option would be to have its own stream:
```
random_generator = np.random.RandomState()
random_generator.randint(0, np.iinfo(np.int32).max)
```

mbohlkeschneider (approved):

PR #1778, opened 2021-11-19
New metrics API.

jaheba:
This is a proof of concept for a new metrics API.

The approach splits metric calculation into two steps: single value `loss` and aggregation of losses. Further, metrics can depend on other metrics.

For example:

```py
# Maybe we should call this Error?
class Loss(Stat):
    name: str = "loss"

    def apply(self, x):
        x.loss = x.target - x.forecast


class MSE(Metric):
    name: str = "mse"
    dependencies = [Loss()]

    def loss(self, x):
        return np.square(x.loss)

    def aggregate(self, loss, axis=None):
        return np.mean(loss, axis=axis)

```

To get the `MSE` on data, we need to evaluate `Loss.loos` first, then can run `MSE.loss`.

The idea is to have simple metric-definitions, but more magic in how these classes are applied to data.

---


src/gluonts/evaluation/new_metrics.py
```diff
@@ -0,0 +1,230 @@
...
+        return self.aggregate(metrics[self.name], axis=axis)
+
+    def loss(self, x):
+        raise NotImplementedError
+
+    def aggregate(self, metrics, axis=None):
+        raise NotImplementedError
+
+
+class DerivedMetric:
```
melopeo, line 47: It seems that the term `DerivedMetric` corresponds to metrics that are available only for the aggregated cases, and not for item-wise cases. Is this interpretation correct?

src/gluonts/evaluation/new_metrics.py
```diff
@@ -0,0 +1,230 @@
...
+        return list(map(WeightedQuantileLoss, self.quantiles))
+
+    def get(self, metrics):
+        return np.mean(
+            [metrics[quantile.name] for quantile in self.dependencies],
+            axis=0,
+        )
+
+
+def merge(a, b):
```
melopeo, line 130: It seems that the functions `merge`, `resolve`, and `ger_order` are necessary to keep metric dependencies in a consistent way. Can we somehow group this functions so that they are clearly standing out on their purpose? Otherwise they might look a bit like new metrics.

src/gluonts/evaluation/new_metrics.py
```diff
@@ -0,0 +1,230 @@
...
+            [1, 2, 3, 4, 5],
+        ]
+    ),
+}
+
+
+get_losses, aggregate, select = get_evaluator([MeanWeightedQuantileLoss()])
+
+value_metrics = get_losses(AttrDict(data))
+item_metrics = aggregate(value_metrics, axis=1)
```
melopeo, line 227: Since the aggregation of data seems to be `axis` dependant, can we include the possibility of computing metrics per _time stamp_? For instance, if we have data in daily frequency, can we compute the metrics per day?

jaheba: Yes, you need to align the data and select the axis accordingly.

kashif:
BTW @melopeo another nice thing to have would be in the evaluator to also get metrics per time-step and per-time-step per time series entity... at the moment the evaluator averages over the timesteps...

melopeo:
@kashif ,
thanks for proposing this functionality. Part of our roadmap on this initiative is indeed to provide metrics per timestamp as you have suggested. This will allow us to compute the standard metrics we have nowadays (item and aggregated metrics), together with their counterpart per time stamp. I think this is interesting as there is a natural interest in observing how this metrics evolve over time.

Please let us know if you have more ideas on how we can improve this initiative.

Issue #1725, opened 2021-10-08
MXNetError when training Deep Renewal Process model


dat181197:
## Description
Hi team, thank you for a great library
I am working a project for demand forecasting on a custom dataset. I have managed to create a custom GluonTS dataset and successfully ran the DeepAR model on it. But when I tried to run the Deep Renewal Process model, I received the following error after one epoch of training. I have been trying to figure out the error, but have not found anything. 

Also I think there has been a change to the GluonTS framework that removed the `decode_fn` of the validation_dataloader in[ #1408](https://github.com/awslabs/gluon-ts/pull/1408), but in the [create_validation_data_loader function](https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/model/renewal/_estimator.py#L242) of the model I still see "decode_fn" being returned which lead to an error when trying to use `model.train(train_ds, test_ds)`

https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/model/renewal/_estimator.py

```python
Starting training...
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100/100 [00:02<00:00, 44.01it/s, epoch=1/30, avg_epoch_loss=0.771]
 65%|███████████████████████████████████████████████████████████████████████████████████▏                                            | 65/100 [00:01<00:00, 41.55it/s, epoch=2/30, avg_epoch_loss=0.341]
Traceback (most recent call last):
  File "src/train_DRP.py", line 82, in <module>
    train(cfg)
  File "src/train_DRP.py", line 54, in train
    predictor = model.train(train_ds)
  File "/data10TB/datduong_research/PlayMaker/playmaker-model-pipeline/venv/lib/python3.8/site-packages/gluonts/mx/model/estimator.py", line 195, in train
    return self.train_model(
  File "/data10TB/datduong_research/PlayMaker/playmaker-model-pipeline/venv/lib/python3.8/site-packages/gluonts/mx/model/estimator.py", line 170, in train_model
    self.trainer(
  File "/data10TB/datduong_research/PlayMaker/playmaker-model-pipeline/venv/lib/python3.8/site-packages/gluonts/mx/trainer/_base.py", line 436, in __call__
    epoch_loss = loop(
  File "/data10TB/datduong_research/PlayMaker/playmaker-model-pipeline/venv/lib/python3.8/site-packages/gluonts/mx/trainer/_base.py", line 356, in loop
    output = net(*batch.values())
[...]

PR #1779, opened 2021-11-19
fix for #1725, bugs in deep renewal process

canerturkmen:
… zero batches

*Issue #, if available:* #1725

*Description of changes:*

Fixes multiple issues in the deep renewal process: (i) breaking changes by the changes in Data Loader, (ii) all-zero batches during training, and (iii) an edge case in the predictor that caused negative indices.


mbohlkeschneider (approved):
Looks good. Thanks!

PR #1780, opened 2021-11-19
adding MQF2 (Multi-horizon)

KelvinKan:
This is a PR for my project Multivariate Quantile Function Forecaster (MQF2). This is the multi-horizon version (multivariate in time-steps).

Main changes:

- Added MQF2 Estimator, Module and Lightning Module, which are based on and are subclasses of the DeepAR model
- Made a small modification to unroll_lagged_rnn of DeepARModel. (so that it also outputs the hidden states)
- Added icnn_utils.py which contains some functions of input convex neural network (icnn) based on the GitHub repo CP-Flow (https://github.com/CW-Huang/CP-Flow)
- Added the installation command of CP-Flow (The repo itself is not a package, but I made a fork and turned it into a package. I have also submitted a PR to the repo. Hopefully they will incorporate it.)


.github/workflows/tests.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         pip install -e ".[shell]"
         pip install -r requirements/requirements-test.txt
         pip install -r requirements/requirements-rotbaum.txt
-        pip install -r requirements/requirements-extras-m-competitions.txt
```
jaheba, line 37: Why remove the `m-competition` requirement?

KelvinKan: It was a mistake when I resolve the conflicts in branch updating. I will correct it in the next commit.

.github/workflows/tests-torch.yml
```diff
@@ -26,6 +26,7 @@ jobs:
         pip install -r requirements/requirements-test.txt
         pip install torch==1.10.0+cpu --no-cache-dir -f https://download.pytorch.org/whl/cpu/torch_stable.html
         pip install -r requirements/requirements-pytorch.txt
+        pip install -r requirements/requirements-extras-cpflows.txt
```
mbohlkeschneider, line 29: The requirements for tests have been revamped a bit. Maybe we can only have this for the specific model as a requirement?

src/gluonts/torch/distributions/mqf2.py
```diff
@@ -0,0 +1,395 @@
...
+        super().__init__(
+            batch_shape=self.batch_shape, validate_args=validate_args
+        )
+
+        self.context_length = (
+            self.hidden_state.shape[-2]
+            if len(self.hidden_state.shape) > 2
+            else 1
+        )
+        self.numel_batch = MQF2Distribution.get_numel(self.batch_shape)
```
mbohlkeschneider, line 92: What is a `numel`?

src/gluonts/torch/distributions/mqf2.py
```diff
@@ -0,0 +1,395 @@
...
+
+        zero = torch.tensor(
+            0, dtype=hidden_state.dtype, device=hidden_state.device
+        )
+        one = torch.ones_like(zero)
+        self.standard_normal = Normal(zero, one)
```
mbohlkeschneider, line 98: Maybe call it `mu` and `sigma`?

src/gluonts/torch/distributions/mqf2.py
```diff
@@ -0,0 +1,395 @@
...
+
+    def loss(self, z: torch.Tensor) -> torch.Tensor:
```
mbohlkeschneider, line 100: Alternatively, one could overload the loss function once (seems cleaner since `is_energy_score` never changes in a model, right?).

src/gluonts/torch/distributions/mqf2.py
```diff
@@ -0,0 +1,395 @@
...
+        if self.is_energy_score:
+            return self.energy_score(z)
+        else:
+            return -self.log_prob(z)
+
+    def unfold_observations(self, z: torch.Tensor) -> torch.Tensor:
```
mbohlkeschneider, line 106: In the end of the day, it is more of a `reshape` than `unfold`. Maybe adapt the name?

src/gluonts/torch/distributions/mqf2.py
```diff
@@ -0,0 +1,395 @@
...
+        z
+            A batch of time series with shape (batch_size, context_length + prediciton_length - 1)
+
+        Returns
+        -------
+        loss
+            Tesnor of shape (batch_size * context_length,)
+        """
+
+        hidden_state = self.hidden_state
```
mbohlkeschneider, line 145: Why do we need local variables here? Can't we use the class variables directly in the function call?

src/gluonts/torch/distributions/mqf2.py
```diff
@@ -0,0 +1,395 @@
...
+        sample_shape
+            Shape of the samples
+
+        Returns
+        -------
+        sample_paths
+            Tesnor of shape (batch_size, *sample_shape, prediction_length)
+        """
+
+        numel_batch = self.numel_batch
```
mbohlkeschneider, line 207: Same comment as above. Why local variabls here?

src/gluonts/torch/distributions/mqf2.py
```diff
@@ -0,0 +1,395 @@
...
+            predicted paths of shape = (batch_shape, prediction_length)
+        """
+
+        picnn = self.picnn
+        if hidden_state is None:
+            hidden_state = self.hidden_state
+
+        normal_quantile = self.standard_normal.icdf(alpha)
+
+        if self.is_energy_score:
```
mbohlkeschneider, line 255: I think some explanation would be useful here.

src/gluonts/torch/distributions/mqf2.py
```diff
@@ -0,0 +1,395 @@
...
+            return TransformedMQF2Distribution(
+                distr, [AffineTransform(loc=loc, scale=scale)]
+            )
+
+    @property
+    def event_shape(self) -> Tuple:
+        return ()
+
+
+class TransformedMQF2Distribution(TransformedDistribution):
```
mbohlkeschneider, line 345: Do the rest of the `TransformedDistribution` functionality actually work here? If no, maybe it is cleaner to NOT inherent from it?

src/gluonts/torch/model/mqf2/lightning_module.py
```diff
@@ -0,0 +1,107 @@
...
+        batch
+            Dictionary containing the (past and future) features and target values in a batch
+
+        Returns
+        -------
+        loss
+            Mean of the loss values
+        """
+
+        feat_static_cat = batch["feat_static_cat"]
```
mbohlkeschneider, line 77: Can we use the standard `FieldNames` here?

src/gluonts/torch/model/mqf2/module.py
```diff
@@ -0,0 +1,254 @@
...
+            past_observed_values=past_observed_values,
+            future_time_feat=future_time_feat,
+            future_target=future_target,
+        )
+
+        hidden_state = hidden_state[:, : self.context_length]
+
+        return hidden_state, scale
+
+    @torch.jit.ignore
```
mbohlkeschneider, line 164: Why?

mbohlkeschneider (dismissed):
Thank  you for this PR. There is a lot of stuff here. I suggest one of the co-authors maybe have an additional review to check correctness. 

One thing I noticed is that the `MQF2DistributionOutput` is not tested separately. Since this is one of the main contributions here, it might make sense to test it (add to the distribution tests) to make sure it is correct?

src/gluonts/torch/distributions/mqf2.py
KelvinKan: Because the GluonTS [DistributionLoss](https://github.com/awslabs/gluon-ts/blob/7e153507beade7d5a9de12375331d752d6d81614/src/gluonts/torch/modules/loss.py#L48-L52) computes negative loss by calling `-input.log_prob`, to keep the consistency, I also define `log_prob` as the positive log likelihood and call `log_prob` in `loss`. 

Since method assignment doesn't support negative sign, i.e. `self.loss=-input.log_prob`, we can either keep it as is or define a `neg_log_prob` and overload the loss. I lean toward the current implementation.

.github/workflows/tests-torch.yml
KelvinKan: I will double check to make sure this is the optimal way to install test requirements after the recent changes.

src/gluonts/torch/distributions/mqf2.py
KelvinKan: number of elements. This abbreviation is also used in PyTorch: https://pytorch.org/docs/stable/generated/torch.numel.html

If you think we should go with a more descriptive name, we can do `num_element`?

src/gluonts/torch/distributions/mqf2.py
KelvinKan: Will do

src/gluonts/torch/distributions/mqf2.py
KelvinKan: It's actually both. It is actually sliding through the tensor and stack the sliding view together. I will rename it as stack_sliding_view so that it is more descriptive.

src/gluonts/torch/distributions/mqf2.py
KelvinKan: It is my practice to prevent myself from messing up the class variables. Would it be fine to keep it since you mentioned you don't have strong opinions on this?

src/gluonts/torch/distributions/mqf2.py
KelvinKan: Yes I will add an explanation

src/gluonts/torch/distributions/mqf2.py
KelvinKan: We do not use the functionality of `TransformedDistribution`. The inheritance is mainly for using the `__init__` of `TransformedDistribution` to do sanity checks.

src/gluonts/torch/model/mqf2/lightning_module.py
KelvinKan: Will do.

src/gluonts/torch/model/mqf2/module.py
KelvinKan: I am actually not sure about this. The torch.DeepAR code has this so I followed.

.github/workflows/tests.yml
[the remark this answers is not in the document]
KelvinKan: It was a mistake when I was resolving conflicts. It has been fixed.

src/gluonts/torch/distributions/mqf2.py
```diff
@@ -0,0 +1,415 @@
...
+
+import torch
+
+from gluonts.core.component import validated
+from gluonts.torch.modules.distribution_output import (
+    Distribution,
+    DistributionOutput,
+)
+
+from torch.distributions import (
```
benidis, line 24: Minor but maybe move the torch dependencies all together before the gluonts ones?

src/gluonts/torch/model/mqf2/lightning_module.py
```diff
@@ -0,0 +1,110 @@
...
+        feat_static_real = batch["feat_static_real"]
+        past_time_feat = batch["past_time_feat"]
+        past_target = batch["past_target"]
+        future_time_feat = batch["future_time_feat"]
+        future_target = batch["future_target"]
+        past_observed_values = batch["past_observed_values"]
```
benidis, line 86: Why don't you use the batch directly in the `unroll_lagged_rnn`? No need to assign everything into new variables.

src/gluonts/torch/model/mqf2/module.py
```diff
@@ -0,0 +1,258 @@
...
+
+        picnn = self.picnn
+
+        hidden_state, scale = self.unroll_lagged_rnn(
+            feat_static_cat,
+            feat_static_real,
+            past_time_feat,
+            past_target,
+            past_observed_values,
+            future_time_feat[:, :1],
```
benidis, line 249: As discussed, this can be improved. We can include all future time features since they are known. Now we are losing information.

src/gluonts/torch/model/mqf2/module.py
```diff
@@ -0,0 +1,258 @@
...
+        Returns
+        -------
+        sample_paths
+            Sample paths (batch_size, num_parallel_samples, prediction_length)
+        """
+
+        if num_parallel_samples is None:
+            num_parallel_samples = self.num_parallel_samples
+
+        picnn = self.picnn
```
benidis, line 241: This is not needed, following the previous comments. I think the two objects will point to the same memory slot so if you don't do a deep copy you are not gaining anything eitherway (you can fact check me on that).

src/gluonts/torch/model/mqf2/module.py
```diff
@@ -0,0 +1,258 @@
...
+        num_parallel_samples
+            Number of parallel sample paths generated for each time series
+
+        Returns
+        -------
+        sample_paths
+            Sample paths (batch_size, num_parallel_samples, prediction_length)
+        """
+
+        if num_parallel_samples is None:
```
benidis, line 238: Why are you passing in the `forward` a parameter that you define in the constructor? If it is something that changes then maybe you can remove this from the constructor entirely and add a default value in the `forward`? Just a suggestion.

benidis: I see that it is needed also in `super`. Is it fine if you pass another value here (shape issues)?

src/gluonts/torch/model/deepar/module.py
```diff
@@ -151,7 +152,7 @@ def unroll_lagged_rnn(
         output, new_state = self.lagged_rnn(prior_input, input, features)
 
         params = self.param_proj(output)
-        return params, scale, static_feat, new_state
+        return params, scale, output, static_feat, new_state
```
lostella, line 155: I think this change suggests that this method is trying to do too much. But we can re-think this after this PR.

src/gluonts/torch/model/mqf2/estimator.py
```diff
@@ -0,0 +1,206 @@
...
+from gluonts.torch.modules.loss import NegativeLogLikelihood, EnergyScore
+from gluonts.torch.distributions.mqf2 import MQF2DistributionOutput
+
+from . import MQF2MultiHorizonLightningModule, MQF2MultiHorizonModel
+
+from gluonts.core.component import validated
+from gluonts.time_feature import TimeFeature
+
+
+class MQF2MultiHorizonEstimator(DeepAREstimator):
```
lostella, line 26: Why subclassing here?

src/gluonts/torch/model/mqf2/module.py
```diff
@@ -0,0 +1,259 @@
...
+            past_time_feat=past_time_feat,
+            past_target=past_target,
+            past_observed_values=past_observed_values,
+            future_time_feat=future_time_feat,
+            future_target=future_target,
+        )
+
+        hidden_state = hidden_state[:, : self.context_length]
+
+        return hidden_state, scale
```
lostella, line 163: It appears like this method is doing very little on top of the one defined in the `super` class (basically just slicing the returned `hidden_state`). It's also changing the signature quite a bit wrt to the `super` method, by reducing the amount of stuff returned.

In my opinion, it would be better not to define this method at all, and instead invoke directly the one from `super` followed by slicing.

src/gluonts/torch/model/mqf2/estimator.py
KelvinKan: My original intention is to reduce duplication and have concise code.

Your suggestion certainly makes sense. We might just copy the code that we need from DeepAR.

src/gluonts/torch/model/mqf2/module.py
KelvinKan: Yes I should return the all the outputs as in the superclass. I will make changes about that. 

Basically this function unrolls the rnn but not all the way until the end of the time series. So I do the slicing to chop off the things I don't need here. I think that it would be easier for the users to not slice the hidden state every time they call the function.

lostella: Isn't it the case that by just not providing the "future" data (`future_time_feat` and `future_target`) you already achieve this? I.e., you only get out the first `context_length` hidden state as output?

I'm not 100% sure, but you may want to check that, to save even more code

lostella (approved):
We can merge this and address some of the comments later if needed

src/gluonts/torch/model/mqf2/module.py
```diff
@@ -0,0 +1,259 @@
...
+    def __init__(
+        self,
+        freq: str,
+        context_length: int,
+        prediction_length: int,
+        num_feat_dynamic_real: int,
+        num_feat_static_real: int,
+        num_feat_static_cat: int,
+        cardinality: List[int],
+        distr_output: DistributionOutput,
```
lostella, line 38: It looks like this *need* to be to a `MQF2DistributionOutput`, right? If so, then there's no need to have this input at construction time, and can just be constructed and passed to the `super` constructor I think

src/gluonts/torch/model/mqf2/module.py
```diff
@@ -0,0 +1,259 @@
...
+            feat_static_cat,
+            feat_static_real,
+            past_time_feat,
+            past_target,
+            past_observed_values,
+            future_time_feat[:, :1],
+        )
+
+        distr = self.output_distribution(
+            self.picnn, hidden_state, inference=True
```
lostella, line 252: Why passing `self.picnn`? `self.output_distribution` can access it directly from `self`

PR #1784, opened 2021-11-21
Add IQN implementation

### LINKED ISSUE #1643: Add IQN probabilistic head to the repo
## Description
After a discussion I had with @jgasthaus and @kashif at the ICML Time Series workshop, I would like to add the code we used for our IQN-RNN model. It is a general probabilistic method, using Implicit Quantile Network as a distribution output. On our preliminary results, it yields competitive KPIs against SQF-RNN. It would be great if you could evaluate it on your side, once it is merged. The code already exists as a part of `pytorch-ts` and is `torch` based. It works with `gluonts` backend already.

## Reference
- Paper: https://arxiv.org/pdf/2107.03743.pdf
- Pytorch-ts demo notebook: https://github.com/zalandoresearch/pytorch-ts/blob/master/examples/Implicit-Quantile-Network-Example.ipynb

kashif:
*Issue #, if available:*

*Description of changes:*

Implemented IQN distribution output head  from the paper https://arxiv.org/abs/2107.03743

fixes #1643


kashif:
@lostella or @jaheba whenever you get a chance, could you kindly review...

src/gluonts/torch/distributions/implicit_quantile_network.py
```diff
@@ -0,0 +1,158 @@
...
+        return self.outputs
+
+    def quantile_loss(self, value):
+        return torch.abs(
+            (self.outputs - value)
+            * ((value <= self.outputs).float() - self.taus)
+        )
+
+
+class ImplicitQuantileNetworkOutput(DistributionOutput):
```
lostella, line 141: Would it be better to keep the name of the two classes aligned to the `X` vs `XOutput` convention? (One has `Network` the other doesn't)

test/torch/model/test_simple_torch_model.py
```diff
@@ -0,0 +1,228 @@
...
+    TestSplitSampler,
+    InstanceSplitter,
+)
+
+
+def mean_abs_scaling(seq, min_scale=1e-5):
+    return seq.abs().mean(1).clamp(min_scale, None).unsqueeze(1)
+
+
+class FeedForwardNetwork(nn.Module):
```
lostella: Did you mean to include this test script? I think it was there and then removed

src/gluonts/torch/distributions/implicit_quantile_network.py
```diff
@@ -0,0 +1,158 @@
...
+
+    @torch.no_grad()
+    def sample(self, sample_shape=torch.Size()):
+        return self.outputs
+
+    def quantile_loss(self, value):
+        return torch.abs(
+            (self.outputs - value)
+            * ((value <= self.outputs).float() - self.taus)
+        )
```
lostella: What is the return type here? It looks like `taus` is a tensor, so the returned value would also be a tensor.

Also I'm not sure whether `quantile_loss` is a good name here, if this is supposed to be the sum of multiple quantile losses (for different quantile levels). In that case, this would be an approximation of CRPS?

lostella (changes requested):
Thanks @kashif! As a general comment, I think it would be great if you could add type annotations

src/gluonts/torch/distributions/implicit_quantile_network.py
kashif: yes so this is actually the quantile loss here where crps would be the sum over quantile losses which is different I believe

src/gluonts/torch/distributions/implicit_quantile_network.py
```diff
@@ -0,0 +1,211 @@
...
+
+    @torch.no_grad()
+    def sample(self, sample_shape=torch.Size()):
+        return self.outputs
+
+    def quantile_loss(self, value) -> torch.Tensor:
+        return torch.abs(
+            (self.outputs - value)
+            * ((value <= self.outputs).float() - self.taus)
+        )
```
lostella, line 138: Here the `abs` is not needed: when the residual is negative, the coefficient is `-taus`, making the result positive

kashif: ah yes let me fix it! thanks!

lostella: I think the most intuitive way to write this is 

```python
return (self.taus - (self.outputs < value)) * (self.outputs - value)
```

which reads “penalize by tau for over-predicting, and by 1-tau for under-predicting”

src/gluonts/torch/distributions/implicit_quantile_network.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def forward(self, tau):  # tau: [B, T]
+        cos_emb_tau = torch.cos(tau.unsqueeze(-1) * self.integers * torch.pi)
+        return self.output_layer(cos_emb_tau)
+
+
+class ImplicitQuantileModule(nn.Module):
+    r"""
+    Implicit Quantile Network from the paper ``IQN for Distributional
+     Reinforcement Learning`` (https://arxiv.org/abs/1806.06923) by
```
lostella: Wooops, mind the initial space

src/gluonts/torch/distributions/implicit_quantile_network.py
```diff
@@ -0,0 +1,211 @@
...
+        concentration0: float = 1.0,
+        cos_embedding_dim: int = 64,
+    ) -> None:
+        super().__init__()
+
+        self.concentration1 = concentration1
+        self.concentration0 = concentration0
+        self.cos_embedding_dim = cos_embedding_dim
+
+        if output_domain in ["Positive", "Unit"]:
```
lostella: Can we use lowercase strings for these encodings?

lostella (changes requested):
@kashif looks good to me, I would ask just for a few minor changes (see also the comment about pinball loss)

src/gluonts/torch/distributions/implicit_quantile_network.py
```diff
@@ -129,9 +129,10 @@ def sample(self, sample_shape=torch.Size()):
...
 
     def quantile_loss(self, value) -> torch.Tensor:
-        return torch.abs(
-            (self.outputs - value)
-            * ((value <= self.outputs).float() - self.taus)
+        # penalize by tau for over-predicting
+        # and by 1-tau for under-predicting
+        return (self.taus - (self.outputs < value).float()) * (
+            self.outputs - value
         )
```
lostella: Sorry, got confused: it should be

```
return (self.taus - (value < self.outputs).float()) * (value - self.outputs)
```

Proof:

```python
import torch
import numpy as np

def pinball_loss(obs, pred, alpha):
    return (alpha - (obs < pred).float()) * (obs - pred)

def subgradient_method(f, p, maxit):
    for k in range(maxit):
        v = f(p)
        v.backward()
        with torch.no_grad():
            p -= 1/(k+1) * p.grad
            p.grad.zero_()
    return p

data = np.random.normal(size=(1000,))
f = lambda q: pinball_loss(torch.from_numpy(data), q, 0.1).mean()
q = subgradient_method(f, torch.tensor(0.0, requires_grad=True), 1000)
print(q)
```

kashif:
awesome thanks @lostella  fixed

src/gluonts/torch/distributions/implicit_quantile_network.py
```diff
@@ -125,13 +127,14 @@ def __init__(self, outputs, taus, validate_args=None):
...
+    def sample(self, sample_shape=torch.Size()) -> torch.Tensor:
         return self.outputs
 
-    def quantile_loss(self, value) -> torch.Tensor:
-        return torch.abs(
-            (self.outputs - value)
-            * ((value <= self.outputs).float() - self.taus)
+    def quantile_loss(self, value: torch.Tensor) -> torch.Tensor:
+        # penalize by tau for over-predicting
+        # and by 1-tau for under-predicting
```
lostella:
```suggestion
        # penalize by tau for under-predicting
        # and by 1-tau for over-predicting
```

lostella (approved):
🚢 thanks @kashif!

kashif:
I thank you!

PR #1786, opened 2021-11-22
Rework tests.

jaheba:
* Create separate tests for nursery.
* Move `gluonts/mx` related tests to `test/mx`.
* Upgrade pip in all test-runners.


.github/workflows/test_release_win32_nightly.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         $tmp=$tmp.psobject.properties.value.trim() 
         git clone https://github.com/awslabs/gluon-ts --branch $tmp
         cd gluon-ts
-        python -m pip install pip==20.2
+        python -m pip install -U pip
```
mbohlkeschneider, line 28: This would just upgrade to latest, no? Does it makes sense to fix the pip version for the tests (as stuff *might* break if there is an update)?

mbohlkeschneider (commented):
Thank you! Doesn't it make sense to fix the pip version? Also there seems to be a few unintended `.typesafe` files. Maybe we want to delete those?

lostella (dismissed):

mbohlkeschneider (dismissed):

jaheba:
> E           RuntimeError: Inference tensors cannot be saved for backward. To work around you can make a clone to get a normal tensor and use it in autograd.

@kashif Is that related to your changes?

kashif:
@jaheba no not that i can think...

kashif:
@jaheba  perhaps set the lightning module as soon as you instantiate the model:

```
    model = DeepARModel(
        freq="1H",
        context_length=context_length,
        num_feat_dynamic_real=num_feat_dynamic_real,
        num_feat_static_real=num_feat_static_real,
        num_feat_static_cat=num_feat_static_cat,
        cardinality=cardinality,
    )

    lightning_module = DeepARLightningModule(model=model)
```

and then its good...

jaheba:
Any idea why this only fails in 3.8?

/cc @borchero

kashif:
@jaheba is the behavior of deepcopy different in 3.8 vs 3.7 etc?

borchero:
@jaheba no clue, this is weird. Maybe replace `with torch.inference_mode()` with `with torch.no_grad()` for the moment, that should solve the problem. PyTorch Lightning also hasn't transitioned to `inference_mode` yet (see https://github.com/PyTorchLightning/pytorch-lightning/pull/9443).

jaheba:
@borchero Thanks, looks good now!

lostella (approved):

PR #1792, opened 2021-11-24
Remove support module.

jaheba:
*Issue #, if available:*

*Description of changes:*


mbohlkeschneider (commented):
Thank you!

Why do we want to get rid of the support module? Do I understand the change right that the support functions have been factored out and put into their respective modules where they fit (for example `pad_to_size` to `mx/batchify`)?

jaheba:
> Why do we want to get rid of the support module? Do I understand the change right that the support functions have been factored out and put into their respective modules where they fit (for example `pad_to_size` to `mx/batchify`)?

The support module is/was a collection of more or less unrelated things. So, yes, my goal is to move the functions to where they are actually used.

mbohlkeschneider (approved):

src/gluonts/transform/convert.py
```diff
@@ -27,6 +26,82 @@
 )
 
 
+def erf(x: np.array) -> np.array:
```
mbohlkeschneider, line 29: This one seems to have only one specific usage throughout the code and I think one could argue that it should be a function for the class that uses it, no? Not sure about other cases here.

jaheba: I think it's fine to have them as standalone functions, since `erf` isn't tied to some object state.

mbohlkeschneider (approved):

PR #1794, opened 2021-11-25
Rework conftest.

jaheba:
* Move `contest.py` to test.
* Remove initial setup of random seeds (there is still a function-level seed)
* Add `require-packages.txt` to mark a test-folder to require packages to be installed, otherwise the folder is skipped.

The idea of `require-packages.txt` is that if I for example only use `mxnet`, I don't really care much about `torch` related tests. With `require-packages` I can just run `pytest` and all `torch` related tests are ignored.


jaheba:
Generally, I would like to remove the requirement to have `mxnet` installed to have `pytest` succeed. However, uses of mxnet are still sprinkled all over.

rshyamsundar (approved):

PR #1802, opened 2021-11-26
Use xdist (n=2) in tests.

jaheba:
Github Action uses workers with 2 CPUs by default. Thus, parallelising tests should make the runs considerably faster.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -6,7 +6,7 @@ markers =
     integration: mark an integration test
     skip_master: mark a test that is temporarily skipped for mxnet master validation.
 
-timeout = 30
+timeout = 60
 
 addopts =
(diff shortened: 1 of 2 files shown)


pytest.ini
```diff
@@ -6,7 +6,7 @@ markers =
     integration: mark an integration test
     skip_master: mark a test that is temporarily skipped for mxnet master validation.
 
-timeout = 30
+timeout = 60
```
mbohlkeschneider, line 9: Why do we double the timeout?

jaheba: Because each tests gets a bit slower when executing things in parallel.

mbohlkeschneider (approved):

PR #1805, opened 2021-11-29
Split tests into one workflow for each framework.

jaheba:
*Issue #, if available:*

*Description of changes:*


src/gluonts/mx/distribution/empirical_distribution.py
```diff
@@ -86,23 +86,19 @@ def stddev(self) -> Tensor:
...
-        self.CDF_sorted = self.F.linspace(
-            start=1 / self.samples.shape[0],
-            stop=1,
-            endpoint=True,
-            num=self.samples.shape[0],
-        )
 
-        # Replace this when mxnet is updated to a version where `searchsorted` is available!
-        import torch
+        ix = [
```
mbohlkeschneider, line 90: Isn't there an `mx.searchsorted` at this point?

jaheba: There is `mx.np.searchsorted`, which AFAIK means that you should be able to do `np.searchsorted` on `mx.nd.array`.

HOWEVER, the behaviour is different and it doesn't work.

jaheba: Also, `torch` behaves different in the sense that you can use 2d arrays, whilst numpy only supports 1d.

mbohlkeschneider: I see. Too bad.

mbohlkeschneider (approved):

PR #1807, opened 2021-11-30
Framework dependent tests.

jaheba:
With these changes, `pytest` should suceed without having `mxnet` installed.


test/core/test_component.py
```diff
@@ -153,35 +153,6 @@ def compare_vals(x, y, z):
     assert baz01 == baz02
 
 
-def test_dynamic_loading():
```
mbohlkeschneider, line 156: Why is this removed? This is not moved anywhere, right? Anything else is moved to a specific `mx` test folder. Why is this not?

jaheba: Good spot! I've removed it since this also only seems to test `load_code` which is also tested elsewhere. I can add it to the mxnet tests though.

mbohlkeschneider: Not sure what the purpose is but I would keep it, if possible.

mbohlkeschneider (approved):

PR #1808, opened 2021-11-30
Add TSBench to Nursery Folder

borchero:
This PR contains a placeholder readme for TSBench, a benchmark of GluonTS forecasting methods that will be published in the coming weeks (via a separate PR).

Merging this PR would to allow for referencing the code with a valid link in the meantime.


jaheba (approved):
That's one way to get more commits :)

PR #1809, opened 2021-12-01
Fixed data leakage bug in implementation of dynamic real and categorical features

zoolhasson:
*Issue #, if available:*

*Description of changes:* The way that the dynamic real and categorical features had been implemented in this code had incorrect logic: it leaked data from the future by taking the entirety of those features (raw, without any adjustments) across all timesteps in each context window. This fixes this problem.

Looking at the rest of the code, I highly suspect that there are still unresolved issues with one hot encoding and whether the notion of cardinality integrates well in this code, but I did not resolve those issues due to time constraints. In particular, the current implementation requires 'feat_static_cat' to be a key in the dataset (even if empty), is this desired behavior?

### DIFF
--- a/src/gluonts/model/rotbaum/_preprocess.py
+++ b/src/gluonts/model/rotbaum/_preprocess.py
@@ -404,7 +404,17 @@ class PreprocessOnlyLagFeatures(PreprocessGeneric):
             feat_static_cat = []
 
         feat_dynamic_real = (
-            [elem for ent in time_series["feat_dynamic_real"] for elem in ent]
+            list(
+                chain(
+                    *[
+                        list(ent[0]) + list(ent[1].values())
+                        for ent in [
+                            self._pre_transform(ts[starting_index:end_index])
+                            for ts in time_series["feat_dynamic_real"]
+                        ]
+                    ]
+                )
+            )
             if self.use_feat_dynamic_real
             else []
         )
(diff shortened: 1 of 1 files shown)


src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -404,12 +404,26 @@ def make_features(self, time_series: Dict, starting_index: int) -> List:
             feat_static_cat = []
 
         feat_dynamic_real = (
-            [elem for ent in time_series["feat_dynamic_real"] for elem in ent]
+            list(
+                chain(
+                    *[
+                        list(ent[0]) + list(ent[1].values())
```
dcmaddix, line 410: what is `ent`? can we use a more descriptive name?

zoolhasson: ent short for entry, presumably. I borrowed the same convention introduced elsewhere in the code. In this particular case, we probably can't be too descriptive: ent here is a tuple; the first coordinate is a transformed window, and the second one is a dictionary of additional features. If we wanted it to be more descriptive it would have to be upstream in the output format of _pre_transform.

dcmaddix (approved):

dcmaddix:
Looks like the docs build is failing so let's check this before merging.

zoolhasson:
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/.mxnet/gluon-ts/datasets/m4_hourly/train/data.json'
 
 Looks unrelated.

PR #1811, opened 2021-12-02
added a readme notebook for LSF

zoolhasson:
*Issue #, if available:*

*Description of changes:* Added a readme notebook for LSF. 

Also fixed two minor bugs: 

1. In LSF it always attempts to combine the last bin with the penultimate bin. This fails if there is only one bin, so I added a check for that.
2. Previously min_bin_size was off by 1, so that if min_bin_size was 100 then all bins contained at least 101 datapoints. I fixed this.


timoschowski (approved):
lgtm

PR #1814, opened 2021-12-03
Compute metrics from user forecasts

melopeo:
*Discussion #1756 , if available:*

*Description of changes:*

We provide the functionality to compute metrics on forecasts/target values provided by the user. By providing the json files containing the data the metrics are automatically computed.

### DIFF
--- a/src/gluonts/evaluation/utils.py
+++ b/src/gluonts/evaluation/utils.py
@@ -97,11 +97,11 @@ def compute_metrics_of_user_provided_forecasts(
     filename_fcsts: str, filename_targets: str, freq: str
 ) -> Tuple[dict, pd.DataFrame]:
 
-    logger.info(f"Loading fcsts from json file")
+    logger.info("Loading fcsts from json file")
     fcsts = load_forecasts_from_json_file(Path(filename_fcsts))
     fcsts = numpify_samples(fcsts)
 
-    logger.info(f"Loading targets from json file")
+    logger.info("Loading targets from json file")
     targets = load_forecasts_from_json_file(Path(filename_targets))
 
     targets_iter = prepare_targets_for_metric_computation(targets, freq)
--- a/src/gluonts/evaluation/utils.py
+++ b/src/gluonts/evaluation/utils.py
@@ -1,0 +1,114 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+from pathlib import Path
+from typing import Iterator, Tuple
+import logging
+import numpy as np
+import pandas as pd
+
+from gluonts.dataset.common import DataEntry, Dataset, ListDataset
+from gluonts.dataset.jsonl import JsonLinesFile
+from gluonts.evaluation import Evaluator
+from gluonts.model.forecast import Forecast, SampleForecast
+from gluonts.dataset.util import forecast_start
+
+logger = logging.getLogger(__name__)
+
+
+def load_forecasts_from_json_file(path: Path) -> list:
+    ds = []
+    for line in JsonLinesFile(path):
+        ds.append(line.content)
+
+    return ds
+
+
+def numpify_samples(fcsts: list) -> list:
+    for item in fcsts:
+        item["SageMakerOutput"]["samples"] = np.array(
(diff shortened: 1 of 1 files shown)


kashif:
@melopeo do you think it would be possible to also have a possibility of adding user defined  multivariate metrics?

melopeo:
Hi @kashif :
thanks for this request. Indeed we are having another initiative going in parallel where we are refurbishing the API of the metrics. Among the positive effects of this is that it will be easier for the user to provide his own metrics. You can follow the discussion and initative here:

- New metrics API: https://github.com/awslabs/gluon-ts/pull/1778
- Discussion on metrics: https://github.com/awslabs/gluon-ts/discussions/1756

Please let us know any insights/requests that you think we should keep in mind to make the most of these initiatives.

src/gluonts/evaluation/utils.py
```diff
@@ -0,0 +1,114 @@
...
+from gluonts.dataset.common import DataEntry, Dataset, ListDataset
+from gluonts.dataset.jsonl import JsonLinesFile
+from gluonts.evaluation import Evaluator
+from gluonts.model.forecast import Forecast, SampleForecast
+from gluonts.dataset.util import forecast_start
+
+logger = logging.getLogger(__name__)
+
+
+def load_forecasts_from_json_file(path: Path) -> list:
```
mbohlkeschneider, line 29: This would have limited scalability of the user has a lot of forecasts. One could use a generator to stream through the file instead.

src/gluonts/evaluation/utils.py
```diff
@@ -0,0 +1,114 @@
...
+        sample_forecasts.append(
+            SampleForecast(
+                samples=samples,
+                start_date=forecast_start(item),
+                freq=freq,
+                item_id=item_id,
+            )
+        )
+
+    return iter(sample_forecasts)
```
mbohlkeschneider, line 93: Why do we need to return an iterator here?

src/gluonts/evaluation/utils.py
```diff
@@ -0,0 +1,114 @@
...
+    ds = []
+    for line in JsonLinesFile(path):
+        ds.append(line.content)
+
+    return ds
+
+
+def numpify_samples(fcsts: list) -> list:
+    for item in fcsts:
+        item["SageMakerOutput"]["samples"] = np.array(
```
mbohlkeschneider, line 39: That seems quite specific (in multiple ways).

1. Why is this a `SageMakerOutput`? What about locally run forecasts? 

2. This assumes that the user would have saved the samples, which I think most users don't do, actually, because it is more efficient to save the quantiles. If quantiles are present, we could load this as a `QuantileForecast` instead and then run the evaluation. This way, we could accomodate all use cases.

src/gluonts/evaluation/utils.py
```diff
@@ -0,0 +1,114 @@
...
+    fcsts = load_forecasts_from_json_file(Path(filename_fcsts))
+    fcsts = numpify_samples(fcsts)
+
+    logger.info("Loading targets from json file")
+    targets = load_forecasts_from_json_file(Path(filename_targets))
+
+    targets_iter = prepare_targets_for_metric_computation(targets, freq)
+    fcsts_iter = prepare_fcsts_for_metric_computation(fcsts, freq)
+
+    agg_metrics, item_metrics = Evaluator(num_workers=0)(
```
mbohlkeschneider, line 110: Why is `num_workers` hard coded?

mbohlkeschneider, line 110: I think we would want to be able to configure the `Evaluator`. Maybe we want to add this functionality? For example, evaluating on more quantiles is quite common.

mbohlkeschneider, line 110: I think supporting the `MultivariateEvaluator` would be a minor change, no?

mbohlkeschneider (changes requested):
Nice! 

I left a few comments and I think this also needs a test. Keep up the good work!

kashif:
thanks @melopeo What I had in mind was a way to add for example the following two multivariate metrics easily: https://github.com/awslabs/gluon-ts/pull/1416#issuecomment-826091683

src/gluonts/evaluation/utils.py
kashif: I believe (unless I am mistaken...) the issue is the `MultivariateEvaluator` which essentially calculates univariate metrics and then some aggregated univariate metric over the multivariate dim... Ideally it would be good to have a way to calculate metrics over the full `[B,S, T,D]` tensor given `[B,T,D]` ground truth

kashif:
another request was to potentially replace the `mean` agg. function in `MSIS` and `MASE` with `nanmean` as the seasonal error can potentially be zero for some time series in a dataset, causing these agg. metrics to be `inf`.

jaheba:
> another request was to potentially replace the `mean` agg. function in `MSIS` and `MASE` with `nanmean` as the seasonal error can potentially be zero for some time series in a dataset, causing these agg. metrics to be `inf`.

Wouldn't this be something we can do independently of this PR? Not sure if we would like to put this feature behind a flag.

kashif:
another nice to have would be to log and get the log-likelihood of the predicted samples...

PR #1821, opened 2021-12-07
Add dynamic code for shell.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/shell/README.md
+++ b/src/gluonts/shell/README.md
@@ -154,3 +154,32 @@ as the environment-variable, instead of `DEEPAR_INFERENCE_CONFIG`.
    ...
 }
 ```
+
+## Dynamic Code
+
+It is possible to pass a `code` channel to training jobs to install code
+dependencies without the need to build a new container.
+
+Code-channel entries will be installed to `/opt/ml/code`, and the process is
+restarted with `/opt/ml/code` being added to `PYTHONPATH`. On the second
+iteration the newly installed modules and packages can be simply imported.
+
+The contents of the code-channel are treated as follows:
+
+Folders with a `setup.py` are assumed to be pip compatible, and are thus
+installed with `pip install`. This is the most powerful option, since this
+means that dependencies of the package are installed as well.
+
+A folder with an `__init__.py` file is treated as a simple python-package,
+which is copied to `/opt/ml/code`.
+
+If neither `setup.py` nor `__init__.py` exists in a folder, all other python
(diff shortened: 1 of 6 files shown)


src/gluonts/shell/sagemaker/dyn.py
```diff
@@ -0,0 +1,108 @@
...
+            ]
+        )
+
+    def install(self, path):
+        if path.is_file():
+            if path.suffix == ".py":
+                self.copy_install(path)
+
+            elif path.suffixes[-2:] == [".tar", ".gz"]:
+                with tarfile.open(path, "r:gz") as archive:
```
RSNirwan, line 68: Is the package automatically tar zipped by the shell or can the user do it himself?

Any reason to hardcode gzip-compression here?
If user is allowed to upload any tarfile, you might use `elif tarfile.is_tarfile(path):` here. That allows other suffixes (like .tgz) and more compression options (bz, bz2, ..).

jaheba: Thanks for the suggestion. The initial idea was to support `.tar.gz` files which are emitted by `python setup.py sdist`. These are pip-installable directly.

However, I didn't want to blankly install any archive, and thus I opted for unpacking and treating the content normally.

I guess we could also think about supporting `.zip`, which we are at it.

src/gluonts/shell/README.md
```diff
@@ -154,3 +154,32 @@ as the environment-variable, instead of `DEEPAR_INFERENCE_CONFIG`.
    ...
 }
 ```
+
+## Dynamic Code
+
+It is possible to pass a `code` channel to training jobs to install code
+dependencies without the need to build a new container.
```
RSNirwan, line 161: As I understood it, you want to upload the directory/file `code` is pointing to to s3.
Would it make sense to run `python setup.py sdist` if the user input is a path to an installable package and upload the resulting `.tar.gz` instead of uploading the full package?
I know many times experiments data/results/scripts/log and much more are also saved n the package directory. It can get really expensive if at every minor change the whole directory is uploaded to s3.

jaheba: Yes, but that's on the client side. Here we need to consume whatever is given to us, which would include both zipped folders, and python source distribution.

On the client side (which is not part of this PR), we would do the `python setup.py sdist` thing to get a smaller archive :)

RSNirwan (approved):

PR #1827, opened 2021-12-11
add SCott code to nursery

EugeneLYC:
Adding the SCott code from ICML'21 to nursery

Adding the experimental code from ICML'21 paper in AWS AI Lab: "Variance Reduced Training with Stratified Sampling for Forecasting Models" to the nursery repo.


lostella:
@EugeneLYC would it be possible to have a summary of what here is being added on top of the `pytorchts` package? Having an entire clone of the repo is not ideal for many reasons (among other things: it interferes with IDEs when searching code, and causes unnecessary documentation being built).

The best approach is to take `pytorchts` as a dependency, and just have the additional code here.

PR #1836, opened 2021-12-17
Fix samples scaling in PyTorch DeepAR

kashif:
In deepAR when sampling we multiply the scaling factor whereas for NB or Poisson we need to scale in the distribution's parameter space. This PR I believe fixes it.


kashif:
closing as I might have a subtle bug here... checking

kashif:
@jaheba can you kindly review this? Its essentially fixing the sampling to use the distribution transforms

kashif:
@lostella so I add repeat params rather than sample with a different shape since its a bit more intuitive and there is no need then to reshape the sample etc. Also for some distribution outputs the sample() method might not take a shape argument, (e.g. the implicit quantile network which doesn't have this concept...)

hopefully the rest is clear?

lostella (approved):
Makes sense, and looks good to me! Thanks for catching this one too! 🚀

kashif:
thanks!

Issue #1192, opened 2020-12-02
Properly check predictor equal types. 


mbohlkeschneider:
## Description
We currently only check for the underlying prediction net to have the same parameters. This can create issues when constructing a `Predictor` using the correct prediction net but other parameters, like `freq` or `transformation`. 

Maybe we can add proper checks the other inputs as well. 

## References
- list reference and related literature
- list known implementations

huibinshen:
Yes, closing it now!

PR #1844, opened 2022-01-03
Check equality of transformations.

huibinshen:
*Issue #1192 

*Support equals check on the transformation*

**Please tag this pr with at least one of these labels to make our release process faster:** enhancement

### DIFF
--- a/test/mx/test_transform_equals.py
+++ b/test/mx/test_transform_equals.py
@@ -10,34 +10,27 @@
 # on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
-from typing import List
 
-import pytest
-import numpy as np
-
-from gluonts.core import serde
+from gluonts.core.serde.flat import clone
 from gluonts import time_feature, transform
 from gluonts.core.component import equals
 from gluonts.dataset.field_names import FieldName
-from gluonts.dataset.loader import Batch
 from gluonts.model.canonical._estimator import CanonicalRNNEstimator
 
 
 def test_map_transformation():
-
-    tran1 = transform.VstackFeatures(
+    tran = transform.VstackFeatures(
         output_field="dynamic_feat",
         input_fields=["age", "time_feat"],
         drop_inputs=True,
     )
 
-    tran2 = serde.flat.clone(tran1)
-    tran3 = serde.flat.clone(tran1, {"drop_inputs": False})
+    assert equals(tran, clone(tran))
+    assert not equals(tran, clone(tran, {"drop_inputs": False}))
 
-    assert equals(tran1, tran2)
-    assert not equals(tran1, tran3)
 
-    tran4 = transform.AddTimeFeatures(
+def test_add_time_features():
+    tran = transform.AddTimeFeatures(
         start_field=FieldName.START,
         target_field=FieldName.TARGET,
         output_field="time_feat",
@@ -49,9 +42,8 @@ def test_map_transformation():
         pred_length=10,
     )
 
-    tran5 = serde.flat.clone(tran4)
-    tran6 = serde.flat.clone(
-        tran4,
+    tran2 = clone(
+        tran,
         {
             "time_features": [
                 time_feature.DayOfWeek(),
@@ -60,12 +52,11 @@ def test_map_transformation():
         },
     )
 
-    assert equals(tran4, tran5)
-    assert not equals(tran4, tran6)
+    assert equals(tran, clone(tran))
+    assert not equals(tran, tran2)
 
 
 def test_filter_transformation():
-
     prediction_length = 10
     tran1 = transform.FilterTransformation(
         lambda x: x["target"].shape[-1] > prediction_length
@@ -82,20 +73,19 @@ def test_filter_transformation():
     assert not equals(tran1, tran3)
 
 
-def test_sampler():
+def test_exp_num_sampler():
     sampler = transform.ExpectedNumInstanceSampler(num_instances=4)
-    sampler2 = serde.flat.clone(sampler, {"num_instances": 5})
-    assert equals(sampler, serde.flat.clone(sampler))
-    assert not equals(sampler, sampler2)
+    assert equals(sampler, clone(sampler))
+    assert not equals(sampler, clone(sampler, {"num_instances": 5}))
 
-    sampler3 = transform.ContinuousTimeUniformSampler(num_instances=4)
-    assert equals(sampler3, serde.flat.clone(sampler3))
-    sampler4 = serde.flat.clone(sampler3, {"num_instances": 5})
-    assert not equals(sampler3, sampler4)
+
+def test_continuous_time_sampler():
+    sampler = transform.ContinuousTimeUniformSampler(num_instances=4)
+    assert equals(sampler, clone(sampler))
+    assert not equals(sampler, clone(sampler, {"num_instances": 5}))
 
 
 def test_instance_splitter():
-
     splitter = transform.InstanceSplitter(
         target_field=FieldName.TARGET,
         is_pad_field=FieldName.IS_PAD,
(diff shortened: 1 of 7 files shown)


huibinshen:
This is currently just for checking the design. Unit tests and formatting will be followed.

jaheba:
Thanks for the PR!

I think we want to make the code part of `transformation` and not add it to `core`.

huibinshen:
> 



> Thanks for the PR!
> 
> I think we want to make the code part of `transformation` and not add it to `core`.

makes sense, will move it to the "transform" folder.

src/gluonts/core/transformation_check.py
```diff
@@ -0,0 +1,114 @@
...
+        if isinstance(this_tran, Batch):
+            equals_batch(this_tran, that_tran)
+        elif isinstance(this_tran, FlatMapTransformation):
+            equals_flatmap_transformation(this_tran, that_tran)
+        elif isinstance(this_tran, SimpleTransformation):
+            equals_simple_transformation(this_tran, that_tran)
+        elif isinstance(this_tran, MapTransformation):
+            equals_map_transformation(this_tran, that_tran)
+        else:
+            equals_transformation(this_tran, that_tran)
```
RSNirwan, line 112: - Isn't a stopping criteria needed to stop the for-loop and return false if any of the equals returns false?
- Do you have a minimal example of 2 chains where the maximum recursion depth is reached?

jaheba:
I wonder, could we get away by defining `__eq__` on these classes?

src/gluonts/core/transformation_check.py
huibinshen: Thanks for the comments!

* Re the early stop the for loop, absolutely, will add that in the next revision, thanks for catching it!
* Re the  recursion, because we decorate the function by "equals", that's why if we call "equals" inside this function, it will certainly reach maximum recursion. Will provide tests in the next revision.

huibinshen:
> I wonder, could we get away by defining `__eq__` on these classes?

I also thought about it in the beginning. It looks to me that `@equals.register` is vastly used in this repo and that's why I try to follow the same fashion for now. 

They both have pros and cons. On the one hand, using `__eq__` inside each class will allow us compare easily in a chain. On the other hand, the equality check logic is scattered in each class and in terms of change, the current way seems cleaner in the sense that not changing the existing code but rather adding extra capabilities. 

What's your thoughts?

src/gluonts/core/transformation_check.py
RSNirwan: Chances are that I have missed something here. That's why I asked for an example.
`equals` is not a function but a generic-function with singledispatch (kind of a set of functions). That means, `equals` forwards (dispatches) its arguments to the specific function based on the type of the first argument. The complete body of your for-loop is equivalent to calling just `equals(this_tran, that_tran)`. You can try this out. 
It would be something like that:
```python
for this_tran, that_tran in zip(this.transformations, that.transformations):
    if not equals(this_tran, that_tran):
        return False
```
I hope I haven't missed anything.

jaheba:
`equals` refers back to `__eq__` by default:

https://github.com/awslabs/gluon-ts/blob/d7679a5f239cbfd9e885dfdfe5312eb292353f73/src/gluonts/core/component.py#L131-L138

Maybe we could define some test cases first so see what works and what not and approach the issue that way?

src/gluonts/core/transformation_check.py
huibinshen: Ah, you are right! I confused myself with a wrong example (where the setting is a bit different with inheritance). This is much better indeed!

test/mx/test_transform_equals.py
```diff
@@ -0,0 +1,219 @@
...
+from gluonts.core import serde
+from gluonts import time_feature, transform
+from gluonts.core.component import equals
+from gluonts.dataset.field_names import FieldName
+from gluonts.dataset.loader import Batch
+from gluonts.model.canonical._estimator import CanonicalRNNEstimator
+
+
+def test_map_transformation():
+
```
jaheba, line 27: Can we remove these empty lines?

test/mx/test_transform_equals.py
```diff
@@ -0,0 +1,219 @@
...
+        drop_inputs=True,
+    )
+
+    tran2 = serde.flat.clone(tran1)
+    tran3 = serde.flat.clone(tran1, {"drop_inputs": False})
+
+    assert equals(tran1, tran2)
+    assert not equals(tran1, tran3)
+
+    tran4 = transform.AddTimeFeatures(
```
jaheba, line 40: Can we split this into its own test? It's generally better to have many smaller tests, since they then can individually fail.

test/mx/test_transform_equals.py
```diff
@@ -0,0 +1,219 @@
...
+
+def test_map_transformation():
+
+    tran1 = transform.VstackFeatures(
+        output_field="dynamic_feat",
+        input_fields=["age", "time_feat"],
+        drop_inputs=True,
+    )
+
+    tran2 = serde.flat.clone(tran1)
```
jaheba, line 34: I think I would write this as

```py
    assert equals(t, serde.flat.clone(t)
    assert not equals(t, serde.flat.clone(t, {"drop_inputs": False})
```
Or maybe shorten it to:

```py
    assert equals(t, clone(t))
    assert not equals(t, clone(t, {"drop_inputs": False}))
```

test/mx/test_transform_equals.py
```diff
@@ -0,0 +1,219 @@
...
+    splitter3 = transform.ContinuousTimeInstanceSplitter(
+        past_interval_length=1,
+        future_interval_length=1,
+        instance_sampler=transform.ContinuousTimePointSampler(min_past=1.0),
+    )
+    assert equals(splitter, splitter2)
+    assert not equals(splitter, splitter3)
+
+
+def test_batch():
```
jaheba, line 146: Shouldn't this be tested as part of the loader?

test/mx/test_transform_equals.py
```diff
@@ -0,0 +1,219 @@
...
+    pred_length = 10
+
+    estimator = CanonicalRNNEstimator("5min", train_length, pred_length)
+
+    assert equals(estimator, serde.flat.clone(estimator))
+    assert not equals(estimator, serde.flat.clone(estimator, {"freq": "1h"}))
+
+
+if __name__ == "__main__":
+    test_filter_transformation()
```
jaheba, line 219: Let's remove this.

There is the `-k` flag in pytest, where you can easily select tests, so you can do `pytest test/mx/test_transform_equals.py -k filter` to just run tests containing the word `filter`.

test/torch/model/test_simple_torch_model.py
```diff
@@ -206,3 +207,23 @@ def test_simple_model():
...
+        ),
+        past_length=context_length,
+        future_length=prediction_length,
+        time_series_fields=[FieldName.OBSERVED_VALUES],
+    )
+
+    another_predictor = net.get_predictor(
+        transformation + another_training_splitter
+    )
+    assert not equals(predictor, another_predictor)
```
jaheba, line 229: I think it would make sense to also check for equality.

jaheba (commented):
Thanks, I have some minor comments.

test/torch/model/test_simple_torch_model.py
huibinshen: added

test/mx/test_transform_equals.py
huibinshen: Removed. I think one can also use `pytest test/mx/test_transform_equals.py::test_filter_transformation`.

test/mx/test_transform_equals.py
huibinshen: Moved.

test/mx/test_transform_equals.py
jaheba: True, but I'm also lazy :)

test/mx/test_transform_equals.py
huibinshen: For object can be created in 1 line, will change it to.
```Python
    assert equals(t, clone(t))
    assert not equals(t, clone(t, {"drop_inputs": False}))
```
For the object can not be created in 1 line, will use `tran = clone(...)`.

test/mx/test_transform_equals.py
huibinshen: Done.

test/mx/test_transform_equals.py
huibinshen: Removed.

jaheba (approved):

PR #1848, opened 2022-01-05
Install gluonts for style checks.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -1,6 +1,6 @@
 holidays>=0.9
 matplotlib~=3.0
-numpy~=1.16
+numpy~=1.21
 pandas~=1.0
 pydantic~=1.1
 tqdm~=4.23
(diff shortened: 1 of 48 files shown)


requirements/requirements.txt
```diff
@@ -1,6 +1,6 @@
 holidays>=0.9
 matplotlib~=3.0
-numpy~=1.16
+numpy~=1.21
```
jaheba, line 3: We probably want to split this change into another PR.

huibinshen: Is this the reason why `test_empirical_distribution` fails? If yes, totally.

jaheba: No, I don't think these are related.

huibinshen (dismissed):
Thanks for fixing these type checks!

huibinshen (approved):

PR #1865, opened 2022-01-31
Add TSBench

borchero:
*Description of changes:* Add TSBench to the nursery folder

**Please tag this pr with at least one of these labels to make our release process faster:** (Future) paper implementation


setup.py
```diff
@@ -143,7 +143,9 @@ def run(self):
         import click
         import black
 
-        black_opts = []
+        black_opts = [
+            "--extend-exclude", "src/gluonts/nursery/tsbench"
```
jaheba, line 147: Let's not do this.

If this code should be part of GluonTS, then it should adhere to its rules.

jaheba:
Let's fix the updated black version here: https://github.com/awslabs/gluon-ts/pull/1867

jaheba (approved):

lostella:
@borchero is the `poetry.lock` file necessary? It appears to be listing some outdated versions of packages, which trigger the potential vulnerability analysis from GitHub

PR #1876, opened 2022-02-07
Start field as datetime fix docs

kashif:
*Issue #, if available:*

*Description of changes:*
I fixed the types and added a test.

### DIFF
--- a/src/gluonts/dataset/common.py
+++ b/src/gluonts/dataset/common.py
@@ -315,12 +315,13 @@ class ProcessStartField(pydantic.BaseModel):
 
     @staticmethod
     @lru_cache(maxsize=10000)
-    def process(
-        string_or_ts: Union[str, pd.Timestamp], freq: str
-    ) -> pd.Timestamp:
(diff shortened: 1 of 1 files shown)


src/gluonts/dataset/common.py
```diff
@@ -315,10 +315,12 @@ def __call__(self, data: DataEntry) -> DataEntry:
 
     @staticmethod
     @lru_cache(maxsize=10000)
-    def process(string: str, freq: str) -> pd.Timestamp:
+    def process(
+        string_or_ts: Union[str, pd.Timestamp], freq: str
```
jaheba, line 319: In theory, a lot of different values should work here. Maybe we just use `Any` here with a comment indicating such?

kashif: sure sounds good to me! I checked the pandas docs and it supports: `datetime-like, str, int, float` types...

jaheba (approved):

kashif:
thank you!

PR #1879, opened 2022-02-12
Fixes for Monash Forecasting Repository datasets

borchero:
*Issue #, if available:* n/a

*Description of changes:* This PR implements two minor (yet important) improvements for datasets from the [Monash Forecasting Repository](https://forecastingdata.org).

**1: TSF Reader**

The `TSFReader` currently uses `glide.imap_unordered` to process the individual lines of the downloaded dataset file. As a result, preprocessing pipelines are nondeterministic which causes a problem if a deterministic order of e.g. test time series is required.

This PR continues to use `glide.imap_unordered`, but passes an additional index to the worker function which enables sorting according to this index after all items have been processed. This way, the dataset is generated in the same way on every invocation.

**2: Generated Keys**

Currently, all time series extracted from Monash datasets only have a `target` and `start_timestamp` field. However, some datasets (which are not explicitly made available via GluonTS but could be used by third parties) do not provide start timestamps for the time series. An optional "fallback" start timestamp can now be provided.

Additionally, a unique `item_id` is generated for every time series. This aligns with the other datasets available in GluonTS.

### DIFF
--- a/src/gluonts/dataset/repository/_tsf_datasets.py
+++ b/src/gluonts/dataset/repository/_tsf_datasets.py
@@ -159,7 +159,12 @@ def save_metadata(
         )
 
 
-def save_datasets(path: Path, data: List[Dict], train_offset: int):
+def save_datasets(
+    path: Path,
+    data: List[Dict],
+    train_offset: int,
+    default_start_timestamp: Optional[str] = None,
+):
     train = path / "train"
     test = path / "test"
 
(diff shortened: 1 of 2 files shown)


kashif:
@borchero do you which time seris don't have a `start_date`?

borchero:
@kashif I'm not sure if there's any datasets that are directly available via GluonTS, but this applies to some datasets on [forecastingdata.org](forecastingdata.org). One example is the "Weather" dataset.

kashif:
@borchero awesome, yes that was my question

jaheba (approved):
Thanks!

Issue #1796, opened 2021-11-25
NaN losses leads to overfitting


sevstafiev:
## Description
I am using a retail store dataset with ~ 3000 time series. Their peculiarity is that there are "special days" when the number of sales grows sharply (on average, 1-2 sales per day, on the day of sale ~ 80). I ran into a problem that with an increase in the number of epochs, the model starts to produce NaN loss over time, overtraining and too often to produce a forecast as for a "special day". Having studied the issues, I decided that adding a validation dataset would help with retraining, since, as I understood, along with the validation, a stop mechanism was also built in. As a validation, I began to take the last 60 days of the train (based on the logic of the context_length = 60 parameter) and 28 days after the train (based on the logic of the prediction_length = 28 parameter) for each time series. I use 28 days after validation as a test. This really helped to track the improvement and deterioration of the quality of the model, however, over time, the model starts to produce NaN loss and the quality on validation drops significantly, that is, the problem has not been solved. Nevertheless, the model does not stop training, but goes through all the remaining batches and epochs, giving NaN loss. As a result, after learning, it begins to give excessively high predictions.

In this issue https://github.com/awslabs/gluon-ts/issues/833 the author was able to overcome the problem with putting a conditional breakpoint in the log_prob() to stop whenever a NaN value is generated. But I did not understand how this can be done and how to get to log_prob() at all. **If you can tell me how to do this, then this will also be a good solution to the problem.**

Since the model refits each time when a new piece is added to the data, it is impossible to guess the optimal number of epochs. If I put a little (1-3), then the model is underfitting and the quality will be poor, however, the fewer epochs, the less chance that NaN loss will occur. If I put a large number of epochs (20), then NaN loss is guaranteed to occur and the predictions will be bad.

## To Reproduce

```python
trainer = Trainer(
      ctx=device,
      epochs=20, 
      learning_rate_decay_factor=0.5,
      patience=3,
      minimum_learning_rate=0.001,
      clip_gradient=1.0,
      weight_decay=1e-08,
      learning_rate=0.01,
      hybridize = False, #True changed nothing but training speed
      batch_size = 32,
  )

deepar_estimator = DeepAREstimator(
    freq="D", 
    prediction_length=h,
    trainer=trainer,
    context_length=60, 
    num_layers=2,
    num_cells=100,
    cell_type="lstm",
    dropout_rate=0.1,
    use_feat_dynamic_real=True,
    use_feat_static_cat=True,
    cardinality=cardinality,
[...]

PR #1890, opened 2022-02-16
Change negative binomial parametrization to failure count and log-odds

### LINKED ISSUE #833: Getting NaN loss after some time with DeepAR and NegativeBinomial
## Description
I have a dataset with 129k time series of demand for products within some branches of a retailer. I'm having trouble with NaN loss ever since I started using GluonTS. At first, I discovered some NaN and Infinity values within my FEAT_DYNAMIC_REAL. But even after solving those issues, I'm still getting NaN loss after some time. With the new release (which included some improvements on the NegativeBinomial, I'm getting a much lower loss value and it's taking longer for the problem to occur, but it still occurs.

I decided to put a conditional breakpoint in the log_prob() to stop whenever a NaN value is generated and used hybridize=False. It's super strange, but the error didn't happen yet (It's far past the point that it usually happens). I'll wait some more time, but it might be the case that it only happens with hybridize=True? If that's the case, it'd be very odd and problematic, since it runs much slower with hybridize=False.

## To Reproduce

Unfortunately, I can't provide the dataset to reproduce the problem. But here are some information:
- Number of Time Series: 128.976
- TARGET: Count values, with a lot of zeros
- FEAT_STATIC_CAT: I'm using categorical features that identify the product (its id, its category, subcategory, and so on) and the branch (its id, its city, state, and so on).
- FEAT_DYNAMIC_REAL: I'm using holidays in one-hot encoding, a binary identifying if there's a special price and the relative discount (calculated using the discount and the price).
- batch_size: 32
- num_batches_per_epoch: 10000
- learning_rate: 1e-4
- clip_gradient: 1.0
- patience: 10
- minimum_learning_rate: 5e-5
- weight_decay: 1e-8
- num_layers: 2
- num_cells: 100
- embedding_dimension: [2, 7, 20, 50, 10, 30, 60, 80, 200]
- context_length: 90
- scaling: True
- freq: 1D

```python
put code here
```

Here is one example where it happened in the first epoch. But sometimes it happened later.

```
  File "/opt/anaconda3/envs/gazin/lib/python3.6/site-packages/luigi/worker.py", line 199, in run
    new_deps = self._run_get_new_deps()
  File "/opt/anaconda3/envs/gazin/lib/python3.6/site-packages/luigi/worker.py", line 141, in _run_get_new_deps
    task_gen = self.task.run()
  File "/root/work/log_inteligente_ai/gazin/task/model/base.py", line 164, in run
    dict(fit=self.fit, cross_validation=self.cross_validate)[self.training_type]()
  File "/root/work/log_inteligente_ai/gazin/task/model/gluonts.py", line 425, in fit
    predictor = self.train_model()
  File "/root/work/log_inteligente_ai/gazin/task/model/gluonts.py", line 416, in train_model
    train_iter=training_data_loader,
  File "/opt/anaconda3/envs/gazin/lib/python3.6/site-packages/gluonts/trainer/_base.py", line 328, in __call__
    "Got NaN in first epoch. Try reducing initial learning rate."
gluonts.core.exception.GluonTSUserError: Got NaN in first epoch. Try reducing initial learning rate.
```

## Environment
- Operating system: Linux
- Python version: 3.6
- GluonTS version: 0.5.0

lostella:
*Issue #, if available:* Should fix #833 and #1796 

*Description of changes:* Changing the negative binomial parametrization from "mean and dispersion" to "number of failures and log-odds". Adapting tests. Examples of the improvement to-be-posted.

### DIFF
--- a/src/gluonts/mx/distribution/neg_binomial.py
+++ b/src/gluonts/mx/distribution/neg_binomial.py
@@ -31,27 +31,29 @@ class NegativeBinomial(Distribution):
 
     Parameters
     ----------
-    mu
-        Tensor containing the means, of shape `(*batch_shape, *event_shape)`.
-    alpha
-        Tensor of the shape parameters, of shape `(*batch_shape, *event_shape)`.
+    count
+        Non-negative Tensor containing the number of failed Bernoulli trials
+        for the process to stop. Shape is `(*batch_shape, *event_shape)`.
+    logit
+        Tensor containing the log-odds of success in the Bernoulli trials.
+        Shape is `(*batch_shape, *event_shape)`.
     F
     """
 
     is_reparameterizable = False
 
     @validated()
-    def __init__(self, mu: Tensor, alpha: Tensor) -> None:
-        self.mu = mu
-        self.alpha = alpha
+    def __init__(self, count: Tensor, logit: Tensor) -> None:
+        self.count = count
+        self.logit = logit
 
     @property
     def F(self):
-        return getF(self.mu)
+        return getF(self.count)
 
     @property
     def batch_shape(self) -> Tuple:
-        return self.mu.shape
+        return self.count.shape
 
     @property
     def event_shape(self) -> Tuple:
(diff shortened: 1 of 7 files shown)


lostella:
Whoops, the zero-inflated needs some rework too, I guess

canerturkmen:
Before the review, why do we need this change to fix NaN losses and were we able to confirm it does? The issue is that this is a breaking change. I know at least renewal models [depend](https://github.com/awslabs/gluon-ts/blob/8132e2462b649455ade4dbc66cc36114df1796bc/src/gluonts/model/renewal/_estimator.py#L90) on the first parameter of the distribution being the conditional mean and I suspect other might as well (this is probably why the interface was designed like this in the first place?).

lostella:
> Before the review, why do we need this change to fix NaN losses and were we able to confirm it does?

So far as I have tested, it does: on a local run on the example discussed in #1796, I'm now getting very similar results with MXNet-DeepAR as with PyTorch-DeepAR. In particular, training consistently goes well. I will post more evidence.

> I know at least renewal models [depend](https://github.com/awslabs/gluon-ts/blob/8132e2462b649455ade4dbc66cc36114df1796bc/src/gluonts/model/renewal/_estimator.py#L90) on the first parameter of the distribution being the conditional mean and I suspect other might as well

That's the point I was looking for, thanks :-) What one could do, is to have different parametrizations, and at distribution construction time we convert to the one used internally for computing log density, sampling, and everything. The same option could exist in the `DistributionOutput` realm.

Issue #1908, opened 2022-02-25
Scaling parameter of Transformer always set to True


lostella:
### Discussed in https://github.com/awslabs/gluon-ts/discussions/1907

<div type='discussions-op-text'>

<sup>Originally posted by **StanislasGuinel** February 24, 2022</sup>
The `scaling` parameter is always set to True when creating the `TransformerTrainingNetwork` or the `TransformerPredictionNetwork` but it is still a parameter of the `TransformerEstimator`.
Is there a reason why ?
</div>

PR #1909, opened 2022-02-25
Fix the use of the scaling parameter in Transformer model

StanislasGuinel:
*Issue #1908*

*Description of changes:*
The Transformer model was always using setting `scaling` to True when creating the `TransformerTrainingNetwork` and the `TransformerPredictionNetwork`.

This PR now uses the `scaling` value set in the `TransformerEstimator` constructor.

Label: bug fix


lostella (approved):
Thanks @StanislasGuinel!

PR #1910, opened 2022-02-28
Clean up README, add table of models

lostella:
Clean up a bit the README, and add a table with a summary of the available models. Ideally, the information this should include is:
* Model name
* Model "type" (e.g. local vs global, univariate vs multivariate, regularly spaced data vs "point process" data, ...)
* High-level architecture summary (e.g. "CNN encoder, MLP decoder" or "Causal convolution, self attention"); essentially what fundamental components the model uses
* Link to references (ideally "the" paper that introduces the model)

But we can agree on a different set of fields, or what exactly they should tell.


dcmaddix (approved):
Looks good to me - thanks for adding Lorenzo.

dcmaddix:
- Should we add SimpleFeedForward as well and maybe explicitly mention ETS and R under RForecast.
I think adding the code link to the model would be helpful too then users could directly access it.

- For GPForecaster, we could link the GPytorch paper which has the loss function https://arxiv.org/abs/1809.11165 or the original book http://www.gaussianprocess.org/gpml/chapters/RW.pdf.

- Not sure if we can mention IQF implementation with MQ-CNN is available and link that paper too.

README.md
```diff
@@ -1,43 +1,78 @@
...
-and then use it to make predictions. As a first step, we need to collect
-some data: in this example we will use the volume of tweets mentioning the
+## Implemented models
+
+Name                             | Local/global | Data layout              | Architecture | Implementation | References
+---------------------------------|--------------|-------------------|--------------|----------------|-----------
+DeepAR                           | Global       | Univariate               | RNN | [MXNet](src/gluonts/model/deepar/_estimator.py), [PyTorch](src/gluonts/torch/model/deepar/estimator.py) | [paper](https://doi.org/10.1016/j.ijforecast.2019.07.001)
+DeepFactor                       | Global       | Univariate               | RNN, state-space model, Gaussian process | [MXNet](src/gluonts/model/deep_factor/_estimator.py) | [paper](https://proceedings.mlr.press/v97/wang19k.html)
+DeepState                        | Global       | Univariate               | RNN, state-space model | [MXNet](src/gluonts/model/deepstate/_estimator.py) | [paper](https://papers.nips.cc/paper/2018/hash/5cf68969fb67aa6082363a6d4e6468e2-Abstract.html)
```
rshyamsundar, line 44: `DeepState` should be above `DeepFactor`; makes sense that way both chronologically and architecture-wise.

lostella: I fell back to alphabetical ordering when in doubt, but what you propose makes sense; maybe the link to the paper could mention the year ("Rangapuram et al., 2018") so that this is also apparent

lostella:
> * Should we add SimpleFeedForward as well and maybe explicitly mention ETS and R under RForecast.
>       I think adding the code link to the model would be helpful too then users could directly access it.

Yes, makes sense to add `SimpleFeedForward`, and ETS or ARIMA could be mentioned in the "architecture" column maybe (which could be called "architecture/method" maybe)

>     * For GPForecaster, we could link the GPytorch paper which has the loss function https://arxiv.org/abs/1809.11165 or the original book http://www.gaussianprocess.org/gpml/chapters/RW.pdf.

We could, but one would be tempted to think that somewhere in those references there is a discussion on the specific model for time series that we implement... so I'm not sure

>     * Maybe we can mention IQF implementation with MQ-CNN is available and link that paper too.

Yes, I'll look into that

dcmaddix:
> > * Should we add SimpleFeedForward as well and maybe explicitly mention ETS and R under RForecast.
> >   I think adding the code link to the model would be helpful too then users could directly access it.
> 
> Yes, makes sense to add `SimpleFeedForward`, and ETS or ARIMA could be mentioned in the "architecture" column maybe (which could be called "architecture/method" maybe)
> 
> > * For GPForecaster, we could link the GPytorch paper which has the loss function https://arxiv.org/abs/1809.11165 or the original book http://www.gaussianprocess.org/gpml/chapters/RW.pdf.
> 
> We could, but one would be tempted to think that somewhere in those references there is a discussion on the specific model for time series that we implement... so I'm not sure
> 
> > * Maybe we can mention IQF implementation with MQ-CNN is available and link that paper too.
> 
> Yes, I'll look into that

Yes, sounds good. Yes I agree the reference for GP for forecasting is not as straightforward.

jgasthaus (approved):

Issue #1912, opened 2022-03-01
Error with multiple-month frequency when data is misaligned


lostella:
## Description
A problem occurs when a dataset with frequency which is multiple months is fed to a model, and the series are not "aligned" according to the frequency. For example (see reproducing snippet below), when using `freq="2M"` and there are two series:
* training works fine if the two series start at, respectively
    * `"2021-01-01"`
    * `"2021-03-01"` (or any other date which is a multiple of 2 months away from the previous)
* training **breaks** if the two series start at, respectively
    * `"2021-01-01"`
    * `"2021-02-01"` (or any other date which is **not** a multiple of 2 months away from the previous)

## To Reproduce

```python
from gluonts.dataset.common import ListDataset
from gluonts.model.deepar import DeepAREstimator
from gluonts.mx.trainer import Trainer

dataset = ListDataset(
    data_iter=[
        {"start": "2021-01-01 00:00:00", "target": [1.0] * 20},
        {"start": "2021-02-01 00:00:00", "target": [1.0] * 20}
    ],
    freq="2M"
)

estimator = DeepAREstimator(
    freq="2M", prediction_length=4, context_length=10,
    trainer=Trainer(epochs=5)
)

predictor = estimator.train(dataset)
```

## Error message or code output

```
Traceback (most recent call last):
  File "pandas/_libs/index.pyx", line 444, in pandas._libs.index.DatetimeEngine.get_loc
  File "pandas/_libs/hashtable_class_helper.pxi", line 1625, in pandas._libs.hashtable.Int64HashTable.get_item
  File "pandas/_libs/hashtable_class_helper.pxi", line 1632, in pandas._libs.hashtable.Int64HashTable.get_item
KeyError: 1614470400000000000

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/xyz/.virtualenvs/gluonts/lib/python3.7/site-packages/pandas/core/indexes/base.py", line 3080, in get_loc
    return self._engine.get_loc(casted_key)
  File "pandas/_libs/index.pyx", line 413, in pandas._libs.index.DatetimeEngine.get_loc
  File "pandas/_libs/index.pyx", line 446, in pandas._libs.index.DatetimeEngine.get_loc
KeyError: Timestamp('2021-02-28 00:00:00', freq='2M')

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

PR #1916, opened 2022-03-02
Fix AddTimeFeatures transformation for multiples of base frequencies

lostella:
*Issue #, if available:* #1912

*Description of changes:*
* Fix the caching mechanism of AddTimeFeatures so that it works also with multiples of base frequencies, fixing #1912
* Reorg transformations tests, add test module specifically to test the correctness of AddTimeFeatures by comparing its output against a stripped-down, non-optimized function that computes the same features


vafl:
Thanks, looks good to me!

This does increase the size of the feature cache, but I don't think it's an issue (unless you use crazy multipliers like `86400s` instead of `1D`).  :+1:

kashif:
BTW would you also be able to test with Month begin "MS" freq as well as "W-MON" etc. ?

vafl (approved):

PR #1917, opened 2022-03-02
Breaking: Fix return types of features

lostella:
*Description of changes:* Type annotation says `np.ndarray`, but they are returning a `pd.Series`. This PR makes sure they return a `np.ndarray` of `float`s.


lostella:
A natural question here is: are the return values to be adjusted to match the type annotations, or are the type annotations to be fixed?

kashif:
i believe its good if its ndarrays, then one can vstack them to temporal covariates  etc. nicely...  or are you asking something else?

lostella:
> i believe its good if its ndarrays, then one can vstack them to temporal covariates etc. nicely... or are you asking something else?

That's exactly what I'm asking, thanks! Strictly speaking, this is a breaking change since features return `pd.Series` prior to this change, that's why I was contemplating changing the return type annotation to `pd.Series` instead.

kashif:
hmm i see... as far as I can see this transformation's output are stacked to any potential covariates as you know, and then things are cast to ndarrays etc.  so I believe its not a big problem for me personally one way or another...

jaheba (approved):

PR #1920, opened 2022-03-14
Fix: use broadcast_lesser in place of comparisons in ISQF

vincentqb:
*Issue #, if available:*

When running in hybridized form, I'm running into shape inference errors where those comparisons are done.

*Description of changes:*

I added explicit `broadcast_like` in those comparisons to resolve this issue.

* There are other comparisons that may need fixing.
* This might be better fixed upstream as part of mxnet.

### DIFF
--- a/src/gluonts/mx/distribution/isqf.py
+++ b/src/gluonts/mx/distribution/isqf.py
@@ -384,7 +384,9 @@ class ISQF(Distribution):
         return quantile
 
     def quantile_spline(
-        self, alpha: Tensor, axis: Optional[int] = None,
+        self,
+        alpha: Tensor,
+        axis: Optional[int] = None,
     ) -> Tensor:
         r"""
         Evaluates the spline functions at the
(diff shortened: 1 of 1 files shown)


lostella:
@vincentqb thanks! Alternatively, it should be possible to use `broadcast_` comparison functions such as https://mxnet.apache.org/versions/1.8.0/api/python/docs/api/ndarray/ndarray.html#mxnet.ndarray.broadcast_lesser

Maybe that would be slightly more readable?

Also make sure to run `black` (22.1.0, see [here](https://github.com/awslabs/gluon-ts/blob/bd78b347b54a75562e684525cc0f494150f42691/requirements/requirements-setup.txt#L1)) over the source files before committing

lostella:
@vincentqb looks like something was off about `black` (maybe a different version?) I fixed it. If you confirm the change looks good to you, I think this can be reviewed.

lostella:
@vincentqb looks like something was off about `black` (maybe a different version?) I fixed it. If you confirm the change looks good to you, I think this can be reviewed.

vincentqb:
LGTM :)

vincentqb:
LGTM :)

lostella (approved):
looks good to me too! 🚀

PR #1923, opened 2022-03-16
gnninf

vgsatorras:
*Description of changes:*
This pull request introduces the multivariate model from [Multivariate Time Series Forecasting with Latent Graph Inference
](https://arxiv.org/abs/2203.03423) coded in Pytorch.

**Label:** New feature

### DIFF
--- a/src/gluonts/torch/model/gnninf/estimator.py
+++ b/src/gluonts/torch/model/gnninf/estimator.py
@@ -1,0 +1,100 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+# If you use this code in your work please cite:
+# Multivariate Time Series Forecasting with Latent Graph Inference (https://arxiv.org/abs/2203.03423)
+
+
+import torch
+from gluonts.dataset.common import Dataset
+from gluonts.model.forecast_generator import DistributionForecastGenerator
+from gluonts.torch.model.predictor import PyTorchPredictor
+from . import trans
+from .module import GNNInfModule
+
+
+class GNNInfEstimator:
+    def __init__(
+        self,
+        freq: str,
+        prediction_length: int,
+        context_length: int,
+        batch_size: int,
+        num_batches_per_epoch: int,
+        epochs: int,
+        lr: float,
+        gnn_name: str,
+        nf: int,
+        gnn_layers: int,
+        device: str='cpu',
+    ) -> None:
+        self.freq = freq
+        self.prediction_length = prediction_length
+        self.context_length = context_length
+        self.batch_size = batch_size
+        self.num_batches_per_epoch = num_batches_per_epoch
+        self.epochs = epochs
+        self.lr = lr
+        self.gnn_name = gnn_name # gnn or bpgnn
+        self.nf = nf
+        self.gnn_layers = gnn_layers
+        self.device = device
+
+    def train_model(self, train_dataset: Dataset, window_loss: int=10) -> GNNInfModule:
+        # Build dataloaders
+        data_loader = trans.wrap_with_dataloader(train_dataset, self.batch_size, self.num_batches_per_epoch,
+                                                 self.prediction_length, self.context_length)
+        n_nodes = len(train_dataset.list_data[0]['target'])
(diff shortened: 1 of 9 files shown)


src/gluonts/torch/model/gnninf/estimator.py
```diff
@@ -0,0 +1,100 @@
...
+
+import torch
+from gluonts.dataset.common import Dataset
+from gluonts.model.forecast_generator import DistributionForecastGenerator
+from gluonts.torch.model.predictor import PyTorchPredictor
+from . import trans
+from .module import GNNInfModule
+
+
+class GNNInfEstimator:
```
lostella, line 26: I think it would make sense to inherit at least from the `Estimator` base class here.

There is also `PyTorchLightningEstimator` [here](https://github.com/awslabs/gluon-ts/blob/8f948c65c25ea888da4b6165fc3ec3fa487fb6b6/src/gluonts/torch/model/estimator.py#L38), but that requires you to set up a `LightningModule` class to wrap the model. The advantage is that you don't need to have your own training loop. But if you need to do something very custom about training, then it make sense to directly implement the `train` method your own way

src/gluonts/torch/model/gnninf/estimator.py
```diff
@@ -0,0 +1,100 @@
...
+                loss = -distr.log_prob(target.to(net.device))
+                loss = loss.mean()
+
+                loss.backward()
+                optimizer.step()
+                loss_arr.append(loss.detach().item())
+                n_samples += 1
+                if idx % window_loss == 0:
+                    avg_loss = sum(loss_arr[- window_loss:]) / len(loss_arr[- window_loss:])
+                    print('Epoch %d \t Iteration %d/%d: \t Avg Loss (10 last iterations): %.4f' % (epoch, idx, data_loader.length, avg_loss))
```
lostella, line 81: I think it's better to use a logger for this, see e.g. https://github.com/awslabs/gluon-ts/blob/8f948c65c25ea888da4b6165fc3ec3fa487fb6b6/src/gluonts/mx/trainer/_base.py#L411-L415

src/gluonts/torch/model/gnninf/estimator.py
```diff
@@ -0,0 +1,100 @@
...
+        return PyTorchPredictor(
+            prediction_length=self.prediction_length,
+            freq=self.freq,
+            input_names=["past_target"],
+            prediction_net=net,
+            batch_size=batch_size,
+            input_transform=input_transform,
+            forecast_generator=DistributionForecastGenerator(net.distr_output),
+            device=net.device,
+        )
```
lostella, line 95: Ah! I see you've been inspired by the `FeedForward` example from the tests

lostella:
Make sure to install `black==22.1.0` and to run

```
black src test
```

before committing your code, so that it takes care of some styling issues

lostella:
There's also a few complaints from `flake8`, see the bottom of [this log file](https://pipelines.actions.githubusercontent.com/LG0kBijBY35Fqw47Be7JAOcSx2cowoyb7WTTLj8BI5oz3i77sX/_apis/pipelines/1/runs/19695/signedlogcontent/2?urlExpires=2022-03-18T11%3A37%3A32.3331116Z&urlSigningMethod=HMACV1&urlSignature=kDh3%2FuBNnmGyS38YJzZCZSU5JQ8%2BYrQSOeS2ivkhFSQ%3D)

src/gluonts/torch/model/gnninf/module.py
```diff
@@ -0,0 +1,94 @@
...
+from typing import Tuple, Any
+
+
+def mean_abs_scaling(
+    context: torch.Tensor, min_scale: float = 1e-3
+) -> torch.Tensor:
+    return context.abs().mean(1).clamp(min_scale, None).unsqueeze(1)
+
+
+class GNNInfModule(nn.Module):
```
lostella, line 33: I think it would be nice to document at least also this class, since it appears like the main entry point for the model in case someone wants to interact with it directly (that is, not through the estimator class)

src/gluonts/torch/model/gnninf/networks/model.py
```diff
@@ -0,0 +1,234 @@
...
+
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+
+class GNNInfNetwork(torch.nn.Module):
+    """
+    Input and Output tensor will have the shape:
+    (batch_size, num_nodes, seq_len, dimensions)
+    :param in_channels: number of channels/dimensions in the input sequence
+    (this value is usually set to 1)
```
lostella, line 31: Better to use [numpy-style docstrings](https://numpydoc.readthedocs.io/en/latest/format.html#style-guide), see for example https://github.com/awslabs/gluon-ts/blob/ae4cfdef539e49f93a87034aa2f2bec194c4b7d8/src/gluonts/model/deepar/_estimator.py#L69-L89

lostella (changes requested):
Sorry for the late reply! I only have comments here and there about docstrings, I think it would be nice to document the estimator and all `nn.Module` classes, at least the ones that could in principle be re-used for other models.

Plus the `lead_time` thing. After this I think this can be merged!

kashif:
@vgsatorras do you have some helpers for the PEMS and METR-LA datasets for gluonts? Thanks!

PR #1924, opened 2022-03-21
Porting SimpleFeedForwardEstimator to PyTorch

lostella:
* Bring the pytorch-based feedforward model from the test folder into the package, with some additional wrapping similarly to the pytorch-based DeepAR implementation: a LightningModule to allow training via PyTorch Lightning, and an estimator class gluing everything together.
* Remove the test where the model was previously defined, and adjusts a bit an existing test where other torch-based estimators are covered.


src/gluonts/torch/model/simple_feedforward/module.py
```diff
@@ -0,0 +1,85 @@
...
+        modules.append(
+            self.__make_lin(
+                dimensions[-1], prediction_length * hidden_dimensions[-1]
+            )
+        )
+
+        self.nn = nn.Sequential(*modules)
+        self.args_proj = self.distr_output.get_args_proj(hidden_dimensions[-1])
+
+    @staticmethod
```
jaheba, line 67: I'm generally against `staticmethod`, couldn't this just be a funcction?

src/gluonts/torch/model/simple_feedforward/estimator.py
```diff
@@ -0,0 +1,213 @@
...
+class SimpleFeedForwardEstimator(PyTorchLightningEstimator):
+    @validated()
+    def __init__(
+        self,
+        freq: str,
+        prediction_length: int,
+        context_length: Optional[int] = None,
+        hidden_dimensions: Optional[
+            List[int]
+        ] = None,  # TODO enforce same defaults to network and estimator
```
jaheba, line 68: Let's put the comment above the line to avoid the line break.

jaheba (approved):

kashif:
@lostella if you are doing this you might as well implement ARNet... which I believe involves a regularizer as well...

lostella:
@kashif maybe separately, I think we can keep this model here mostly as an instructional example on how a network in framework X (for X \in {MXNet, PyTorch}) can be made interact with the rest of GluonTS

kashif:
cool cool! no worries!

Issue #1928, opened 2022-03-22
Support for multiple workers in the validation data loader


lostella:
The data loader used for validation data does not use multiple workers like the one for training data. When training models on a GPU, multiple workers are useful to prevent GPU starvation. As a consequence, the validation loop ends up slowing down the training process considerably.

PR #1929, opened 2022-03-23
Add multiprocessing to validation data loader

lostella:
Addresses #1928 

Fixes multiprocessing iterator so that it can be iterated over multiple times. This requires a little more synchronization between workers and main process, but it’s pretty simple.

To do:
- [x] Propagate multiprocessing options appropriately in the estimator types
- [x] Run experiment to confirm the speedup

### DIFF
--- a/src/gluonts/dataset/loader.py
+++ b/src/gluonts/dataset/loader.py
@@ -77,7 +77,7 @@ def worker_fn(
 DataLoader = Iterable[DataBatch]
 
 
-class MultiProcessIterable(DataLoader):
+class MultiProcessLoader(DataLoader):
     def __init__(
         self,
         dataset: Dataset,
@@ -208,12 +208,13 @@ def TrainDataLoader(
     transformed_dataset = transform.apply(dataset, is_train=True)
 
     if num_workers is not None:
-        return iter(MultiProcessIterable(
+        loader = MultiProcessLoader(
             transformed_dataset,
             decode_fn=decode_fn,
             num_workers=num_workers,
             max_queue_size=num_prefetch,
-        ))
+        )
+        batches = iter(loader)
     else:
         batches = iter(transformed_dataset)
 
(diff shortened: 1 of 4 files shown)


lostella:
I tested this on both macOS and Linux (with GPU), and everything works fine. The one issue I see now is that spawning processes at each validation loop adds a few seconds of overhead. I'll see if there's a clean way to avoid that, and have processes spawned at construction time only.

jaheba:
Do we need to create a new loader for each loop iteration? How do we handle this for the training loader?

src/gluonts/dataset/loader.py
```diff
@@ -89,42 +94,39 @@ def __init__(
...
-            mp.Process,
-            target=worker_fn,
-            kwargs={
-                "dataset": dataset,
-                "result_queue": self.result_queue,
-                "num_workers": num_workers,
-            },
-        )
+        self.queue_timeout_seconds = queue_timeout_seconds
+        self.output_queue = mp.Manager().Queue(maxsize=max_queue_size)
```
jaheba, line 98: Does it matter if we create multiple managers?

Would something change if we just use one?

lostella: Good spot, I'll update this. We were using just one queue so far, that's why it was created on the spot

lostella:
Training batch composition is unaffected by the change (the following example is with 2 workers)

Before:

image

After:

image

jaheba:
What's the dataset used?

lostella:
> What's the dataset used?

Electricity

jaheba (approved):
LGTM, thanks!

kashif:
@lostella did you also test via jupyter? sometimes I have the feeling with a lot of multiprocessors it slows things down.... anyways looking forward to testing this! Thanks!

lostella:
@kashif no I didn't, and I'm not sure how multiprocessing is dealt with in the case of notebooks. It may be that spawning processes there adds even more significant overhead, I'm not sure. I tested this using scripts, and the speedup is apparent whenever the network would otherwise starve (e.g. CNNs on GPU).

Arguably, notebooks are not an ideal environment where to run performance-critical code (like training loops that need to run for several hours). Or do you have some specific use-case in mind?

Issue #1861, opened 2022-01-20
Predictor error with lead_time  >0


FocusLiwen:
## Description
(A clear and concise description of what the bug is.)
When I used lead_time = 10 to train deepar, or mqcnn model, I got errors which calling 
forecast_it, ts_it = make_evaluation_predictions(val, predictor=predictor).
The whole time series is 1261 long for val.

## To Reproduce
(Please provide minimal example of code snippet that reproduces the error. For existing examples, please provide link.)

```python
put code here
```
        predictor = estimator.train(training_data=train)
        print("predictor lead time after train before set up:", predictor.lead_time)
        predictor.lead_time = self.estimator_config.get("lead_time")
        print("predictor lead time after  set up:", predictor.lead_time)
        forecast_it, ts_it = make_evaluation_predictions(train, predictor=predictor)
        
        tss = list(ts_it)
        forecasts = list(forecast_it)

## Error message or code output
(Paste the complete error message, including stack trace, or the undesired output that the above snippet produces.)

```
put error or undesired output here
```

The error messages are:
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)

/opt/conda/lib/python3.7/site-packages/gluonts/mx/model/predictor.py in predict(self, dataset, num_samples, num_workers, num_prefetch, **kwargs)
    169                 freq=self.freq,
    170                 output_transform=self.output_transform,
--> 171                 num_samples=num_samples,
    172             )
    173 

/opt/conda/lib/python3.7/site-packages/gluonts/model/forecast_generator.py in __call__(self, inference_data_loader, prediction_net, input_names, freq, output_transform, num_samples, **kwargs)
    131         **kwargs
    132     ) -> Iterator[Forecast]:
--> 133         for batch in inference_data_loader:
    134             inputs = [batch[k] for k in input_names]
    135             outputs = predict_to_numpy(prediction_net, inputs)

/opt/conda/lib/python3.7/site-packages/gluonts/transform/_base.py in __iter__(self)
    102     def __iter__(self) -> Iterator[DataEntry]:
    103         yield from self.transformation(
--> 104             self.base_dataset, is_train=self.is_train
    105         )
    106 

/opt/conda/lib/python3.7/site-packages/gluonts/transform/_base.py in __call__(self, data_it, is_train)
    121         self, data_it: Iterable[DataEntry], is_train: bool
    122     ) -> Iterator:
--> 123         for data_entry in data_it:
    124             try:
[...]

PR #1932, opened 2022-03-30
Adding lead_time to torch DeepAR and trivial predictor

raffienficiaud:
*Issue #1861 (related but not solving)*

*Description of changes:*

* adding lead_time to DeepAR pytorch implementation. 
* adding lead_time to trivial/`ConstantValuePredictor`

I am pretty new to GluonTS and it is likely I am missing some parts for implementing this properly. Those involve my current understanding of the framework as well as the DeepAR model (as per [this comment](https://github.com/awslabs/gluon-ts/issues/1861#issuecomment-1040102782)).
The changes "just work" on my setup. 

Note: I am not able to run `mypy` as the number of failures is high and they are unrelated to the changes I am making. Same for `pytest`. `black` has be run over the changes.


lostella:
Thanks for starting this! (And for the related deep dive in #1936!)

> The changes "just work" on my setup.

I understand the changes to the splitters, which make room for the additional `lead_time` time point at the end when iterating and splitting the data. At a high level those appear to be the required ones.

However, for DeepAR, I'm not sure that the right thing is happening under the hood: thinking about how the underlying RNN is used, what I mentioned in [my comment](https://github.com/awslabs/gluon-ts/issues/1861#issuecomment-1040102782) should translate into the DeepAR module doing the following:
* have a `lead_time` parameter too
* the `forward` method producing `lead_time + prediction_length` many samples (along the time axis), only to chop-off the first `lead_time`
* related changes to the lightning module used for training

Given this is it's not a trivial change, I think it would be nice to have some evidence showing that the estimator and module are doing the right thing with `lead_time` (for training and prediction, respectively). A simple example with some artificial, periodic data could be enough maybe.

Note that this results in the model doing all the work for the entire `lead_time + prediction_length` range, only discarding part of it. Arguably, a simpler way to get the same result is to leave the code as it is, set the estimator's `prediction_length` to `lead_time + prediction_length` when constructing it, and just post-process the forecasts that come out of the model (ignore their initial `lead_time` length).

PR #1935, opened 2022-04-01
added support for pandas dataframes

### LINKED ISSUE #418: Pandas-backed Dataset types
Pandas is the go-to library to do lots of things with data, and GluonTS should probably provide easy ways to use its data structures directly in models.

One way to do this is to provide an appropriate `Dataset` type that feeds from Pandas structures, and produces `DataEntry` objects (the dictionaries that GluonTS models use) when iterated over.

### A single DataFrame as a Dataset

The simplest use case would be to be able to use a single `DataFrame` as `Dataset`, i.e. to consider each column in it as time series to train, test, or make predictions on. One could even specify a subset of the series to be used, using indices or column names for example.

This has some apparent restrictions:
* All time series share the same index
* It's probably hard to specify per-series features

On the other hand, this would allow doing stuff like

```python
df = pd.read_csv("data.csv")
training_dataset = PandasDataset(df)
estimator.train(training_dataset)
```

which I think would be quite nice.

### One DataFrame per DataEntry

This would allow to overcome some limitations of the previous approach: each `DataFrame` has its own index, so each `DataEntry` spans its own time range, and one could indicate which column(s) should be used as target, which as dynamic features.

Some limitations here would remain:
* How to provide static features?

### Other observations

Given that we assume uniformly spaced data, accepted `DataFrame`s would probably need to have a `DatetimeIndex` type of index, such as the ones returned by `pd.date_range` (see [here](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.date_range.html?highlight=date_range#pandas.date_range)).

### LINKED ISSUE #1930: Pandas-based datasets
Many users have tabular datasets in csv, parquet, excel, ... . By having a Dataset-class which works directly on tabular data users do not need to convert data to gluonts-dataentry. That would make it way easier for new users to use gluonts. Also old users can skip the conversion part and can directly start modeling.

# User-facing API (Example)
Here is how an user can interact with gluonts:
```python
import pandas as pd
from gluonts.dataset.common import LongDataset, WideDataset # see long/wide-table definition below

df = pd.read_csv("long_df.csv")
dataset = LongDataset(data=df, target="target", timestamp="time", item_id="item", freq="1h")
# dataset = WideDataset(data=df, freq="1h")

model = DeepAREstimator(freq="1h", prediction_length=20)
predictor = model.train(dataset)
predictions = list(predictor.predict(dataset))
```

# Dataframe examples

The tabular data the user has can have many different formats. In the following I ll discuss two of them. Long- and the Wide-format, which will cover 99% of the cases, I guess.

## Long-table
A DataFrame with timestamp, target and item_id columns.

|time|target|item|stat_cat_1|dyn_real_1|
|----|----|----|----|----|
|1750-01-01 00:00:00|-0.21|A|0|0.79|
|1750-01-01 01:00:00|-0.33|A|0|0.59|
|1750-01-01 02:00:00|-0.33|A|0|0.39|
|1750-01-01 01:00:00|-1.24|B|1|-0.60|
|1750-01-01 02:00:00|-1.37|B|1|-0.91|

## Wide-table
A DataFrame with timestamps as index and each column corresponds to a time series. item_id is the column name.

### Target
||A|B|
|---|---|---|
|1750-01-01 00:00:00|-0.21|NaN|
|1750-01-01 01:00:00|-0.33|1.94|
|1750-01-01 02:00:00|-0.33|2.28|

### Dynamic features (cat and real)
Dynamic features are provided in a separate dataframes of the same size as the target-dataframe. For mulitple features multiple dataframes are provided.
||A|B|
|---|---|---|
|1750-01-01 00:00:00|0.79|NaN|
|1750-01-01 01:00:00|0.59|-0.60|
|1750-01-01 02:00:00|0.39|-0.91|

### Static features (cat and real)
Static features are also provided in a separate dataframe. For multiple features we have multiple rows in the same dataframe.
||A|B|
|---|---|---|
|static_cat_1|0|1|

# API

As in the example above we can provide a LongDataset and a WideDataset.

For the LongDataset we have all in one dataframe and the user only need to provide information about what the values in a column represent (target, timestamp, item_id, feature):
```python
dataset = LongDataset(
  data: pd.DataFrame,
  target: str,
  timestamp: str,
  item_id: str,
  freq: str,
  feat_dynamic_real: List[str] = None,
  feat_dynamic_cat: List[str] = None,
  feat_static_real: List[str] = None,
  feat_static_cat: List[str] = None,
  transform_data: bool = False,  # see the section 'handling prediction-data' below
)
```

For the WideDataset we split targets and features into different dataframes and provide the appropriate dataframe directly as input (remember that timestamps are the index and each column corresponds to a time series):
```python
dataset = WideDataset(
  target: pd.DataFrame,
  freq: str,
  feat_dynamic_real: List[pd.Datframe] = None,
  feat_dynamic_cat: List[pd.Dataframe] = None,
  feat_static_real: List[pd.DataFrame] = None,  # List[pd.Series] is also fine
  feat_static_cat: List[pd.DataFrame] = None,  # List[pd.Series] is also fine
  transform_data: bool = False, # see the section 'handling prediction-data' below
)
```

# Gluonts-internal representation

We can implement one Dataset class for the long-format and one for wide-format.
Inheriting from `gluonts.dataset.common.Dataset` and implementing `__iter__` and `__len__` will allow a direct interaction with models and predictors. There is no need for the user to know about `DataEntry`. All conversions are taken care of by the new Datasets-classes.

For long format the `__iter__` is more or less `map(long_to_dataentry, df.groupby(item_id))`.
For wide format the `__iter__` is more of less `map(wide_to_dataentry, target.iteritems(), dynfeat1.iteritems(), statfeat1.iteritems(), ...)`.

`long_to_dataentry` and `wide_to_dataentry` will contain all the logic needed for the conversion, handling missing values and other type of pre/post processing.

# Handling prediction-data (transform-data)
Prediction-data here is some data where target length is not the same as dynamic feature length.

In the case there are no dynamic features. We don't have to do anything in addition to the above mentioned procedure.
If dynamic features are present, the processing is slightly different for prediction-data. The flag `transform_data` (is there any standardized name for this?) will indicate if the 'slightly different' processing is triggered.

# Todos:
- In the first PR we can support target only for train, test and prediction.
- We can extend that in the next PR to support dynamic and static features for train and test.
- After that, we can discuss and implement the design for supporting dynamic features for predictions.

RSNirwan:
PR contains implementation for the LongDataset format proposed in #1930.
It works for:
 - targets (univariate and multivariate)
 - dynamic/static categorical/real features
 - prediction data (specifically target and past_feat_dynamic_real) is taken care of
 
resolves #1930, resolves #418 


src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,226 @@
...
+        return self.process(dataentry)
+
+    def __iter__(self) -> Iterator[DataEntry]:
+        def func(x):
+            dataentry = self._dataentry(x[0], x[1])
+            if not self.prediction_length:
+                return dataentry
+            return prepare_prediction_data(dataentry, self.prediction_length)
+
+        return map(func, self.data.groupby(self.item_id))
```
jaheba, line 115: Couldn't you just use `yield` here?

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,226 @@
...
+        item_id: str,
+        freq: str,
+        feat_dynamic_real: List[str] = None,
+        feat_dynamic_cat: List[str] = None,
+        feat_static_real: List[str] = None,
+        feat_static_cat: List[str] = None,
+        past_feat_dynamic_real: List[str] = None,
+        prediction_length: int = None,
+    ) -> None:
+        self.data = data.copy()
```
jaheba, line 76: Is there an actual need to copy the data?

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,226 @@
...
+from typing import Iterator, List, Union
+from datetime import datetime
+
+import pandas as pd
+
+from gluonts.dataset.common import Dataset, DataEntry, ProcessDataEntry
+from gluonts.dataset.field_names import FieldName
+
+
+class DataFrameDataset(Dataset):
```
jaheba, line 23: How do you feel about using a `dataclass`?

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,226 @@
...
+        self.freq = freq
+        self.prediction_length = prediction_length
+        self.feat_dynamic_real = feat_dynamic_real
+        self.feat_dynamic_cat = feat_dynamic_cat
+        self.feat_static_real = feat_static_real
+        self.feat_static_cat = feat_static_cat
+        self.past_feat_dynamic_real = past_feat_dynamic_real
+
+        self.one_dim_target = not isinstance(self.target, list)
+        self.process = ProcessDataEntry(
```
jaheba, line 89: I don't like how we do it in the other dataset classes, so I'm also not a fan having it here. Since it just does some type conversions (AFAIR), is there truly a need for this here, or could we just leverage pandas for this?

src/gluonts/dataset/tabular.py
RSNirwan: No, we only read from `data`. I removed the copy.

src/gluonts/dataset/tabular.py
RSNirwan: done. Is much cleaner now. Thanks!

src/gluonts/dataset/tabular.py
RSNirwan: The main reason, I use it, is to have the 'start' field processing in one place rather than making the DataFrameDataset to have its own logic for conversion. Meanwhile, it also does the conversions for other fields.
ProcessDataEntry is standardizing Dataset-outputs, which is quite nice. Whether to have the ProcessDataEntry as stadardizing mechanism or to implement a new one can maybe be discussed on another thread. What do you think?

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,213 @@
...
+            feat_dynamic_cat=self.feat_dynamic_cat,
+            feat_static_real=self.feat_static_real,
+            feat_static_cat=self.feat_static_cat,
+            past_feat_dynamic_real=self.past_feat_dynamic_real,
+        )
+        dataentry["item_id"] = item_id
+        return dataentry
+
+    def __iter__(self) -> Iterator[DataEntry]:
+        for item_id, df in self.data.groupby(self.item_id):
```
lostella, line 96: How about this class is constructed directly with a collection of `DataFrame` objects? For example, a dictionary `{item_id: df}` (or a list in case one doesn't need or want to associate an `item_id` to each series). And then to handle the long format we add a thin wrapper that does the `groupby`?

Something like
* `DataFramesDataset` (note the `s`)
* `LongDataFrameDataset`

where essentially

```python
def LongDataFrameDataset(df, ..., item_id_column):
    return DataFramesDataset(list(df.groupby(item_id_column)))
```

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    Removes `prediction_length` values from `target` and `past_feat_dynamic_real`.
+    Works in univariate and multivariate case.
+
+    >>> prepare_prediction_data({"target": [1., 2., 3., 4.]}, prediction_length=2)
+    {'target': [1.0, 2.0]}
+    """
+
+    def cut(
+        x: Union[List[float], List[List[float]]]
```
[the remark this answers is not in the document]
RSNirwan: It was a list before. Know I do the conversion to `np.ndarray` before entering this function and cut became a trivial indexing `[..., :-prediction_length]`. Thanks!

src/gluonts/dataset/tabular.py
RSNirwan: Done. DataFramesDataset supports both. `List` and `Dict` now.

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+from typing import Iterator, List, Union, Dict
+
+import pandas as pd
+
+from gluonts.dataset.common import Dataset, DataEntry, ProcessDataEntry
+from gluonts.dataset.field_names import FieldName
+
+
+@dataclass
+class DataFramesDataset(Dataset):
```
jaheba, line 25: Couldn't we name this `TabularDataset`?

Isn't the dataframe aspect an internal implementation detail?

lostella: > Isn't the dataframe aspect an internal implementation detail?

The object is constructed by giving it a `DataFrame`, so...

babaniyi:
Thanks for the enhancement @RSNirwan, I look forward to when it will be approved.
I am curious about what happens if we have a dataframe that doesn't fit into (pandas) memory, what do you think could be done? (The only solution coming to mind is using generators or something similar). 

@lostella your input will be appreciated.

lostella:
@babaniyi a generator of some kind would be the way to go: for example you could use the built-in [`FileDataset`](https://github.com/awslabs/gluon-ts/blob/925d44c71b7ed85f0154afe22198a9cfc30713eb/src/gluonts/dataset/common.py#L141) class.

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+
+    This class is constructed with a collection of pd.DataFrame-objects where each
+    DataFrame is representing one time series. A target and a timestamp columns are
+    essential. Furthermore, static/dynamic real/categorical features can be specified.
+
+    Parameters
+    ----------
+    dataframes: Union[List[pd.DataFrame], Dict[str, pd.DataFrame]]
+        List or Dict of 'pd.DataFrame's containing at least timestamp and target columns.
+        If a Dict is provided, the key will be the associated 'item_id'.
```
lostella, line 37: Double-backticks to get `monospace` font in sphinx-generated docs

```suggestion
        List or Dict of ``pd.DataFrame``s containing at least timestamp and target columns.
        If a Dict is provided, the key will be the associated ``item_id``.
```

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+    target: str or List[str]
+        Name of the column that contains the target time series.
+        For multivariate targets `target` is expecting a list of column names.
```
lostella, line 40:
```suggestion
        For multivariate targets, a list of column names should be provided.
```

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+    """
+    A pd.DataFrame-based dataset type.
+
+    This class is constructed with a collection of pd.DataFrame-objects where each
+    DataFrame is representing one time series. A target and a timestamp columns are
+    essential. Furthermore, static/dynamic real/categorical features can be specified.
+
+    Parameters
+    ----------
+    dataframes: Union[List[pd.DataFrame], Dict[str, pd.DataFrame]]
```
lostella, line 35: Would it make sense to also support a single `DataFrame`? That would be a dataset with a single entry: while probably unsuited for training large models, this may be something people often want to use (I have a single series, I want predictions for it)

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+        For target and past dynamic features last `prediction_length` elements are
+        removed when iterating over the data set. This becomes important when the
+        predictor is called.
+
+    Returns
+    -------
+    DataFramesDataset
+        A dataset based on a collection of pd.DataFrame-objects.
+    """
+    dataframes = dict(list(dataframe.groupby(item_id)))
```
lostella, line 162: Nice!

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+    feat_dynamic_real: List[str]
+        List of column names that contain dynamic real features.
+    feat_dynamic_cat: List[str]
+        List of column names that contain dynamic categorical features.
+    feat_static_real: List[str]
+        List of column names that contain static real features.
+    feat_static_cat: List[str]
+        List of column names that contain static categorical features.
+    past_feat_dynamic_real: List[str]
+        List of column names that contain dynamic real features only for the history.
```
lostella, line 54: I understand the need for this: we want to cover the case where some dynamic feature is only available in the past, say, up to time `T`. But then in that case, wouldn't one better provide both features and target up to `T` and set `prediction_length=0`?

lostella (commented):
Thanks @RSNirwan! This looks great, I only have some comments here and there. Do you think we should add some example notebook to the documentation for this feature in this PR, or would you do it in a separate one?

I think adding documentation material is a great way to find out otherwise-hidden limitations or pain points. The biggest "issue" seems to be the target-vs-feature length story at prediction time, which adds a little complexity, but solving that may require thinking about this problem more in general.

src/gluonts/dataset/tabular.py
jaheba: The `list` in there looks unnecessary.

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+import pandas as pd
+
+from gluonts.dataset.common import Dataset, DataEntry, ProcessDataEntry
+from gluonts.dataset.field_names import FieldName
+
+
+@dataclass
+class DataFramesDataset(Dataset):
+    """
+    A pd.DataFrame-based dataset type.
```
jaheba, line 27: I think it's better to write out `pandas`.

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+
+    This class is constructed with a collection of pd.DataFrame-objects where each
+    DataFrame is representing one time series. A target and a timestamp columns are
```
jaheba, line 30: Can you put things like "target" and "timesamp" into back-ticks?

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+        List of column names that contain dynamic real features.
+    feat_dynamic_cat: List[str]
+        List of column names that contain dynamic categorical features.
+    feat_static_real: List[str]
+        List of column names that contain static real features.
+    feat_static_cat: List[str]
+        List of column names that contain static categorical features.
+    past_feat_dynamic_real: List[str]
+        List of column names that contain dynamic real features only for the history.
+    prediction_length: int
```
jaheba, line 55: This looks wrong. A dataset should be independent from a given `prediction_length`.

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+    freq: str
+    feat_dynamic_real: List[str] = field(default_factory=list)
+    feat_dynamic_cat: List[str] = field(default_factory=list)
+    feat_static_real: List[str] = field(default_factory=list)
+    feat_static_cat: List[str] = field(default_factory=list)
+    past_feat_dynamic_real: List[str] = field(default_factory=list)
+    prediction_length: int = 0
+
+    def __post_init__(self) -> None:
+        self.one_dim_target = not isinstance(self.target, list)
```
jaheba, line 73: What about a list with one element?

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+        self.process = ProcessDataEntry(
+            self.freq, one_dim_target=self.one_dim_target
+        )
+        if isinstance(self.dataframes, dict):
```
jaheba, line 77: Can you add a comment how we store dataframes? What is the key of the dict used for?

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+                dataentry = prepare_prediction_data(
+                    dataentry, self.prediction_length
+                )
+            yield dataentry
+
+    def __len__(self) -> int:
+        return len(self._dataframes)
+
+
+def LongDataFrameDataset(
```
jaheba, line 111: If this is a function, why does it use camel-case?

src/gluonts/dataset/tabular.py
jaheba: I think that's not a very good argument -- we generally don't just call things by what we pass them as arguments.

Why call the module `tabular`, if it exposes this class?

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+        Frequency of observations in the time series. Must be a valid pandas frequency.
+    feat_dynamic_real: List[str]
+        List of column names that contain dynamic real features.
+    feat_dynamic_cat: List[str]
+        List of column names that contain dynamic categorical features.
+    feat_static_real: List[str]
+        List of column names that contain static real features.
+    feat_static_cat: List[str]
+        List of column names that contain static categorical features.
+    past_feat_dynamic_real: List[str]
```
jaheba, line 53: I hate this -- but I don't know a better solution at the moment.

We have hard-coded these field-names, but it would be great if we could support arbitrary fields.

One way to solve this could be to take a dict of fields instead

```py
DataSet(
    features={"feat_dynamic_real": [...])
)
```

src/gluonts/dataset/tabular.py
jaheba: We could have an alternative constructor for `DataFramesDataset` instead. What's the advantage to make this standalone?

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+    target: Union[str, List[str]],
+    timestamp: str,
+    feat_dynamic_real: List[str] = [],
+    feat_dynamic_cat: List[str] = [],
+    feat_static_real: List[str] = [],
+    feat_static_cat: List[str] = [],
+    past_feat_dynamic_real: List[str] = [],
+) -> DataEntry:
+    """
+    Converts a single time series (uni- or multi-variate) that is given in
```
jaheba, line 188: Should be "Convert ..."

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+        feat_dynamic_real=feat_dynamic_real,
+        feat_dynamic_cat=feat_dynamic_cat,
+        feat_static_real=feat_static_real,
+        feat_static_cat=feat_static_cat,
+        past_feat_dynamic_real=past_feat_dynamic_real,
+        prediction_length=prediction_length,
+    )
+
+
+def dataframeTS_to_dataentry(
```
jaheba, line 177: I think this could just be called `as_dataentry`

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+    set_field(FieldName.FEAT_STATIC_CAT, feat_static_cat, lambda x: x[0])
+    set_field(FieldName.PAST_FEAT_DYNAMIC_REAL, past_feat_dynamic_real)
+    return dataentry
+
+
+def prepare_prediction_data(
+    dataentry: DataEntry, prediction_length: int
+) -> DataEntry:
+    """
+    Removes `prediction_length` values from `target` and `past_feat_dynamic_real`.
```
jaheba, line 242: "Remove ..."

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+    Works in univariate and multivariate case.
+
+    >>> prepare_prediction_data({"target": np.array([1., 2., 3., 4.])}, prediction_length=2)
+    {'target': array([1., 2.])}
+    """
+    for fname in [FieldName.TARGET, FieldName.PAST_FEAT_DYNAMIC_REAL]:
+        if fname in dataentry:
+            dataentry[fname] = dataentry[fname][..., :-prediction_length]
```
jaheba, line 250: This looks like it's doing it in-place. What happens when you iterate twice over the dataset?

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+    return dataentry
+
+
+def check_timestamps(
```
jaheba, line 254: Can we find a better name? `Check` is soo generic.

src/gluonts/dataset/tabular.py
```diff
@@ -0,0 +1,269 @@
...
+    timestamps: List[Union[str, datetime]], freq: str
+) -> None:
```
jaheba, line 256: I think it's better to return a `bool` and then use `assert` at the call-site.

jaheba (approved):

src/gluonts/dataset/tabular.py
lostella: I disagree: what is being provided here is a very specific type of dataset, which sources data from a pandas dataframe; if anything, `FileDataset` is a misnomer since it doesn’t support just “any file” but a specific format (JSON lines).

The module could be renamed to something else: `pandas` or `dataframe`?

src/gluonts/dataset/tabular.py
lostella: But at this point there is still need for this, so maybe for now we can have a better name? `ignore_last_n_target` or something (seems too verbose)

src/gluonts/dataset/tabular.py
RSNirwan: The casting to list is indeed needed. Otherwise there is a TypeError.
```python
import numpy as np; import pandas as pd
a = pd.DataFrame(np.random.normal(size=(5,2)))
a["x"] = 3*['a'] + 2*['b']
#dict(a.groupby('x')) # not working
dict(list(a.groupby('x'))) # working
```

RSNirwan:
Thanks for the comments @jaheba and @lostella !  I hope I have addressed all comments as you intended. If misunderstood something, let me know.

src/gluonts/dataset/dataframe.py
```diff
@@ -0,0 +1,300 @@
...
+    feat_dynamic_real: List[str] = field(default_factory=list)
+    feat_dynamic_cat: List[str] = field(default_factory=list)
+    feat_static_real: List[str] = field(default_factory=list)
+    feat_static_cat: List[str] = field(default_factory=list)
+    past_feat_dynamic_real: List[str] = field(default_factory=list)
+    ignore_last_n_targets: int = 0
+
+    def __post_init__(self) -> None:
+        if isinstance(self.target, list) and len(self.target) == 1:
+            self.target = self.target[0]
```
RSNirwan, line 83: This will not allow for Multivariate-1-dimensional target. I.e. no support for matrix of size (1, T). It will always be represented as a vector of length T. Is that desired?

lostella: I think this is ok

src/gluonts/dataset/dataframe.py
```diff
@@ -0,0 +1,300 @@
...
+    def __len__(self) -> int:
+        return len(self._dataframes)
+
+
+@dataclass
+class LongDataFrameDataset(Dataset):
+    """
+    This dataset uses the long format for each variable. Target time series
+    values, for example, are stacked on top of each other rather than
+    side-by-side. The same is true for other dynamic or categorical features.
```
lostella, line 133: I think @jaheba meant to use an alternative constructor for the class above. Something like

```python
@classmethod
def from_long_dataframe(cls, df: pd.DataFrame, item_id: str, **kwargs) -> "DataFramesDataset":
    return cls(dataframes=dict(list(df.groupby(item_id))), **kwargs)
```

should allow saving the definition of this class entirely

lostella (changes requested):
Thanks @RSNirwan, I added just one comment which I believe should be addressed.

Otherwise it looks OK for now, and could be merged with that change. I think there is more room for improvement, but we can address it later when you add some documentation notebook for this.

For example, I think it would be nice to make most of the constructor arguments optional: if the provided dataframes have a `DatetimeIndex` index with known `freq`, then one could avoid specifying both the `timestamp` column and the `freq`. Maybe that could even be a strict requirement when going through the main constructor?

When converting from a long dataframe, however, timestamp column and frequency would be required I guess, since the `DatetimeIndex` would need to be constructed in that case (but the frequency can also be inferred, see https://pandas.pydata.org/docs/reference/api/pandas.infer_freq.html)

lostella (approved):

PR #1946, opened 2022-04-13
Breaking: Split incremental quantile output into separate class

### LINKED ISSUE #1944: Incorrect order of quantiles in Temporal Fusion Transformer
## Description
The values in the QuantileForecasts generated by the temporal fusion transformer do not coincide with the forecast_keys. I've looked into the issue and found that the QuantileOutput used in the network sorts the quantiles,

https://github.com/awslabs/gluon-ts/blob/cf58988b53e545962b35631cc0fa943498461fbf/src/gluonts/mx/block/quantile_output.py#L277

however the estimator does not sort the quantiles when creating the QuantileForecastGenerator

https://github.com/awslabs/gluon-ts/blob/cf58988b53e545962b35631cc0fa943498461fbf/src/gluonts/model/tft/_estimator.py#L446-L448

One potential solution could be to sort the quantiles in the QuantileForecastGenerator 
```python
forecast_generator=QuantileForecastGenerator(
                quantiles=sorted([str(q) for q in prediction_network.quantiles])` 
```
The above solution worked for me but I'm not sure if the final fix should be made there or elsewhere.
Note: The estimator internally generates quantiles from num_outputs, and they are disorderd. .e.g, `[0.5, 0.1, 0.9]` for `num_outputs=3`, or `[0.5, 0.1, 0.9, 0.2, 0.8]` for `num_outputs=5`.

## To Reproduce
I took this example from the readme and adapted it simply to show that the order of the forecast_keys is incorrect.

```python
import pandas as pd
from gluonts.model.tft import TemporalFusionTransformerEstimator


url = "https://raw.githubusercontent.com/numenta/NAB/master/data/realTweets/Twitter_volume_AMZN.csv"
df = pd.read_csv(url, header=0, index_col=0)

training_data = ListDataset(
    [{"start": df.index[0], "target": df.value[:"2015-04-05 00:00:00"]}],
    freq = "5min"
)

estimator = TemporalFusionTransformerEstimator(freq="5min", 
                                               prediction_length=2, 
                                               trainer=Trainer(epochs=10))
predictor = estimator.train(training_data=training_data)

test_data = ListDataset(
    [{"start": df.index[0], "target": df.value[:"2015-04-15 00:00:00"]}],
    freq = "5min"
)

forecast_it = predictor.predict(test_data)
print(next(iter(forecast_it)))
print(predictor.prediction_net.output.quantiles)
```

```
QuantileForecast(array([[43.30777 , 43.03252 ],
       [48.941864, 48.626392],
       [56.510143, 56.666008]], dtype=float32)), start_date=Timestamp('2015-04-15 00:00:00', freq='5T'), freq='5min', forecast_keys=['0.5', '0.1', '0.9'], item_id=None, info=None)
[0.1, 0.5, 0.9]
```

## Environment
- Operating system: Ubuntu 18.04.6 LTS
- Python version: 3.9.6
- GluonTS version: 0.9.3
- MXNet version: 1.9.0

lostella:
*Issue #, if available:* Fixes #1944

*Description of changes:* #1693 modified the behavior of `QuantileOutput` causing #1944. This PR recovers the former behavior, and splits the "incremental quantiles" case into a separate `IncrementalQuantileOutput` class (instead of having the `is_iqf` option). Surrounding changes to related classes are also there (projection layers used, MQ-DNN models).

This is a bit of a sensible change, so I'd like to add tests. Opening nevertheless to get early feedback @dcmaddix @youngsuk0723 

I also think that `IncrementalQuantileOutput` should avoid sorting the quantile levels, or at least allow the projection layer to sort them back, which would make the interface clearer: if I construct the class with quantile levels `[0.9, 0.1]`, maybe I expect the projection layer to output the 90% and 10% percentile in this order (principle of least surprise); but maybe we can address this separately, or in this PR after proper discussion. ***Edit:** sorting back the quantile values appears to be non-trivial with mxnet*

### DIFF
--- a/src/gluonts/model/seq2seq/_mq_dnn_estimator.py
+++ b/src/gluonts/model/seq2seq/_mq_dnn_estimator.py
@@ -240,14 +240,12 @@ class MQCNNEstimator(ForkingSeq2SeqEstimator):
             prefix="decoder_",
         )
 
-        if self.quantiles:
-            quantile_output = (
-                IncrementalQuantileOutput(self.quantiles)
-                if is_iqf
-                else QuantileOutput(self.quantiles)
-            )
-        else:
+        if not self.quantiles:
             quantile_output = None
+        elif is_iqf:
+            quantile_output = IncrementalQuantileOutput(self.quantiles)
+        else:
+            quantile_output = QuantileOutput(self.quantiles)
 
         super().__init__(
             encoder=encoder,
@@ -401,14 +399,12 @@ class MQRNNEstimator(ForkingSeq2SeqEstimator):
             prefix="decoder_",
         )
 
-        if self.quantiles:
-            quantile_output = (
-                IncrementalQuantileOutput(self.quantiles)
-                if is_iqf
-                else QuantileOutput(self.quantiles)
-            )
-        else:
+        if not self.quantiles:
             quantile_output = None
+        elif is_iqf:
+            quantile_output = IncrementalQuantileOutput(self.quantiles)
+        else:
+            quantile_output = QuantileOutput(self.quantiles)
 
         super().__init__(
             encoder=encoder,
--- a/src/gluonts/model/seq2seq/_mq_dnn_estimator.py
+++ b/src/gluonts/model/seq2seq/_mq_dnn_estimator.py
@@ -26,7 +26,10 @@ from gluonts.mx.block.encoder import (
     HierarchicalCausalConv1DEncoder,
     RNNEncoder,
 )
-from gluonts.mx.block.quantile_output import QuantileOutput
+from gluonts.mx.block.quantile_output import (
+    QuantileOutput,
+    IncrementalQuantileOutput,
+)
 from gluonts.mx.distribution import DistributionOutput
 from gluonts.mx.trainer import Trainer
 
@@ -237,11 +240,14 @@ class MQCNNEstimator(ForkingSeq2SeqEstimator):
             prefix="decoder_",
         )
 
-        quantile_output = (
-            QuantileOutput(self.quantiles, is_iqf=self.is_iqf)
-            if self.quantiles
-            else None
-        )
+        if self.quantiles:
+            quantile_output = (
+                IncrementalQuantileOutput(self.quantiles)
+                if is_iqf
+                else QuantileOutput(self.quantiles)
+            )
+        else:
+            quantile_output = None
 
         super().__init__(
             encoder=encoder,
@@ -395,11 +401,14 @@ class MQRNNEstimator(ForkingSeq2SeqEstimator):
             prefix="decoder_",
         )
 
-        quantile_output = (
-            QuantileOutput(self.quantiles, is_iqf=self.is_iqf)
-            if self.quantiles
-            else None
-        )
+        if self.quantiles:
+            quantile_output = (
+                IncrementalQuantileOutput(self.quantiles)
+                if is_iqf
+                else QuantileOutput(self.quantiles)
+            )
+        else:
+            quantile_output = None
 
         super().__init__(
             encoder=encoder,
--- a/src/gluonts/mx/block/quantile_output.py
+++ b/src/gluonts/mx/block/quantile_output.py
@@ -33,15 +33,15 @@ def crps_weights_pwl(quantile_levels: List[float]) -> List[float]:
 
     Under the assumption of linear interpolation
 
-        CRPS = sum_{i=0}^{n-1} 0.5 * (q_{i+1} - q_{i}) * (z_{i+1} + z_{i})
+    .. math:: CRPS = sum_{i=0}^{n-1} 0.5 * (q_{i+1} - q_{i}) * (z_{i+1} + z_{i})
 
-    where ``z_i`` is the i-th quantile prediction ``q_i``.
+    where :math:`z_i` is the i-th quantile prediction :math:`q_i`.
     The inner terms cancel due to the telescoping sum property and we obtain
 
-        CRPS = sum_{i=1}^n w_i z_i
+    .. math:: CRPS = sum_{i=1}^n w_i z_i
 
-    with the weights ``w_i = (q_{i+1}-q_{i-1})/2`` for ``i = 1, ..., n-1``,
-    ``w_0 = (q_1-q_0)/2`` and ``w_n = (w_n - w_{n-1})/2``.
+    with the weights :math:`w_i = (q_{i+1}-q_{i-1})/2` for :math:`i = 1, ..., n-1`,
+    :math:`w_0 = (q_1-q_0)/2` and :math:`w_n = (w_n - w_{n-1})/2`.
     """
     num_quantiles = len(quantile_levels)
 
--- a/src/gluonts/mx/block/quantile_output.py
+++ b/src/gluonts/mx/block/quantile_output.py
@@ -183,74 +183,6 @@ class QuantileLoss(Loss):
         return quantile_weights
 
 
-class ProjectParams(nn.HybridBlock):
-    """
-    Defines a dense layer to compute the projection weights into the quantile
-    space.
-
-    Parameters
-    ----------
-    num_quantiles
-        number of quantiles to compute the projection.
-    is_iqf
-        determines whether to use IQF or QF.
-    """
-
-    @validated()
-    def __init__(self, num_quantiles: int, is_iqf: bool = False, **kwargs):
-        super().__init__(**kwargs)
-
-        self.num_quantiles = num_quantiles
-        self.is_iqf = is_iqf
-        with self.name_scope():
-            if self.is_iqf:
-                self.proj_intrcpt = nn.Dense(1, flatten=False)
-                if self.num_quantiles > 1:
-                    self.proj_incrmnt = nn.Dense(
-                        self.num_quantiles - 1,
-                        flatten=False,
-                        activation="relu",
-                    )  # increments between quantile estimates
-            else:
-                self.projection = nn.Dense(
-                    units=self.num_quantiles, flatten=False
-                )
-
-    # noinspection PyMethodOverriding,PyPep8Naming
-    def hybrid_forward(self, F, x: Tensor) -> Tuple[Tensor, Tensor, Tensor]:
-        """
-
-        Parameters
-        ----------
-        F
-            A module that can either refer to the Symbol API or the NDArray
-            API in MXNet.
-        x
-            input tensor
-
-        Returns
-        -------
-        Tensor
-            output of the projection layer
-        """
-        return (
-            (
-                self.proj_intrcpt(x)
-                if self.num_quantiles == 1
-                else (
-                    F.cumsum(
-                        F.concat(
-                            self.proj_intrcpt(x), self.proj_incrmnt(x), dim=-1
-                        ),
-                        axis=3,
-                    )
-                )
-            )
-            if self.is_iqf
-            else self.projection(x)
-        )
-
-
 class QuantileOutput:
     """
     Output layer using a quantile loss and projection layer to connect the
@@ -263,9 +195,6 @@ class QuantileOutput:
 
     quantile_weights
         weights of the quantiles.
-
-    is_iqf
-        determines whether to use IQF or QF.
     """
 
     @validated()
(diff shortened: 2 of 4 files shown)


src/gluonts/model/seq2seq/_mq_dnn_estimator.py
```diff
@@ -237,11 +240,14 @@ def __init__(
...
 
-        quantile_output = (
-            QuantileOutput(self.quantiles, is_iqf=self.is_iqf)
-            if self.quantiles
-            else None
-        )
+        if self.quantiles:
+            quantile_output = (
+                IncrementalQuantileOutput(self.quantiles)
+                if is_iqf
```
jaheba, line 246: Not a fan of this inline `if` here, can't we do

```py
if not self.quantiles:
    quantile_output = None
elif is_iqf:
    quantile_output = IncrementalQuantileOutput(self.quantiles)
else:
    quantile_output = QuantileOutput(self.quantiles)
```

src/gluonts/model/seq2seq/_mq_dnn_estimator.py
```diff
@@ -395,11 +401,14 @@ def __init__(
...
```
jaheba, line 407: Same as above.

src/gluonts/model/seq2seq/_mq_dnn_estimator.py
lostella: Sure, why not

src/gluonts/model/seq2seq/_mq_dnn_estimator.py
```diff
@@ -237,11 +240,12 @@ def __init__(
...
 
-        quantile_output = (
-            QuantileOutput(self.quantiles, is_iqf=self.is_iqf)
-            if self.quantiles
-            else None
-        )
+        if not self.quantiles:
+            quantile_output = None
+        elif is_iqf:
+            quantile_output = IncrementalQuantileOutput(self.quantiles)
```
dcmaddix, line 246: Thanks this is nice and cleaner to have a separate class.

src/gluonts/mx/block/quantile_output.py
```diff
@@ -183,103 +183,104 @@ def compute_quantile_weights(self) -> List:
...
+        return QuantileLoss(
+            quantiles=self.quantiles,
+            quantile_weights=self.quantile_weights,
+            is_equal_weights=True,
+        )
 
-    # noinspection PyMethodOverriding,PyPep8Naming
-    def hybrid_forward(self, F, x: Tensor) -> Tuple[Tensor, Tensor, Tensor]:
-        """
+    def get_quantile_proj(self, **kwargs) -> nn.HybridBlock:
```
dcmaddix, line 217: Does this class still need the `hybrid_forward` returning `self.projection(x)` on line 250?

dcmaddix (commented):
Thanks for cleaning this up, Lorenzo! I just left some comments.

src/gluonts/mx/block/quantile_output.py
```diff
@@ -183,103 +183,104 @@ def compute_quantile_weights(self) -> List:
...
         quantiles: List[float],
         quantile_weights: Optional[List[float]] = None,
-        is_iqf: bool = False,
     ) -> None:
-        self.quantiles = sorted(quantiles)
-        self.num_quantiles = len(self.quantiles)
-        self.quantile_weights = quantile_weights
-        self.is_iqf = is_iqf
-        self.is_equal_weights = False if self.is_iqf else True
+        super().__init__(sorted(quantiles), quantile_weights)
```
dcmaddix, line 283: ok so we still for the sorting here. I think the quantile knots for interpolation need to be sorted, but we can handle that inside of the `InterpolationClass` in `utils`

lostella: Yes. Still I think it would be nice for the output of the projection (quantile values) to be in the originally intended order (the one specified by who constructed `IncrementalQuantileOutput`), and for that we could just reshuffle the quantiles back at the end of the projection.

src/gluonts/model/seq2seq/_mq_dnn_estimator.py
lostella: Yes I think this is better than a boolean flag. By the way, we could use this also in TFT to get rid of quantile crossing there too.

src/gluonts/mx/block/quantile_output.py
lostella: I believe that special case is still needed, because `self.proj_incrmnt` won't even be assigned as attribute in the constructor whenever `num_outputs == 1`

src/gluonts/model/seq2seq/_mq_dnn_estimator.py
dcmaddix: Yes, this should be used a generic class that can be combined on the last layer of any model, such as TFT, to resolve quantile crossing, not just specific to MQ-CNN.

src/gluonts/mx/block/quantile_output.py
```diff
@@ -183,103 +183,104 @@ def compute_quantile_weights(self) -> List:
...
+            else (
+                F.cumsum(
+                    F.concat(
+                        self.proj_intrcpt(x), self.proj_incrmnt(x), dim=-1
+                    ),
+                    axis=3,
                 )
             )
-            if self.is_iqf
-            else self.projection(x)
```
dcmaddix, line 250: Yes, it looks like this line is missing in `hybrid_forward` for the `QuantileOutput` class.

src/gluonts/mx/block/quantile_output.py
dcmaddix: yes, I agree that we should return it in the specified order, and just internally sort inside of the `InterpolationClass` where it is required.

src/gluonts/mx/block/quantile_output.py
```diff
@@ -20,13 +20,46 @@
 from gluonts.mx import Tensor
 
 
+def uniform_weights(x: list) -> List[float]:
```
jaheba, line 23: Can we add a more concrete type?

And let's try to avoid single letter variables. Especially `x` should not be a list.

src/gluonts/mx/block/quantile_output.py
```diff
@@ -20,13 +20,46 @@
...
+    return [1.0 / len(x) for _ in x]
```
jaheba, line 24:
```suggestion
    return [1.0 / len(x)] * len(x)
```

src/gluonts/mx/block/quantile_output.py
```diff
@@ -20,13 +20,46 @@
...
+
+
+def crps_weights_pwl(quantile_levels: List[float]) -> List[float]:
+    """
+    Computes the quantile loss weights corresponding to CRPS under linear
```
jaheba, line 29: Pedantic, but this should be "Compute ..."

src/gluonts/mx/block/quantile_output.py
```diff
@@ -20,13 +20,46 @@
...
+    interpolation assumption.
+
+    Quantile levels are assumed to be sorted in increasing order.
+
+    CRPS = sum_{i=0}^{n-1} 0.5 * (q_{i+1} - q_{i}) * (z_{i+1} + z_{i})
```
jaheba, line 34: Can we use backticks (\`) so that the comment reads more nicely in the generated docs?

src/gluonts/mx/block/quantile_output.py
```diff
@@ -20,13 +20,46 @@
...
+    under the assumption of linear interpolation, where z_i is the
+    ith quantile prediction q_i. The inner terms cancel due to the
+    telescoping sum property and we obtain CRPS = sum_{i=1}^n w_i z_i, with
+    the weights w_i = (q_{i+1}-q_{i-1})/2 for i = 1, ..., n-1,
+    w_0 = (q_1-q_0)/2 and w_n = (w_n - w_{n-1})/2.
+    """
+    num_quantiles = len(quantile_levels)
+    assert num_quantiles >= 0
```
jaheba, line 42:
```suggestion
    assert quantile_levels
```

src/gluonts/mx/block/quantile_output.py
```diff
@@ -20,13 +20,46 @@
...
+
+    if num_quantiles < 2:
+        return [1.0] * num_quantiles
```
jaheba, line 45:
```suggestion
    if num_quantiles == 1:
        return [1.0]
```

lostella: A test is in place for the edge case `num_quantiles == 0`, so I opted to keep the behaviour as it is there

src/gluonts/mx/block/quantile_output.py
lostella: Same here

src/gluonts/mx/block/quantile_output.py
jaheba: Ah, misread the code.

src/gluonts/mx/block/quantile_output.py
lostella: > Can we add a more concrete type?

We could, but it would be an artificial restriction in my opinion. In this context `x: List[float]` but could be a list of anything

src/gluonts/mx/block/quantile_output.py
jaheba: Actually, this check doesn't make sense at all. Are we ensuring that `len` decides to return negative values?

lostella: One is never too sure! (Joking, I'll remove it)

src/gluonts/mx/block/quantile_output.py
jaheba: Right. Could we then maybe add a docstring explaining what this does and have a quick example like:

```
>>> uniform_weights(["a", "b", "c", "d"])
[0.25, 0.25, 0.25, 0.25]
```

src/gluonts/mx/block/quantile_output.py
lostella: I added directives for inline and displayed math throughout the docstring

src/gluonts/mx/block/quantile_output.py
```diff
@@ -96,21 +143,19 @@ def hybrid_forward(
...
+            self.quantiles, self.quantile_weights, y_pred_all
+        ):
+            qt_loss.append(
+                weight * self.compute_quantile_loss(F, y_true, y_pred_q, level)
             )
-            qt_loss.append(weighted_qt)
         stacked_qt_losses = F.stack(*qt_loss, axis=-1)
         sum_qt_loss = F.mean(
             stacked_qt_losses, axis=-1
         )  # avg across quantiles
```
lostella, line 155: @dcmaddix is `F.mean` correct here? Shouldn't this be `F.sum`? Since we are already weighting the quantile loss terms

src/gluonts/mx/block/quantile_output.py
```diff
@@ -20,47 +20,95 @@
...
+
+    Returns
+    -------
+    List[float]
+        List of weights.
+    """
+    return [1.0 / len(objects)] * len(objects)
+
+
+def crps_weights_pwl(quantile_levels: List[float]) -> List[float]:
```
lostella, line 43: @dcmaddix as @jaheba pointed out, weights produced by this function don't sum to one in general. Only if 0 and 1 are included as quantile levels they do:

```
In [4]: sum(crps_weights_pwl([0.0, 0.1, 0.9, 0.95, 1.0]))
Out[4]: 1.0

In [5]: sum(crps_weights_pwl([0.1, 0.9, 0.95]))
Out[5]: 0.8499999999999999
```

dcmaddix: Yes, I think 0 and 1 should be included here since the CRPS is the integral over the quantiles from 0 to 1, which this function approximates. Best to confirm with @youngsuk0723.

src/gluonts/mx/block/quantile_output.py
dcmaddix: Yes, I'm not sure if we need `F.mean` since you're right we take the weighted sum on line 150. Best to confirm with @youngsuk0723.

src/gluonts/mx/block/quantile_output.py
dcmaddix: yes

dcmaddix (approved):
Looks good to me, thanks Lorenzo!

src/gluonts/mx/block/quantile_output.py
youngsuk0723: @dcmaddix is correct. @lostella The sum is 1 when to include 0, 1. Otherewise, it is less than 1, which is also similar to the way of approximaying CRPS in original MQ-CNN.

PR #1962, opened 2022-04-29
Add support for beta-NLL loss

kashif:
from the paper [On the Pitfalls of Heteroscedastic  Uncertainty Estimation with Probabilistic Neural Networks](https://openreview.net/pdf?id=aPOpXlnV1T) by Seitzer et al. 2022

official code repo: https://github.com/martius-lab/beta-nll

*Issue #, if available:*

*Description of changes:*


src/gluonts/torch/modules/loss.py
```diff
@@ -48,10 +49,31 @@ def forward(
 
 
 class NegativeLogLikelihood(DistributionLoss):
+    def __init__(self, beta: float = 0.0):
+        """
```
lostella, line 53: Could you move this docstring to the class directly, instead of the constructor? So it’s consistent with the rest

src/gluonts/torch/modules/loss.py
```diff
@@ -48,10 +49,31 @@ def forward(
...
+        self.beta = beta
+
     def forward(
         self, input: torch.distributions.Distribution, target: torch.Tensor
     ) -> torch.Tensor:
-        return -input.log_prob(target)
+        nll = -input.log_prob(target)
+        if self.beta > 0.0:
+            if type(input) == TransformedDistribution:
+                variance = input.base_dist.variance
```
lostella, line 72: Is this right?

src/gluonts/torch/modules/loss.py
```diff
@@ -48,10 +49,31 @@ def forward(
...
 class NegativeLogLikelihood(DistributionLoss):
+    def __init__(self, beta: float = 0.0):
+        """
+        Compute the negative log likelihood loss.
+
+        Parameters
+        ----------
+        beta
+            beta parameter from the paper: "On the Pitfalls of Heteroscedastic
+            Uncertainty Estimation with Probabilistic Neural Networks" Seitzer et al. 2022
```
lostella, line 60: Flake8 complaining about too long lines (limit is 79 characters)

src/gluonts/torch/modules/loss.py
```diff
@@ -48,10 +49,31 @@ def forward(
...
 
 class NegativeLogLikelihood(DistributionLoss):
+    def __init__(self, beta: float = 0.0):
+        """
+        Compute the negative log likelihood loss.
+
+        Parameters
+        ----------
+        beta
+            beta parameter from the paper: "On the Pitfalls of Heteroscedastic
```
lostella, line 59: Maybe you can suggest the range (0, 1) for this hyperparameter, and reasonable values: 0 recovers NLL, 0.5 appears to be a reasonable choice (at least from the results in the paper)

lostella (changes requested):
Straight out of ICLR, thanks for this! I have left some comments

kashif:
yes so I have to figure out how the method would work with TransformerDistribution with an affine transform coming from the scale... 

let me move this to draft for now...

lostella:
> yes so I have to figure out how the method would work with TransformerDistribution with an affine transform coming from the scale...

Maybe it’s fine to support the simple case where only one affine transformation is there (multiple affine transformations should be collapsed into a single one anyway).

kashif:
right... but the issue is i believe with the scale the variance needs to be scaled too... I am working through the maths... initial experiments are giving me nans for Gaussian and Student-t...

lostella:
> right... but the issue is, i believe with the scale, the variance needs to be scaled too... 

Yes, it does. For MXNet distributions we have an explicit type for affinely-transformed distributions: https://github.com/awslabs/gluon-ts/blob/925d44c71b7ed85f0154afe22198a9cfc30713eb/src/gluonts/mx/distribution/transformed_distribution.py#L227 

This allows to avoid having to handle these (not so) special cases explicitly elsewhere

kashif:
yup! that is what i did but getting nans... let me push it so you can have a look

src/gluonts/torch/modules/distribution_output.py
```diff
@@ -141,11 +141,16 @@ def distribution(
...
         else:
+            scale = 1.0 if scale is None else scale
             transform = AffineTransform(
                 loc=0.0 if loc is None else loc,
-                scale=1.0 if scale is None else scale,
+                scale=scale,
             )
-            return TransformedDistribution(distr, [transform])
+            transformed_distr = TransformedDistribution(distr, [transform])
+            TransformedDistribution.variance = property(
```
lostella, line 150: This seems a bit weird…

lostella (changes requested):
How about something like:

```python
def variance(distr):
    if isinstance(distr, TransformedDistribution):
        # assert it’s transformed by single affine
        # return scaled variance
    return distr.variance
```

src/gluonts/torch/modules/distribution_output.py
kashif: you are right! I've fixed it I hope! please kindly review whenever you get time!

src/gluonts/torch/distributions/affine_transformed.py
```diff
@@ -0,0 +1,50 @@
...
+    AffineTransform,
+    Distribution,
+)
+
+
+class AffineTransformed(TransformedDistribution):
+    def __init__(self, base_distribution: Distribution, loc=None, scale=None):
+
+        self.scale = 1.0 if scale is None else scale
+        self.loc = 0.0 if loc is None else loc
```
lostella, line 25:
```suggestion
    def __init__(self, base_distribution: Distribution, loc=0, scale=1):
        self.scale = scale
        self.loc = loc
```

src/gluonts/torch/modules/loss.py
```diff
@@ -48,10 +48,29 @@ def forward(
...
+    """
+    Compute the negative log likelihood loss.
+
+    Parameters
+    ----------
+    beta: float in range (0, 1)
+        beta parameter from the paper: "On the Pitfalls of Heteroscedastic
+        Uncertainty Estimation with Probabilistic Neural Networks" by
+        Seitzer et al. 2022
+        https://openreview.net/forum?id=aPOpXlnV1T"""
```
lostella, line 60:
```suggestion
        https://openreview.net/forum?id=aPOpXlnV1T
    """
```

lostella (changes requested):
Looks good to me, there are only a couple of minor things, plus what seems to be a circular dependency that is breaking the docs build: https://github.com/awslabs/gluon-ts/runs/6255414029?check_suite_focus=true

kashif:
@lostella do you think the circular dependency is due to my changes? I cannot seem to figure out where its coming from...

lostella:
> @lostella do you think the circular dependency is due to my changes? I cannot seem to figure out where its coming from...

I think so, I’ll try to sort it out

lostella (approved):
Looks good to me, thanks @kashif! 🚀

kashif:
thanks @lostella 

My initial experiments were not too conclusive in the benefit of this scheme... my personal intuition would be to have the beta parameter decay down to zero or something small as training progresses... 

In any case, I would need more time to investigate this...

PR #1969, opened 2022-05-07
Fix: torch DeepAR observed indicator in multivariate case

kashif:
*Issue #, if available:*

*Description of changes:*
pytorch min return min and min_indicies so only use the min


lostella:
Then I guess the first sentence and example in the docs is wrong? https://pytorch.org/docs/stable/generated/torch.min.html#torch.min

lostella:
Oh I see, it returns a tuple if `dim` is provided? That’s a bit confusing…

kashif:
right... but this code block is never used since its univariate emission model...

lostella (approved):
🚀 thanks!

PR #1970, opened 2022-05-09
Updated GluonTS_SageMaker_SDK_Tutorial

mr-1993:
Due to some outdated package versions and Jupyter kernel, some commands need to be updated.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/examples/GluonTS_SageMaker_SDK_Tutorial.ipynb
+++ b/examples/GluonTS_SageMaker_SDK_Tutorial.ipynb
@@ -11,7 +11,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "***This notebook is meant to be uploaded to a SageMaker notebook instance and executed there. As a kernel choose `conda_mxnet_p36`***"
+    "***This notebook is meant to be uploaded to a SageMaker notebook instance and executed there. As a kernel choose `conda_mxnet_latest_p37`***"
    ]
   },
   {
@@ -42,7 +42,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "!pip install --upgrade mxnet==1.6  git+https://github.com/awslabs/gluon-ts.git#egg=gluonts[dev]"
+    "!pip install --upgrade mxnet==1.6  git+https://github.com/awslabs/gluon-ts.git#egg=gluonts[dev]\n",
+    "!pip install s3fs==0.4.2\n",
+    "!pip install --upgrade pip "
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


mr-1993:
Hi there, just would like to enquire whether this PR is ready to be merged with master? Thx for your support!

examples/GluonTS_SageMaker_SDK_Tutorial.ipynb
```diff
@@ -42,7 +42,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "!pip install --upgrade mxnet==1.6  git+https://github.com/awslabs/gluon-ts.git#egg=gluonts[dev]"
+    "!pip install --upgrade mxnet==1.6  git+https://github.com/awslabs/gluon-ts.git#egg=gluonts[dev]\n",
```
lostella, line 45: This is outdated, a better one here would be 
```suggestion
    "!pip install --upgrade mxnet~=1.9  git+https://github.com/awslabs/gluon-ts.git#egg=gluonts[dev]\n",
```

mr-1993: This breaks the remainder of the tutorial as the screenshot below shows: Do you know what might be the reason there? (apparently, freq does not exist anymore)

image

lostella: Yes, just remove `freq` from the constructor of `SimpleFeedForwardEstimator` (it's not required anymore since it's not actually used by this specific model).

mr-1993: Thank you very much :) The next problem we run into (I think that was also the reason we decided to stick to gluonts 1.6 back then and s3fs 0.4.2) is an issue with aiobotocore do you have an idea how to solve this? The error persists if we upgrade s3fs to the current version

image

lostella: I've never seen this issue myself, but a quick search seems to suggest that recent versions of `s3fs` got rid of this problem, see eg:
- https://stackoverflow.com/questions/68864939/s3fs-suddenly-stopped-working-in-google-colab-with-error-attributeerror-module
- https://ittutoria.net/solve-the-error-attributeerror-module-aiobotocore-has-no-attribute-aiosession/

lostella: nevermind, just saw this

> The error persists if we upgrade s3fs to the current version

mr-1993: There seems to be a problem with conflicting version dependencies which I cannot resolve immediately, need to have a closer look

image

image

PR #1973, opened 2022-05-12
Irregular time series support

kashif:
*Issue #, if available:*

*Description of changes:*

Adding support to irregular time series where the datatimes are given by an INDEX key in the dataset

### DIFF
--- a/src/gluonts/dataset/util.py
+++ b/src/gluonts/dataset/util.py
@@ -14,16 +14,10 @@ import logging
 import multiprocessing
 import os
 from pathlib import Path
-from typing import (
-    Callable,
-    Iterator,
-    List,
-    NamedTuple,
-    Tuple,
-    TypeVar,
-)
+from typing import Callable, Iterator, List, NamedTuple, Tuple, TypeVar
 
 import pandas as pd
+from gluonts.dataset.field_names import FieldName
 
 T = TypeVar("T")
 
@@ -132,11 +126,15 @@ def to_pandas(instance: dict, freq: str = None) -> pd.Series:
     pandas.Series
         Pandas time series object.
     """
-    target = instance["target"]
-    start = instance["start"]
-    if not freq:
-        freq = start.freqstr
-    index = pd.date_range(start=start, periods=len(target), freq=freq)
+    target = instance[FieldName.TARGET]
+    start = instance[FieldName.START]
+
+    if FieldName.INDEX in instance:
+        index = instance[FieldName.INDEX]
+    else:
+        if not freq:
+            freq = start.freqstr
+        index = pd.date_range(start=start, periods=len(target), freq=freq)
     return pd.Series(target, index=index)
 
 
--- a/src/gluonts/dataset/field_names.py
+++ b/src/gluonts/dataset/field_names.py
@@ -44,3 +44,5 @@ class FieldName:
     FORECAST_START = "forecast_start"
 
     TARGET_DIM_INDICATOR = "target_dimension_indicator"
+
+    INDEX = "index"
--- a/src/gluonts/evaluation/backtest.py
+++ b/src/gluonts/evaluation/backtest.py
@@ -20,6 +20,7 @@ import pandas as pd
 import gluonts  # noqa
 from gluonts.core.serde import load_code
 from gluonts.dataset.common import DataEntry, Dataset
+from gluonts.dataset.field_names import FieldName
 from gluonts.dataset.stat import (
     DatasetStatistics,
     calculate_dataset_statistics,
@@ -71,13 +72,16 @@ def make_evaluation_predictions(
     ) -> Iterator[DataEntry]:
         for data_entry in data_iterator:
             data = data_entry.copy()
-            index = pd.date_range(
-                start=data["start"],
-                freq=freq,
-                periods=data["target"].shape[-1],
-            )
+            if FieldName.INDEX in data:
+                index = data[FieldName.INDEX]
+            else:
+                index = pd.date_range(
+                    start=data[FieldName.START],
+                    freq=freq,
+                    periods=data[FieldName.TARGET].shape[-1],
+                )
             data["ts"] = pd.DataFrame(
-                index=index, data=data["target"].transpose()
+                index=index, data=data[FieldName.TARGET].transpose()
             )
             yield data
 
@@ -87,11 +91,11 @@ def make_evaluation_predictions(
 
     def truncate_target(data):
         data = data.copy()
-        target = data["target"]
+        target = data[FieldName.TARGET]
         assert (
             target.shape[-1] >= prediction_length
         )  # handles multivariate case (target_dim, history_length)
-        data["target"] = target[..., : -prediction_length - lead_time]
+        data[FieldName.TARGET] = target[..., : -prediction_length - lead_time]
         return data
 
     # TODO filter out time series with target shorter than prediction length
--- a/src/gluonts/torch/model/deepar/estimator.py
+++ b/src/gluonts/torch/model/deepar/estimator.py
@@ -189,6 +189,7 @@ class DeepAREstimator(PyTorchLightningEstimator):
                     output_field=FieldName.FEAT_TIME,
                     time_features=self.time_features,
                     pred_length=self.prediction_length,
+                    index_field=FieldName.INDEX,
                 ),
                 AddAgeFeature(
                     target_field=FieldName.TARGET,
--- a/src/gluonts/transform/feature.py
+++ b/src/gluonts/transform/feature.py
@@ -342,6 +342,7 @@ class AddTimeFeatures(MapTransformation):
         output_field: str,
         time_features: List[TimeFeature],
         pred_length: int,
+        index_field: Optional[List[pd.Timestamp]] = None,
         dtype: Type = np.float32,
     ) -> None:
         self.date_features = time_features
@@ -349,6 +350,7 @@ class AddTimeFeatures(MapTransformation):
         self.start_field = start_field
         self.target_field = target_field
         self.output_field = output_field
+        self.index_field = index_field
         self._freq_base: Optional[pd.offsets.DateOffset] = None
         self._min_time_point: Optional[pd.Timestamp] = None
         self._max_time_point: Optional[pd.Timestamp] = None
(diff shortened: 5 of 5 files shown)


jaheba:
To understand it correctly, "irregular" means that the values are not spaced apart by the same time offset?

kashif:
right! When then implies we need to specify the time index for each point of the target... I believe i have training working... stuck at the inference part... I would appreciate some help! Let me push the latest changes and example to test...

kashif:
@jaheba I am using the following snippet to test:

```py
import random
import pandas as pd

from gluonts.dataset.common import Dataset, ListDataset
from gluonts.evaluation import make_evaluation_predictions, Evaluator

freq_str="1T"
prediction_length = 10

index = sorted(random.sample(list(pd.date_range("2021-01-01 02:12:06", periods=400, freq=freq_str)), 50+prediction_length))


train_ds = ListDataset(
    [
        {
        "start": index[0], 
        "target": [float(i) for i in range(50)],
        "item_id": "0",
        "index": index[:50],
        },
    ],
    freq=freq_str,
) 

test_ds = ListDataset(
    data_iter=[
        {
        "start": index[0], 
        "target": [float(i) for i in range(50+prediction_length)],
        "item_id": "0",
        "index": index[:50+prediction_length],
        }
    ],
    freq=freq_str,
) 

estimator = DeepAREstimator(
    freq=freq_str,
    scaling=False,
    trainer_kwargs={"max_epochs": 1},
)

predictor = estimator.train(train_ds)

    dataset=test_ds,
    predictor=predictor
)

forecasts = list(forecast_it) # the predictions are not the right size in time
tss = list(ts_it)

evaluator = Evaluator()
agg_metrics, ts_metrics = evaluator(iter(tss), iter(forecasts), num_series=len(test_ds)) #  crash here
```

jaheba:
Can we add the snippet as a test?

src/gluonts/mx/model/forecast.py
```diff
@@ -61,6 +61,7 @@ def __init__(
         freq: str,
         item_id: Optional[str] = None,
         info: Optional[Dict] = None,
+        index: Optional[Union[List[pd.Timestamp], pd.DatetimeIndex]] = None,
```
jaheba, line 64: We could create a type-alias for this.

kashif:
sure... let me do that...

src/gluonts/dataset/common.py
```diff
@@ -287,7 +288,7 @@ class Config:
         arbitrary_types_allowed = True
 
     freq: Union[str, pd.DateOffset]
-    name: str = "start"
+    name: str = FieldName.START
```
jaheba, line 291: Could we split these renaming-changes off into a separate PR, which we quickly merge?

kashif: ok

jaheba: Thanks!

src/gluonts/mx/model/forecast.py
kashif: so this is was initial thought... but i believe i only need the index to be `pd.DatetimeIndex`...

jaheba:
Something I'm wondering: If we merge this PR we break the assumption that time-series are regular. Is that an assumption we rely on in other places? /cc @lostella

kashif:
so the one big missing piece is in the `InstanceSplitter` which currently is setting the wrong `forecast_start_field` field

kashif:
@jaheba #1975 is the change to use FieldName. Once merged I'll fix up this PR wrt the conflicts.

So I believe, basically, this functionality is working... could you kindly see if there might be some edge cases? One I believe is the padding in the instance splitters...

kashif:
@lostella I believe I have taken care of all the 3 issues you so kindly listed and I believe I am able to train and do inference on irregular time series data now.

The only edge case I have in my mind right now is the padding in the Splitters... if you could kindly have a closer look at that, I would appreciate it!

Thanks!

src/gluonts/dataset/util.py
```diff
@@ -132,11 +126,15 @@ def to_pandas(instance: dict, freq: str = None) -> pd.Series:
...
-    target = instance["target"]
-    start = instance["start"]
-    if not freq:
-        freq = start.freqstr
-    index = pd.date_range(start=start, periods=len(target), freq=freq)
+    target = instance[FieldName.TARGET]
+    start = instance[FieldName.START]
+
+    if FieldName.INDEX in instance:
+        index = instance[FieldName.INDEX]
```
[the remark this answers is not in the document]
kashif: yes sure, both can be provided and are used actually since I get the freq sometimes from the start datetime rather than the datetime index which doesn't have a freq in the case of irregular dates.

src/gluonts/evaluation/_base.py
```diff
@@ -310,7 +310,7 @@ def extract_past_data(
         # Remove the prediction range
         # If the prediction range is not in the end of the time series,
         # everything after the prediction range is truncated
-        date_before_forecast = forecast.index[0] - forecast.index[0].freq
+        date_before_forecast = forecast.index[0] - forecast.start_date.freq
```
kashif, line 313: @lostella this change is needed because irregular datetime index does not have a freq, thus I need to use the freq assigned to the start date

kashif:
@jaheba another potential issue i have is if i run the code snippet above and then call `forecasts[0]` in jupyter I get the following error in the serializer:

```py
forecasts[0]

---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
File ~/.env/pytorch/lib/python3.8/site-packages/IPython/core/formatters.py:707, in PlainTextFormatter.__call__(self, obj)
    700 stream = StringIO()
    701 printer = pretty.RepresentationPrinter(stream, self.verbose,
    702     self.max_width, self.newline,
    703     max_seq_length=self.max_seq_length,
    704     singleton_pprinters=self.singleton_printers,
    705     type_pprinters=self.type_printers,
    706     deferred_pprinters=self.deferred_printers)
--> 707 printer.pretty(obj)
    708 printer.flush()
    709 return stream.getvalue()

File ~/.env/pytorch/lib/python3.8/site-packages/IPython/lib/pretty.py:410, in RepresentationPrinter.pretty(self, obj)
    407                         return meth(obj, self, cycle)
    408                 if cls is not object \
    409                         and callable(cls.__dict__.get('__repr__')):
--> 410                     return _repr_pprint(obj, self, cycle)
    412     return _default_pprint(obj, self, cycle)
    413 finally:

File ~/.env/pytorch/lib/python3.8/site-packages/IPython/lib/pretty.py:778, in _repr_pprint(obj, p, cycle)
    776 """A pprint that just redirects to the normal repr function."""
    777 # Find newlines and replace them with p.break_()
--> 778 output = repr(obj)
    779 lines = output.splitlines()
    780 with p.group():

File ~/gluon-ts-PR/src/gluonts/core/component.py:310, in validated.<locals>.validator.<locals>.validated_repr(self)
    309 def validated_repr(self) -> str:
--> 310     return dump_code(self)

File ~/gluon-ts-PR/src/gluonts/core/serde/_repr.py:115, in dump_code(o)
     95 def dump_code(o: Any) -> str:
     96     """
     97     Serializes an object to a Python code string.
     98 
   (...)
    112         Inverse function.
    113     """
--> 115     return as_repr(encode(o))

File /usr/lib/python3.8/functools.py:875, in singledispatch.<locals>.wrapper(*args, **kw)
    871 if not args:
    872     raise TypeError(f'{funcname} requires at least '
    873                     '1 positional argument')
--> 875 return dispatch(args[0].__class__)(*args, **kw)

File ~/gluon-ts-PR/src/gluonts/core/serde/_base.py:220, in encode(v)
    212 if hasattr(v, "__getnewargs_ex__"):
    213     args, kwargs = v.__getnewargs_ex__()  # mypy: ignore
    215     return {
    216         "__kind__": Kind.Instance,
    217         "class": fqname_for(v.__class__),
    218         # args need to be a list, since we encode tuples explicitly
    219         "args": encode(list(args)),
--> 220         "kwargs": encode(kwargs),
    221     }
    223 try:
    224     # as fallback, we try to just take the path of the value
    225     fqname = fqname_for(v)

File /usr/lib/python3.8/functools.py:875, in singledispatch.<locals>.wrapper(*args, **kw)
    871 if not args:
    872     raise TypeError(f'{funcname} requires at least '
    873                     '1 positional argument')
--> 875 return dispatch(args[0].__class__)(*args, **kw)

File ~/gluon-ts-PR/src/gluonts/core/serde/_base.py:207, in encode(v)
    204     return list(map(encode, v))
    206 if isinstance(v, dict):
--> 207     return valmap(encode, v)
    209 if isinstance(v, type):
    210     return {"__kind__": Kind.Type, "class": fqname_for(v)}

File ~/.env/pytorch/lib/python3.8/site-packages/toolz/dicttoolz.py:83, in valmap(func, d, factory)
     72 """ Apply function to values of dictionary
     73 
     74 >>> bills = {"Alice": [20, 15, 30], "Bob": [10, 35]}
   (...)
     80     itemmap
     81 """
     82 rv = factory()
---> 83 rv.update(zip(d.keys(), map(func, d.values())))
     84 return rv

File /usr/lib/python3.8/functools.py:875, in singledispatch.<locals>.wrapper(*args, **kw)
    871 if not args:
    872     raise TypeError(f'{funcname} requires at least '
    873                     '1 positional argument')
--> 875 return dispatch(args[0].__class__)(*args, **kw)

File ~/gluon-ts-PR/src/gluonts/core/serde/_base.py:243, in encode(v)
    240 except AttributeError:
    241     pass
--> 243 raise RuntimeError(bad_type_msg.format(fqname_for(v.__class__)))

RuntimeError: Cannot serialize type pandas.core.indexes.datetimes.DatetimeIndex. See the documentation of the `encode` and
`validate` functions at

```

kashif:
@jaheba ok I added a DatetimeIndex encoder to fix the above issue I believe...

kashif:
I will update this I suppose when the PR #1980 is merged as I can then base my changes based on that

kashif:
@rsnirwan would you have some time to quickly see how I can also use the new pandas datasets with irregular time indices in this PR?

kashif:
@rsnirwan would you have some time to quickly see how I can also use the new pandas datasets with irregular time indices in this PR?

AjinkyaBankar:
@kashif @jaheba @lostella Is this resolved? Can we use irregular time series forecasting using DeepAR? This is highly required for some practical cases.

kashif:
hmm @AjinkyaBankar kind-of, things were refactored and this branch kind of got left behind... i would love to get it up to speed but it might require a new PR

AjinkyaBankar:
@kashif I am excited to use this as soon as possible. Let me know if I can help with this.

Issue #1939, opened 2022-04-12
Suppress Pandas FutureWarning: Timestamp.freq is deprecated


lostella:
We're getting numerous warnings from Pandas about this, this should be addressed:

```
FutureWarning: Timestamp.freq is deprecated and will be removed in a future version
```

jaheba:
I'm not sure if this would break anything. The other change we would need to make is to replace `pd.date_range` with `pd.period_range`.

PR #1980, opened 2022-05-19
Use `pd.Period` instead of `pd.Timestamp`.

jaheba:
*Issue #, if available:*

*Description of changes:*


lostella:
Fixes #1939?

jaheba:
> Fixes #1939?

Yes!

kashif:
@jaheba with this change would it be possible to remove the requirement of specifying `freq` arguments to the forecast generators eg. SampleForecast and QuantileForecasts? I believe the freq can be inferred from the start_date period and this will also help me train models without a freq string specified...

jaheba:
> @jaheba with this change would it be possible to remove the requirement of specifying `freq` arguments to the forecast generators eg. SampleForecast and QuantileForecasts? I believe the freq can be inferred from the start_date period and this will also help me train models without a hard-coded freq string specified...

In short, no, this doesn't change anything about the required `freq` problem.

But I agree, we should make it at least optional, since in many cases `freq` is just carried around because it has to.

kashif:
sorry, I didnt mean we should remove it but yes make it optional since e.g. here: 

https://github.com/awslabs/gluon-ts/blob/master/src/gluonts/model/forecast.py#L507-L508

we should be able to infer the freq from the start period now... thus the freq string can be made optional

kashif:
if you want i can send that PR after you are done with this one?

jaheba:
`Forecast` is something I generally want to rework. I guess we can make it optional first and then think about how to do it right.

I think we should generally remove `start` and use an index in cases where we want to associate a value with a time instead.

src/gluonts/data/__init__.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jaheba, line 1: I should remove this file.

src/gluonts/core/ty.py
```diff
@@ -17,6 +17,16 @@
 
 import pydantic
 
+from typing import TypeVar, Optional
+
+T = TypeVar("T")
+
+
+def unwrap(x: Optional[T]) -> T:
```
jaheba, line 25: I guess we can add this in a different PR

src/gluonts/dataset/common.py
```diff
@@ -165,10 +150,14 @@ def __init__(
         freq: str,
         one_dim_target: bool = True,
         cache: bool = False,
+        use_timestamp: bool = False,
```
lostella, line 153: This needs to be added to the docstring

src/gluonts/dataset/common.py
```diff
@@ -242,8 +231,10 @@ def __init__(
         data_iter: Iterable[DataEntry],
         freq: str,
         one_dim_target: bool = True,
+        use_timestamp: bool = False,
```
lostella, line 234: Add to docstring

src/gluonts/dataset/common.py
```diff
@@ -288,58 +273,22 @@ class Config:
         arbitrary_types_allowed = True
 
     freq: Union[str, pd.DateOffset]
+    use_timestamp: bool = False
```
lostella, line 276: Add to docstring

src/gluonts/dataset/common.py
```diff
@@ -288,58 +273,22 @@ class Config:
...
     name: str = FieldName.START
-    tz_strategy: TimeZoneStrategy = TimeZoneStrategy.error
 
     def __call__(self, data: DataEntry) -> DataEntry:
         try:
-            timestamp = ProcessStartField.process(data[self.name], self.freq)
+            if self.use_timestamp:
+                data[self.name] = pd.Timestamp(data[self.name])
+            else:
+                data[self.name] = pd.Period(data[self.name], self.freq)
```
lostella, line 284: Why do we need this option?

jaheba: For point-processes datasets, where `start` needs to be a `Timestamp`. These also have an `end` field which needs to be a timestamp, which we don't handle.

It's a hack.

src/gluonts/dataset/artificial/_base.py
```diff
@@ -143,10 +157,16 @@ def __init__(
         self.is_promotions = is_promotions
         self.holidays = holidays
 
+        if isinstance(self.freq, Week):
+            self.freq = Week(
+                self.freq.n, weekday=pd.Timestamp(start).weekday()
+            )
+        self.start = cast(pd.Period, pd.Period(start, self.freq))
```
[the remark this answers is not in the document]
jaheba: Because pyright complained, because of this:

```py
class Period(PeriodMixin):
    ordinal: int  # int64_t
    freq: BaseOffset

    # error: "__new__" must return a class instance (got "Union[Period, NaTType]")
    def __new__(  # type: ignore[misc]
        cls,
        value=...,
        freq: int | str | BaseOffset | None = ...,
        ordinal: int | None = ...,
        year: int | None = ...,
        month: int | None = ...,
        quarter: int | None = ...,
        day: int | None = ...,
        hour: int | None = ...,
        minute: int | None = ...,
        second: int | None = ...,
    ) -> Period | NaTType: ...
```


https://github.com/pandas-dev/pandas/blob/9289c46e166533dfafd18843137f101624cf41a5/pandas/_libs/tslibs/period.pyi#L62-L79

examples/anomaly_detection.py
```diff
@@ -97,8 +97,8 @@
         i = k // nll.shape[1]
         t = k % nll.shape[1]
 
-        time_index = pd.date_range(
-            pd.Timestamp(data_entry["forecast_start"][i]),
+        time_index = pd.period_range(
+            data_entry["forecast_start"][i],
             periods=context_length + prediction_length,
         )
```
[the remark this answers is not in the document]
jaheba: It takes the frequency of the `start` field.

lostella (approved):

kashif:
@jaheba testing models after this commit I have the feeling that the time feature creation has become a bit slower... meaning its taking longer to make batches for training... the CPU also starts to throttle up when making batches.... did you have that feeling on your end?

jaheba:
Having tested it. I think we now create a new `period_range` for each instance, where we used a cache before. It's something I wanted to address later.

Do you have rough numbers?

jaheba:
Also, I've removed the cache for time-stamp creation which might also slow things down a bit.

kashif:
@jaheba another issue I see is that the ground truth data seems to be a bit screwed up see before traffic:
download-5

and now there seem to be spurious glitches:

download-6

can you kindly confirm you also see this?

kashif:
the data is traffic... i do not believe i am doing anything to the ground truth data with the model... e.g. pytorch deepAR i am just doing the standard prediction and then plotting...

jaheba:
Screenshot 2022-06-08 at 17 08 45

This is how it looks for me.

kashif:
hmmm ok good to know... wonder what's going on at my end.... perhaps i am going crazy....

PR #1982, opened 2022-05-20
Add Uber-TLC dataset to dataset repository.

Hongqing-work:
Description of changes:
    I added a new dataset in `gluonts.dataset.repository`. The raw data is from [uber-tlc-foil-response](https://github.com/fivethirtyeight/uber-tlc-foil-response). There are two kinds of frequency and users can get them using 'uber_daily', 'uber_hourly' as parameters in the method `get_dataset`.

Note: There is also a test for this new dataset in [uber-time-series-prediction-using-gluonts](https://github.com/Hongqing-work/uber-time-series-prediction-using-gluonts).

Please tag this pr with at least one of these labels to make our release process faster: BREAKING, new feature, bug fix, other change, dev setup


src/gluonts/dataset/repository/_uber.py
```diff
@@ -0,0 +1,82 @@
...
+
+import pandas as pd
+
+from gluonts.dataset.repository._util import metadata
+
+
+def generate_uber_dataset(
+    dataset_path: Path, uber_freq: str, prediction_length: Optional[int] = None
+):
+    from gluonts.dataset.repository.datasets import default_dataset_path
```
jaheba, line 14: Couldn't we just have this import at the top of the file?

src/gluonts/dataset/repository/_uber.py
```diff
@@ -0,0 +1,82 @@
...
+
+    uber_csv_path = default_dataset_path / "uber-raw-data-janjune-15.csv"
+    if not os.path.exists(uber_csv_path):
```
jaheba, line 17: Since it's already a `pathlib.Path`, you can just do `uber_csv_path.exists()`.

src/gluonts/dataset/repository/_uber.py
```diff
@@ -0,0 +1,82 @@
...
+        raise RuntimeError(
+            "The uber data is available at "
+            "https://github.com/fivethirtyeight/uber-tlc-foil-response/blob/"
```
jaheba, line 20: Can you check if we automatically download other datasets? This would make things much easier for users.

src/gluonts/dataset/repository/datasets.py
```diff
@@ -209,6 +210,12 @@ def get_download_path() -> Path:
             prediction_length=28,
             m5_file_path=get_download_path() / "m5",
         ),
+        "uber_daily": partial(
```
jaheba, line 213: Can we name these `uber_tlc_daily` and `uber_tlc_hourly`? That makes it a bit more explicit what data we are using.

src/gluonts/dataset/repository/_uber.py
```diff
@@ -0,0 +1,82 @@
...
+        usecols=["Pickup_date", "locationID"],
+        index_col=0,
+    )
+    # We divide the raw data according to locationID. Each json line represents
+    # a time series of a loacationID. The targets are numbers of pickup-events
+    # during a day or an hour.
+    time_series_of_locations = list(uber_df.groupby(by="locationID"))
+    os.makedirs(dataset_path, exist_ok=True)
+    os.makedirs(dataset_path / "train", exist_ok=True)
+    os.makedirs(dataset_path / "test", exist_ok=True)
```
jaheba, line 45: Instead of `os.makedirs`, we can just use `pathlib.Path.mkdir`: https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

src/gluonts/dataset/repository/_uber.py
```diff
@@ -0,0 +1,82 @@
...
+    test_file = dataset_path / "test" / "data.json"
+    with open(train_file, "w") as o_train, open(test_file, "w") as o_test:
+        for locationID, df in time_series_of_locations:
+            df.sort_index()
+            df.index = pd.to_datetime(df.index)
+
+            count_series = df.resample(rule=freq_setting).size()
+            start_time = pd.Timestamp(df.index[0]).strftime("%Y-%m-%d %X")
+            target = count_series.values.tolist()
+            feat_static_cat = [locationID] * len(target)
```
jaheba, line 56:
```suggestion
            feat_static_cat = [locationID]

```

As the name suggests static features are static for a given time-series, and thus we only need to store one value.

jaheba:
Thanks for the PR! I've added some minor comments.

src/gluonts/dataset/repository/_uber.py
Hongqing-work: That's because the `gluonts.dataset.repository.datasets` have to include `gluonts.dataset.repository._uber`.

src/gluonts/dataset/repository/_uber.py
Hongqing-work: The csv file is very large and we don't have access to the raw data and can only download a zip file. I will try to add a function to automatically download a zip file and read the raw data.

jaheba: `pd.read_csv` might be able to handle it for us already.

src/gluonts/dataset/repository/_uber.py
```diff
@@ -9,29 +8,20 @@
...
     subsets = {"daily": "1D", "hourly": "1H"}
-
     assert (
         uber_freq.lower() in subsets
     ), f"invalid uber_freq='{uber_freq}'. Allowed values: {subsets.keys()}"
-
     freq_setting = subsets[uber_freq.lower()]
+
+    uber_url_path = (
+        "https://github.com/Hongqing-work/uber-tlc-foil-response/releases"
```
jaheba, line 20: Couldn't you just use the link from the `fivethirtyeight` repo?

https://github.com/fivethirtyeight/uber-tlc-foil-response/raw/63bb878b76f47f69b4527d50af57aac26dead983/uber-trip-data/uber-raw-data-janjune-15.csv.zip

src/gluonts/dataset/repository/_uber.py
```diff
@@ -40,11 +30,13 @@ def generate_uber_dataset(
...
-    os.makedirs(dataset_path / "test", exist_ok=True)
-    train_file = dataset_path / "train" / "data.json"
-    test_file = dataset_path / "test" / "data.json"
+    dataset_path.mkdir(exist_ok=True)
+    train_path = dataset_path / "train"
+    test_path = dataset_path / "test"
+    train_path.mkdir(exist_ok=True)
+    test_path.mkdir(exist_ok=True)
+    train_file = train_path / "data.json"
+    test_file = test_path / "data.json"
```
jaheba, line 39: Can we add empty lines before and after to visually declutter the code?

src/gluonts/dataset/repository/_uber.py
```diff
@@ -0,0 +1,100 @@
...
+            "https://raw.githubusercontent.com/fivethirtyeight/"
+            "uber-tlc-foil-response/master/uber-trip-data/"
+            "uber-raw-data-janjune-15.csv.zip"
+        )
+        request.urlretrieve(uber_url_path, temp_zip_path)
+        with zipfile.ZipFile(temp_zip_path) as zf:
+            zf.extractall(path=temp_dir_path)
+        uber_file_path = (
+            temp_dir_path / "uber-dataset" / "uber-raw-data-janjune-15.csv"
+        )
```
jaheba, line 48:
```suggestion
        uber_file_path = temp_dir_path / "uber-raw-data-janjune-15.csv"
```

I think the `uber-dataset` part needs to be removed.

jaheba (approved):

PR #1986, opened 2022-05-25
Fix OffsetSplitter for negative offsets

### LINKED ISSUE #1984: OffsetSplitter fails when splitting at the very end
## Description
I suspect an off-by-one issue is there in `OffsetSplitter` (and maybe shared by `DateSplitter`)

## To Reproduce

```python
from gluonts.dataset.split import DateSplitter, OffsetSplitter

dataset = ListDataset([
    {"item_id": 0, "start": "2021-03-04", "target": [1.0] * 300},
    {"item_id": 1, "start": "2021-03-04", "target": [2.0] * 300},
], freq="D")

split = OffsetSplitter(prediction_length=7, split_offset=-7).split(dataset)

print([len(t["target"]) for t in split.train])

print([len(t["target"]) for t in split.test])
```

```
  assert train.end + train.end.freq * prediction_length <= test.end
  File "issues/data_split/split_and_eval.py", line 9, in <module>
    split = OffsetSplitter(prediction_length=7, split_offset=-7).split(dataset)
  File "/Users/stellalo/gluon-ts/src/gluonts/dataset/split/splitter.py", line 230, in split
    assert train.end + train.end.freq * prediction_length <= test.end
  File "/Users/stellalo/gluon-ts/src/gluonts/dataset/split/splitter.py", line 134, in end
    return self.target.index[-1]
  File "/Users/stellalo/.virtualenvs/gluonts/lib/python3.7/site-packages/pandas/core/indexes/extension.py", line 279, in __getitem__
    result = self._data[key]
  File "/Users/stellalo/.virtualenvs/gluonts/lib/python3.7/site-packages/pandas/core/arrays/datetimelike.py", line 322, in __getitem__
    result = super().__getitem__(key)
  File "/Users/stellalo/.virtualenvs/gluonts/lib/python3.7/site-packages/pandas/core/arrays/_mixins.py", line 194, in __getitem__
    result = self._ndarray[key]
IndexError: index -1 is out of bounds for axis 0 with size 0
```

while I would expect this to print

```
[293, 293]
[300, 300]
```

lostella:
*Issue #, if available:* Fixes #1984 

### DIFF
--- a/src/gluonts/dataset/split/splitter.py
+++ b/src/gluonts/dataset/split/splitter.py
@@ -274,6 +274,7 @@ class OffsetSplitter(pydantic.BaseModel, AbstractBaseSplitter):
         If given, all entries in the *test*-set have a max-length of
         `max_history`. This can be used to produce smaller file-sizes.
     """
+
     prediction_length: int
     split_offset: int
     max_history: Optional[int] = None
@@ -292,7 +293,8 @@ class OffsetSplitter(pydantic.BaseModel, AbstractBaseSplitter):
 
 class DateSplitter(AbstractBaseSplitter, pydantic.BaseModel):
     """
-    A splitter that slices training and test data based on a ``pandas.Timestamp``.
+    A splitter that slices training and test data based on a
+    ``pandas.Timestamp``.
 
     Training entries obtained from this class will be limited to observations
     up to (including) the given ``split_date``.
--- a/src/gluonts/dataset/split/splitter.py
+++ b/src/gluonts/dataset/split/splitter.py
@@ -189,16 +189,6 @@ class AbstractBaseSplitter(ABC):
             `max_history`. This can be used to produce smaller file-sizes.
     """
 
-    # @property
-    # @abstractmethod
-    # def prediction_length(self) -> int:
-    #     pass
-
-    # @property
-    # @abstractmethod
-    # def max_history(self) -> Optional[int]:
-    #     pass
-
     @abstractmethod
     def _train_slice(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
         pass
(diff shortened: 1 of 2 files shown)


src/gluonts/dataset/split/splitter.py
```diff
@@ -281,6 +281,8 @@ def _test_slice(
         self, item: TimeSeriesSlice, offset: int = 0
     ) -> TimeSeriesSlice:
         offset_ = self.split_offset + offset + self.prediction_length
+        if self.split_offset < 0 and offset_ >= 0:
```
jaheba, line 284: Why do you check for `offset_ >= 0:`?

jaheba:
Should we also adjust the documentation? Negative offset values is probably what one wants in most cases.

src/gluonts/dataset/split/splitter.py
lostella: So that when `offset_ > 0` it gets increased, exceeds `len(item)` and fails the assertion that follows (which is what we want in this case).

Maybe instead of an assertion we raise a `ValueError` with a nice message?

lostella: Or should we let the final indexing operation surface the out-of-bounds index?

src/gluonts/dataset/split/splitter.py
```diff
@@ -267,7 +257,22 @@ def rolling_split(
 
 class OffsetSplitter(pydantic.BaseModel, AbstractBaseSplitter):
     """
-    Requires uniform data.
+    A splitter that slices training and test data based on an integer offset.
```
jaheba, line 260: Maybe we can point out that it's a "fixed integer offset"?

src/gluonts/dataset/split/splitter.py
```diff
@@ -267,7 +257,22 @@ def rolling_split(
...
+
+    A positive offset indicates how many observations since the start of each
+    series should be in the training slice; a negative offset indicates how
+    many observations before the end of each series should be excluded from
+    the training slice.
```
jaheba, line 265: Maybe it's easier to think about slicing from the left and right to decide what goes into which set.

jaheba (approved):
Minor comments, but I'm happy to take this version.

jaheba (approved):

PR #1997, opened 2022-06-01
Remove `freq` argument from `Forecast`.

kashif:
So the `freq` string in the estimators and model modules are used for 2 things: 
1. to create a list of time feature classes and 
1. to calculate the `lags_seq` array. 

However, if I manually supply the estimators with the time features and lag sequences I am able to train models without specifying the freq of the data. But at inference time gluonts still expects the forecasts to have a freq string which can now be deduced from the start date of the dataset.

*Description of changes:*
Removed the notion of freq from Forecasts. This way one is able to train/predict model with data from different freq.


jaheba:
Thanks! I think this makes sense as an intermediate step.

I want to be even more radical and make the prediction interface much simpler: #1998

src/gluonts/model/forecast.py
```diff
@@ -350,6 +350,8 @@ def __init__(
         ), "start_date should be a pandas Period object"
         self.start_date = start_date
 
+        if freq is None:
+            freq = self.start_date.freqstr
```
lostella, line 354: Is the idea to still accept `freq` as input for backward compatibility? Because since it's actually redundant, it may lead to confusion. Should we deprecate it?

src/gluonts/evaluation/backtest.py
```diff
@@ -74,7 +73,7 @@ def add_ts_dataframe(
             data = data_entry.copy()
             index = pd.period_range(
                 start=data[FieldName.START],
-                freq=freq,
+                freq=data[FieldName.START].freq,
```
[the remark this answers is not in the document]
kashif: ah yes could be let me confirm!

src/gluonts/model/forecast.py
kashif: I believe its not needed... In all my PR I try to make sure I do not break anything but yes we can slowly deprecate the freq arg here

kashif: so i believe the Predictor also uses it but I did not chase up why the Predictor needs the freq, its most probably also not needed... but I didn't want to break anything...

jaheba: I think I'm fine with removing it, do tests pass though?

kashif: hmm i can check...

kashif: @jaheba yes I believe i can make things work if I remove freq... should i do it?

jaheba: Let's break things!

kashif:
@jaheba the evaluator uses the forecaster's freq in the `calculate_seasonal_error` helper... so checking now how to fix that

src/gluonts/evaluation/_base.py
```diff
@@ -332,7 +332,7 @@ def get_metrics_per_ts(
 
         median_fcst = forecast.quantile(0.5)
         seasonal_error = calculate_seasonal_error(
-            past_data, forecast, self.seasonality
+            past_data, forecast.start_date.freqstr, self.seasonality
```
jaheba, line 335: How about we create a property on `Forecast` that returns `start_data.freq`?

kashif: ah yes... sometimes i get confused when to use `freq` and when to use `freqstr`... but good idea let me do that

kashif:
ok so now the real breaking changes... I will remove the freq arg. from the `Predictor`...

kashif:
@jaheba should I make another PR to remove the freq from the Predictors?

lostella (approved):
Thanks @kashif! This is a very nice change in my opinion. I'd ask @jaheba to cross-check, given it's a breaking change, worst case we'll let this rest over the long weekend 🙃

lostella:
> @jaheba should I make another PR to remove the freq from the Predictors?

In my opinion yes, we should do a separate one

kashif:
awesome! @lostella will send that PR once this is merged! I believe with this change and the potentially irregular PR change, it opens up a bunch of nice possibilities! looking forward to using it once released!

src/gluonts/model/tpp/forecast.py
```diff
@@ -107,21 +107,24 @@ def __init__(
         self.start_date = start_date
 
         assert isinstance(freq, str), "freq should be a string"
-        self.freq = freq
+        self.interarrival_freq = freq
```
jaheba, line 110: Why not directly use `freq`?

kashif: right I thought it was a better name for this arg. rather than the data freq.... I can revert if you like?

jaheba: I guess either are fine. Let's just merge this and think about it later (maybe).

jaheba (approved):

kashif:
awesome! please merge and then i can test better

kashif:
@jaheba i believe its the dataset... can you kindly send me a plot of

```py
dataset = get_dataset("traffic")

index = 123
test = list(dataset.test)
plt.plot(test[index]["target"][-60:])
```
download

kashif:
@jaheba yes I believe something is wrong with gluonts 0.9.4 I get:

download

jaheba:
image

That's what I'm getting on `master` -- haven't tried. `0.9.4`.

kashif:
hmm... really strange! I must be going mad! haha

jaheba:
Can imagine! Can you share the raw values of the strange plot?

kashif:
of course this morning i do not see it haha 😠

kashif:
ok i believe i can reproduce the issue now... the issue occurs if i re-read the dataset in the same jupyter session:

```
traffic = get_dataset("traffic")
test = list(traffic.test)

test[index]["target"][-60:]
array([0.0991, 0.1109, 0.1059, 0.1046, 0.0994, 0.1018, 0.0951, 0.0823,
       0.0817, 0.0736, 0.0662, 0.0647, 0.068 , 0.0697, 0.0472, 0.0298,
       0.0198, 0.0146, 0.0105, 0.0119, 0.0195, 0.0263, 0.0362, 0.0511,
       0.0771, 0.1023, 0.1   , 0.1045, 0.0941, 0.0901, 0.0903, 0.077 ,
       0.0747, 0.0703, 0.0655, 0.0692, 0.0647, 0.0608, 0.0431, 0.0227,
       0.0162, 0.0138, 0.0121, 0.0216, 0.0537, 0.0945, 0.1789, 0.2191,
       0.1024, 0.1265, 0.1061, 0.1025, 0.1024, 0.1023, 0.1074, 0.1159,
       0.1239, 0.1131, 0.0784, 0.0765], dtype=float32)
```

and if I redo this:

```
traffic = get_dataset("traffic")
test = list(traffic.test)

test[index]["target"][-60:]

array([0.0991, 0.109 , 0.1059, 0.1046, 0.0994, 0.108 , 0.0951, 0.0823,
       0.087 , 0.0736, 0.0662, 0.0647, 0.068 , 0.0697, 0.0472, 0.0298,
       0.098 , 0.046 , 0.005 , 0.019 , 0.095 , 0.0263, 0.0362, 0.051 ,
       0.0771, 0.1023, 0.1   , 0.1045, 0.0941, 0.0901, 0.0903, 0.077 ,
       0.0747, 0.0703, 0.0655, 0.0692, 0.0647, 0.0608, 0.0431, 0.0227,
       0.062 , 0.038 , 0.021 , 0.026 , 0.0537, 0.0945, 0.1789, 0.291 ,
       0.1024, 0.1265, 0.1061, 0.1025, 0.1024, 0.1023, 0.1074, 0.159 ,
       0.1239, 0.131 , 0.0784, 0.0765], dtype=float32)
```

note some values are not the same! and then if i re-run it the issue disappears.... crazy!!

jaheba:
Shouldn't the dataset be already on disk? This indeed look very strange!

lostella:
@kashif could you put everything in a single snippet and open a separate issue? This is worrying

Issue #1000, opened 2020-08-21
Proposal: Dataset NG


jaheba:
Goal: Replace current input-data handling with a composed approach. 

## Schema

We introduce `Schema`s and each `Forecater` has to define a data schema that it
supports.

This is based on #776.

```python
deepar = DeepAREstimator(freq="D", prediction_length=7)
deepar.get_schema()
```

## Datasource

The `Dataset` abstraction gets replaced with `DataSource`. A data source is
just an `Iterator[dict]` and thus a list can be also used as a data source:

```python
    xs = [{"start": "2020", "target": [1, 2, 3, 4, 5]}]
```

Mapping a Schema to a data source produces what we used to describe as a
dataset:

```python
dataset = Map(deepar.get_schema(), datasource)
```

## Loader

We replace data loaders with pipelines using `glide.ParApply`,

```python
parts = partition(datasource, N)

train_data_loader = ParApply(
    compose(
        cycle,
        apply_schema,
        apply_transformation,
        shuffle,
        stack,
    ),
    parts
)
```

PR #2001, opened 2022-06-03
Dataset schema reconciliation

Hongqing-work:
Issue #1000: Proposal: Dataset NG

Description of changes:

Add get_schema method for DeepAR model.

### DIFF
--- a/src/gluonts/model/deepar/_estimator.py
+++ b/src/gluonts/model/deepar/_estimator.py
@@ -11,14 +11,17 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
+from dataclasses import field
 from functools import partial
+from itertools import starmap
 from typing import List, Optional, Type
 
 import numpy as np
 from mxnet.gluon import HybridBlock
 
 from gluonts.core.component import validated
-from gluonts.dataset.common import Dataset
+from gluonts.dataset.common import Dataset, NumpyArrayField, PandasTimestampField
+from gluonts.dataset.common import Schema
 from gluonts.dataset.field_names import FieldName
 from gluonts.dataset.loader import (
     DataLoader,
@@ -489,3 +492,15 @@ class DeepAREstimator(GluonEstimator):
             ctx=self.trainer.ctx,
             dtype=self.dtype,
         )
+
+    def get_schema(self, one_dim_target = True) -> Schema:
+        fields = dict()
+        fields["start"] = PandasTimestampField(freq=self.freq,is_optional=False)
+        fields["target"] = NumpyArrayField(dtype=self.dtype, ndim=1 if one_dim_target else 2, is_optional=False)
+        if self.use_feat_dynamic_cat:
+            fields["feat_dynamic_cat"] = NumpyArrayField(dtype=np.int32,ndim=2,is_optional=False)
+        if self.use_feat_dynamic_real:
+            fields["feat_dynamic_real"] = NumpyArrayField(dtype=self.dtype,is_optional=False)
+        if self.use_feat_static_real:
+            fields["feat_static_real"] = NumpyArrayField(dtype=self.dtype,ndim=2,is_optional=False)
+        return Schema(fields)
--- a/src/gluonts/model/deepar/_estimator.py
+++ b/src/gluonts/model/deepar/_estimator.py
@@ -11,10 +11,8 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-from dataclasses import field
 from functools import partial
-from itertools import starmap
-from typing import List, Optional, Type
+from typing import List, Optional, Type, Dict, Any
 
 import numpy as np
 from mxnet.gluon import HybridBlock
@@ -65,7 +63,7 @@ from gluonts.transform.feature import (
 )
 
 from ._network import DeepARPredictionNetwork, DeepARTrainingNetwork
-from ._schema import Schema, PandasTimestampField, NumpyArrayField
+from ._schema import Schema, FieldType, PandasPeriodField, NumpyArrayField
 
 
 class DeepAREstimator(GluonEstimator):
@@ -494,13 +492,15 @@ class DeepAREstimator(GluonEstimator):
         )
 
     def get_schema(self) -> Schema:
-        fields = dict()
-        fields["start"] = PandasTimestampField(freq=self.freq)
+        fields: Dict[str, Any] = dict()
+        fields["start"] = PandasPeriodField(freq=self.freq)
         fields["target"] = NumpyArrayField(dtype=self.dtype, ndim=1)
         if self.use_feat_dynamic_real:
             fields["feat_dynamic_real"] = NumpyArrayField(dtype=self.dtype)
         if self.use_feat_static_cat:
-            fields["feat_static_cat"] = NumpyArrayField(dtype=np.int32,ndim=2)
+            fields["feat_static_cat"] = NumpyArrayField(dtype=np.int32, ndim=2)
         if self.use_feat_static_real:
-            fields["feat_static_real"] = NumpyArrayField(dtype=self.dtype,ndim=2)
+            fields["feat_static_real"] = NumpyArrayField(
+                dtype=self.dtype, ndim=2
+            )
         return Schema(fields)
--- a/src/gluonts/model/deepar/_estimator.py
+++ b/src/gluonts/model/deepar/_estimator.py
@@ -20,8 +20,7 @@ import numpy as np
 from mxnet.gluon import HybridBlock
 
 from gluonts.core.component import validated
-from gluonts.dataset.common import Dataset, NumpyArrayField, PandasTimestampField
-from gluonts.dataset.common import Schema
+from gluonts.dataset.common import Dataset
 from gluonts.dataset.field_names import FieldName
 from gluonts.dataset.loader import (
     DataLoader,
@@ -66,6 +65,7 @@ from gluonts.transform.feature import (
 )
 
 from ._network import DeepARPredictionNetwork, DeepARTrainingNetwork
+from ._schema import Schema, PandasTimestampField, NumpyArrayField
 
 
 class DeepAREstimator(GluonEstimator):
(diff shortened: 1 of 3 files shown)


src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,161 @@
...
+        return f"PandasPeriodField(freq={self.freq!r})"
+
+    def __call__(self, value: Any) -> pd.Period:
+        period = PandasPeriodField._process(value, self.freq)
+        return period
+
+    @staticmethod
+    @lru_cache(maxsize=10000)
+    def _process(string: str, freq: pd.DateOffset) -> pd.Timestamp:
+        return pd.Period(string, freq=freq)
```
jaheba, line 113: I think we can just remove this. We used to do more complicated calculations here, but `pd.Period` is a lot faster than `pd.Timestamp`. So `__call__` should just be `return pd.Period(value, self.freq)`.

src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+class NumpyArrayField(FieldType[np.ndarray]):
+    def __init__(
+        self,
+        dtype: Type = np.float32,
+        ndim: Optional[int] = None,
+    ) -> None:
+        self.dtype = dtype
+        self.ndim = ndim
```
jaheba, line 46: I think we can simplify this by using a `dataclass`.

src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def __eq__(self, other):
+        if not isinstance(other, NumpyArrayField):
+            return False
+        return self.dtype == other.dtype and self.ndim == other.ndim
```
jaheba, line 51:
```suggestion
        if not isinstance(other, NumpyArrayField):
            return False

        return self.dtype == other.dtype and self.ndim == other.ndim
```

I generally prefer using visual blocks in these cases.

src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def __repr__(self):
+        return f"NumpyArrayField(dtype={self.dtype!r}, ndim={self.ndim!r}"
```
jaheba, line 54: When using a `dataclass` we can also remove this.

src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,161 @@
...
+            )
+        return value
+
+    def is_compatible(self, value: Any) -> bool:
+        if not isinstance(value, (list, tuple, np.ndarray)):
+            return False
+
+        try:
+            x = np.asarray(value)
+        except:
```
jaheba, line 71: We should not use blanket excepts -- at least we should do `except Exception`

src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,161 @@
...
+        return True
+
+
+class Schema:
+    def __init__(self, fields: Dict[str, FieldType]) -> None:
+        self.fields = fields
+        # select fields that should be handled in the loop, because
+        # - they are non optional
+        # - or they are not AnyFields
+        self._fields_for_processing = {k: f for k, f in self.fields.items()}
```
jaheba, line 131: I think this can be removed since we don't have optional fields anymore.

src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def __eq__(self, other):
+        if self.fields.keys() != other.fields.keys():
+            return False
+        return all(self.fields[k] == other.fields[k] for k in self.fields)
```
jaheba, line 136: Couldn't this just be `self.fields == other.fields`?

src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def __repr__(self):
+        return (
+            "Schema(fields={"
+            + ", ".join(f"'{k}':{v}" for k, v in self.fields.items())
+            + "})"
+        )
+
+    def __call__(self, d: Dict[str, Any]) -> None:
```
jaheba, line 145: Maybe we can have an `in_place` parameter, which allows us to either mutate the passed value or otherwise returns a copy.

src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,161 @@
...
+        if not isinstance(other, NumpyArrayField):
+            return False
+        return self.dtype == other.dtype and self.ndim == other.ndim
+
+    def __repr__(self):
+        return f"NumpyArrayField(dtype={self.dtype!r}, ndim={self.ndim!r}"
+
+    def __call__(self, value: Any) -> np.ndarray:
+        value = np.asarray(value, dtype=object)
+        value = value.astype(self.dtype)
```
jaheba, line 58: Why do this in two steps?

src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,149 @@
...
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+from typing import Any, Dict, Optional, Union, Generic, TypeVar, Type
+
+import numpy as np
+import pandas as pd
+from dataclasses import dataclass
```
jaheba, line 18: Minor: We want to keep the imports in three blocks: 1st party (provided by python), 3rd party (our dependencies that we installed, and then imports from `gluonts`.

jaheba: In this case this would look like:

```py
from dataclasses import dataclass
from typing import Any, Dict, Optional, Union, Generic, TypeVar, Type

import numpy as np
import pandas as pd
from pandas.tseries.frequencies import to_offset

from gluonts.exceptions import GluonTSDataError
```

src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,149 @@
...
+        if not isinstance(v, (str, pd.Period)):
+            return False
+        try:
+            self(v)
+        except Exception:
+            return False
+        return True
+
+
+class Schema:
```
jaheba, line 111: Let's also make this a dataclass.

src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,149 @@
...
+    def __init__(self, fields: Dict[str, FieldType]) -> None:
+        self.fields = fields
+
+    def __eq__(self, other):
+        return self.fields == other.fields
```
jaheba, line 116: Another benefit of dataclasses, it implements `__eq__` for us automatically.

src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,149 @@
...
+            + "})"
+        )
+
+    def __call__(self, d: Dict[str, Any], inplace: bool):
+        """
+        inplace
+            applies the schema to a data dict if True. The dictionary is updated in place.
+            return a new data dictionary if False.
+        """
+        copied_data: Dict[str, Any] = dict()
```
jaheba, line 131: You can do something like:

```suggestion
        if inplace:
            out = d
        else:
            out: Dict[str, Any] = {}
```

And then always write to `out`.

src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,149 @@
...
+                if inplace:
+                    d[field_name] = field_type(value)
+                else:
+                    copied_data[field_name] = field_type(value)
+            except Exception as e:
+                raise GluonTSDataError(
+                    f"Error when processing field {field_name} using {field_type}"
+                ) from e
+        if not inplace:
+            return copied_data
```
jaheba, line 149: Let's always return something.

src/gluonts/model/deepar/_schema.py
Hongqing-work: By doing so, we can convert string type to our target type. `asarray` can’t handle it.

src/gluonts/model/deepar/_estimator.py
```diff
@@ -489,3 +451,23 @@ def create_predictor(
             ctx=self.trainer.ctx,
             dtype=self.dtype,
         )
+
+    def get_schema(self) -> Schema:
+        fields: Dict[str, Any] = dict()
+        default_values: Dict[str, Any] = dict()
```
jaheba, line 457: Minor: I prefer `{}` over `dict`.

Also, do you need to declare the types for `mypy` to not complain?

src/gluonts/model/deepar/_estimator.py
```diff
@@ -489,3 +451,23 @@ def create_predictor(
...
+        fields["start"] = PandasPeriodField(freq=self.freq)
+        # in the following line, we add 1 for the time dimension
+        fields["target"] = NumpyArrayField(
+            dtype=self.dtype, ndim=1 + len(self.distr_output.event_shape)
+        )
+        if self.use_feat_dynamic_real:
+            fields["feat_dynamic_real"] = NumpyArrayField(
+                dtype=self.dtype, ndim=1
+            )
+        fields["feat_static_cat"] = NumpyArrayField(dtype=self.dtype, ndim=1)
```
jaheba, line 467: This looks like we are always requiring `feat_static_cat`, even if it is set to not be used.

If I'm missing something, can you add some comments to why you do things like you do?

Last thing: I would prefer if each field is handled one after another. So this check should be together with the field definition:

```py
        if not self.use_feat_static_cat:
            default_values["feat_static_cat"] = [0.0]
```

src/gluonts/model/deepar/_schema.py
jaheba: Can you provide an example?

src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,155 @@
...
+                    )
+            try:
+                out[field_name] = field_type(value)
+            except Exception as e:
+                raise GluonTSDataError(
+                    f"Error when processing field {field_name} using "
+                    "{field_type}"
+                ) from e
+        if inplace:
+            remove_field_names = []
```
jaheba, line 149: I'm not sure we want to do this, but I also see why we would want to.

Maybe have it as an option?

src/gluonts/model/deepar/_schema.py
```diff
@@ -0,0 +1,155 @@
...
+            for field_name in out.keys():
```
jaheba, line 150: I think this can be simplified to:

```py
for key in out.keys() - self.fields.keys():
    del out[key]
```

src/gluonts/model/deepar/_estimator.py
Hongqing-work: In `hybrid_forward`, deepAR has `feat_static_cat: Tensor` and `feat_static_real: Tensor` as input. `get_hybrid_forward_input_names` can always return what a model needs for network forward process.

src/gluonts/model/deepar/_schema.py
Hongqing-work: I think this doesn't work too. What I what to solve is converting float string data like '1.2' to int data 1.
```
>>> a = np.asarray(['1.2', '2.3'], dtype=float)
>>> print(a)
[1.2 2.3]
>>> a = np.asarray(['1.2', '2.3'], dtype=int)
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '1.2'
>>> a = np.asarray(['1.2', '2.3'], dtype=object)
>>> print(a.astype(int))
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '1.2'
```

src/gluonts/model/deepar/_schema.py
Hongqing-work: Yes, `SelectFields` will do it later.
```
return TrainDataLoader(
            dataset=data,
            transform=instance_splitter + SelectFields(input_names),
            batch_size=self.batch_size,
            stack_fn=partial(batchify, ctx=self.trainer.ctx, dtype=self.dtype),
            decode_fn=partial(as_in_context, ctx=self.trainer.ctx),
            **kwargs,
        )
```
So I think I can remove it here.

PR #2002, opened 2022-06-03
add ND for item_metrics

melopeo:
In this PR we compute ND for item_metrics, as currently it is only available for aggregated metrics.


lostella (approved):

lostella:
@melopeo just a note: the value of this metric (like others I believe) may be `nan` or `inf` if the target is zero

melopeo:
@lostella that's correct. It is expected that several metrics have this behaviour in that case.

PR #2004, opened 2022-06-07
Refactor splitter assertion and add error message

rsnirwan:
When the splitter fails to split an entry, there is no information on why and where to look for. This PR adds a massage on what can be potentially wrong and extends docstrings to give more information on how `split_date` and `split_offset` should be set. 

### DIFF
--- a/src/gluonts/dataset/split/splitter.py
+++ b/src/gluonts/dataset/split/splitter.py
@@ -220,7 +220,7 @@ class AbstractBaseSplitter(ABC):
 
             prediction_length = getattr(self, "prediction_length")
 
-            self._check_split_length(train.end, test.end, prediction_length)
+            _check_split_length(train.end, test.end, prediction_length)
             split._add_test_slice(test)
 
         return split
@@ -249,23 +249,21 @@ class AbstractBaseSplitter(ABC):
                 )
                 prediction_length = getattr(self, "prediction_length")
 
-                self._check_split_length(
-                    train.end, test.end, prediction_length
-                )
+                _check_split_length(train.end, test.end, prediction_length)
                 split._add_test_slice(test)
 
         return split
 
-    @staticmethod
-    def _check_split_length(
-        train_end: pd.Period, test_end: pd.Period, prediction_length: int
-    ) -> None:
-        msg = (
-            "Not enough observations after the split point to construct"
-            " the test instance; consider using longer time series,"
-            " or splitting at an earlier point."
-        )
-        assert train_end + prediction_length <= test_end, msg
+
+def _check_split_length(
+    train_end: pd.Period, test_end: pd.Period, prediction_length: int
+) -> None:
+    msg = (
+        "Not enough observations after the split point to construct"
+        " the test instance; consider using longer time series,"
+        " or splitting at an earlier point."
+    )
+    assert train_end + prediction_length <= test_end, msg
 
 
 class OffsetSplitter(pydantic.BaseModel, AbstractBaseSplitter):


src/gluonts/dataset/split/splitter.py
```diff
@@ -250,13 +249,24 @@ def rolling_split(
...
         return split
 
+    @staticmethod
+    def _check_split_length(
+        train_end: pd.Period, test_end: pd.Period, prediction_length: int
+    ) -> None:
+        msg = (
+            "Number of left over values is too small to create the test"
+            + " window. Either the timeseries is too short or"
+            + " ``split_offset`` or ``split_date`` is not set properly."
```
lostella, line 266: I think this goes into details of the subclasses, which is not ideal. Maybe just saying “Not enough observations after the split point to construct the test instance; consider using longer time series, or splitting at an earlier point”

src/gluonts/dataset/split/splitter.py
```diff
@@ -266,7 +276,9 @@ class OffsetSplitter(pydantic.BaseModel, AbstractBaseSplitter):
     A positive offset indicates how many observations since the start of each
     series should be in the training slice; a negative offset indicates how
     many observations before the end of each series should be excluded from
-    the training slice.
+    the training slice. Please make sure that the number of excluded values
+    is enough for the test case, i.e., at least ``prediction_length`` (for
+    ``rolling_split`` multiple of ``prediction_length``) values are left off.
```
lostella, line 281: Minor: why is this in the class docstring for `OffsetSplitter` but in the argument description for `DateSplitter`? :-)

RSNirwan: I was asking myself too :D Rolling_split was explained in the main docstring instead of in the parameters section. I assume whoever did that had a reason...
I'll put it into the parameters section now.

src/gluonts/dataset/split/splitter.py
```diff
@@ -250,13 +249,24 @@ def rolling_split(
...
         return split
 
+    @staticmethod
+    def _check_split_length(
+        train_end: pd.Period, test_end: pd.Period, prediction_length: int
+    ) -> None:
+        msg = (
+            "Not enough observations after the split point to construct"
+            + " the test instance; consider using longer time series,"
+            + " or splitting at an earlier point."
```
jaheba, line 266:
```suggestion
            "Not enough observations after the split point to construct"
            " the test instance; consider using longer time series,"
            " or splitting at an earlier point."
```

You can leave the plusses out. Some people don't like it, but I do :)

src/gluonts/dataset/split/splitter.py
```diff
@@ -250,13 +249,24 @@ def rolling_split(
...
-                assert (
-                    train.end + train.end.freq * prediction_length <= test.end
+                self._check_split_length(
+                    train.end, test.end, prediction_length
                 )
                 split._add_test_slice(test)
 
         return split
 
+    @staticmethod
```
jaheba, line 259: Not sure I like being a `staticmethod`. I prefer having a standalone function instead.

jaheba (commented):
Thanks for the PR?

Should we have an ignore option that would ignore time-series that are not long enough?

src/gluonts/dataset/split/splitter.py
RSNirwan: Yes, an ignore options would be nice. I ll do that towards the end of the week/next week. I also want to make `item_id` optional. Currently it's mandatory.

lostella (approved):
🚀

PR #2005, opened 2022-06-07
Rework `itertools`, add col-to-row and row-to-col functions.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/itertools.py
+++ b/src/gluonts/itertools.py
@@ -47,8 +47,8 @@ def prod(xs):
     return p
 
 
-@dataclass  # type: ignore[misc]
-class Cyclic(Collection):
+@dataclass
+class Cyclic:
     """
     Like `itertools.cycle`, but does not store the data.
     """
@@ -87,8 +87,8 @@ def batcher(iterable: Iterable[T], batch_size: int) -> Iterator[List[T]]:
     return iter(get_batch, [])
 
 
-@dataclass  # type: ignore[misc]
-class Cached(Collection):
+@dataclass
+class Cached:
     """
     An iterable wrapper, which caches values in a list the first time it is
     iterated.
@@ -116,8 +116,8 @@ class Cached(Collection):
         return len(self.iterable)
 
 
-@dataclass  # type: ignore[misc]
-class PseudoShuffled(Collection):
+@dataclass
+class PseudoShuffled:
     """
     Yields items from a given iterable in a pseudo-shuffled order.
     """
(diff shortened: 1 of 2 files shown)


jaheba:
Maybe r2c for row-2-col is a better name?

src/gluonts/itertools.py
```diff
@@ -129,18 +149,79 @@ def __len__(self) -> int:
...
+    def next(self):
+        self.current = list(itertools.islice(self.iterable, self.length))
 
     def __iter__(self):
-        return itertools.islice(self.iterable, self.length)
+        yield from self.current
+        self.next()
 
     def __len__(self) -> int:
-        return len(self.iterable)
```
lostella, line 146: So this was wrong?

src/gluonts/itertools.py
```diff
@@ -129,18 +149,79 @@ def __len__(self) -> int:
...
+        return len(self.current)
+
+
+K = TypeVar("K")
+V = TypeVar("V")
+
+
+def rows_to_columns(
```
lostella, line 189: These two are very simple, but we should probably have a few test cases

lostella (changes requested):

src/gluonts/itertools.py
lostella: Right, I'm wondering if it makes even sense to have `__len__` here. Also, we probably lacked tests for it.

src/gluonts/itertools.py
jaheba: They have a simple test case in case of the doctest!

src/gluonts/itertools.py
jaheba: I'm not so sure either, I guess I can remove it and check what happens?

src/gluonts/itertools.py
lostella: Do those run? I cannot remember honestly :-)

src/gluonts/itertools.py
lostella: Yes, let's do that

src/gluonts/itertools.py
```diff
@@ -129,18 +149,79 @@ def __len__(self) -> int:
...
-    over multiple times.
+    over multiple times:
+
+        >>> isl = IterableSlice([1, 2, 3, 4, 5], 3)
+        >>> list(isl)
+        [1, 2, 3]
+        >>> list(isl)
+        [4, 5]
+        >>> list(isl)
+        []
```
lostella, line 164: I'm not sure this is correct, I would expect `isl` to materialize to the same list every time

lostella: `IterableSlice(iter([1, 2, 3, 4, 5]), 3)` would give what the docstring says

lostella (approved):

PR #2006, opened 2022-06-07
Added support of "long" RTS, making short RTS be "past_feat_dynamic_real"

zoolhasson:
*Issue #, if available:*

*Description of changes:*

I added support for related time series (RTS) that are accessible in the forecast horizon. I changed the syntax so that 'feat_dynamic_real' now means "long" RTS (having info in the forecast horizon), and 'past_feat_dynamic_real' refers to "short" RTS (having no info in the forecast horizon). This is a change of syntax from before in the sense that before there was only support for 'feat_dynamic_real', which was assumed to be short, not long.


youngsuk0723 (approved):
Approve

dcmaddix:
Let’s call them past_feat_dynamic_real to be consistent with MQ-CNN.

dcmaddix (approved):

PR #2012, opened 2022-06-10
Move `dataset.repository` to `lab`.

jaheba:
*Issue #, if available:*

*Description of changes:*


jaheba:
Something I want to achieve is to separate the library stuff from more interactive utilities. Things that belong to `lib` should be happily imported across different gluon-ts modules, but stuff that is `lab` is only used by end-users and in tests.

I think for `lib` longer import paths can be a good thing, since they are more precise about where things come from.

However, for `lab` we could be more aggressive and do things like:

```py
from gluonts.lab import get_dataset
```

PR #2015, opened 2022-06-10
Fix mean in `AffineTransformedDistribution`

stailx:
Incorrect mean property in AffineTransformedDistribution.
Scaling was not taken into account in an Affine Transform.
Mathematic proof: mean(y)= mean(scale * x+loc ) = scale * mean(x) + loc.


src/gluonts/mx/distribution/transformed_distribution.py
```diff
@@ -210,9 +210,9 @@ def __init__(
     @property
     def mean(self) -> Tensor:
         return (
-            self.base_distribution.mean
+            self.base_distribution.mean * self.scale
```
lostella, line 213: This does not account for the case `self.scale = None`. You could use `scale` defined as

```python
scale = self.scale if self.scale is not None else 1
```

lostella (changes requested):
@stailx thanks for spotting this! I only have one comment to be addressed

lostella (approved):
@stailx thanks again for spotting this! I took the liberty of pushing this over the finish line, I hope you don't mind

🚀

kashif:
well spotted @stailx

PR #2017, opened 2022-06-12
Remove mandatory `freq` attribute of `Predictor`.

kashif:
*Issue #, if available:*

*Description of changes:*

Follow-up changes to #1997 


kashif:
@lostella and @jaheba do you have any idea how to get rid of the typecheck errors?

lostella:
@kashif all of the complaints appear to be of the form:

```
src/gluonts/torch/model/deepar/module.py:116: error: Argument "freq_str" to "get_lags_for_frequency" has incompatible type "Optional[str]"; expected "str"
```

It looks like you made `freq` of type `Optional` in a few places, and downstream code is not happy with it (expects `str`). In the specific case of DeepAR above, one could make the explicit lags mandatory in case `freq` is not given.

kashif:
@lostella thanks! yes I believe i have added asserts so that at the point these functions are called they freq is not None...

kashif:
@canerturkmen do you have any code to test the `DeepTPPEstimator`?  I believe I can remove the `freq` arg. from it but I cannot currently test it as there don't seem to be tests for this estimator...

kashif:
@jaheba this PR also needs a a breaking tag

kashif:
@lostella  can you kindly help me with this last type issue? Everything I try doesn't seem to work haha thanks!

src/gluonts/model/deepstate/_estimator.py
```diff
@@ -79,7 +79,9 @@
 }
 
 
-def longest_period_from_frequency_str(freq_str: str) -> int:
+def longest_period_from_frequency_str(freq_str: Optional[str] = None) -> int:
+    if freq_str is None:
+        freq_str = "T"
```
jaheba: Can you add an explanation of why you set this to `minute`?

src/gluonts/model/deepvar/_estimator.py
```diff
@@ -114,8 +118,11 @@ def time_features_from_frequency_str(freq_str: str) -> List[TimeFeature]:
 
 
 def get_lags_for_frequency(
-    freq_str: str, num_lags: Optional[int] = None
+    freq_str: Optional[str] = None, num_lags: Optional[int] = None
 ) -> List[int]:
+    if freq_str is None:
+        return [[1]]
+
```
jaheba: Not sure this is the best way to do it. I think `freq` should be mandatory, but we can do the check elsewhere.

src/gluonts/model/tpp/predictor.py
```diff
@@ -31,7 +31,7 @@
 
 
 class PointProcessForecastGenerator(ForecastGenerator):
-    def __call__(
+    def __call__(  # type: ignore
```
jaheba, line 34: Why is this needed?

src/gluonts/model/transformer/_estimator.py
```diff
@@ -182,8 +182,13 @@ def __init__(
         assert (
             num_parallel_samples > 0
         ), "The value of `num_parallel_samples` should be > 0"
+        assert (
+            freq is not None or lags_seq is not None
+        ), "Either `freq` or `lags_seq` should be set"
+        assert (
+            freq is not None or time_features is not None
+        ), "Either `freq` or `time_features` should be set"
```
jaheba: Looking at this again and again, can we maybe unify this check somehow?

src/gluonts/time_feature/lag.py
```diff
@@ -27,7 +27,9 @@ def _make_lags(middle: int, delta: int) -> np.ndarray:
 
 
 def get_lags_for_frequency(
-    freq_str: str, lag_ub: int = 1200, num_lags: Optional[int] = None
+    freq_str: Optional[str] = None,
```
jaheba: Same as above, I think this should be further mandatory.

src/gluonts/model/deepstate/_estimator.py
kashif: right so no where is this function being called without a freq argument, however to make it  robust when freq is None, I set the resolution to be the finest that is currently supported... which at the moment is "T"... I would much rather fix the mypy issue but as mentioned I couldn't get anything working apart from this...

src/gluonts/model/deepvar/_estimator.py
kashif: you are right... here too no where is this function called without a valid freq. but mypy complains...

src/gluonts/model/tpp/predictor.py
kashif: this was another place mypy complains... I believe TPP also doesn't really need the freq arg. so i can get rid of it and the mypy error will go away... but currently I have no code to test the TPP Estimator... I have asked @canerturkmen

src/gluonts/model/transformer/_estimator.py
kashif: right so these checks are in all the models where freq is not mandatory... i am happy to figure out a better way here... and the github copilot suggested this and I went with it haha

src/gluonts/time_feature/lag.py
kashif: i agree! again a change to make the mypy happy...

src/gluonts/model/deepstate/_estimator.py
jaheba: Can you change it back and then we look at the `mypy` issue?

kashif:
sure @jaheba let me change it back to the nice clean version with mypy error give me a sec

kashif:
@jaheba It was at this point where I started to hit the code with the ugly stick... have a look at the failing type checks...

jaheba:
@kashif How about we keep `freq` mandatory for the models that make use of the field and look at making it optional for them later?

kashif:
@jaheba ok... sure let me do undo that... that will leave only the TPPEstimator issue...

kashif:
@jaheba I see a failing test in the shell... I am not familiar with that... any idea how to deal with that?  I believe the ListDataset should be made using either the configuration or the dataset's properties rather than the freq from the predictor...

jaheba:
> @jaheba I see a failing test in the shell... I am not familiar with that... any idea how to deal with that? I believe the ListDataset should be made using either the configuration or the dataset's properties rather than the freq from the predictor...

Looks like we depend on the `Predictor` frequency, to create the dataset. This will require some breaking changes.

I've discussed with @lostella that it would make sense to force the `freq` into each time-series, so you have something like this:

```json
{"start": "2020-01-01", "freq": "D", ...}
```

But that would mean that we have a new dataset format, which may annoy some people.

kashif:
@jaheba the configuration when it is a Forecast does have a freq which can be used... however sometimes in the test it's the base Config and so configuration.freq doesn't exist...

src/gluonts/shell/serve/app.py
```diff
@@ -99,7 +99,7 @@ def handle_predictions(predictor, instances, configuration):
     # create the forecasts
     forecasts = ThrougputIter(
         predictor.predict(
-            ListDataset(instances, predictor.freq),
+            ListDataset(instances, configuration.freq),
```
jaheba, line 102: This means that users have to pass `freq` in `configuration` when making requests.

It's fine for now, but we should revisit this.

kashif:
@jaheba `hyperparam.env` is also sometimes `None`

kashif:
ok @jaheba  the only issue is the ForecastGenerator for TPPEstimator... I believe it doesn't even require the freq str. but I wanted to check with the author...

jaheba:
I've just added a `type` ignore for now. We can reinvestigate this later.

kashif:
thank you! I will be happy when this is merged haha!

src/gluonts/model/r_forecast/_predictor.py
```diff
@@ -99,11 +99,15 @@ def __init__(
...
+        period: Optional[int] = None,
         trunc_length: Optional[int] = None,
         params: Optional[Dict] = None,
     ) -> None:
-        super().__init__(freq=freq, prediction_length=prediction_length)
+        super().__init__(prediction_length=prediction_length)
+
+        assert (
+            freq is not None or period is not None
+        ), "Either `freq` or `period` should be set"
```
lostella: According to the type annotation, `freq` should never be `None`

src/gluonts/model/san/_estimator.py
```diff
@@ -86,7 +86,10 @@ def __init__(
         batch_size: int = 32,
     ):
         super().__init__(trainer=trainer, batch_size=batch_size)
-        self.freq = freq
+        assert (
+            freq is not None or time_features is not None
+        ), "Either `freq` or `time_features` should be set"
+
```
lostella: Same here

src/gluonts/model/seasonal_naive/_predictor.py
```diff
@@ -53,13 +53,15 @@ def __init__(
...
     ) -> None:
-        super().__init__(freq=freq, prediction_length=prediction_length)
+        super().__init__(prediction_length=prediction_length)
 
         assert (
             season_length is None or season_length > 0
         ), "The value of `season_length` should be > 0"
+        assert (
+            freq is not None or season_length is not None
+        ), "Either `freq` or `season_length` should be set"
```
lostella: Same

src/gluonts/model/r_forecast/_predictor.py
kashif: I know... but i would like to use it with `freq=None` haha...  that is why i kept it...

jaheba:
@lostella any other concerns?

lostella (dismissed):

jaheba (approved):

jaheba:
I've removed all the unnecessary checks.

kashif:
thank you!

PR #2018, opened 2022-06-13
Remove multiprocessing dataloader.

jaheba:
*Issue #, if available:*

*Description of changes:*


lostella (approved):
LGTM but we should probably keep a mechanism around to speed up the transformation chain in case that's the bottleneck. Something like [this](https://github.com/awslabs/gluon-ts/blob/dev/src/gluonts/nursery/glide/parallel.py) may work?

PR #2019, opened 2022-06-13
Rework filedataset

jaheba:
*Issue #, if available:*

*Description of changes:*


src/gluonts/dataset/common.py
```diff
@@ -145,6 +170,44 @@ class FileDataset(Dataset):
         Indicates whether the dataset should be cached or not.
     """
 
+    def __new__(  # type: ignore
```
lostella: Defining `__new__` seems a little odd. Does this need to be a class now? Could it be a function, or an alternative constructor for `DatasetCollection` (something like `from_files`)?

For example I'm not sure when `FileDataset.__iter__` is invoked with this change

src/gluonts/dataset/common.py
```diff
@@ -54,6 +65,30 @@ def __len__(self) -> int:
         raise NotImplementedError
 
 
+@dataclass
+class DatasetCollection(Dataset):
+    """Flattened access to a colleciton of datasets."""
```
lostella:
```suggestion
    """Flattened access to a collection of datasets."""
```

src/gluonts/itertools.py
```diff
@@ -174,6 +175,21 @@ def __iter__(self):
         yield from itertools.islice(self.iterable, self.length)
 
 
+class Map:
+    def __init__(self, fn, iterable: SizedIterable):
```
lostella, line 179: This could be even more generic, accept also potentially a `Collection`, and expose `__getitem__`. But that complicates it, so maybe we do that if/when needed

lostella (dismissed):

src/gluonts/itertools.py
jaheba: But `__getitem__` would only work if `iterable` is `Collection`. I don't think that's expressible in Python's type system.

lostella: I was thinking of just not annotating it, but again this is not important as of now

src/gluonts/dataset/util.py
```diff
@@ -10,57 +10,17 @@
...
-
-
-def _list_files(directory: Path) -> Iterator[Path]:
-    for dirname, _, filenames in os.walk(directory):
-        for filename in filenames:
-            yield Path(dirname, filename)
-
-
-def true_predicate(*args) -> bool:
-    return True
```
lostella, line 49: This one was correct!

lostella (approved):

PR #2025, opened 2022-06-13
Make HP tuning tutorial more accurate

lostella:
*Description of changes:* Training model for a little longer in the tutorial, reducing the number of trials. I think this makes the notebook a little slower to run, but at the advantage that the displayed results are more meaningful (prediction looks better in the plotted example).

### DIFF
--- a/docs/tutorials/advanced_topics/hp_tuning_with_optuna.md
+++ b/docs/tutorials/advanced_topics/hp_tuning_with_optuna.md
@@ -34,7 +34,6 @@ print(f"Available datasets: {list(dataset_recipes.keys())}")
 dataset = get_dataset("m4_hourly")
 ```
 
-
 ### Extract and split training and test data sets
 
 In general, the datasets provided by GluonTS are objects that consists of three things:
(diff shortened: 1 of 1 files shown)


docs/tutorials/advanced_topics/hp_tuning_with_optuna.md
```diff
@@ -45,31 +44,22 @@ In general, the datasets provided by GluonTS are objects that consists of three
...
-test_subset = list(islice(dataset.test, 15))
-```
-
-
 This is what the data looks like (first training series, first two weeks of data)
 
+
 ```python
-to_pandas(train_subset[0])[:14 * 24].plot()
+to_pandas(next(iter(dataset.train)))[:14 * 24].plot()
```
jaheba, line 57: There is `first` in `toolz` but it just does the same: https://toolz.readthedocs.io/en/latest/_modules/toolz/itertoolz.html#first

lostella: I think it’s better to avoid the dependency since it’s just the same

jaheba: The dependency is already there!

lostella: is it? https://github.com/awslabs/gluon-ts/blob/dev/requirements/requirements-docs.txt

jaheba (approved):
Improve is such a wishy washy word.

docs/tutorials/advanced_topics/hp_tuning_with_optuna.md
jaheba: https://github.com/awslabs/gluon-ts/blob/483bc5a8aba6a4a92adc0b8b356fe1f6bccb6e00/requirements/requirements.txt#L7

PR #2029, opened 2022-06-13
Add support for nan values in Rotbaum

zoolhasson:
*Issue #, if available:*

*Description of changes:*

I added support for nan values. Since in the featurization process Rotbaum removes the mean, having nan values make all of the coordinates into nans. I have done the following:
1. Change all "mean"s and "std"s in the featurization to np.nanmean and np.nanstd.
2. Allow for a feature that counts number of nans in the context windows. You can turn it on by setting count_nans=True.
3. Allow for not removing the mean value in the featurization process. You can do this by setting subtract_mean=False.

### DIFF
--- a/src/gluonts/model/rotbaum/_predictor.py
+++ b/src/gluonts/model/rotbaum/_predictor.py
@@ -117,6 +117,8 @@ class TreePredictor(RepresentablePredictor):
         max_workers: Optional[int] = None,
         method: str = "QRX",
         quantiles=None,  # Used only for "QuantileRegression" method.
+        subtract_mean: bool = True,
+        count_nans: bool = False,
         model=None,
         seed=None,
     ) -> None:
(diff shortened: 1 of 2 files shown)


lostella:
> Since in the featurization process Rotbaum removes the mean, having nan values make all of the coordinates into nans

@zoolhasson does this qualify as a bugfix as well then?

zoolhasson:
> > Since in the featurization process Rotbaum removes the mean, having nan values make all of the coordinates into nans
> 
> @zoolhasson does this qualify as a bugfix as well then?

Does it?

lostella:
@zoolhasson the `mean` -> `nanmean` change in a way may be considered that; in any case it's fine to mark it as "new feature" since new options for the models are there. The change looks good to me!

lostella (approved):

PR #2030, opened 2022-06-14
add schema for dataset abstractions

Hongqing-work:
*Issue #, if available:*

*Description of changes:*

Add schema with fieldtypes: NumpyArrayField, PandasPeriodField.

### DIFF
--- a/src/gluonts/dataset/schema.py
+++ b/src/gluonts/dataset/schema.py
@@ -12,7 +12,7 @@
 # permissions and limitations under the License.
 
 from dataclasses import dataclass
-from typing import Any, Dict, ClassVar, Optional, Union, Generic, TypeVar, Type
+from typing import Any, Dict, Optional, Union, Generic, TypeVar, Type
 
 import numpy as np
 import pandas as pd
--- a/src/gluonts/dataset/schema.py
+++ b/src/gluonts/dataset/schema.py
@@ -37,7 +37,6 @@ class FieldType(Generic[T]):
 
 @dataclass
 class NumpyArrayField(FieldType[np.ndarray]):
-    default: ClassVar[list] = [0.0]
     dtype: Type = np.float32
     ndim: Optional[int] = None
 
@@ -112,6 +111,15 @@ class PandasPeriodField(FieldType[pd.Period]):
 
 
 @dataclass
+class FieldWithDefault:
+    field_type: Any
+    default_value: Any
+
+    def __call__(self, value: Any) -> Any:
+        return self.field_type(value)
+
+
+@dataclass
 class Schema:
     fields: Dict[str, Any]
 
@@ -133,8 +141,8 @@ class Schema:
             try:
                 value = d[field_name]
             except KeyError:
-                if "default" in dir(field_type):
-                    value = field_type.default
+                if isinstance(field_type, FieldWithDefault):
+                    value = field_type.default_value
                 else:
                     raise GluonTSDataError(
                         f"field {field_name} does not occur in the data"
--- a/src/gluonts/dataset/schema.py
+++ b/src/gluonts/dataset/schema.py
@@ -11,13 +11,14 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Any, Dict, Optional, Union, Generic, TypeVar, Type
 
 import numpy as np
 import pandas as pd
 import pyarrow as pa
 from pandas.tseries.frequencies import to_offset
+from torch import layout
 
 from gluonts.exceptions import GluonTSDataError
 
@@ -124,7 +125,7 @@ class PandasPeriodField(FieldType[pd.Period]):
 class Schema:
     fields: Dict[str, FieldType]
     default_values: Dict[str, Any]
-    targert_layout: Dict[str, str] = {}
+    targert_layout: Dict[str, str] = field(default_factory=dict)
 
     # convert functions
     def ct_to_tc(xs):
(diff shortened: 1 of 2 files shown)


src/gluonts/dataset/schema.py
```diff
@@ -0,0 +1,267 @@
...
+        return self.ndim is None or x.ndim == self.ndim
+
+
+Freq = Union[str, pd.DateOffset]
+
+
+class PandasPeriodField(FieldType[pd.Period]):
+    def __init__(
+        self,
+        freq: Freq,
```
jaheba: If you remove the trailing `,` this should be condensed into one line.

src/gluonts/dataset/schema.py
```diff
@@ -0,0 +1,267 @@
...
+    ) -> None:
+        self.freq = to_offset(freq) if isinstance(freq, str) else freq
```
jaheba: I think you can always call `to_offset` since it just returns the value if it is already an offset.

src/gluonts/dataset/schema.py
```diff
@@ -0,0 +1,267 @@
...
+import pandas as pd
+import pyarrow as pa
+from pandas.tseries.frequencies import to_offset
+from torch import layout
+
+from gluonts.exceptions import GluonTSDataError
+
+
+@dataclass
+class PyArray:
```
jaheba: How is this different from `NumpyArrayField`?

src/gluonts/dataset/schema.py
```diff
@@ -0,0 +1,267 @@
...
+        except Exception:
+            return False
+        return True
+
+
+@dataclass
+class Schema:
+    fields: Dict[str, FieldType]
+    default_values: Dict[str, Any]
+    targert_layout: Dict[str, str] = field(default_factory=dict)
```
jaheba: I think we shouldn't try to make this do too many things. The `Schema` should just represent what the model requires and doesn't have more logic in it.

src/gluonts/dataset/schema.py
```diff
@@ -0,0 +1,267 @@
...
+        return xs.transpose()
+
+    def tc_to_ct(xs):
+        return xs.transpose()
+
+    def t_to_tc(xs):
+        return xs.reshape((-1, 1))
+
+    def t_to_ct(xs):
+        return xs.reshape((1, -1))
```
jaheba: I think it's better to remove these from this PR and then add support if needed later.

src/gluonts/dataset/schema.py
```diff
@@ -0,0 +1,267 @@
...
+            self(v)
+        except Exception:
+            return False
+        return True
+
+
+@dataclass
+class Schema:
+    fields: Dict[str, FieldType]
+    default_values: Dict[str, Any]
```
jaheba: I was thinking that maybe it's better not to have `default_values` but enable it through the field-types. So there could be a special `PrivateField`, which means that the field is internal and not taken from user data.

`pydantic` for instance does it similarly: https://pydantic-docs.helpmanual.io/usage/models/#private-model-attributes

src/gluonts/dataset/schema.py
```diff
@@ -0,0 +1,267 @@
...
+
+            except Exception as e:
+                raise GluonTSDataError(
+                    f"Error when processing field {field_name} using "
+                    "{field_type}"
+                ) from e
+        return out
+
+    @staticmethod
+    def infer(entry: Dict[str, Any]) -> "Schema":
```
jaheba: Let's remove this for this PR and add it in a later step.

jaheba (commented):
Thanks! I've added some comments.

Let's focus on just the `Schema` class build things step by step. We should try to keep it as simple as possible and ideally have some tests.

test/dataset/test_schema.py
```diff
@@ -0,0 +1,185 @@
...
+    )
+
+    assert output_data["start"] == pd.Period("2022-12-16", freq="D")
+
+    expected_target = np.asarray([1, 2], dtype=int)
+    assert (output_data["target"] == expected_target).all()
+    assert output_data["target"].dtype == expected_target.dtype
+
+    expected_cat = np.asarray([6], dtype=int)
+    assert (output_data["feat_static_cat"] == expected_cat).all()
```
jaheba: I think for these there is:

```py
from numpy.testing import assert_equal
```

test/dataset/test_schema.py
```diff
@@ -0,0 +1,185 @@
...
+
+    expected_cat = np.asarray([0.0], dtype=float)
+    assert (output_data["feat_static_cat"] == expected_cat).all()
+    assert output_data["feat_static_cat"].dtype == expected_cat.dtype
+
+
+def test_call_schema_not_using_default_value():
+    schema = Schema(
+        {
+            "start": PandasPeriodField(freq="D"),
```
jaheba, line 109: I wonder if we want to define the `freq` in the schema.

@lostella Thoughts?

src/gluonts/dataset/schema.py
```diff
@@ -0,0 +1,160 @@
...
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+from dataclasses import dataclass
+from typing import Any, Dict, Optional, Union, Generic, TypeVar, Type
+
+import numpy as np
+import pandas as pd
+import pyarrow as pa
```
jaheba: `pyarrow` is an optional dependency, so we should not rely that it is installed.

src/gluonts/dataset/schema.py
```diff
@@ -0,0 +1,160 @@
...
+        raise NotImplementedError()
+
+
+@dataclass
+class NumpyArrayField(FieldType[np.ndarray]):
+    dtype: Type = np.float32
+    ndim: Optional[int] = None
+
+    def __call__(self, value: Any) -> np.ndarray:
+        if isinstance(value, pa.Array):
```
jaheba: I wonder, when would that be the case. `gluonts.dataset.arrow` is designed to unpack the Arrow types into Python/numpy values, so ideally we would not need to be concerned about this.

src/gluonts/dataset/schema.py
```diff
@@ -0,0 +1,160 @@
...
+            value = value.to_numpy()
+            value = value.astype(self.dtype)
+        else:
+            value = np.asarray(value, dtype=self.dtype)
+        if self.ndim is not None and self.ndim != value.ndim:
```
jaheba, line 56: Let's add an empty line here.
```suggestion

        if self.ndim is not None and self.ndim != value.ndim:
```

PR #2031, opened 2022-06-14
Breaking: Update data splitters to return `(input, output)` pairs in the test split

npnv:
*Issue #, if available:* Addresses part of #2041

*Description of changes:*


jaheba:
What is the advantage of using Timestamp over Period?

src/gluonts/dataset/split/splitter.py
```diff
@@ -28,23 +28,23 @@
 
     splitter = DateSplitter(
         prediction_length=24,
-        split_date=pd.Period('2018-01-31', freq='D')
+        split_date=pd.Timestamp('2018-01-31', freq='D')
```
lostella: I think these can stay as `pd.Period`

src/gluonts/dataset/split/splitter.py
```diff
@@ -77,8 +77,8 @@ def from_data_entry(
         if freq is None:
             freq = item["start"].freq
 
-        index = pd.period_range(
-            start=item["start"], freq=freq, periods=len(item["target"])
+        index = pd.date_range(
+            start=item["start"].to_timestamp(), freq=freq, periods=len(item["target"])
```
lostella: Also this one needs not be changed. You probably started this branch from before #1980 was merged?

test/dataset/split/test_split.py
```diff
@@ -44,11 +44,13 @@ def test_splitter():
     prediction_length = dataset.metadata.prediction_length
     splitter = DateSplitter(
         prediction_length=prediction_length,
-        split_date=pd.Period("1750-01-05 04:00:00", freq="h"),
+        split_date=pd.Timestamp("1750-01-05 04:00:00"),
```
lostella: `Period` should stay here

lostella (changes requested):

src/gluonts/dataset/split/splitter.py
```diff
@@ -199,116 +187,151 @@ def _train_slice(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
...
                 test = self._trim_history(
                     self._test_slice(item, offset=offset)
                 )
-                prediction_length = getattr(self, "prediction_length")
 
-                _check_split_length(train.end, test.end, prediction_length)
-                split._add_test_slice(test)
+                assert (
+                    train.end + train.end.freq * prediction_length <= test[1].end
+                )
```
lostella: `_check_split_length` is actually an improved version of this, so you can keep that

src/gluonts/dataset/split/splitter.py
```diff
@@ -199,116 +187,151 @@ def _train_slice(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
...
         Length of the prediction interval in test data.
     split_offset
         Offset determining where the training data ends.
-        A positive offset indicates how many observations since the start of
-        each series should be in the training slice; a negative offset
-        indicates how many observations before the end of each series should
-        be excluded from the training slice. Please make sure that the number
-        of excluded values is enough for the test case, i.e., at least
-        ``prediction_length`` (for ``rolling_split`` multiple of
-        ``prediction_length``) values are left off.
```
lostella: Also this change is unrelated, right?

src/gluonts/dataset/split/splitter.py
```diff
@@ -199,116 +187,151 @@ def _train_slice(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
...
+                item[self.split_offset : offset_])
 
 
-class DateSplitter(AbstractBaseSplitter, pydantic.BaseModel):
+@dataclass
+class DateSplitter(AbstractBaseSplitter):
     """
     A splitter that slices training and test data based on a
-    ``pandas.Period``.
+    ``pandas.Timestamp``.
```
lostella: `Period` can stay

src/gluonts/dataset/split/splitter.py
```diff
@@ -318,19 +341,16 @@ class DateSplitter(AbstractBaseSplitter, pydantic.BaseModel):
     prediction_length
         Length of the prediction interval in test data.
     split_date
-        Period determining where the training data ends. Please make sure
-        at least ``prediction_length`` (for ``rolling_split`` multiple of
-        ``prediction_length``) values are left over after the ``split_date``.
```
lostella: No need to change this one too

lostella (changes requested):

src/gluonts/dataset/split/splitter.py
```diff
@@ -107,7 +109,7 @@ def from_data_entry(
     def to_data_entry(self) -> DataEntry:
         ret = {
             FieldName.START: self.start,
-            FieldName.ITEM_ID: self.item,
+            FieldName.ITEM_ID: self.item_id,
```
rsnirwan: Thanks for the changes! just a minor comment: no need to add `item_id` if its None. You can just remove that line and add after the assignment:
```
if self.item: ret[FieldName.ITEM_ID] = self.item
```

test/dataset/split/test_split.py
```diff
@@ -44,11 +44,13 @@ def test_splitter():
...
     )
     train, validation = splitter.split(dataset.train)
-    assert len(train[1][0][FieldName.TARGET]) + prediction_length == len(
-        validation[1][0][FieldName.TARGET]
+    train = iter(train)
+    validation = iter(validation)
+    assert len(next(train)[FieldName.TARGET]) == len(
+        next(validation)[0][FieldName.TARGET]
```
lostella: We can check all of them, right?

```python
for train_entry, valid_pair in zip(train, validation):
    assert len(train_entry[FieldName.TARGET]) == len(valid_pair[0][FieldName.TARGET])
    assert len(valid_pair[1]) == prediction_length
    # TODO also check that "label" index starts where we expect it to start
```

We can even check that training and validation data are the same amount:

```python
assert len(list(train)) == len(list(validation))
```

test/dataset/split/test_split.py
```diff
@@ -58,39 +60,63 @@ def test_splitter():
         max_history=max_history,
     )
     train, validation = splitter.split(dataset.train)
-    assert len(validation[1][0][FieldName.TARGET]) == max_history
-    assert len(train[1][0][FieldName.TARGET]) == 4 * prediction_length
+    train = iter(train)
+    validation = iter(validation)
+    a = next(validation)
+    assert len(a[0][FieldName.TARGET]) + len(a[1]) == max_history
+    assert len(next(train)[FieldName.TARGET]) == 4 * prediction_length
```
lostella: Similar comments as above

test/dataset/split/test_split.py
```diff
@@ -58,39 +60,63 @@ def test_splitter():
...
+    a = next(validation)
     for i in range(3):
-        assert len(validation[1][i][FieldName.TARGET]) == max_history
+        assert len(a[0][FieldName.TARGET]) + len(a[1]) == max_history
 
 
 def test_split_mult_freq():
     splitter = DateSplitter(
-        prediction_length=1, split_date=pd.Period("2021-01-01", "2h")
+        prediction_length=1, split_date=pd.Timestamp("2021-01-01")
```
lostella: No need to change this

test/dataset/split/test_split.py
```diff
@@ -58,39 +60,63 @@ def test_splitter():
...
     )
 
     splitter.split(
         [
             {
                 "item_id": "1",
                 "target": pd.Series([0, 1, 2]),
-                "start": pd.Period("2021-01-01", freq="2H"),
+                "start": pd.Timestamp("2021-01-01", freq="2H"),
```
lostella: Also this one

lostella (changes requested):

src/gluonts/dataset/split/splitter.py
```diff
@@ -158,27 +163,15 @@ def __getitem__(self, slice_: slice) -> "TimeSeriesSlice":
...
-    def _add_train_slice(self, train_slice: TimeSeriesSlice) -> None:
-        # is there any data left for training?
-        if train_slice:
-            self.train.append(train_slice.to_data_entry())
-
-    def _add_test_slice(self, test_slice: TimeSeriesSlice) -> None:
-        self.test.append(test_slice.to_data_entry())
-
-
+@dataclass
```
lostella: This can probably be removed (causing typecheck errors)

src/gluonts/dataset/split/splitter.py
```diff
@@ -199,60 +192,73 @@ def _train_slice(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
...
-
-    def rolling_split(
+    def _generate_test_slices(
         self,
         items: List[DataEntry],
-        windows: int,
-        distance: Optional[int] = None,
-    ) -> TrainTestSplit:
+        windows: Optional[int],
+        distance: Optional[int],
```
lostella: Here `windows` and `distance` can just be `int`

src/gluonts/dataset/split/splitter.py
```diff
@@ -289,23 +326,26 @@ class OffsetSplitter(pydantic.BaseModel, AbstractBaseSplitter):
         `max_history`. This can be used to produce smaller file-sizes.
     """
 
-    prediction_length: int
     split_offset: int
+    prediction_length: int
+    windows: Optional[int] = 1
+    distance: Optional[int] = None
```
lostella: Also no `Optional` here, just default `windows = 1` and `distance = prediction_length` somehow

src/gluonts/dataset/split/splitter.py
```diff
@@ -326,11 +366,10 @@ class DateSplitter(AbstractBaseSplitter, pydantic.BaseModel):
...
     """
 
-    class Config:
-        arbitrary_types_allowed = True
-
-    prediction_length: int
     split_date: pd.Period
+    prediction_length: int
+    windows: Optional[int] = 1
+    distance: Optional[int] = None
```
lostella: Same as above

lostella (changes requested):

src/gluonts/dataset/split/splitter.py
npnv: the `distance` can't be directly assign with `prediction_length` because when initializing splitter objective, the `prediction_length` is not defined. So I think `distance` remains `Optional[int]` is OK.

src/gluonts/dataset/split/splitter.py
```diff
@@ -266,7 +268,38 @@ def _check_split_length(
...
+            yield pairs[0]
+
+    @property
+    def labels(self):
+        for pairs in self:
+            yield pairs[1]
+
+
+@dataclass
+class TrainingDataset(Iterable[DataEntry]):
```
lostella:
```suggestion
class TrainingDataset(Dataset):
```

lostella (changes requested):

test/dataset/split/test_split.py
```diff
@@ -11,6 +11,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
+from posixpath import split
```
lostella:
```suggestion
```

This is probably not the `split` you wanted to import; it was auto-added by VSCode I believe

src/gluonts/dataset/split/splitter.py
```diff
@@ -36,13 +36,16 @@
 
     splitter = DateSplitter(
         prediction_length=24,
-        split_date=pd.Period('2018-01-31', freq='D')
+        split_date=pd.Period('2018-01-31', freq='D'), 
+        windows=7
     )
-    train, test = splitter.rolling_split(whole_dataset, windows=7)
+    train, test = splitter.split(whole_dataset)
```
lostella, line 42: Just a reminder to also update docstrings (this one, but also those of classes) once you are done with the changes

src/gluonts/dataset/split/splitter.py
```diff
@@ -36,13 +36,16 @@
...
 """
 
 from abc import ABC, abstractmethod
-from typing import List, Optional, cast
+from posixpath import split
```
lostella:
```suggestion
```

src/gluonts/dataset/split/splitter.py
```diff
@@ -326,21 +343,30 @@ class DateSplitter(AbstractBaseSplitter, pydantic.BaseModel):
...
+            item[: self.split_date],
+            item[
+                self.split_date
+                + item.start.freq : self.split_date
+                + (prediction_length + offset) * item.start.freq
+            ],
+        )
+
+
+def split(dataset, *, split_offset=None, split_date=None):
```
lostella: I would just call these arguments `offset` and `date`

src/gluonts/dataset/split/splitter.py
```diff
@@ -199,60 +192,64 @@ def _train_slice(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
...
-            distance = getattr(self, "prediction_length")
-        assert distance is not None
-
-        split = TrainTestSplit()
+    def _generate_test_slices(
+        self, items: List[DataEntry], prediction_length: int, *args, **kwargs
+    ):
+        max_history = args[0] if len(args) else None
+        windows = kwargs.get("windows", 1)
+        distance = kwargs.get("distance", prediction_length)
```
lostella:
```suggestion
    def _generate_test_slices(
        self, items: Dataset, prediction_length: int, windows=1, distance=None, max_history=None,
    ):
        if distance is None:
            distance=prediction_length
```

src/gluonts/dataset/split/splitter.py
```diff
@@ -266,7 +263,28 @@ def _check_split_length(
...
-class OffsetSplitter(pydantic.BaseModel, AbstractBaseSplitter):
+@dataclass
+class TestTemplate:
+    dataset: Iterable[DataEntry]
+    splitter: AbstractBaseSplitter
+
+    def generate_instances(self, prediction_length, *args, **kwargs):
+        yield from self.splitter._generate_test_slices(
+            self.dataset, prediction_length, *args, **kwargs
+        )
```
lostella:
```suggestion
    def generate_instances(self, prediction_length: int, windows=1, distance=None, max_history=None):
        yield from self.splitter._generate_test_slices(
            self.dataset,
            prediction_length=prediction_length,
            windows=windows,
            distance=distance,
            max_history=max_history,
        )
```

lostella: You can also add a docstring for this method, see eg https://github.com/awslabs/gluon-ts/blob/dev/src/gluonts/transform/feature.py#L217-L232

src/gluonts/dataset/split/splitter.py
```diff
@@ -266,7 +263,28 @@ def _check_split_length(
...
+
+
+@dataclass
+class TrainingDataset(Iterable[DataEntry]):
```
lostella:
```suggestion
class TrainingDataset(Dataset):
```

lostella (changes requested):

src/gluonts/dataset/split/splitter.py
```diff
@@ -198,61 +183,69 @@ def _train_slice(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
...
+        self, item: TimeSeriesSlice, label_len: int, max_history
+    ) -> TimeSeriesSlice:
+        if max_history is not None:
+            return item[-max_history + label_len :]
         else:
             return item
 
-    def split(self, items: List[DataEntry]) -> TrainTestSplit:
-        split = TrainTestSplit()
+    def split(self, items: List[DataEntry]):
```
lostella: I think we can rename this and fix its type

```suggestion
    def split(self, dataset: Dataset):
```

src/gluonts/dataset/split/splitter.py
```diff
@@ -266,15 +259,73 @@ def _check_split_length(
...
 
-class OffsetSplitter(pydantic.BaseModel, AbstractBaseSplitter):
+@dataclass
+class TestTemplate:
+    """
+    A class used for generating test data.
+
+    Parameters
+    ----------
+    splitter:
```
lostella, line 262: You also need the description for `dataset` here

src/gluonts/dataset/split/splitter.py
```diff
@@ -266,15 +259,73 @@ def _check_split_length(
...
+
+        Parameters
+        ----------
+        prediction_length
+            Length of the prediction interval in test data.
+        windows
+            Windows is used especially when doing ``rolling_split``, otherwise,
+            it should be ``1`` for a simple ``split``.
+        distance
+            Length of each window. If set to None, the prediction_length is
```
lostella: This is rather the difference between the start of each test window generated, for each of the original dataset entries

src/gluonts/dataset/split/splitter.py
```diff
@@ -266,15 +259,73 @@ def _check_split_length(
...
+        """
+        Generate an iterator of test dataset, which includes input part and
+        label part.
+
+        Parameters
+        ----------
+        prediction_length
+            Length of the prediction interval in test data.
+        windows
+            Windows is used especially when doing ``rolling_split``, otherwise,
```
lostella: “Indicates how many test windows to generate for each original dataset entry”

lostella (changes requested):

test/dataset/split/test_split.py
```diff
@@ -39,50 +39,150 @@ def test_ts_slice_to_item():
...
+    if max_history is None:
+        assert len(train_entry[FieldName.TARGET]) == len(
+            valid_pair[0][FieldName.TARGET]
+        )
+    elif not is_rolling_split:
+        assert (
+            len(valid_pair[0][FieldName.TARGET]) + len(valid_pair[1])
+            == max_history
+        )
+    assert len(valid_pair[1]) == prediction_length * window
```
lostella: Shouldn’t it just be 

```suggestion
    assert len(valid_pair[1]) == prediction_length
```

test/dataset/split/test_split.py
```diff
@@ -39,50 +39,150 @@ def test_ts_slice_to_item():
...
+    train_end = (
+        train_entry[FieldName.START]
+        + len(train_entry[FieldName.TARGET])
+        * train_entry[FieldName.START].freq
+    )
+    if cutoff_date:
+        assert train_end == cutoff_date + train_entry[FieldName.START].freq
+    if split_offset:
+        assert len(train_entry[FieldName.TARGET]) == split_offset
```
lostella: I think this will only work if `split_offset >= 0` but we also allow it to be negative (and slice the data “from the right”)

src/gluonts/dataset/split/splitter.py
```diff
@@ -315,32 +362,36 @@ class DateSplitter(AbstractBaseSplitter, pydantic.BaseModel):
...
-        self, item: TimeSeriesSlice, offset: int = 0
-    ) -> TimeSeriesSlice:
-        return item[
-            : self.split_date
-            + (self.prediction_length + offset) * item.start.freq
-        ]
+        self, item: TimeSeriesSlice, prediction_length: int, offset: int = 0
+    ) -> Tuple[TimeSeriesSlice, TimeSeriesSlice]:
+        return (
+            item[: self.split_date],
```
lostella: I think this should end at
```suggestion
            item[: self.split_date + offset * item.start.freq],
```

src/gluonts/dataset/split/splitter.py
```diff
@@ -315,32 +362,36 @@ class DateSplitter(AbstractBaseSplitter, pydantic.BaseModel):
...
+            item[
+                self.split_date
+                + item.start.freq : self.split_date
```
lostella: And this should be
```suggestion
                + (offset + 1) * item.start.freq : self.split_date
```

src/gluonts/dataset/split/splitter.py
```diff
@@ -284,28 +335,24 @@ class OffsetSplitter(pydantic.BaseModel, AbstractBaseSplitter):
...
     def _train_slice(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
         return item[: self.split_offset]
 
     def _test_slice(
-        self, item: TimeSeriesSlice, offset: int = 0
-    ) -> TimeSeriesSlice:
-        offset_ = self.split_offset + offset + self.prediction_length
+        self, item: TimeSeriesSlice, prediction_length: int, offset: int = 0
+    ) -> Tuple[TimeSeriesSlice, TimeSeriesSlice]:
+        offset_ = self.split_offset + offset + prediction_length
```
lostella: If you set this to
```suggestion
        offset_ = self.split_offset + offset
```

src/gluonts/dataset/split/splitter.py
```diff
@@ -284,28 +335,24 @@ class OffsetSplitter(pydantic.BaseModel, AbstractBaseSplitter):
...
         if self.split_offset < 0 and offset_ >= 0:
             offset_ += len(item)
-        return item[:offset_]
+        return (item[: self.split_offset], item[self.split_offset : offset_])
```
lostella: Then this should probably be
```suggestion
        return (item[: offset_], item[offset_: offset_ + prediction_length])
```

lostella (changes requested):
@npnv I noticed something strange in the test assertions (see comment in the test), for which I think some of the slicing logic needs to be adjusted (see suggestions in the comments)

src/gluonts/dataset/split/splitter.py
```diff
@@ -36,18 +36,21 @@
...
+        windows=7
     )
-    train, test = splitter.rolling_split(whole_dataset, windows=7)
+    train, test = splitter.split(whole_dataset)
 """
 
 from abc import ABC, abstractmethod
-from typing import List, Optional, cast
+from typing import List, Optional, cast, Iterable, Tuple
+from dataclasses import dataclass
```
jaheba: Minor: Imports should be alphabetically ordered.

src/gluonts/dataset/split/splitter.py
```diff
@@ -36,18 +36,21 @@
...
 
 import numpy as np
 import pandas as pd
 import pydantic
 
+from gluonts.dataset import Dataset
```
jaheba: I think we also directly expose `DataEntry` as well.

test/dataset/split/test_split.py
```diff
@@ -11,6 +11,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
+from tkinter.messagebox import NO
```
jaheba:
```suggestion
```

test/dataset/split/test_split.py
```diff
@@ -39,50 +39,145 @@ def test_ts_slice_to_item():
...
 
 
+def check_training_validation(
+    train_entry,
+    valid_pair,
+    prediction_length: int,
+    cutoff_date=None,
+    max_history=None,
+    split_offset=None,
+    is_rolling_split=False,
```
lostella: I think this is not needed

test/dataset/split/test_split.py
```diff
@@ -39,50 +39,145 @@ def test_ts_slice_to_item():
...
+    assert train_entry[FieldName.ITEM_ID] == valid_pair[0][FieldName.ITEM_ID]
+    if max_history is None:
+        assert len(train_entry[FieldName.TARGET]) == len(
+            valid_pair[0][FieldName.TARGET]
+        )
+    elif not is_rolling_split:
+        assert (
+            len(valid_pair[0][FieldName.TARGET]) + len(valid_pair[1])
+            == max_history
+        )
```
lostella: I think this check should be always done (i.e., `elif not is_rolling_split` can be removed)

test/dataset/split/test_split.py
```diff
@@ -39,50 +39,145 @@ def test_ts_slice_to_item():
...
+    assert train_end <= valid_pair[1].index[0]
+    valid_end = (
+        valid_pair[0][FieldName.START]
+        + len(valid_pair[0][FieldName.TARGET])
+        * valid_pair[0][FieldName.START].freq
+    )
+    assert valid_end == valid_pair[1].index[0]
+
+
 def test_splitter():
```
lostella: This test function got pretty long: you can improve it by encoding a general logic in it, that applies to all possible cases, and have `date, offset, windows, distance, max_history` as arguments to the function. You can then provide all test cases for the above parameters with `pytest`, see for example how [this test](https://github.com/awslabs/gluon-ts/blob/9dd1450bfbf15626ca93ebcb2a73fe74cfe75604/test/test_itertools.py#L60-L80) does it.

I think the testing logic should essentially be:
```python
def test_split_m4_hourly(date, offset, windows, distance, max_history)
    prediction_length = dataset.metadata.prediction_length
    
    train, test_template = split(dataset.train, date=date, offset=offset)

    train = list(train)
    validation = list(test_template.generate_instances(
        windows=windows,
        distance=distance,
        max_history=max_history,
    ))

    assert len(train) == len(dataset)
    assert len(validation) == windows * len(train)

    k = 0
    for train_entry in train:
        for _ in range(windows):
            check_training_validation(...)
            k += 1
```

This way the testing logic is pretty clear and written only once (although some details are hidden inside `check_training_validation`, which is OK).

Also note that I'm using directly `split` in the test body, instead of `DateSplitter` and `OffsetSplitter`, since that makes things simpler.

lostella (changes requested):

src/gluonts/dataset/split/splitter.py
```diff
@@ -198,61 +182,69 @@ def _train_slice(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
...
-    ) -> TimeSeriesSlice:
+        self, item: TimeSeriesSlice, prediction_length: int, offset: int = 0
+    ) -> Tuple[TimeSeriesSlice, TimeSeriesSlice]:
         pass
 
-    def _trim_history(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
-        if getattr(self, "max_history") is not None:
-            return item[-getattr(self, "max_history") :]
+    def _trim_history(
+        self, item: TimeSeriesSlice, label_len: int, max_history
```
lostella:
```suggestion
        self, item: TimeSeriesSlice, label_len: int, max_history: Optional[int]
```

src/gluonts/dataset/split/splitter.py
```diff
@@ -198,61 +182,69 @@ def _train_slice(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
...
-        items: List[DataEntry],
-        windows: int,
-        distance: Optional[int] = None,
-    ) -> TrainTestSplit:
-        # distance defaults to prediction_length
+        items: Dataset,
+        prediction_length: int,
+        windows,
+        distance,
+        max_history,
```
lostella:
```suggestion
        windows: int = 1,
        distance: Optional[int] = None,
        max_history: Optional[int] = None,
```

src/gluonts/dataset/split/splitter.py
```diff
@@ -198,61 +182,69 @@ def _train_slice(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
...
+                df_label = df_label.set_index(
+                    pd.period_range(
+                        start=label["start"],
+                        periods=len(label["target"]),
+                        freq=train.end.freq,
+                    )
+                )
 
-        return split
+                yield input, df_label
```
lostella: I think it's better to `yield label` directly, so that the output here is a pair of `DataEntry` objects. Conversion to `pandas` can happen later wherever needed

src/gluonts/dataset/split/splitter.py
```diff
@@ -198,61 +182,69 @@ def _train_slice(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
...
-        windows: int,
-        distance: Optional[int] = None,
-    ) -> TrainTestSplit:
-        # distance defaults to prediction_length
+        items: Dataset,
+        prediction_length: int,
+        windows,
+        distance,
+        max_history,
+    ):
```
lostella:
```suggestion
    ) -> Generator[Tuple[DataEntry, DataEntry], None, None]:
```

src/gluonts/dataset/split/splitter.py
```diff
@@ -266,15 +258,75 @@ def _check_split_length(
...
+
+    dataset: Dataset
+    splitter: AbstractBaseSplitter
+
+    def generate_instances(
+        self,
+        prediction_length: int,
+        windows=1,
+        distance=None,
+        max_history=None,
```
lostella, line 275:
```suggestion
        windows=1,
        distance=None,
        max_history=None,
```

src/gluonts/dataset/split/splitter.py
```diff
@@ -266,15 +258,75 @@ def _check_split_length(
...
+    ):
```
lostella:
```suggestion
    ) -> Generator[Tuple[DataEntry, DataEntry], None, None]:
```

lostella (changes requested):

test/dataset/split/test_split.py
```diff
@@ -105,21 +103,72 @@ def test_negative_offset_splitter():
...
+
+@pytest.mark.parametrize(
+    "date, offset, windows, distance, max_history",
+    [
+        (pd.Period("1750-01-05 04:00:00", freq="h"), None, 1, None, None),
+        (pd.Period("1750-01-05 04:00:00", freq="h"), None, 1, None, 96),
+        (pd.Period("1750-01-05 04:00:00", freq="h"), None, 3, None, 96),
+        (None, 192, 1, None, 96),
+        (None, 192, 3, None, 96),
+    ],
```
lostella: We should add test cases with:
- negative `offset` values
- positive `windows` but `max_history=None`

src/gluonts/dataset/split/splitter.py
```diff
@@ -198,61 +182,61 @@ def _train_slice(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
...
         pass
 
-    def _trim_history(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
-        if getattr(self, "max_history") is not None:
-            return item[-getattr(self, "max_history") :]
+    def _trim_history(
+        self, item: TimeSeriesSlice, label_len: int, max_history: Optional[int]
+    ) -> TimeSeriesSlice:
+        if max_history is not None:
+            return item[-max_history + label_len :]
```
lostella:
```suggestion
    def _trim_history(
        self, item: TimeSeriesSlice, max_history: Optional[int]
    ) -> TimeSeriesSlice:
        if max_history is not None:
            return item[-max_history:]
```

src/gluonts/dataset/split/splitter.py
```diff
@@ -198,61 +182,61 @@ def _train_slice(self, item: TimeSeriesSlice) -> TimeSeriesSlice:
...
+
+                _check_split_length(
+                    train.end, test[1].end, train.end.freq * prediction_length
                 )
-                prediction_length = getattr(self, "prediction_length")
 
-                _check_split_length(train.end, test.end, prediction_length)
-                split._add_test_slice(test)
+                input = self._trim_history(
+                    test[0], len(test[1]), max_history
```
lostella:
```suggestion
                    test[0], max_history
```

test/dataset/split/test_split.py
```diff
@@ -39,50 +40,47 @@ def test_ts_slice_to_item():
...
+    max_history,
+    date,
+    offset,
+) -> None:
+    assert train_entry[FieldName.ITEM_ID] == valid_pair[0][FieldName.ITEM_ID]
+    if max_history is None:
+        assert len(train_entry[FieldName.TARGET]) == len(
+            valid_pair[0][FieldName.TARGET]
+        )
+    else:
```
lostella: I think the first case needs not to be checked, so we can simply do

```suggestion
    if max_history is not None:
```

test/dataset/split/test_split.py
```diff
@@ -39,50 +40,47 @@ def test_ts_slice_to_item():
...
+        assert (
+            len(valid_pair[0][FieldName.TARGET])
+            + len(valid_pair[1][FieldName.TARGET])
+            == max_history
+        )
```
lostella:
```suggestion
        assert (
            len(valid_pair[0][FieldName.TARGET])
            == max_history
        )
```

lostella (changes requested):

lostella:
Looks good to me @npnv! Thanks! 🚀

lostella (dismissed):

lostella:
There seemed to be an issue with using `m4_hourly` in the test, I replaced it with a small artificial dataset

lostella (approved):

PR #2035, opened 2022-06-14
Simplify lag values computation in torch DeepAR

lostella:
*Description of changes:* The code related to lags is a bit involved, when it can be a simple function to be called right before invoking the RNN. No wonder that, even after introducing an additional module with all the overhead, this results in saving lines.


jaheba (dismissed):

src/gluonts/torch/model/deepar/util.py
```diff
@@ -0,0 +1,78 @@
...
+    sequence: torch.Tensor,
+    features: Optional[torch.Tensor] = None,
+) -> Tuple[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]:
+    """
+    Constructs an array of lagged values from a given sequence.
+
+    Parameters
+    ----------
+    indices
+        Indices of the lagged observations that the RNN takes as input. For
```
lostella: Docstring needs to be fixed, there’s no RNN at this point

src/gluonts/torch/model/deepar/util.py
```diff
@@ -0,0 +1,78 @@
...
+        A tensor of shape ``(N, T, L)``: if ``I = len(indices)``,
+        ``sequence.shape = (N, T, C)``, and ``features.shape = (N, T, F)``,
+        then ``L = C * I + F``.
+    """
+    full_sequence = torch.cat((prior_sequence, sequence), dim=1)
+    full_sequence_length = full_sequence.shape[1]
+
+    output_sequence_length = sequence.shape[1]
+
+    assert max(indices) + output_sequence_length <= full_sequence_length, (
```
lostella: Also this assertion could be simplified I think

test/torch/model/test_deepar_modules.py
```diff
@@ -17,62 +17,44 @@
...
             torch.ones((4, 8, 3)),
-            torch.ones((4, 5, 2)),
-            torch.ones((4, 5, 3)),
+            (4, 8, 13),
         ),
     ],
 )
-def test_lagged_lstm(
-    model: LaggedLSTM,
+def test_prepare_lagged_inputs(
```
lostella: Test should be improved; only assert shapes, we could assert output values

kashif:
the lag seq helper is useful I believe for other recurrent models, do you think it makes sense to move it somewhere else? In particular, the [S4 model](https://github.com/kashif/pytorch-transformer-ts/blob/main/s4/s4.ipynb) I have follows a similar structure to this...  but yes it's fine here too...

kashif:
actually @lostella every model i implement I have to copy the `get_lagged_subsequences` helper... could we move this helper to the ~~features~~ torch/utils.py?

lostella:
@kashif sorry just saw your previous comment. Yes it's probably useful more in general, let me move it

kashif:
this is a nice helper which is somewhat unique to the time series situation... essentially I think of it as trading off between sequence length and feature size... I do not think any other modality as far as I can tell allows such a tradeoff... and hence its useful for every model I believe

lostella:
Yes, and having it as a simple function is much clearer I think. Plus it seems that it's saving us ~140~ ~80 lines which is always good

kashif:
yes for some reason it always feels nicer when you delete code than when you add it haha... BTW I suppose you will re-add the test in its own file?

kashif:
thank you! this is super helpful I believe!

jaheba (approved):

PR #2039, opened 2022-06-15
Fix README wording

lostella:
*Issue #, if available:*

*Description of changes:*


jaheba:
I was thinking, should we maybe recommend to install `gluonts[pro]` and then say that you can also install it without extras?

lostella:
I think I'd be very upfront with useful installation options and even suggest immediately `mxnet` or `torch` (that's what most people likely look for)

jaheba (approved):

PR #2042, opened 2022-06-16
Add auto-inference of `timestamp` and `freq` in DataFramesDatasets

### LINKED ISSUE #2026: `DataFramesDataset` should take `index` for timestamp field if available.
Currently, we require that the `timestamp` field is a column in the passed datasets, but don't support it if it is the `index` of the `DataFrame`.

Ideally, it would try to use the `index` for it by default, raising an error if neither `timestamp` was passed nor the index can be used for it. 

- [ ] #2042

rsnirwan:
closes #2026 

### DIFF
--- a/src/gluonts/dataset/dataframe.py
+++ b/src/gluonts/dataset/dataframe.py
@@ -35,32 +35,32 @@ class DataFramesDataset(Dataset):
 
     Parameters
     ----------
-    dataframes:
+    dataframes
         List or Dict of ``DataFrame``s or a single ``DataFrame`` containing
         at least ``timestamp`` and ``target`` columns. If a Dict is provided,
         the key will be the associated ``item_id``.
-    target:
+    target
         Name of the column that contains the ``target`` time series.
         For multivariate targets, a list of column names should be provided.
-    timestamp:
+    timestamp
         Name of the column that contains the timestamp information.
-    freq:
+    freq
         Frequency of observations in the time series. Must be a valid pandas
         frequency.
-    item_id:
+    item_id
         Name of the column that contains the item_id information.
-    feat_dynamic_real:
+    feat_dynamic_real
         List of column names that contain dynamic real features.
-    feat_dynamic_cat:
+    feat_dynamic_cat
         List of column names that contain dynamic categorical features.
-    feat_static_real:
+    feat_static_real
         List of column names that contain static real features.
-    feat_static_cat:
+    feat_static_cat
         List of column names that contain static categorical features.
-    past_feat_dynamic_real:
+    past_feat_dynamic_real
         List of column names that contain dynamic real features only for the
         history.
-    ignore_last_n_targets:
+    ignore_last_n_targets
         For target and past dynamic features last ``ignore_last_n_targets``
         elements are removed when iterating over the data set. This becomes
         important when the predictor is called.
@@ -158,13 +158,13 @@ class DataFramesDataset(Dataset):
 
         Parameters
         ----------
-        dataframe:
+        dataframe
             pandas.DataFrame containing at least ``timestamp``, ``target`` and
             ``item_id`` columns.
-        item_id:
+        item_id
             Name of the column that, when grouped by, gives the different time
             series.
-        **kwargs:
+        **kwargs
             Additional arguments. Same as of DataFramesDataset class.
 
         Returns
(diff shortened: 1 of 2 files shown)


src/gluonts/dataset/dataframe.py
```diff
@@ -68,9 +69,10 @@ class DataFramesDataset(Dataset):
     dataframes: Union[
         List[pd.DataFrame], Dict[str, pd.DataFrame], pd.DataFrame
     ]
-    target: Union[str, List[str]]
-    timestamp: str
-    freq: str
+    target: Union[str, List[str]] = "target"
+    timestamp: str = ""
+    freq: str = ""
+    item_id: str = ""
```
lostella: Should these be `None` instead by default?

src/gluonts/dataset/dataframe.py
```diff
@@ -82,17 +84,38 @@ def __post_init__(self) -> None:
...
+
+        if not self.timestamp:  # infer timestamp from index
+            for i, (_, df) in enumerate(self._dataframes):
+                assert isinstance(
+                    df.index, DatetimeIndexOpsMixin
+                ), "Please set ``timestamp`` or provide a DatetimeIndex."
+                df = df.reset_index()
+                df.columns = ["_timestamp"] + df.columns.tolist()[1:]
+                self._dataframes[i] = (self._dataframes[i][0], df)
+            self.timestamp = "_timestamp"
```
lostella: I'm thinking: wouldn't it be better to do the opposite? If `timestamp` is set, then we `.set_index(timestamp)`, and then use the index from here on; if `timestamp` is not set, then we just use the index. I feel like that way, some of the logic here would be simpler (but I didn't think it through)

src/gluonts/dataset/dataframe.py
rsnirwan: I had them initially as `None`s. But then the type checker complains about the None type and I have to manually cast (typing.cast) all occurrences of `timestamp`, `freq` and `item_id` in all methods to `str`. Do you want to go for `None`? For me both are fine.

src/gluonts/dataset/dataframe.py
rsnirwan: Yes, it got a bit complicated here. I'll try your suggestion tomorrow and report afterwards.

src/gluonts/dataset/dataframe.py
rsnirwan: `Union[str, None]` would be the input to the initializer. In the initializer they are set to `str`. So, in other methods of the class we can assume their types to be `str`. Now,  the input to a method is of type `Optional[str]` (inferred by the checker) but the method only accepts `str`. This is why the checker complains.
Here, e.g., `item_id` is of type `str` but in init it would be `Union[str, None]`:
```python
def _dataentry(self, item_id: str, df: pd.DataFrame) -> DataEntry:
```
This seems to be in general a problem of the `Union` type :shrug:.

jaheba: I think the issue is that `dataclass` mixes the type of the passed argument with the object's attribute.

I guess we can either use an alternative constructor or not use a `dataclass`.

lostella: @RSNirwan maybe there are some other issues, but in the `_dataentry` example you mentioned, I guess `_dataentry` should just accept `item_id: Optional[str]` and it would be fine

src/gluonts/dataset/dataframe.py
```diff
@@ -170,24 +191,25 @@ def as_dataentry(
...
+    data:
         pandas.DataFrame containing at least ``timestamp``, ``target`` and
         ``item_id`` columns.
-    target: str or List[str]
+    target:
         Name of the column that contains the ``target`` time series.
         For multivariate targets ``target`` is expecting a list of column
         names.
-    timestamp: str
+    timestamp:
```
jaheba: Did you remove the type-annotations intentionally?

And wouldn't you then also need to remove the `:` as well?

jaheba (commented):
I think there is also the option to override `__new__` to do the logic that you have in `__post_init__`. So instead of post-processing the data, we would pre-process it.

src/gluonts/dataset/dataframe.py
rsnirwan: Thanks for spotting. Yes, I removed them intentionally. They already have an annotation with the type-hints. I removed them from the docstring to avoid mismatches between two different annotations and to follow the rest of gluonts code basis.

lostella (approved):
Thanks @RSNirwan! 🚀

PR #2044, opened 2022-06-16
Update README.md

geoalgo:
On a mac, `pip install gluonts[mxnet, pro]` yields the error: `ERROR: Invalid requirement: 'gluonts[mxnet,'.
Without the space, there is no error.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ GluonTS is a Python package for probabilistic time series modeling, focusing on 
 GluonTS requires Python 3.6 or newer, and the easiest way to install it is via `pip`:
 
 ```bash
-pip install gluonts[mxnet, pro]  # support for mxnet models, faster datasets
-pip install gluonts[torch, pro]  # support for torch models, faster datasets
+pip install gluonts[mxnet,pro]  # support for mxnet models, faster datasets
+pip install gluonts[torch,pro]  # support for torch models, faster datasets
 ```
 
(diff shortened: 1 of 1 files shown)


jaheba:
I think the problem might be shell -- at least it is for me. `[mxnet, pro]` is valid, HOWEVER the empty space throws the shell off. Also zsh interprets `[` and `]` specially.

I think we should change this to:

```sh
pip install "gluonts[mxnet, pro]"
```

lostella:
Does this really work? I thought the latest release (0.9.5) did not have the `pro` extras defined (cc @jaheba)

jaheba:
```
WARNING: gluonts 0.9.5 does not provide the extra 'mxnet'
WARNING: gluonts 0.9.5 does not provide the extra 'pro'
```

It is just ignored :)

jaheba:
Which gets me thinking: Should we add a disclaimer to the readme that it might not represent the release but only the dev-branch?

lostella:
> Which gets me thinking: Should we add a disclaimer to the readme that it might not represent the release but only the dev-branch?

I think we should have instructions on installing the stable versions; since those depend on what is on PyPI, they should work with the latest published release on PyPI. I think this is what the majority of the README readers look for.

We could have another line explaining how to install the package for development.

jaheba (approved):

jaheba:
Let's merge this now and then rethink what the job of the README is.

Issue #2041, opened 2022-06-16
Improving train/test splitters and test datasets


lostella:
This is related to #1447, but a more concrete proposal.
Work in progress on part of this is in #2031

## Context

In GluonTS, data and tools provided for evaluating models accuracy on test data work a bit backwards (in a bad sense). Datasets are provided in terms of “train” and “test” portions, where test data is essentially (simplifying a bit here) a “continuation” of training data. So we have things like:

```
training entry = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
test entry     = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13]
```

There is an implicit assumption on how test data will be used: the final portion of it (the “label” if you want) will be taken out, the rest will be given as input to a predictor producing a forecast, and the forecast can then be compared back to the label for evaluation.

This implicit assumption is so pervasive that `make_evaluation_prediction`, the main entry point for getting evaluation out of models, has it baked in (it splits test entries into input and label internally).

### Representing test data better

*Edit: refer to the discussion below for updated details on this*

A better way to represent test data is to have `(input, label)` pairs explicitly exposed, like in any supervised learning setting:

```python
@dataclass
class TestDataset:
    input_label_pairs: Iterable[Tuple[DataEntry, pd.Series]]  # can use something else other than a pd.Series

    def __iter__(self):
        return input_output_pairs

    @property
    def inputs(self):
        for pair in self.input_output_pairs:
            yield pair[0]

    @property
    def labels(self):
        for pair in self.input_output_pairs:
            yield pair[1]
```

### Splitters to construct test data

*Edit: refer to the discussion below for updated details on this*

Once one has a `Dataset`, intended as an `Iterable[DataEntry]`, they can come up with various kinds of test scenarios using [splitters](https://github.com/awslabs/gluon-ts/blob/dev/src/gluonts/dataset/split/splitter.py):

```
training_data, test_data = OffsetSplitter(
    prediction_length=24
    split_offset=-24,
).split(data)
```

Here `training_data: Dataset` and `test_data: TestDataset` as defined above. Training data will satisfy the property that it does not overlap with any of the labels in `test_data`. One can define more complex test cases for rolling evaluation as follows:

```
training_data, test_data = OffsetSplitter(
    prediction_length=24
    split_offset=-24,
    windows=7,
    distance=24, # 7 non-overlapping test windows of length 24, out of each original series
).split(data)
```

[...]

PR #2045, opened 2022-06-17
Add `Evaluator.evaluate`.

jaheba:
This should make `make_evaluation_predictions` obsolete in most cases.

```py
from gluonts.model.npts import NPTSPredictor

npts = NPTSPredictor(prediction_length=12, freq="H")
ev = Evaluator()

metrics = ev.evaluate(npts, get_dataset("electricity").test)
```

### DIFF
--- a/src/gluonts/evaluation/_base.py
+++ b/src/gluonts/evaluation/_base.py
@@ -179,6 +179,20 @@ class Evaluator:
         self.aggregation_strategy = aggregation_strategy
         self.ignore_invalid_values = ignore_invalid_values
 
+    def evaluate(
+        self, predictor, test_dataset, **kwargs
+    ) -> Tuple[Dict[str, float], pd.DataFrame]:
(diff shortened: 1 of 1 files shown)


jaheba:
Would `.backtest` a better name?

jaheba:
Something I'm wondering: Is there any reason to keep `make_evaluation_predictions`? As I see it, it is used to call the `Evaluator`, but I think it's not something people enjoy doing.

There is also `backtest_metrics`, which does the same as this one, but adds some logging.

In other words, `evaluation/backtest.py` could be removed entirely.

lostella:
> Is there any reason to keep make_evaluation_predictions?

Eventually no, I think. Right now it's a shortcut to get backetesting forecasts so you can plot them for example, instead of directly evaluating their accuracy.

But in light of what is being discussed in #2041, this should be as easy as doing

```python
forecasts = model.predict(dataset.test.inputs)
```

PR #2050, opened 2022-06-17
Add concepts section to docs.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -23,17 +23,17 @@ forecasts = predictor.predict(data)
 The reason to split `Estimator` and `Predictor` into two classes is that many
 models require a dedicated training step to generate a global model. This
 global model is only trained once, but is used to make predictions for all
-time-serires.
+time-series.
 
 This is in contrast to local models, which are fitted on individual time-series
 and therefore try to capture the characteristics of each time-series but not
 the dataset in its entirety.
 
-Training a global model usually takes a lot of time: often hours, but sometimes
-even days. Thus, it is not feasable to train the model "offline" (on demand).
-In contrast, a local model is usually fitted "oline" as part of the request.
+Training a global model can take a lot of time: up to hours, but sometimes
+even days. Thus, it is not feasible to train the model "offline" (on demand).
+In contrast, a local model is usually fitted "online" as part of the prediction.
 
-In GluonTS models that are local are available as predictors, whilst global
+In GluonTS, local models are directly available as predictors, whilst global
 models are offered as estimators:
 
 ```py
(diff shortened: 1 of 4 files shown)


docs/concepts.md
```diff
@@ -0,0 +1,79 @@
...
+
+```py
+predictor = estimator.train(train_data)
+forecasts = predictor.predict(data)
+```
+
+The reason to split `Estimator` and `Predictor` into two classes is that many
+models require a dedicated training step to generate a global model. This
+global model is only trained once, but is used to make predictions for all
+time-serires.
```
lostella:
```suggestion
time-series.
```

docs/concepts.md
```diff
@@ -0,0 +1,79 @@
...
+
+This is in contrast to local models, which are fitted on individual time-series
+and therefore try to capture the characteristics of each time-series but not
+the dataset in its entirety.
+
+Training a global model usually takes a lot of time: often hours, but sometimes
+even days. Thus, it is not feasable to train the model "offline" (on demand).
+In contrast, a local model is usually fitted "oline" as part of the request.
```
lostella:
```suggestion
Training a global model can take a lot of time: up to hours, but sometimes
even days. Thus, it is not feasible to train the model "offline" (on demand).
In contrast, a local model is usually fitted "online" as part of the prediction.
```

However, I don't understand what this sentence means:

```
Thus, it is not feasible to train the model "offline" (on demand)
```

docs/concepts.md
```diff
@@ -0,0 +1,79 @@
...
+
+In GluonTS models that are local are available as predictors, whilst global
```
lostella:
```suggestion
In GluonTS, local models are directly available as predictors, whilst global
```

docs/concepts.md
jaheba: The second part of the sentence is missing. It's not feasible to do it for each prediction but needs to do be done separately.

lostella: Do you want to add that?

jaheba: Yes

docs/concepts.md
```diff
@@ -51,7 +53,8 @@ In GluonTS a `Dataset` is a collection of time-series objects. Each of these
 objects has columns (or fields) which represent attributes of the
 time-series.
 
-In most models the `target`-field is the column that we want to predict:
+Most odels use the `target`-column to indicate the time-series that we want to
```
lostella:
```suggestion
Most models use the `target`-column to indicate the time-series that we want to
```

lostella (changes requested):
Just a typo, then it can go! 🚀

lostella (approved):

PR #2053, opened 2022-06-21
Add `dataset_writer` to `get_dataset`.

Hongqing-work:
*Issue #, if available:*

*Description of changes:*
Add arrow support for data repositories.


jaheba:
Thanks for the PR!

To keep things more flexible we could pass some `dataset_writer` to `get_dataset` which is able to write the given dataset to disk.

That way we give users more control over what dataset format should be generated, which can also be a custom one. It further should make dependency management a bit easier.

We would need to think about some interface that defines the dataset-writer API.

jaheba:
I think what we might want is a `Writer` class for each of our dataset classes:

```py
get_dataset("electricity, writer=JsonLinesWriter())
get_dataset("electricity, writer=ArrowWriter(stream=False))
get_dataset("electricity, writer=ParquetWriter())
```

This would require us to define a `Writer` interface that these writers would implement.

jaheba:
#2061 adds the writers described above.

Looks like it's going to be some work to port all datasets to it.

src/gluonts/dataset/artificial/generate_synthetic.py
```diff
@@ -117,17 +114,7 @@ def generate_sf2s_and_csv(
...
-        os.makedirs(os.path.dirname(file_path))
-    freq = artificial_dataset.metadata.freq
-    train_set = artificial_dataset.train
-    generate_sf2(file_path + "train.json", train_set, is_missing, num_missing)
-    test_set = artificial_dataset.test
-    generate_sf2(file_path + "test.json", test_set, is_missing, num_missing)
-    with open(file_path + "input_to_forecast.csv", "w") as csv_file:
-        # Test set has training set with the additional values to predict
-        write_csv_row(test_set, freq, csv_file, is_missing, num_missing)
+    pass
```
jaheba: Should we just remove `generate_sf2s_and_csv`? It doesn't to be used anywhere.

src/gluonts/dataset/repository/_gp_copula_2019.py
```diff
@@ -148,24 +150,23 @@ def save_metadata(
         )
 
 
-def save_dataset(dataset_path: Path, ds_info: GPCopulaDataset):
+def save_dataset(
+    dataset_path: Path, ds_info: GPCopulaDataset, ds_writer: DatasetWriter
```
jaheba: I think we should be consistent. Let's use either `writer`, `ds_writer` or `dataset_writer` -- but only one of them.

src/gluonts/dataset/repository/_gp_copula_2019.py
```diff
@@ -148,24 +150,23 @@ def save_metadata(
...
+):
     dataset = list(FileDataset(dataset_path, freq=ds_info.freq))
-    shutil.rmtree(dataset_path)
```
jaheba, line 153: Why remove this?

src/gluonts/dataset/repository/_lstnet.py
```diff
@@ -135,6 +137,10 @@ def generate_lstnet_dataset(
     )
 
     os.makedirs(dataset_path, exist_ok=True)
+    train_path = dataset_path / "train"
+    test_path = dataset_path / "test"
+    train_path.mkdir(exist_ok=True)
+    test_path.mkdir(exist_ok=True)
```
jaheba: Where these missing?

src/gluonts/dataset/repository/_m3.py
```diff
@@ -192,10 +196,12 @@ def normalize_category(c: str):
             )
         )
 
-    train_file = dataset_path / "train" / "data.json"
-    test_file = dataset_path / "test" / "data.json"
+    train_path = dataset_path / "train"
+    test_path = dataset_path / "test"
+    train_path.mkdir(exist_ok=True)
+    test_path.mkdir(exist_ok=True)
```
jaheba: Since we do this over and over again, maybe we can create a utility for this.

Maybe something like this:

```py
def create_dataset_paths(base_path: Path, channel_names: List[str]):
    base_path.mkdir(exists_ok)

    channels = {}

    for channel_name in channel_names:
        channel = base_path / channel_name
        channel.mkdir(exists_ok=True)

        channels[channel_name] = channel

    return channels



paths = create_dataset_paths(path, ["train", "test"])
dataset_writer.write_to_folder(train_data, paths["train"])
dataset_writer.write_to_folder(test_data, paths["test"])
```

src/gluonts/dataset/repository/_m4.py
```diff
@@ -73,28 +75,29 @@ def generate_m4_dataset(
...
+        to_dict(
+            target_values=target,
+            start=mock_start_dataset,
+            cat=[cat],
+            item_id=cat,
+        )
+        for cat, target in enumerate(train_target_values)
+    ]
+    test_data = [
+        to_dict(
```
jaheba: I think `to_dict` might not be needed since we do some conversion in the `DatasetWriter`. It might actually break the arrow dataset, since it assumes that you pass numpy arrays.

src/gluonts/dataset/repository/_m3.py
jaheba: Ah, it looks like we already have this in `TrainDatasets`.

src/gluonts/dataset/artificial/generate_synthetic.py
```diff
@@ -1,241 +0,0 @@
...
-                    ts[FieldName.FEAT_DYNAMIC_REAL]
-                ):
-                    ts[FieldName.FEAT_DYNAMIC_REAL][i] = feat_dynamic_real[
-                        : len(ts[FieldName.TARGET])
-                    ]
-            json.dump(ts, json_file)
-            json_file.write("\n")
-
-
-def generate_sf2s_and_csv(
```
jaheba, line 113: @lostella @jgasthaus Is this something we still want to use? I don't think we have much usage for csv files.

jaheba:
Looks really good!

I wonder, is there a way we can easily test that all datasets still work?

lostella:
@jaheba we could have a format-agnostic checksum of some kind, and compute it before/after the change for each dataset

jaheba:
> @jaheba we could have a format-agnostic checksum of some kind, and compute it before/after the change for each dataset

Not sure how you would create a good checksum, but I guess you could just calculate the stats and they should be pretty robust. One issue is that floats might be stored slightly differently giving non-exact results.

src/gluonts/dataset/common.py
```diff
@@ -110,7 +110,7 @@ def save(
         if overwrite:
             shutil.rmtree(path, ignore_errors=True)
 
-        metadata = path / "metadata"
```
jaheba, line 113: I'm not sure we want to change this. I think it's fine to have a dedicated `metadata` folder -- this then would also work on SageMaker.

jaheba (dismissed):

jaheba (dismissed):

src/gluonts/dataset/common.py
jaheba: Actually, we never used this :|

jaheba (approved):

PR #2055, opened 2022-06-21
Remove `holidays` and `matplotlib` from core dependencies.

jaheba:
*Issue #, if available:*

*Description of changes:*


requirements/requirements-articial-dataset.py
```diff
@@ -0,0 +1 @@
+holidays >= 0.9
```
lostella, line 1: Why `>=`? Would be good to have an upper bound

jaheba: I've just copied it from the requirements. I don't think this should be an issue.

lostella (approved):

PR #2057, opened 2022-06-21
Add optional parameter `time_axis` to `forecast_start`.

melopeo:
We add the optional parameter `time_axis` to `forecast_start`. This is particularly useful for hierarchical time series, where we have multivariate data. The default value of the optional parameter `time_axis` is set to `-1` to make it consistent with the previous case, which assumes that the target is univariate.

This PR is related to the following ongoing PR: 
https://github.com/awslabs/gluon-ts/pull/1685#discussion_r902732323_

### DIFF
--- a/src/gluonts/dataset/util.py
+++ b/src/gluonts/dataset/util.py
@@ -12,13 +12,16 @@
 # permissions and limitations under the License.
 from typing import Optional
 
+import numpy as np
 import pandas as pd
 
 from .field_names import FieldName
 
 
-def forecast_start(entry):
-    return entry[FieldName.START] + len(entry[FieldName.TARGET])
+def forecast_start(entry, time_axis: int = -1):
+    return (
+        entry[FieldName.START] + np.shape(entry[FieldName.TARGET])[time_axis]
+    )
 
 
 def to_pandas(instance: dict, freq: Optional[str] = None) -> pd.Series:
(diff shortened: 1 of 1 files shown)


src/gluonts/dataset/util.py
```diff
@@ -12,13 +12,16 @@
...
 import pandas as pd
 
 from .field_names import FieldName
 
 
-def forecast_start(entry):
-    return entry[FieldName.START] + len(entry[FieldName.TARGET])
+def forecast_start(entry, time_axis: int = -1):
+    return (
+        entry[FieldName.START] + np.shape(entry[FieldName.TARGET])[time_axis]
```
lostella:
```suggestion
        entry[FieldName.START] + entry[FieldName.TARGET].shape[time_axis]
```

lostella:
What layout does multivariate/hierarchical data have? Is `-1` or `0` the best default for `time_axis`?

melopeo:
@lostella I think the `time_axis` default value was discussed here: https://github.com/awslabs/gluon-ts/pull/1685#discussion_r716010101, where it seems that it was agreed that `time_axis = -1` was the way to go.

src/gluonts/dataset/util.py
melopeo: Suggestion addressed in a new commit.

lostella (approved):

jaheba (approved):
Makes me think that we should revisit the idea of having named axis.

PR #2062, opened 2022-06-22
Fix type annotations for `predict_to_numpy`

lostella:
*Issue #, if available:*

*Description of changes:*


jaheba (approved):
Fix meaning weakening?

lostella:
> Fix meaning weakening?

Not really, they were wrong since `mx` or `torch` arrays is not what we put there. We really put there lists, or anything that can be "splatted" in a function call's arguments.

PR #2065, opened 2022-06-22
Rename `DataFramesDataset` to `PandasDataset`

### LINKED ISSUE #2056: Rename `DataFramesDataset` to `PandasDataset`
After some discussion with @lostella we want to 

- [x] rename `DataFramesDataset` to `PandasDataset`
- [x] support `pandas.Series` objects as a dataset with only a `target` column
- [x] treat a `DataFrame` with just one column like a `Series` object

As part of this we should also

- [x] fix naming used in the documentation notebook from #2051

rsnirwan:
resolves #2056


jaheba:
Would it make sense to rename `dataframe.py` to something else? `pandas.py` for example.

Also can you check that you change all references? I've updated `README.md` to use the dataset and it looks it's missing from your PR.

rsnirwan:
ah yes .. I recursively checked in `src`, `test` and `doc` and that way missed the top level files. thanks!

test/dataset/test_pandas.py
```diff
@@ -15,7 +15,7 @@
 import numpy as np
 import pytest
 
-from gluonts.dataset import dataframe
+from gluonts.dataset import pandas
```
jaheba, line 18: Now this looks a bit weird! Maybe we should rename this to `pandas_dataset` , or just `pd` -- oh wait!

Let's leave it like that.

jaheba (approved):

docs/index.rst
```diff
@@ -41,7 +41,7 @@ See :doc:`community/devsetup` for setup instructions in case you want to develop
 
    tutorials/forecasting/quick_start_tutorial
    tutorials/forecasting/extended_tutorial
-   tutorials/data_manipulation/dataframesdataset
+   tutorials/data_manipulation/pandasdataframes
```
lostella, line 44: This did not need to change (since the tutorial file name did not change)

lostella: Wondering why the docs build did not fail?

jaheba: Because the file was actually renamed? ;)

lostella: Ok, didn't see that

PR #2071, opened 2022-06-23
Only publish docs for dev and releases.

jaheba:
*Issue #, if available:*

*Description of changes:*


.github/workflows/docs.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         make docs
 
     - name: Configure AWS Credentials
-      if: ${{ github.event_name == 'push' }}
+      if: ${{ github.event_name == 'push' && ( github.ref_name == 'dev' || startsWith(github.ref_name, 'v0.') }}
```
lostella, line 38: This will hit us back when we release `v1.0.0`

lostella (approved):

.github/workflows/docs.yml
jaheba: Who says we ever will?!

PR #2074, opened 2022-06-24
Add support for second frequency.

### LINKED ISSUE #367: Use DeepAR for second-level data
When using `DeepAREstimator` for second-level data, both `get_lags_for_frequency` and `time_features_from_frequency_str` do not support `freq = 'S'`. It is possible to support second-level data in these two functions? Or is it possible to disable `lags_seq` and `time_features` in `DeepAREstimator`?

kashif:
*Description of changes:*

Added support for Second level freq.

fixes #367 

### DIFF
--- a/src/gluonts/time_feature/lag.py
+++ b/src/gluonts/time_feature/lag.py
@@ -53,6 +53,12 @@ def get_lags_for_frequency(
 
     # Lags are target values at the same `season` (+/- delta) but in the
     # previous cycle.
+    def _make_lags_for_second(multiple, num_cycles=3):
+        # We use previous ``num_cycles`` hours to generate lags
+        return [
+            _make_lags(k * 60 // multiple, 2) for k in range(1, num_cycles + 1)
+        ]
+
     def _make_lags_for_minute(multiple, num_cycles=3):
         # We use previous ``num_cycles`` hours to generate lags
         return [
@@ -121,6 +127,15 @@ def get_lags_for_frequency(
             + _make_lags_for_hour(offset.n / 60.0)
             + _make_lags_for_day(offset.n / (60.0 * 24))
             + _make_lags_for_week(offset.n / (60.0 * 24 * 7))
+        )
+    # second
+    elif offset_name == "S":
+        lags = (
+            _make_lags_for_second(offset.n)
+            + _make_lags_for_minute(offset.n / 60.0)
+            + _make_lags_for_hour(offset.n / (60.0 * 60.0))
+            + _make_lags_for_day(offset.n / (60.0 * 60.0 * 24))
+            + _make_lags_for_week(offset.n / (60.0 * 60.0 * 24 * 7))
         )
     else:
         raise Exception("invalid frequency")
--- a/src/gluonts/time_feature/lag.py
+++ b/src/gluonts/time_feature/lag.py
@@ -134,7 +134,6 @@ def get_lags_for_frequency(
             _make_lags_for_second(offset.n)
             + _make_lags_for_minute(offset.n / 60)
             + _make_lags_for_hour(offset.n / (60 * 60))
-            + _make_lags_for_day(offset.n / (60 * 60 * 24))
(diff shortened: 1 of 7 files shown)


src/gluonts/time_feature/seasonality.py
```diff
@@ -14,12 +14,14 @@
...
 
 import pandas as pd
+
 from gluonts.time_feature import norm_freq_str
 
 logger = logging.getLogger(__name__)
 
 
 DEFAULT_SEASONALITIES = {
+    "S": 86400,
```
jaheba: This looks very high. I wonder, why do we always want to refer to the same value a day ago?

src/gluonts/time_feature/lag.py
```diff
@@ -122,6 +128,15 @@ def _make_lags_for_month(multiple, num_cycles=3):
...
             + _make_lags_for_week(offset.n / (60.0 * 24 * 7))
         )
+    # second
+    elif offset_name == "S":
+        lags = (
+            _make_lags_for_second(offset.n)
+            + _make_lags_for_minute(offset.n / 60.0)
+            + _make_lags_for_hour(offset.n / (60.0 * 60.0))
+            + _make_lags_for_day(offset.n / (60.0 * 60.0 * 24))
+            + _make_lags_for_week(offset.n / (60.0 * 60.0 * 24 * 7))
```
jaheba: I think this is excessive. We already saw issue for `minute` frequency.

Going back 4 weeks blows up the memory requirements.

jaheba (commented):
Thanks!

I think we shouldn't go back so much for the lagged value.

src/gluonts/time_feature/lag.py
```diff
@@ -122,6 +128,15 @@ def _make_lags_for_month(multiple, num_cycles=3):
             + _make_lags_for_day(offset.n / (60.0 * 24))
             + _make_lags_for_week(offset.n / (60.0 * 24 * 7))
         )
+    # second
+    elif offset_name == "S":
+        lags = (
+            _make_lags_for_second(offset.n)
+            + _make_lags_for_minute(offset.n / 60.0)
```
jaheba: Why are we using floats here at all? 🤔

kashif: I believe `_make_lags_for_x` methods expect floats...

src/gluonts/time_feature/lag.py
kashif: yes you are right! removed!

src/gluonts/time_feature/seasonality.py
kashif: made it 60*60

src/gluonts/time_feature/lag.py
jaheba: But `/` will already return a float, so the `0` is unneeded. I assume it's because this once was written for Py2.

src/gluonts/time_feature/lag.py
```diff
@@ -111,16 +117,24 @@ def _make_lags_for_month(multiple, num_cycles=3):
...
+            + _make_lags_for_day(offset.n / (60 * 24))
+            + _make_lags_for_week(offset.n / (60 * 24 * 7))
+        )
+    # second
+    elif offset_name == "S":
+        lags = (
+            _make_lags_for_second(offset.n)
+            + _make_lags_for_minute(offset.n / 60)
+            + _make_lags_for_hour(offset.n / (60 * 60))
+            + _make_lags_for_day(offset.n / (60 * 60 * 24))
```
jaheba: I think we can even remove `_make_lags_for_day` since it goes back 4 weeks in time, which is a lot, about 2.5 million time-steps.

`_make_lags_for_hour` max lagged value is going back `7` days, which should be plenty.

jaheba (approved):

kashif:
thank you dear @jaheba I appreciate it!

PR #2081, opened 2022-06-27
Fix: m3 dataset checking.

Hongqing-work:
*Issue #, if available:*

*Description of changes:*
We use `pd.Period` now, so the `start` field checking is changed as this result.


src/gluonts/dataset/repository/_m3.py
```diff
@@ -55,20 +54,19 @@ def check_dataset(dataset_path: Path, length: int, sheet_name):
 
         assert ts_train["start"] == ts_test["start"]
         start = ts_train["start"]
-        regex = r"^(\d{4})-(\d{2})-(\d{2})( 00:00(:00)?)?$"
-        m = re.match(regex, str(start))
-        assert m
-        month, day = m.group(2), m.group(3)
+        assert isinstance(start, pd.Period)
```
jaheba: Do we need to check this? I think we can just assume that this is the case, and otherwise the below code will fail.

src/gluonts/dataset/repository/_m3.py
```diff
@@ -55,20 +54,19 @@ def check_dataset(dataset_path: Path, length: int, sheet_name):
...
+        month, day = start.month, start.day
```
jaheba: I think we don't need to rebind these values.

src/gluonts/dataset/repository/_m3.py
```diff
@@ -55,20 +54,19 @@ def check_dataset(dataset_path: Path, length: int, sheet_name):
...
         if sheet_name in ["M3Quart", "Other"]:
-            assert f"{month}-{day}" in [
-                "03-31",
-                "06-30",
-                "09-30",
-                "12-31",
-            ], f"Invalid time stamp `{month}-{day}`"
+            assert (month, day) in [
```
jaheba:
```suggestion
            assert (start.month, start.day) in [
```

src/gluonts/dataset/repository/_m3.py
```diff
@@ -55,20 +54,19 @@ def check_dataset(dataset_path: Path, length: int, sheet_name):
...
+                (3, 31),
+                (6, 30),
+                (9, 30),
+                (12, 31),
+            ], f"Invalid time stamp {month}-{day}"
         elif sheet_name == "M3Year":
-            assert (
-                f"{month}-{day}" == "12-31"
+            assert (month, day) == (
```
jaheba:
```suggestion
            assert (start.month, start.day) == (
```

src/gluonts/dataset/repository/_m3.py
jaheba:
```suggestion
```

src/gluonts/dataset/repository/_m3.py
jaheba:
```suggestion
```

jaheba (commented):
Thanks! Did you try running these?

src/gluonts/dataset/repository/_m3.py
Hongqing-work: OK.

Hongqing-work:
> 

Yes, this part was called when generating m3 dataset. It can already pass the assert now.

jaheba (approved):
Thanks, LGTM.

PR #2092, opened 2022-06-27
Fix: DistributionLoss not encodable.

### LINKED ISSUE #2090: Pytorch models are not serializable
## Description
Printing or serializing pytorch deepar estimator results into `RuntimeError`.
The problem is that `gluonts.torch.modules.loss.DistributionLoss` is not serializable.

## To Reproduce
```python
from gluonts.torch.model.deepar.estimator import DeepAREstimator

model = DeepAREstimator(
    freq=dataset.metadata.freq,
    prediction_length=dataset.metadata.prediction_length,
    context_length=dataset.metadata.prediction_length,
)
print(model)
```

```
  File "p.py", line 11, in <module>
    print(model)
  File "/Users/nirwar/Dev/gitrepos/gluon-ts/src/gluonts/core/component.py", line 310, in validated_repr
    return dump_code(self)
  File "/Users/nirwar/Dev/gitrepos/gluon-ts/src/gluonts/core/serde/_repr.py", line 115, in dump_code
    return as_repr(encode(o))
  File "/Users/nirwar/.runtimes/Python38/lib/python3.8/functools.py", line 875, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/Users/nirwar/Dev/gitrepos/gluon-ts/src/gluonts/core/serde/_base.py", line 220, in encode
    "kwargs": encode(kwargs),
  File "/Users/nirwar/.runtimes/Python38/lib/python3.8/functools.py", line 875, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/Users/nirwar/Dev/gitrepos/gluon-ts/src/gluonts/core/serde/_base.py", line 207, in encode
    return valmap(encode, v)
  File "/Users/nirwar/.runtimes/Python38/lib/python3.8/site-packages/toolz/dicttoolz.py", line 83, in valmap
    rv.update(zip(d.keys(), map(func, d.values())))
  File "/Users/nirwar/.runtimes/Python38/lib/python3.8/functools.py", line 875, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/Users/nirwar/Dev/gitrepos/gluon-ts/src/gluonts/core/serde/_base.py", line 243, in encode
    raise RuntimeError(bad_type_msg.format(fqname_for(v.__class__)))
RuntimeError: Cannot serialize type gluonts.torch.modules.loss.NegativeLogLikelihood. See the documentation of the `encode` and
`validate` functions at

shell returned 1
```

## Environment
- Operating system: macOS
- Python version: 3.8
- GluonTS version: current dev (9322d73)
- MXNet version: -

jaheba:
Fixes: #2090 

*Description of changes:*

### DIFF
--- a/src/gluonts/torch/modules/loss.py
+++ b/src/gluonts/torch/modules/loss.py
@@ -17,27 +17,39 @@ from pydantic import create_model
 
 class DistributionLoss(torch.nn.Module):
     def __init_subclass__(cls):
+        # If a class doesn't have annotations, it inherits the ones from its
+        # parent. We don't want that!
         if cls.__annotations__ is torch.nn.Module.__annotations__:
-            return
+            annotations = {}
+        else:
+            annotations = cls.__annotations__
 
+        # We create a new pydantic model to check the input args
         cls.__pydantic_model__ = create_model(
             cls.__name__ + "Model",
             **{
                 name: (ty, cls.__dict__.get(name, ...))
-                for name, ty in cls.__annotations__.items()
+                for name, ty in annotations.items()
                 if not name.startswith("_")
             },
         )
 
     def __new__(cls, **kwargs):
+        # First we need to create the object.
         obj = object.__new__(cls)
+
+        # Now we check the **kwargs using the generated model.
         values = cls.__pydantic_model__(**kwargs).dict()
+
+        # Update the __dict__, which is like assigning the values to `obj`
         obj.__dict__.update(values)
-        obj.__newargs_ex__ = values
+
+        # Store the values so we can return them for the pickle protocol.
+        obj.__init_kwargs__ = values
         return obj
 
     def __getnewargs_ex__(self):
-        return (), self.__newargs_ex__
+        return (), self.__init_kwargs__
 
     def __init__(self, **kwargs):
         torch.nn.Module.__init__(self)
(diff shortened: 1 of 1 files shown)


jaheba:
Other methods (pydantic BaseModel or dataclass) didn't work, because `torch.nn.Module` defines `__getattr__` and `__setattr__` which clashes with the more elaborated approaches.

lostella:
Another approach could be that of having `DistributionLoss` not inherit from `torch.nn.Module` since there is no need for it (loss functions are just functions, they don't hold learnable parameters).

If/when there is need for it (but I cannot think of any such case right now) we can reconsider this.

src/gluonts/torch/modules/loss.py
```diff
@@ -12,15 +12,16 @@
 # permissions and limitations under the License.
 
 import torch
+from pydantic import BaseModel
 
 
-class DistributionLoss(torch.nn.Module):
+class DistributionLoss(BaseModel):
```
jaheba, line 18: Not sure we want to make this a `BaseModel` since we then force all of the child classes to use it as well. Maybe it's better to just do it on all children?

lostella: Maybe yes, but also not a big deal at the moment

lostella (approved):

PR #2104, opened 2022-06-28
Add more use-cases for `PandasDataset`

### LINKED ISSUE #2089: Improving `PandasDataset` tutorial
The tutorial on `PandasDataset` should be improved to cover common workflows like:
- loading data from a single "long" CSV file
- doing a `groupby` on it to then cast the groups to have a certain frequency using `asfreq` (and filling the gaps)
- loading data from a "wide" CSV file, where each column is a separate univariate target

Essentially, it would be nice to expose all common ways one can bring data from a file to GluonTS via Pandas, with common adjustments/preprocessing that might be needed, instead of randomly generating the `DataFrame` objects directly.

rsnirwan:
closes #2089 

### DIFF
--- a/docs/tutorials/data_manipulation/pandasdataframes.md.input
+++ b/docs/tutorials/data_manipulation/pandasdataframes.md.input
@@ -9,7 +9,7 @@ static and dynamic features and how to do train/test splits.
 
 ## Introduction
 
-`PandasDataset` is meant to work on `pandas.Dataset`, `pandas.Series` and
+`PandasDataset` is meant to work on `pandas.DataFrame`, `pandas.Series` and
 any collection of both.
 The minimal requirement to start modelling using GluonTS's dataframe dataset
 is a list of monotonically increasing timestamps with a fixed frequency
--- a/docs/tutorials/data_manipulation/pandasdataframes.md.input
+++ b/docs/tutorials/data_manipulation/pandasdataframes.md.input
@@ -155,7 +155,6 @@ dataframe to fill in desired value.
 ```python
 from gluonts.dataset.pandas import PandasDataset
 
-# get the latest date. we also fill forward if latest entries are missing.
 max_end = max(df.groupby("item_id").apply(lambda _df: _df.index[-1]))
 dfs_dict = {}
 for item_id, gdf in df.groupby("item_id"):
--- a/docs/tutorials/data_manipulation/pandasdataframes.md.input
+++ b/docs/tutorials/data_manipulation/pandasdataframes.md.input
@@ -1,15 +1,128 @@
 # `pandas.DataFrame` based dataset
 
+
+This tutorial covers how to use GluonTS's pandas DataFrame based dataset
+`PandasDataset`. After going through a few common use cases we explain in
+detail what data formats `PandasDataset` can work with, how to include
+static and dynamic features and how to do train/test splits.
+
+
+
+## Use case 1 - Loading data from a `long` dataframe and training `DeepAR`
+
+In the first use case we are given multiple time series stacked on top
+of each other in a dataframe with an `item_id` column that distinguishes
+different series.
+
+```python
+import pandas as pd
+
+url = (
+    "https://gist.githubusercontent.com/rsnirwan/a8b424085c9f44ef2598da74ce43e7a3"
+    "/raw/b6fdef21fe1f654787fa0493846c546b7f9c4df2/ts_long.csv"
+)
+df = pd.read_csv(url, index_col=0, parse_dates=True)
+df.head()
+```
+
+After reading the data into a `pandas.DataFrame` we can easily convert
+it to `gluonts.dataset.pandas.PandasDataset` and train an estimator an get forecasts.
+
+```python
+from gluonts.dataset.pandas import PandasDataset
+
+ds = PandasDataset.from_long_dataframe(df, target="target", item_id="item_id")
+```
+
+```python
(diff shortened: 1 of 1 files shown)


docs/tutorials/data_manipulation/pandasdataframes.md.input
```diff
@@ -354,3 +355,62 @@ for index, ax in zip(indices, axes):
...
+we can easily pivot from wide to long using `pd.melt`.
+
+```python
+import numpy as np
+import pandas as pd
+from gluonts.dataset.pandas import PandasDataset
+
+N, T = 3, 10
+index = pd.date_range("2021-01-01", freq="1D", periods=T)
+wide = pd.DataFrame(np.random.normal(size=(T, N)), columns=["A", "B", "C"], index=index)
```
lostella: Also here I would just load a CSV from somewhere, since that's how one would want to use it

docs/tutorials/data_manipulation/pandasdataframes.md.input
```diff
@@ -354,3 +355,62 @@ for index, ax in zip(indices, axes):
...
+print(wide)
+
```
lostella: Maybe we split the cells here?

docs/tutorials/data_manipulation/pandasdataframes.md.input
```diff
@@ -354,3 +355,62 @@ for index, ax in zip(indices, axes):
...
+
+## Use case 2 - Loading data from a single "long" CSV file and filling missing rows
+
+In this section we load data from a long-dataframe format, where each time series is
+stacked on top of each other and has gaps that need to be filled.
+
+To fill the gaps (to include new rows), we group by the `item_id` and reindex the
+dataframe according to our needs. New rows will be created with `NaN`s which
+we fill with -1.
```
lostella: Is this something one would do? Why not just leaving the `nan`s there? In a sense, reindexing to get the complete time index already "fills the gaps" with `nan`s, then whether `nan`s there make sense or another assumption should be made (like missing data meaning `0`) depends on the context. However `-1` seems a little arbitrary

lostella (commented):
Thanks @rsnirwan! I would double down on this, and put these examples on top since they are simpler and probably more common use cases. My concern is that the top section of the notebook contains a lot of stuff before we show simple examples.

In my mind, the structure could be:
1. Loading data from a long CSV, where nothing needs to be done
2. Long CSV with gaps, where one needs to group by, reindex, and construct the dataset from the groups. For this one can take the previous file, remove random rows, assume those are "missing"
3. Wide CSV file
4. Anything else

In 1-3 I would avoid any data generation, and just start from a CSV file. Doing the wide format as first example could also be an option, if we had one `from_wide_dataframe` constructor that does the one-liner.

docs/tutorials/data_manipulation/pandasdataframes.md.input
rsnirwan: filling -1 for missing values when the rest is positive makes it more visible when printing.. that was the only reason :D . but you are right, no need to fill in any value here.

docs/tutorials/data_manipulation/pandasdataframes.md.input
```diff
@@ -1,15 +1,16 @@
...
-GluonTS estimators.
+This tutorial covers how to use GluonTS's pandas DataFrame based dataset
+`PandasDataset`. After going through a few common use cases we explain in
+detail what data formats `PandasDataset` can work with, how to include
+static and dynamic features and how to do train/test splits.
+
 
+## Introduction
 
+`PandasDataset` is meant to work on `pandas.Dataset`, `pandas.Series` and
```
lostella:
```suggestion
`PandasDataset` is meant to work on `pandas.DataFrame`, `pandas.Series` and
```

lostella (changes requested):

lostella (approved):
🚀

docs/tutorials/data_manipulation/pandasdataframes.md.input
```diff
@@ -54,9 +55,126 @@ In this example, we have two time series, one indicated by item `A` and the othe
...
+
+Now, we group by the `item_id` and reindex each of the grouped dataframes.
+Reindexing, as it is done below, will add new rows with `NaN` values where
+the data is missing. The user can then use the `fillna()` method on each
+dataframe to fill in desired value.
+
+```python
+from gluonts.dataset.pandas import PandasDataset
+
+# get the latest date. we also fill forward if latest entries are missing.
```
lostella: Maybe this is redundant

```suggestion
```

docs/tutorials/data_manipulation/pandasdataframes.md.input
```diff
@@ -54,9 +55,126 @@ In this example, we have two time series, one indicated by item `A` and the othe
...
+max_end = max(df.groupby("item_id").apply(lambda _df: _df.index[-1]))
+dfs_dict = {}
+for item_id, gdf in df.groupby("item_id"):
+    new_index = pd.date_range(gdf.index[0], end=max_end, freq="1D")
+    dfs_dict[item_id] = gdf.reindex(new_index).drop("item_id", axis=1)
+
+# here we use the dict of dataframes directly to instantiate `PandasDataset`
```
lostella:
```suggestion
```

lostella (changes requested):

lostella (approved):

PR #2106, opened 2022-06-29
New readme example

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ plt.plot(test.index, test.values, label="Test")
 plt.legend(loc="upper left")
 ```
 
-![[train-test]](https://ts.gluon.ai/static/README/train-test.png)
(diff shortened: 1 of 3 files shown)


src/gluonts/model/forecast.py
```diff
@@ -392,7 +392,7 @@ def alpha_for_percentile(p):
 
         if show_mean:
             mean_data = np.mean(self._sorted_samples, axis=0)
-            pd.Series(data=mean_data, index=self.index).plot(
+            pd.Series(data=mean_data, index=self.index.to_timestamp()).plot(
```
lostella: Should we have this separately?

lostella: What I mean is: it would be nice to understand the issue with `self.index` better, so a dedicated PR may be worth it

jaheba: Probably.

lostella (dismissed):

PR #2107, opened 2022-06-29
Fix: Use timestamps for forecast plotting.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -387,12 +387,12 @@ class Forecast:
         i_p50 = len(percentiles_sorted) // 2
 
         p50_data = ps_data[i_p50]
-        p50_series = pd.Series(data=p50_data, index=self.index)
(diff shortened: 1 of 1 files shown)


src/gluonts/model/forecast.py
```diff
@@ -404,7 +404,7 @@ def alpha_for_percentile(p):
             ptile = percentiles_sorted[i]
             alpha = alpha_for_percentile(ptile)
             plt.fill_between(
-                self.index,
+                self.index.to_timestamp(),
```
lostella, line 407: Should this be done once and for all? If `DatetimeIndex` is what is really needed here?

lostella (commented):
Do we know why this is needed?

src/gluonts/model/forecast.py
jaheba: Maybe, but I think we want to replace this entire thing in `0.11` anyway.

lostella (approved):

PR #2121, opened 2022-06-30
Change default behavior for TrainDatasets overwrite

### LINKED ISSUE #2120: TrainDataset.save default behaviour deletes everything
## Description
TrainDataset.save deletes all files in a folder as default behaviour (i.e. overwrite = True by default). Providing "." as path deletes all files in cwd which can cause lots of trouble for someone unfamiliar with the code. Proposal would be to set default behaviour to overwrite = False.

## To Reproduce

```python
from gluonts.dataset.repository.datasets import get_dataset  # type: ignore
from pathlib import Path

dataset_name = "m4_hourly"
dataset = get_dataset(dataset_name, regenerate=False)
dataset.save(str(".")) 
```

natashaklingenbrunn:
Fixes #2120 

*Description of changes:*
Default behaviour for TrainDatasets is no longer to overwrite, as this can delete the entire folder.

### DIFF
--- a/src/gluonts/dataset/common.py
+++ b/src/gluonts/dataset/common.py
@@ -93,7 +93,7 @@ class TrainDatasets(NamedTuple):
         self,
         path_str: str,
         writer: DatasetWriter,
-        overwrite=True,
+        overwrite=False,
     ) -> None:
         """
         Saves an TrainDatasets object to a JSON Lines file.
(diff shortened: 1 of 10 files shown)


jaheba (approved):

jaheba:
Thanks!

PR #2129, opened 2022-07-04
Add 'Background' section to docs.

jaheba:
*Issue #, if available:*

*Description of changes:*


docs/getting_started/introduction.md
```diff
@@ -0,0 +1,71 @@
+
+
+# What is Time-Series Forecasting?
```
lostella:
```suggestion
# What is time-series forecasting?
```

unless we're writing German!

docs/getting_started/introduction.md
```diff
@@ -0,0 +1,71 @@
...
+
+Generally speaking, forecasting just means making predictions about events in
+the feature. Trivially, while weather forecasting tries to predict the
```
huibinshen: If one encode the weather (sunny, cloudy, rainy) into numbers and we it is also a forecasting problem right? Maybe demand forecasting is a more common example?

docs/getting_started/introduction.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+Naturally, it's impossible to forecast the unpredictable. For instance, in 2019
+it was virtually impossible to account for the possibility of travel
+restrictions due to the Covid-pandemic when trying to forecacst travel demand
+for 2020.
+
+Thus, forecasting operates on the caveat that the underlying factors that
+generate the time-series values don't fundamentally change in the future. It is
+a tool to predict the ordinary and the surprising.
```
lostella: This seems contradictory with the previous sentence: you shouldn't be able to predict something that is too surprising, given the knowledge on past behaviour, right?

docs/getting_started/introduction.md
```diff
@@ -0,0 +1,71 @@
...
+
+To look at this another way: Models are actually trained to predict the past
+and it is only us that uses models to forecast into the future.
+
```
huibinshen: Since this is getting_started, I found that in forecasting, many concepts are not the same as other domain. For example, the feature in tabular dataset is just input while the features in forecasting means other relevant information that might be co-observed with the time series of interest. Something like this might be helpful: https://developer-blogs.nvidia.com/wp-content/uploads/2022/02/Time-Series-Forecasting_Pic-1-625x481.png

docs/getting_started/introduction.md
jaheba: The idea was to name something which isn't time-series forecasting.

docs/getting_started/introduction.md
jaheba: No, the question is whether the is should be also capitalized. 

Just take this article as an example 

https://www.nytimes.com/2022/07/03/health/covid-ppe-masks-health-care.html

docs/getting_started/introduction.md
jaheba: Features is another section I wanted to add.

docs/getting_started/introduction.md
jaheba: This is missing a not! 🫣

lostella (dismissed):

huibinshen (approved):
thanks Jasper!

Issue #2128, opened 2022-07-04
Interrupting training should lead to end of training instead of error.


jaheba:
When training is running, a CTRL-C interrupt should end training and return a result instead of bubbling the error.

We would need to catch `KeyboardInterrupt` in our trainers.

lostella:
I thought we had this at some point, but I cannot find it. The nice thing is that this would make our own training loop consistent with PyTorch Lightning's, see note in https://pytorch-lightning.readthedocs.io/en/stable/common/trainer.html

PR #2131, opened 2022-07-05
Interrupting `mx.Trainer` stops training.

Hongqing-work:
*Issue #2128 , if available:*

*Description of changes:*
Catch the KeyboardInterrupt and save the model when training.


jaheba:
Looks like flake8 and black are complaining -- otherwise I think this should be good to go.

The trainer is something we might want to rework more thoroughly, then we can also revisit whether to refine the interrupt handling.

jaheba (approved):

Issue #2133, opened 2022-07-05
get_dataset function is broken with python3.9


rsnirwan:
## Description
A call to `get_dataset` with any dataset doesn't work with python3.9, but works with python 3.8.
The Error is in [DatasetCollection](https://github.com/awslabs/gluon-ts/blob/dev/src/gluonts/dataset/__init__.py#L37).
It might be the interplay of inheriting from `Protocol` and `dataclass` decorator as reported [here](https://stackoverflow.com/questions/69962929/passing-the-arguments-to-dataclass-causes-typeerror).

## To Reproduce

```python
from gluonts.dataset.repository.datasets import get_dataset
get_dataset("m4_hourly")
```

## Error message or code output
```
warnings.warn(
Traceback (most recent call last):
  File "/local/home/nirwar/tests/gts39error/p.py", line 2, in <module>
    get_dataset("m4_hourly")
  File "/local/home/nirwar/tests/gts39error/lib/python3.9/site-packages/gluonts/dataset/repository/datasets.py", line 332, in get_dataset
    return load_datasets(
  File "/local/home/nirwar/tests/gts39error/lib/python3.9/site-packages/gluonts/dataset/common.py", line 446, in load_datasets
    train_ds = FileDataset(
  File "/local/home/nirwar/tests/gts39error/lib/python3.9/site-packages/gluonts/dataset/common.py", line 183, in FileDataset
    return DatasetCollection(
TypeError: DatasetCollection() takes no arguments
```

## Environment
- Operating system: Amazon Linux and macOS
- Python version: 3.9.7 (3.8.13 works)
- GluonTS version: 0.10.0
- MXNet version: -

jaheba:
There is no need to inherit from Dataset, since that's the idea behind Protocols: implicit membership. 

On the other hand the docs explicitly state that this is possible.

PR #2141, opened 2022-07-06
Fix `PandasDataset` for Python 3.9

### LINKED ISSUE #2140: TypeError: PandasDataset() takes no arguments for PandasDataset.from_long_dataframe
Following as described in https://ts.gluon.ai/stable/tutorials/data_manipulation/pandasdataframes.html#Use-case-3---Loading-data-with-missing-values I could not get it to work with my own DF nor the example given:

```
ds = PandasDataset.from_long_dataframe(df, target="target", item_id="item_id")

  File "C:\\AppData\Local\Temp/ipykernel_33648/388385624.py", line 1, in <module>
    ds = PandasDataset.from_long_dataframe(dfx, target="target", item_id="item_id")

  File "C:\\Anaconda3\lib\site-packages\gluonts\dataset\pandas.py", line 182, in from_long_dataframe
    return cls(dataframes=dict(list(dataframe.groupby(item_id))), **kwargs)

TypeError: PandasDataset() takes no arguments
```

Also wanted to know if there are any plans of adding a methodology to achieve the same result starting from a wide DF, but (unlike the wide df example) incorporating a way to pass a list of feat_dynamic_cat, feat_static_cat, and so on (in same shape as the width of the wide df).

lostella:
*Issue #, if available:* Fixes #2140

*Description of changes:* See discussion in #2133 

### DIFF
--- a/src/gluonts/dataset/pandas.py
+++ b/src/gluonts/dataset/pandas.py
@@ -24,7 +24,7 @@ from gluonts.dataset.field_names import FieldName
 
 
 @dataclass
-class PandasDataset(Dataset):
+class PandasDataset:
     """
     A pandas.DataFrame-based dataset type.
 
--- a/src/gluonts/dataset/pandas.py
+++ b/src/gluonts/dataset/pandas.py
@@ -19,7 +19,7 @@ import pandas as pd
 from pandas.core.indexes.datetimelike import DatetimeIndexOpsMixin
 from toolz import valmap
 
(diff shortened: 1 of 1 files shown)


lostella:
Makes me think: we don't have tests that raise the issue on Python 3.9? Are we sure the issue is related to Python 3.9? @rsnirwan @jaheba

jaheba:
Looks like this is deliberate: https://github.com/python/cpython/pull/27545

I don't think I agree with that decision.

jaheba (approved):

Issue #1852, opened 2022-01-06
Deprecate `validated`


jaheba:
# `validated` Deprecation

Many classes of GluonTS use the `validated` decorator, which offers the following benefits:

* Validation of input arguments.
* Providing an implementation of `__getnewargs_ex__`, which is then used by `serde`.
* Adds `__init_args__` to instance, which is also used for equality checks.
* Implements custom `__repr__` (via serde) to have a descriptive representation.

However, there are some disadvantages. `validated` is not that pythonic and confuses especially new users. It doesn't play well when a super-class uses `validated` but a subclass doesn't.

Of the above features the interplay with `serde` is the most important. It allows us to easily serialise objects to disk and load them later, for example when training a model in SageMaker and then using it for prediction in two distinct steps.

I would consider type checking and validation as "nice to have" but they are not strictly required.

## dataclasses

Python 3.7 introduced dataclasses, which offer a consice way to define classes:

```py
from dataclasses import dataclass

@dataclass
class Point:
    x: int
    y: int

# vs

class Point:
    def __init_(self, x, y):
        self.x = x
        self.y = y
```

In addition it also provides better representations and comparison:

```py
>>> p = Point(1, 2)
>>> p
Point(x=1, y=2)
>>> p == Point(1, 2)
True
>>> asdict(p)
{'x': 1, 'y': 2}
```

In addition, pydantic also provides a `dataclass` decorator, which adds type-checking.

## Serialisation

As we can see, dataclasses offer a concise interface to define new types and provide similar benefits as `validated`. However, there is an area where both differ: serialisation.

`validated` builds up on the `__getnewargs_ex__` interface, which during deserialisation just passes the init arguments to the constructor:

```py
args, kwargs = Point(x=1, y=2).__getnewargs_ex__()
# args: == ()
# kwargs: {"x": 1, "y": 2}

Point(*args, **kwargs)
```

In contrast dataclasses use the stateful pickle API:

```py
# not how dataclasses work, but close enough
state = asdict(Point)

point = Point.__new__(Point)
point.__dict__.update(state)
```

There are some advantages by just passing the construction arguments to the constructor. The arguments can get validated, spotting potential problems when the given class has changed. When just setting the state it is very difficult to verify that the newly constructed object is consistent, especially in cases where the public API of a given class hasn't changed, but internally there are differences.

Implemting a stateless serialisation for dataclasses is not straightforward, since passed arguments need not be assigned:

```py
@dataclass
class Example:
[...]

PR #2146, opened 2022-07-08
Set __getnewargs_ex__ on the Class Instance rather than the Class itself

aatifjiwani:
## Issue and Reproducible Case
I currently use `gluon-ts` to train a `DeepAREstimator` and predict on new data across different processes. This involves the following workflow:
- In one Python process, create an untrained `DeepAREstimator` and serialize it to the file system
- In a second Python process (possibly on a different machine), `deser` the untrained estimator, train the estimator, and then serialize it to the file system. 
- Finally, in a third Python process, `deser` the trained `Predictor` and predict on new data.

Assume these steps execute sequentially and all the files required for `deser` are copied to the respective machine (if needed). Unfortunately, the attempt to serialize the trained `DeepAREstimator` (a `Predictor`) fails during the execution of the second step. The execution stops when it attempts to serialize `gluonts.time_feature._base.MinuteOfHour` because there is no attribute `__getnewargs_ex__` on the object. 

### Reproducible Case (best reproduced via Jupyter Notebook):
In one Python process, execute:
```py
import dill
dar_estimator = DeepAREstimator(
    prediction_length=10,
    freq='15min',
    trainer=Trainer(epochs=1)
)
with open("model.dill", "wb") as f:
    f.write(dill.dumps(dar_estimator))
```

In a new Python process in the same location, execute:
```py
import dill
with open("nested/model.dill", 'rb') as f:
    dar_estimator = dill.loads(f.read())

trained_estimator = dar_estimator.train(<sample-data>)
with open("trained_model.dill", "wb") as f:
    f.write(dill.dumps(trained_estimator))
```

You should get the following `RuntimeError`:
```
RuntimeError: Cannot serialize type gluonts.time_feature._base.MinuteOfHour. See the documentation of the `encode` and
`validate` functions at

```

Also, executing 
```py
from gluonts.core.serde._base import encode
encode(trained_estimator)
```
results in the same `RuntimeError`. The same goes if we tried to execute `encode(dar_estimator)` on the untrained estimator. Interestingly, if we execute `dill.dumps(dar_estimator)` in the new process, it properly serializes the estimator (while `encode` does not). 

## Technical Explanation of Issue
This issue stems from the fact that for classes like `MinuteOfHour`, there is no explicit definition of `__getnewargs_ex__` (whether that be on the class itself or through its super classes). Instead, the attribute `__getnewargs_ex__` is added _retroactively_ to `obj.__class__` during the execution of the `@validated` decorator. In `component.py`, we do the following:

https://github.com/awslabs/gluon-ts/blob/94247a9c0d4768aeb4a17a8bb44252706c519a6a/src/gluonts/core/component.py#L342

Because `__getnewargs_ex__` is not explicitly defined and is instead set on the class dynamically, `dill` will fail to capture this member method during serialization and will not be able to load it back on deserialization. In fact, the following happens:

```py
import dill
with open("nested/model.dill", 'rb') as f:
    dar_estimator = dill.loads(f.read())

# the MinuteOfHour object
dar_estimator.time_features[0].__getnewargs_ex__()
>>> AttributeError: 'MinuteOfHour' object has no attribute '__getnewargs_ex__'

# the same goes for all gluon-ts objects
dar_estimator.__getnewargs_ex__()
>>> AttributeError: 'DeepAREstimator' object has no attribute '__getnewargs_ex__'
```

This issue prevents us from being able to fully serialize/deserialize estimators and predictors in different Python processes. The only way around this without changing the source code is to trigger the `@validated` decorator via calling the constructor. For example, the following code block will _not_ raise an error:
```py
import dill
from gluonts.time_feature._base import MinuteOfHour
with open("nested/model.dill", 'rb') as f:
    dar_estimator = dill.loads(f.read())

# the MinuteOfHour object
MinuteOfHour()
dar_estimator.time_features[0].__getnewargs_ex__()
```

## The Solution
To reiterate, the issue is that `dill` is not capturing `__getnewargs_ex__` on objects where this attribute was set retroactively on `obj.__class__`. The solution given in this PR is to instead set `__getnewargs_ex__` on the object itself and bind it as a member method. This way, `dill` will capture this member method on serialization and will load it back during deserialization. 

### DIFF
--- a/src/gluonts/core/component.py
+++ b/src/gluonts/core/component.py
@@ -17,6 +17,7 @@ import logging
 from collections import OrderedDict
 from functools import singledispatch
 from typing import Any, Type, TypeVar
+from types import MethodType
 
 import numpy as np
 from pydantic import BaseConfig, BaseModel, ValidationError, create_model
@@ -339,8 +340,11 @@ def validated(base_model=None):
                         if not skip_encoding(arg)
                     }
                 )
-                self.__class__.__getnewargs_ex__ = validated_getnewargs_ex
                 self.__class__.__repr__ = validated_repr
+
+                # bind validated_getnewargs_ex to the object such that the
+                # __getnewargs_ex__ method is captured during serialization
+                self.__getnewargs_ex__ = MethodType(validated_getnewargs_ex, self)
 
             return init(self, **all_args)
 


src/gluonts/core/component.py
```diff
@@ -339,9 +340,12 @@ def init_wrapper(*args, **kwargs):
                         if not skip_encoding(arg)
                     }
                 )
-                self.__class__.__getnewargs_ex__ = validated_getnewargs_ex
                 self.__class__.__repr__ = validated_repr
```
aatifjiwani, line 343: Notice that we do not have to make changes to setting `__repr__` because this method is explicitly defined on every Python class by default.

aatifjiwani:
Hmm, the CI shows that I've introduced some failures. Digging through the logs shows that the main point of failure is 
```py
AttributeError: '<Class>' object has no attribute 'validated_getnewargs_ex'
```

I'm confused why `pickle` is trying to call the `validated_getnewargs_ex` method. Even before this PR, we were never setting `validated_getnewargs_ex` as an attribute on a class, we are setting `__getnewargs_ex__`. Can a maintainer/contributor help me understand the test failures? (@lostella @jaheba)

aatifjiwani:
Ah after some experimentation, it seems to just be the behavior of `pickle`. It seems `dill` is robust to `types.MethodType` but `pickle` is not... ugh.

jaheba:
Thanks for the detailed report!

The issue appears to be with `Predictor` and not `Estimator. Note that to serialize the predictor (what you call `trained_estimator`) you should use `predictor.serialize(<path>)`.

---

Looking into this, it appears the problem is in `validated` and the fact that we assign `__getnewargs_ex__` when invoking `__init__`.

This is bad, because it requires `__init__` to be called before we can serialize the object. There is no easy fix, `validated` should probably have been a class decorator.

`pickle` and `dill` will not invoke `__init__` when deserializing, meaning `__getnewargs_ex__` will not be assigned leading to above issue.

I think #1852 is the best way out.

jaheba:
@aatifjiwani In general using `serde.encode` should work:

```py
import json
dar_estimator = DeepAREstimator(
    prediction_length=10,
    freq='15min',
    trainer=Trainer(epochs=1)
)
with open("estimator.json", "w") as f:
    json.dump(serde.encode(dar_estimator), f)
```

And:


```py
with open("nested/estimator.json", 'r') as f:
    dar_estimator = serde.decode(json.load(f))
```

c3-aatifjiwani:
Thank you @jaheba! Will try this out and report back if there are more issues

PR #2150, opened 2022-07-11
Docs: Simplify wide `DataFrame` example

lostella:
*Description of changes:* Just using `dict` on the wide `DataFrame` appears to be faster then `melt` + `from_long_dataframe`.

```python
from timeit import default_timer
import pandas as pd

class Timer:
    def __init__(self, verbose=True):
        self.verbose = verbose
        self.timer = default_timer
        
    def __enter__(self):
        self.start = self.timer()
        return self
        
    def __exit__(self, *args):
        end = self.timer()
        self.elapsed_secs = end - self.start
        self.elapsed = self.elapsed_secs * 1000  # millisecs
        if self.verbose:
            print(f'elapsed time: {self.elapsed} ms')

df = pd.DataFrame(
    {
        k: [1.0] * 5000
        for k in range(200)
    },
    index=pd.period_range("2005-01-01", periods=5000, freq="2H")
)

with Timer():
    ds1 = PandasDataset(dict(df))

with Timer():
    df_long = pd.melt(df.reset_index(), id_vars="index", value_vars=df.columns)
    ds2 = PandasDataset.from_long_dataframe(
        df_long, item_id="variable", target="value", timestamp="index"
    )

for entry1, entry2 in zip(ds1, ds2):
    assert entry1["start"] == entry2["start"]
    assert (entry1["target"] == entry2["target"]).all()
    assert entry1['item_id'] == entry2['item_id']
```

gives

```
elapsed time: 81.61166999999992 ms
elapsed time: 3277.0063850000006 ms
```

### DIFF
--- a/docs/tutorials/data_manipulation/pandasdataframes.md.template
+++ b/docs/tutorials/data_manipulation/pandasdataframes.md.template
@@ -93,10 +93,8 @@ predictions = predictor.predict(ds)
 
 ## Use case 2 - Loading data from a `wide` dataframe
 
-Here, instead of a `long` dataset, we are given data in the `wide` format,
-where time series are stacked side-by-side. To convert the dataset
-to `PandasDataset` we first convert it to the `long` format and then use the
-`PandasDataset.from_long_dataframe` constructor.
+Here, we are given data in the `wide` format, where time series are stacked side-by-side in a `DataFrame`.
+We can simply turn this into a dictionary of `Series` objects with `dict`, and construct a `PandasDataset` with it:
 
 ```python
 import pandas as pd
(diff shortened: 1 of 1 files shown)


rsnirwan (dismissed):
Yes, thanks! that makes way more sense. No need to convert to the long format first.

rsnirwan (approved):

jaheba:
What is the advantage of having `from_long_dataframe` if `dict` does the job?

rsnirwan:
`dict`, here, only converts a wide-dataset to a list of `pd.Series`. If data is given in long-dataformat we would still need the `PandasDatset.from_long_dataframe` constructor.

lostella:
> What is the advantage of having `from_long_dataframe` if `dict` does the job?

That's in case someone has data originally in the long format (i.e. data from multiple series interleaved in a single dataframe). Only going through it in the "wide" case appears not to be the fastest way

PR #2153, opened 2022-07-12
Remove deprecated args from `mx.Trainer`; rework LRS.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/mx/trainer/learning_rate_scheduler.py
+++ b/src/gluonts/mx/trainer/learning_rate_scheduler.py
@@ -13,14 +13,14 @@
 
 from dataclasses import field
 from typing import Dict, Any
+
+from pydantic.dataclasses import dataclass
 from typing_extensions import Literal
 
 import numpy as np
 import mxnet as mx
 from mxnet import gluon
 import mxnet.gluon.nn as nn
-
-from pydantic.dataclasses import dataclass, Field
 
 from gluonts.core.component import validated
 
--- a/src/gluonts/mx/trainer/learning_rate_scheduler.py
+++ b/src/gluonts/mx/trainer/learning_rate_scheduler.py
@@ -106,13 +106,14 @@ class MetricAttentiveScheduler(mx.lr_scheduler.LRScheduler):
         assert (
             self.base_lr > 0
         ), f"base_lr should be positive, got {self.base_lr}"
-        assert (
-            self.base_lr > self.min_lr
-        ), f"base_lr should greater than min_lr, {self.base_lr} <= {self.min_lr}"
+        assert self.base_lr > self.min_lr, (
+            "base_lr should greater than min_lr, "
+            f"{self.base_lr} <= {self.min_lr}"
+        )
 
         assert (
             0 < self.decay_factor < 1
-        ), f"decay_factor factor should be between 0 and 1, got {self.decay_factor}"
+        ), f"decay_factor:  0 < {self.decay_factor} < 1"
 
         assert (
             self.patience >= 0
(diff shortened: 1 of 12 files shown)


src/gluonts/mx/trainer/learning_rate_scheduler.py
```diff
@@ -215,32 +237,4 @@ def on_epoch_end(
...
-        should_continue = self.lr_scheduler.step(metric_value=epoch_loss)
-        if not should_continue:
-            print(
-                "Early stopping based on learning rate scheduler callback"
-                " (min_lr was reached)."
-            )
-            return False
-
-        pre_step_learning_rate = trainer.learning_rate
-        trainer.optimizer.set_learning_rate(
```
jaheba, line 227: This should never alter `trainer.learning_rate`:


```py
In [3]: mx.optimizer.Adam.set_learning_rate??
Signature: mx.optimizer.Adam.set_learning_rate(self, lr)
Source:   
    def set_learning_rate(self, lr):
        """Sets a new learning rate of the optimizer.

        Parameters
        ----------
        lr : float
            The new learning rate of the optimizer.
        """
        if self.lr_scheduler is not None: # pylint: disable=no-else-raise
            raise UserWarning("LRScheduler of the optimizer has already been "
                              "defined. Note that set_learning_rate can mutate "
                              "the value of the learning rate of the optimizer "
                              "only when the LRScheduler of the optimizer is "
                              "undefined.")
        else:
            self.lr = lr
```

src/gluonts/mx/trainer/learning_rate_scheduler.py
```diff
@@ -11,21 +11,55 @@
...
-from gluonts.exceptions import GluonTSUserError
+from gluonts.core.component import validated
 
 from .callback import Callback
 
 
-class MetricAttentiveScheduler(mx.lr_scheduler.LRScheduler):
-    r"""
+@dataclass
+class Metric:
```
lostella: Is `Metric` the right name? Seems like the sub-classes do not define actual metrics, but only a strategy for what value to keep track of. Maybe `Objective` is a better term?

src/gluonts/mx/trainer/learning_rate_scheduler.py
```diff
@@ -11,21 +11,55 @@
...
+    best: float
+
+    def update(self, metric: float) -> bool:
+        if self.should_update(metric):
+            self.best = metric
+            return True
+        return False
+
+    def should_update(self, metric: float) -> bool:
+        raise NotImplementedError
```
lostella, line 48: Minor, no real need to change this but: I would put `return True` here

test/mx/trainer/test_learning_rate_scheduler.py
```diff
@@ -113,18 +113,12 @@ def test_PatientMetricAttentiveScheduler(
         # We construct the scheduler with a different rate, because we need
         # to test that the optimizer can override this setting: this is
         # consistent with how schedulers and optimizers operate in MXNet.
-        base_lr=10 * base_lr,
+        learning_rate=base_lr,
```
lostella: I guess the comment above does not apply anymore?

jaheba: True.

lostella (approved):

lostella:
@jaheba since this was merged, the following warning started appearing:

```python
[...] src/gluonts/mx/trainer/learning_rate_scheduler.py:67: RuntimeWarning: fields may not start with an underscore, ignoring "_current_patience"
  @dataclass
```

Is this intended, or otherwise worth fixing?

lostella:
sorry, just saw #2166

jaheba:
@lostella I think we should still fix this separately. I think while `dataclasses.dataclass` support underscore, the pydantic version does not.

PR #2155, opened 2022-07-13
add encode for pydantic dataclass

Hongqing-work:
*Issue #1852, if available:*

*Description of changes:*
Add encode for `pydantic` dataclass. Later we can use `dataclass` to replace `validated`.

### DIFF
--- a/src/gluonts/core/serde/_base.py
+++ b/src/gluonts/core/serde/_base.py
@@ -220,6 +220,15 @@ def encode(v: Any) -> Any:
             "kwargs": encode(kwargs),
         }
 
+    if hasattr(v, "__pydantic_model__"):
+        member = v.__dict__
+        del member["__initialised__"]
+        return {
+            "__kind__": Kind.Instance,
+            "class": fqname_for(v.__class__),
+            "kwargs": encode(member),
+        }
+
     try:
         # as fallback, we try to just take the path of the value
         fqname = fqname_for(v)
--- a/test/core/test_serde.py
+++ b/test/core/test_serde.py
@@ -19,6 +19,7 @@ from typing import NamedTuple
 import numpy as np
 import pandas as pd
 import pytest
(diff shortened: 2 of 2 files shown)


jaheba:
Thank you!

I think we should add some more complex examples.


```py
from pydantic.dataclasses import dataclass
from dataclasses import InitVar, field

@dataclass
class X:
    a: int
    b: InitVar[int]
    c: int = field(init=False)

    def __post_init__(self, b):
        self.c = self.a + b
```

jaheba:
It might be that we need to define our own `dataclass` decorator to store the init-args.

Hongqing-work:
> It might be that we need to define our own `dataclass` decorator to store the init-args.

So what kind of new features we want to have in our `dataclass` decorator compared with that of `pydantic`?

jaheba:
Something which we aimed for was to not serialize the object's state, but it's init args instead. The idea was to capture how something was called, and then replicate the result, instead of copying the state.

This has some advantages: The result is less noisy, since we don't encode members which use their default values. Also, we need to encode less data, in cases were we generate some data -- imagine something like `np.arange(10_000)`.

That said, getting this done with dataclasses could be painful. We could just store and encode the unhandled init args, relying on the fact that they worked when the object was created for the first time. Alternatively, we could create the pydantic model ourselves, similar to how we do it in `validated`. But then we also need to provide a custom `dataclass` decorator.

I think I tend to the latter approach, even though it will be much more work.


```py
from gluonts.core.sede import dataclass
from dataclasses import InitVar, field

@dataclass
class X:
    a: int
    b: InitVar[int]
    c: int = field(init=False)

    def __post_init__(self, b):
        self.c = self.a + b


x = X(1, 2)
serde.encode(x) # returns something like {"a": 1, "b": 2}
```

Maybe we can use `frozen=True` by default, making serialization more reliable.

PR #2166, opened 2022-07-14
Add `serde.dataclass`.

jaheba:
*Issue #, if available:*

*Description of changes:*


lostella:
This is nice and should make many of the classes (especially estimators) more concise. I'm wondering how much of a burden you think maintaining our own `dataclass` decorator will be?

jaheba:
Maintaining it shouldn't be so hard, since we offload the heavy lifting to pydantic's dataclass.

However, there are a few things we should think about.

We have this concept of capturing the init-args to re-create the instances. This can lead to issues when the object is mutated, since then syncs get out of sync. I'm not sure we want to have some checks that ensure that the object is not altered, or just rely on it through documentation. Maybe we could try to make it hashable, since that forces that nothing really changes. Still, this is a possible source of errors.

The other question is if we want to just capture the passed init-args, or these args plus the default arguments. The first would make serialisation output much smaller. We could have both in `serde`, so you could do something like `serde.encode(x, minimal=True)` to just capture the actually passed arguments.

In this draft, we convert args + kwargs into kwargs only, which could be prone for errors, and also requires us to generate good error messages. We could force use of kwargs only.

There can be a mismatch between public interface and internal state (this is true for any dataclass implementation). For example `context_length` is `Optional`, but if not set, it will default to `prediction_length`. Thus, once initialised the type is `int` and not `Optional[int]` anymore. I'm not sure how to support these cases yet.

jaheba:
I've added support for inferring values (not sure the name is good).

But the idea is this:


```py
@serde.dataclass
class X:
    x: int
    y: int = serde.INFER

    def __infer__(self, y):
        y.map(self.x * 2)


assert X(1).y == 2
```

The nice thing is that we can use `y: int` as a type here, opposed to `y: Optional[int]`.

I think I'm happy with the approach, except for naming.

jaheba:
Current design allows for something like this:

```py

@serde.dataclass
class Estimator:
    prediction_length: int
    context_length: int = serde.OrElse(
        lambda prediction_length: prediction_length * 2
    )

    use_feat_static_cat: bool = True
    cardinality: List[int] = serde.EVENTUAL

    def __eventually__(self, cardinality: serde.Eventual[List[int]]):
        if not self.use_feat_static_cat:
            cardinality.set([1])
        else:
            cardinality.set_default([1])


x = Estimator(
    prediction_length=12, cardinality=[1, 2, 3], use_feat_static_cat=False
)
```

lostella (approved):

PR #2170, opened 2022-07-18
Apply Serde dataclass

Hongqing-work:
*Issue #2166, #1852 , if available:*

*Description of changes:*
Fix some bug for the draft PR and apply it on more models to avoid using `validated`.


src/gluonts/dataset/stat.py
```diff
@@ -67,6 +59,8 @@ def add(self, target_values):
         if len(target_values) > 0:
             bucket = self.bucket_index(target_values)
             self.bin_counts[bucket] = self.bin_counts[bucket] + 1
+            # TODO: Hongqing, check if there is another way
+            self.__init_kwargs__["bin_counts"] = self.bin_counts
```
jaheba: I think this class doesn't lend itself to be used with `serde.dataclass`. There is mutable state, which should follow a different model.

I'm not sure if and how we want to support these cases with dataclass.

src/gluonts/model/estimator.py
```diff
@@ -30,14 +32,12 @@ class Estimator:
     __version__: str = gluonts.__version__
 
     prediction_length: int
-    lead_time: int
+    lead_time: int = 0
```
jaheba: Something that should work is to do 
```suggestion
    lead_time: int = field(default=0, metadata={"ge": 0})
```

Then pydantic does the bounds checking for us.

src/gluonts/model/forecast.py
```diff
@@ -505,32 +506,30 @@ class SampleForecast(Forecast):
...
-    ) -> None:
+    samples: np.ndarray
+    start_date: pd.Period
+    item_id: Optional[str] = None
+    info: Optional[Dict] = None
+
+    def __post_init_post_parse__(self):
         assert isinstance(
-            samples, np.ndarray
+            self.samples, np.ndarray
```
jaheba: I wonder if this is needed -- shouldn't pydantic take care of this check?

src/gluonts/model/forecast.py
```diff
@@ -505,32 +506,30 @@ class SampleForecast(Forecast):
...
+
         self._sorted_samples_value = None
         self._mean = None
         self._dim = None
-        self.item_id = item_id
-        self.info = info
 
         assert isinstance(
-            start_date, pd.Period
+            self.start_date, pd.Period
```
jaheba: Same thing here.

src/gluonts/model/naive_2/_predictor.py
```diff
@@ -133,30 +134,21 @@ class Naive2Predictor(RepresentablePredictor):
...
 
-        self.freq = freq
-        self.prediction_length = prediction_length
-        self.season_length = (
-            season_length
-            if season_length is not None
-            else get_seasonality(freq)
-        )
+        if self.season_length is None:
+            self.season_length = get_seasonality(self.freq)
```
jaheba, line 152: This we should be able to simplify using the `OrElse` approach I've added.

The idea is that you just specify this:

```py
season_length: int = serde.OrElse(lambda freq: get_seasonality(self.freq))
```

If one passes a value we take that, "or else" we use `get_seasonality`.

src/gluonts/dataset/stat.py
```diff
@@ -39,22 +40,18 @@ class ScaleHistogram:
...
+
+    def __post_init_post_parse__(self):
         self.bin_counts = defaultdict(
-            int, {} if bin_counts is None else bin_counts
+            int, {} if self.bin_counts is None else self.bin_counts
         )
-        self.empty_target_count = empty_target_count
-        self.__init_args__ = dict(
-            base=self._base,
+        self.__init_kwargs__ = dict(
```
Hongqing-work, line 51: If the value is mutable, we should set the `__init_kwargs__` in `__post_init_post_parse__` and use `self.xxx`

Hongqing-work:
I've add some examples and guide in `/examples`.

Issue #2067, opened 2022-06-23
Multivariate Hierarchical Timeseries Forecasting


AkshtaChoudhary:
Trying to add features to hierarchical model (Multivariate time series)  using DeepVARHierarchicalEstimator)

I have 3 dataframes , data_df, feature1, feature2 which are of the size (21 x 36). There are 36 time periods and 21 are the individual time series including TOTAL

I have created training dataset in the following format

# CODE 

--# Forecast Length
forecast_length = 6

--# Frequency
freq = 'M'

-- #Dataframes
train_df = data_df.iloc[:,:-forecast_length]
f1_train_df = f1_df.iloc[:,:-forecast_length]
f2_train_df = f2_df.iloc[:,:-forecast_length]

 --# Creating Dictionary for training data 
traindict ={}

for x in train_df.index:
    traindict[x] = {'target': train_df[train_df.index=='Total'],
                           'start' : train_df.columns[0],
                           'features': np.vstack((f1_train_df [f1_train_df .index==x].values.flatten(),
                                                            f2_train_df [f2_train_df .index==x].values.flatten()))}

--# Training List Dataset
train_ds = ListDataset([{'target': traindict[df]['target'].values.flatten(), 'start':  traindict[df]['start'],
                                     'feat_dynamic_real': traindict[df]['features'] } for df in traindict.keys()],
                                    freq=freq,one_dim_target=False)  

--# S Matrix 
total_s = np.ones((1,len(brand_list))) -- #Creating 1's for the total level data
id_mat = np.eye(len(brand_list)) -- # Identity matrix of RegionState level combination
sum_matrix_rand =np.concatenate((total_s,id_mat)) --# Creating the S matrix for all the levels

--# Estimator
hier_estimator = DeepVARHierarchicalEstimator(freq = freq,
                                              prediction_length = forecast_length,
                                              target_dim = data_df.shape[0],
                                              S = sum_matrix_rand,
                                              trainer=Trainer(epochs=25, hybridize=False, learning_rate=0.1),
                                              context_length=10,
                                              num_layers=10)

--# Predictor
hier_predictor = hier_estimator.train(training_data=train_ds)
 
# ERROR   
I am getting the below error :

---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
~\AppData\Local\Temp\ipykernel_46940\2690552091.py in <module>
      1 # Predictor
----> 2 hier_predictor = hier_estimator.train(training_data=train_ds)

[...]

Issue #1967, opened 2022-05-05
DeepVARHierarchicalEstimator returning AssertionError: Argument lhs must have NDArray type


rhameysayed:
## Description
I am trying to fit a hierarchical VAR model. The dataset (dollar_sales_hier) has 70 time-series, with 324 weekly observations, which combine into 10 categories, and one grand total - there are 81 time series. Each row in the dataset is a time series and the columns are indexed by date. I have made sure the dimensions of my matrices are similar, specifically for the summation matrix and the data set. I double-checked my input dataset and summation matrix by looking here: https://github.com/rshyamsundar/gluonts-hierarchical-ICML-2021/tree/master/experiments/data/tourism . Also, as shown in the paper here (https://www.amazon.science/publications/end-to-end-learning-of-coherent-probabilistic-forecasts-for-hierarchical-time-series) the S matrix (s_mat) is [S_sum | I (70x70)]

## To Reproduce

```python
import pandas as pd
import numpy as np
from gluonts.dataset.common import ListDataset
from gluonts.model.deepvar_hierarchical import DeepVARHierarchicalEstimator
from gluonts.mx.trainer import Trainer

groups = list('ABCDEFGHIJ')
group_list = list('ABCDEFGHIJ'*7)
time_line = pd.date_range("2016-01-10",freq= 'w',periods = 324)
dollar_sales_hier_rand = pd.DataFrame(np.random.randint(0,300,size=(70, 324)))
dollar_sales_hier_rand.columns = time_line
dollar_sales_hier_rand["Group"] = group_list
dollar_sales_hier_rand.set_index("Group",inplace = True)
dollar_sales_group_rand = dollar_sales_hier_rand.groupby("Group").agg('sum')
dollar_sales_total_rand = dollar_sales_hier_rand.agg('sum')
dollar_sales_hier_rand = pd.concat([dollar_sales_total_rand,dollar_sales_group_rand,dollar_sales_hier_rand])

total_s = np.ones((1,70))
id_mat = np.eye(70)
group_sum_mat = pd.DataFrame(index = groups,columns = group_list)
group_sum_mat = group_sum_mat.fillna(0)
for g in groups:
  group_sum_mat.loc[group_sum_mat.index.isin([g]),group_sum_mat.columns.isin([g])] = 1

sum_matrix_rand =np.concatenate((total_s,id_mat,group_sum_mat))

[...]

PR #2206, opened 2022-08-10
Add data preparation utility for hierarchical time series 

rshyamsundar:
*Issue #, if available:*

*Description of changes:*

This PR implements the missing functionality for handling hierarchical/grouped time series data and training hierarchical models correctly without data-preprocessing issues (see #1967 and #2067).  It also provides a tutorial on loading hierarchical time series and training and evaluation of the hierarchical model [`DeepVARHierarchical`](https://github.com/awslabs/gluon-ts/blob/dev/src/gluonts/mx/model/deepvar_hierarchical/_estimator.py#L95).

**Details:**
* The user has to provide only the bottom-level time series (pandas dataframe) and the aggregation matrix (numpy ndarray), which are the minimal requirements in constructing hierarchical or even grouped time series.  The aggregated time series are then internally constructed and all the time series in the hierarchy are available as a pandas dataframe.

* The hierarchical time series thus constructed can be easily converted to a `Dataset` on which mini-batch training can be done, thanks to the newly implemented [`PandasDataset`](https://github.com/awslabs/gluon-ts/blob/dev/src/gluonts/dataset/pandas.py#L27).

* The tutorial explains the procedure for creating hierarchical time series, training the model and doing evaluation via backtesting by creating the train-test split explicitly (i.e., not using `make_evaluation_predictions` or `backtest_metrics`).


lostella:
@rshyamsundar you also need to add the tutorial to the table of contents in the `index.md` in the same directory as the `.md.template` file

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
```diff
@@ -0,0 +1,177 @@
...
+
+# Make sure the dataframe has `PeriodIndex`.
+ts_at_bottom_level = ts_at_bottom_level.to_period()
+
+# Load the aggregation matrix `S`.
+S_csv = (
+    "https://gist.githubusercontent.com/rshyamsundar/17084fd1f28021867bcf6f2d69d9b73a/raw/"
+    "32780ca43f57a78f2d521a75e73b136b17f34a02/example_agg_mat.csv"
+)
+S = pd.read_csv(S_csv).values
```
lostella, line 58: Would it make sense to break this cell into multiple ones, to stop and explain what is being dealt with? For example, maybe showing what `S` contains and what it means would make this easier to follow or adapt to one's use case.

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
```diff
@@ -0,0 +1,177 @@
...
+We assume that the hierarchical time series `hts` of type `HierarchicalTimeSeries` has already been
+constructed as described above.
+The first step is to convert this hierarchical time series to a `Dataset` on which
+mini-batch training can be run.
+Here we convert it into `gluonts.dataset.pandas.PandasDataset`.
+
+```python
+from gluonts.dataset.hierarchical import to_pandas_dataset
+
+dataset = to_pandas_dataset(hts=hts)
```
lostella: This could be a method of `HierarchicalTimeSeries`:

```python
hts.to_pandas_dataset()
```

I'm wondering whether `pandas` should be in the name here though: in the end, what matters to the user is that this is converting a `HierarchicalTimeSeries` object into a `Dataset`, so maybe `to_dataset` is a better name.

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
```diff
@@ -0,0 +1,177 @@
...
+
+# Create `HierarchicalTimeSeries` object.
+hts = HierarchicalTimeSeries(
+    ts_at_bottom_level=ts_at_bottom_level,
+    S=S,
+)
+
+# You can access all the time series of the hierarchy using `ts_at_all_levels` property.
+hts.ts_at_all_levels.head()
+```
```
lostella: Also here: instead of using comments, you could break the cell and explain this in a text section

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
```diff
@@ -0,0 +1,177 @@
...
+ts_at_bottom_level = ts_at_bottom_level.to_period()
+
+# Load the aggregation matrix `S`.
+S_csv = (
+    "https://gist.githubusercontent.com/rshyamsundar/17084fd1f28021867bcf6f2d69d9b73a/raw/"
+    "32780ca43f57a78f2d521a75e73b136b17f34a02/example_agg_mat.csv"
+)
+S = pd.read_csv(S_csv).values
+
+# Create `HierarchicalTimeSeries` object.
```
lostella: This comment seems a bit redundant :-)

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
```diff
@@ -0,0 +1,177 @@
...
+    "raw/f02f5aeadad73e3f3e9cf54478606d3507826492/example_bottom_ts.csv"
+)
+ts_at_bottom_level = pd.read_csv(
+    ts_at_bottom_level_csv,
+    index_col=0,
+    parse_dates=True,
+)
+
+# Make sure the dataframe has `PeriodIndex`.
+ts_at_bottom_level = ts_at_bottom_level.to_period()
```
lostella:
```suggestion
ts_at_bottom_level = pd.read_csv(
    ts_at_bottom_level_csv,
    index_col=0,
    parse_dates=True,
).to_period()
```

The fact that a `PeriodIndex` is needed could be stressed in the text

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
```diff
@@ -0,0 +1,177 @@
...
+hts = HierarchicalTimeSeries(
+    ts_at_bottom_level=ts_at_bottom_level,
+    S=S,
+)
+
+# You can access all the time series of the hierarchy using `ts_at_all_levels` property.
+hts.ts_at_all_levels.head()
+```
+
+## Use case 1 - Simple Model Training and Forecasting
```
lostella: Not sure these two are orthogonal use cases, maybe we can just title this 

```suggestion
## Model Training and Forecasting
```

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
```diff
@@ -0,0 +1,177 @@
...
+predictor = estimator.train(dataset)
+
+forecast_it = predictor.predict(dataset)
+
+# There is only one element in `forecast_it` containing forecasts for all the time series.
+forecasts = next(forecast_it)
+```
+
+
+## Use case 2 - Model Evaluation via Backtesting
```
lostella: Similarly to above:

```suggestion
## Model Evaluation via Backtesting
```

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
rshyamsundar: ya, actually it is better it if it a method of `HierarchicalTimeSeries`.

I tried being explicit about to which dataset I am converting to (`PandasDataset`). `to_dataset` is also fine and we don't imagine converting this to to other types of datasets anyway right?

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
rshyamsundar: I had your suggested version initially and thought people might miss `to_period` call if it is in a single statement and raise unnecessary issues around this point. So made that explicit. I will change and stress it in the text.

src/gluonts/dataset/hierarchical.py
```diff
@@ -0,0 +1,217 @@
...
+    Parameters
+    ----------
+    ts_at_bottom_level
+        A single pandas dataframe whose columns are the time series
+        corresponding to the leaves of the hierarchy.
+    S
+        Summation or aggregation matrix whose ordering should be consistent
+        with the orderig of the columns of `ts_at_all_levels`.
+        In particular, the bottom `k x k` sub-matrix should be identity matrix,
+        where `k` is the number of leaves of the hierarchy.
```
lostella, line 47: If this is strictly required, then maybe it's fine to omit the bottom `k x k` submatrix and leave it implicit?

That would mean that `S` encodes the non-leaf nodes in the hierarchy only. I understand this may go against common mathematical notation, but it seems better on the coding side of things.

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
rshyamsundar: `S` is the standard summation matrix that anyone using the hierarchical models is familiar with. Also it is the user input and not an artifact of `gluonts` code. That's why I didn't want to dwell on it too much. We can provide this [reference link (Eq. 10.3)](https://otexts.com/fpp2/hts.html#fig:HierTree) to make it clear the expected format of the summation matrix.

src/gluonts/dataset/hierarchical.py
rshyamsundar: Didn't want to introduce yet another notation and users typically have this matrix available for their own processing tasks as well: multiplying the forecasts at bottom level by this matrix gives them "all" forecasts in one go. There is a good reason for the introduction of `S` in the literature and from the user-perspective, it makes sense to use this standard matrix.

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
rshyamsundar: Now I remember why I moved `to_pandas_dataset` out of the class. We are not converting just the `hts` to `Dataset` here. We are taking external features and the target time series (happened to be `hts`) to create a dataset. So I thought `to_dataset(hts, features)` is better than `hts.to_dataset(features)`. Your call!

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
rshyamsundar: Done

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
rshyamsundar: Removed.

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
rshyamsundar: Done

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
rshyamsundar: Done

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
rshyamsundar: Done

melopeo:
I think this PR is in great shape. One thing that could make it even better is to include an example (maybe like a 3rd section after `Model Training and Forecasting`, and `Model Evaluation via Backtesting`), where `feat_dynamic_real` are included. I suggest this as it is one element from `gluonts.dataset.hierarchical.to_pandas_dataset` that remains a tiny bit unclear how to use, and one can expect questions about this in the future.

docs/tutorials/forecasting/hierarchical_model_tutorial.md.template
lostella: Right, but features are optional, so I’d say making it a method is fine

rshyamsundar: Moved.

src/gluonts/dataset/hierarchical.py
lostella: I still think it's strange to require some input if part of it is expected to have specific values. But we can go on with this and maybe revisit it later.

lostella (approved):

rshyamsundar:
> I think this PR is in great shape. One thing that could make it even better is to include an example (maybe like a 3rd section after `Model Training and Forecasting`, and `Model Evaluation via Backtesting`), where `feat_dynamic_real` are included. I suggest this as it is one element from `gluonts.dataset.hierarchical.to_pandas_dataset` that remains a tiny bit unclear how to use, and one can expect questions about this in the future.

Hi @melopeo Here is the [PR](https://github.com/awslabs/gluon-ts/pull/2253) that addresses your comments. Thanks!

PR #2207, opened 2022-08-10
Revert "Fix `aggregate_valid` function to ignore non-numeric columns"

lostella:
This reverts awslabs/gluon-ts#2204 since it was pointed out by @kashif that it creates further issues.

### DIFF
--- a/src/gluonts/evaluation/_base.py
+++ b/src/gluonts/evaluation/_base.py
@@ -23,10 +23,10 @@ from typing import (
     Iterable,
     Iterator,
     List,
-    Mapping,
     Optional,
     Tuple,
     Union,
+    Mapping,
     cast,
 )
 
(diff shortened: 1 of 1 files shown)


kashif:
so i am thinking:

```
metric_per_ts = metric_per_ts.drop(FieldName.ITEM_ID, axis=1).apply(
        np.ma.masked_invalid
)
```

lostella:
@kashif let's first revert the change, then we see separately.

Thinking about it, I'm not sure `aggregate_valid` should even be there: what is it supposed to do? Aggregate only metric values that are finite? That will silently sweep under the rug the fact that some metric may not make sense in some test case (relative metrics on mostly-zero data), or that the model performs badly or is buggy (like forecasts containing `nan` or `inf`).

In my opinion, a model gets applied to test cases for evaluation, then numbers come out: if the numbers look weird, then one should either adjust the set of metrics to evaluate on, or the test cases to consider, or fix the model.

PR #2216, opened 2022-08-15
Fix `r_forecast` wrapper to shift start date when truncating time series

abdulfatir:
*Issue #, if available:*

*Description of changes:*

Fixed the r_forecast wrapper to shift the start date of the time series accordingly when truncating the time series. Currently, it uses the the original start date even for the truncated time series which results in an incorrect forecast start date.

### DIFF
--- a/src/gluonts/model/r_forecast/_predictor.py
+++ b/src/gluonts/model/r_forecast/_predictor.py
@@ -249,6 +249,9 @@ class RForecastPredictor(RepresentablePredictor):
 
         for data in dataset:
             if self.trunc_length:
+                data["start"] = data["start"] + (
+                    data["target"].shape[-1] - self.trunc_length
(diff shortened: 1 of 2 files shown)


lostella:
Thanks @abdulfatir! I think we should probably also update [this test](https://github.com/awslabs/gluon-ts/blob/ff2497154fe694554aa4142eaa8435cb1a61e115/test/model/r_forecast/test_r_predictor.py#L63-L65) to cover also non-default values of `trunc_length`. I guess that test could have caught this, right?

abdulfatir:
@lostella Updated test.

lostella (approved):
Thanks @abdulfatir! Looks good to me, it seems like tests for these models are not running in the GitHub workflows, but we will fix this separately.

Issue #2195, opened 2022-08-08
Significant Speed-up of Parquet Data Reader


mr-1993:
## Description

@emptymalei , @julian-sieber and myself were investigating a potential speed-up of your Parquet data reader since we use a similar implementation in a project of ours. What we found out is that there is a large potential in speeding it up. The difference in the implementation is essentially in the _arrow_to_py methods seen below. Instead of using a numpy-mediated conversion, we rely on pandas. In fact, this gives us a significant speed-up as the benchmarking section of the [attached notebook](https://gist.github.com/mr-1993/ee7dc378ba33d7f7f978be40af2cd8b0) suggests.  If you agree that this is a significant improvement of your current implementation, we would be happy to create a PR and implement the necessary changes.

https://github.com/awslabs/gluon-ts/blob/f0743d944a291d06d3a80561a0ec7368491e47b0/src/gluonts/dataset/arrow/dec.py#L23-L41

jaheba:
Thanks! Yes, that would be very helpful :)

One thing we need to check is to translate the flattened arrays.

mr-1993:
Hi @jaheba, thank you very much for your interest in this enhancement! Just as a clarification so that we can run more experiments: what exactly do you mean with "flattened arrays"? do you have a small example? thank you!

jaheba:
We use an optimisation to support arrays with 2 or more dimensions:

https://github.com/awslabs/gluon-ts/blob/4c78ecfac2357935ab523e93d6e857cd8749aad6/src/gluonts/dataset/arrow/enc.py#L61-L70

This is something we also need to handle during decoding:

https://github.com/awslabs/gluon-ts/blob/4c78ecfac2357935ab523e93d6e857cd8749aad6/src/gluonts/dataset/arrow/dec.py#L74-L78

This enables zero-copy reads for arrays with 2 or more dimensions.

jaheba:
@mr-1993 @emptymalei If you need any additional help, let me know! We can have a quick chat if you want.

mr-1993:
That would actually be great, thank you! I am having a bit of trouble understanding how the schema is generated. Would you have a few minutes between 11:30 am and 1 pm?

jaheba:
11:30 sounds good.

PR #2217, opened 2022-08-15
Speedup data reader

mr-1993:
*Issue #2195, if available:*

*Description of changes:* @emptymalei @julian-sieber and I have implemented a significant speedup of the data reader. @jaheba Please have a look. We significantly simplified the from_schema method in the decoder by using dictionaries directly instead of the need to infer the idx of the ndarray columns (maybe you have further ideas to simplify the code? we did not do this yet in order not to break anything). For the moment, the implementation is only for data frames the rows of which contain 1D- or 2D-arrays. 

**Please tag this pr with at least one of these labels to make our release process faster:** other change

### DIFF
--- a/src/gluonts/dataset/arrow/dec.py
+++ b/src/gluonts/dataset/arrow/dec.py
@@ -60,8 +60,7 @@ class ArrowDecoder:
             row = {}
             for column_name in self.columns:
                 value = raw_row[column_name]
-                shape_col_name = f"{column_name}._np_shape"
-                shape = raw_row.get(shape_col_name)
+                shape = raw_row.get(f"{column_name}._np_shape")
 
                 if shape is not None:
                     value = np.stack(value).reshape(shape)
--- a/src/gluonts/dataset/arrow/dec.py
+++ b/src/gluonts/dataset/arrow/dec.py
@@ -48,13 +48,8 @@ class ArrowDecoder:
 
     @classmethod
     def from_schema(cls, schema):
-        columns = []
+        return cls([column.name for column in schema if not column.name.endswith("._np_shape")])
 
-        for column in schema:
-            if not column.name.endswith("._np_shape"):
-                columns.append(column.name)
-
-        return cls(columns)
 
     def decode(self, batch, row_number):
         for row in self.decode_batch(batch.slice(row_number, row_number + 1)):
--- a/src/gluonts/dataset/arrow/dec.py
+++ b/src/gluonts/dataset/arrow/dec.py
@@ -41,15 +41,19 @@ def _arrow_to_py_list_scalar(scalar: pa.ListScalar):
     return arr
 
 
-
 @dataclass
 class ArrowDecoder:
     columns: Dict[str, int]
 
     @classmethod
     def from_schema(cls, schema):
-        return cls([column.name for column in schema if not column.name.endswith("._np_shape")])
-
+        return cls(
+            [
+                column.name
+                for column in schema
+                if not column.name.endswith("._np_shape")
+            ]
+        )
 
     def decode(self, batch, row_number):
         for row in self.decode_batch(batch.slice(row_number, row_number + 1)):
@@ -64,11 +68,12 @@ class ArrowDecoder:
 
                 if shape is not None:
                     value = np.stack(value).reshape(shape)
-                if isinstance(value, np.ndarray) and isinstance(value[0], np.ndarray) and len(value.shape) == 1:
+                if (
+                    isinstance(value, np.ndarray)
+                    and isinstance(value[0], np.ndarray)
+                    and len(value.shape) == 1
+                ):
                     value = np.stack(value)
                 row[column_name] = value
 
             yield row
-
-
-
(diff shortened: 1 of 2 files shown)


lostella:
@mr-1993 looks like style checks are failing; we use `black` to enforce coding styles, you can fix the issue by installing it and running

```
black src test
```

and committing the changes it applies.

mr-1993:
Thanks @lostella! Should be fixed by now

mr-1993:
Some tests fail atm, but not related to our implementation (numerical instabilities)

src/gluonts/dataset/arrow/dec.py
```diff
@@ -44,39 +44,36 @@ def _arrow_to_py_list_scalar(scalar: pa.ListScalar):
...
-
-                shape_idx = self.ndarray_columns.get(column_name)
-
-                if shape_idx is not None:
-                    shape = _arrow_to_py(raw_row[shape_idx])
-                    value = value.reshape(shape)
-
+            for column_name in self.columns:
+                value = raw_row[column_name]
+                shape = raw_row.get(f"{column_name}._np_shape")
```
jaheba: I don't think we want to do this check for every entry.

Rather, I would keep storing `ndarray_columns` and then just itereate over them.

src/gluonts/dataset/arrow/dec.py
```diff
@@ -44,39 +44,36 @@ def _arrow_to_py_list_scalar(scalar: pa.ListScalar):
...
+
+                if shape is not None:
+                    value = np.stack(value).reshape(shape)
```
jaheba: Why stack the value here?

src/gluonts/dataset/arrow/dec.py
```diff
@@ -44,39 +44,36 @@ def _arrow_to_py_list_scalar(scalar: pa.ListScalar):
...
+                if (
+                    isinstance(value, np.ndarray)
```
jaheba: Couldn't we just check the shape of the array?

jaheba (commented):
Thanks for the PR.

Do you have any comparison numbers on how much this improves decoding?

src/gluonts/dataset/arrow/dec.py
mr-1993: Thanks for the comment, does not seem to be necessary, will delete it

src/gluonts/dataset/arrow/dec.py
emptymalei: Hi @jaheba ,

The problem is inline with what we discussed last time. What comes out of the pandas conversion for 2d array is something like

```python
np.array([
    np.array([1,2,3]),
    np.array([4,5,6])
])
```

then if we check the shape, we would only get 1d array. This is why we have 

```python
and len(value.shape) == 1
```

here.

But would be glad to adapt if you have a better idea.

mr-1993: To the best of our knowledge this does not work. The problem lies in the pandas conversion of multi-dimensional arrays. One obtains arrays of arrays (instead of multi dim arrays)

mr-1993:
> Thanks for the PR.
> 
> Do you have any comparison numbers on how much this improves decoding?

Atm, we have just tested it on our internal data which we are not allowed to share (but which contains a lot of 2D arrays). We will run some consecutive experiments on open source data, but for benchmarking, we already provide this screenshot with the results:
image

image

src/gluonts/dataset/arrow/dec.py
mr-1993: We would have to run some additional experiments to see whether using `ndarray_columns` really speeds things up significantly. The codebase without it seems to be more elegant. We will provide you with benchmarks as soon as we have them.

jaheba:
Thanks! These are massive improvements. Can you share a bit what shape the data has? I assume the larger each row is the bigger the performance difference gets. If I remember correctly, I saw a 2x improvement on some simpler case I tested, but this looks much more impressive!

src/gluonts/dataset/arrow/dec.py
jaheba: Oh, I see. Thanks for the clarification.

I think neither arrow nor pandas support arrays with 2 or more dimensions, thus resulting in these nested arrays, with `dtype=object`.

That's why the re-shaping approach makes things so much easier.

Playing around it looks like `np.stack` is indeed able to handle nested arrays just fine, even with dimensions higher than 2.

However, I think we can enable stacking only on arrays which `dtype == object`.

mr-1993:
> Thanks! These are massive improvements. Can you share a bit what shape the data has? I assume the larger each row is the bigger the performance difference gets. If I remember correctly, I saw a 2x improvement on some simpler case I tested, but this looks much more impressive!

Our dataset contains 500 rows with: 12 columns that contain strings or integers, and 11 columns that contain arrays. 6 of these arrays are one-dimensional and 5 are two-dimensional. They have varying sizes, the largest of the one-dim arrays contain around 270 elements, the largest two-dim arrays contain around 15x270 entries. The file itself is 2.5MB large.

jaheba:
@mr-1993 I went ahead and implemented the changes myself. Hope you don't feel overlooked!

Thanks a lot for the PR 🎉

lostella (approved):
Thanks @mr-1993 @emptymalei !

mr-1993:
@jaheba @lostella Thank you very much for the small collaboration! Hope to contribute other things as well! Also, thank you for implementing the last changes and sorry for our delay in that regard...we were blocked last week.

PR #2157, opened 2022-07-13
Add tutorial with data splitting examples

npnv:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Please tag this pr with at least one of these labels to make our release process faster:** BREAKING, new feature, bug fix, other change, dev setup

PR #2223, opened 2022-08-19
Update `gluonts.dataset.split` code, test, docs

lostella:
*Description of changes:* Follow up to #2031, #2157 with further adjustments:
- Rename `TestIterable` into `TestDataset`
- Add `input` and `label` properties to `TestDataset`
- Rename `split_date` and `split_offset`, in `DateSplitter` and `OffsetSplitter`, to `date` and `offset` respectively
- Update the module docstring and make it a (running) doctest
- Remove subpackage in favor of the module directly
- Add/fix type annotations where needed
- Change tests and documentation notebooks accordingly


src/gluonts/dataset/split.py
```diff
@@ -15,36 +15,65 @@
 Train/test splitter
 ~~~~~~~~~~~~~~~~~~~
 
+.. testsetup:: *
```
jaheba, line 18: What is this?

lostella: It's to add doctest snippets that run, but are not displayed, see: https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-testsetup

For example for imports or some other setup that is used by following doctest snippets.

jaheba (approved):

PR #2225, opened 2022-08-19
Fix bug in `to_quantile_forecast`

eugeneteoh:
*Issue #, if available:*

*Description of changes:*
Fixing some minor bugs with QuantileForecast. See changes for more details.

### DIFF
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -636,11 +636,14 @@ class SampleForecast(Forecast):
             ]
         )
 
-    def to_quantile_forecast(
-        self, quantiles: List[str]
-    ) -> "QuantileForecast":
+    def to_quantile_forecast(self, quantiles: List[str]) -> "QuantileForecast":
         return QuantileForecast(
-            forecast_arrays=np.array([self.quantile(q) if q != "mean" else self.mean() for q in quantiles]),
+            forecast_arrays=np.array(
+                [
+                    self.quantile(q) if q != "mean" else self.mean()
+                    for q in quantiles
+                ]
+            ),
             start_date=self.start_date,
             forecast_keys=quantiles,
             item_id=self.item_id,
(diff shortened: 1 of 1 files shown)


src/gluonts/model/forecast.py
```diff
@@ -637,10 +637,10 @@ def __repr__(self):
         )
 
     def to_quantile_forecast(
-        self, quantiles: List[Union[float, str]]
+        self, quantiles: List[str]
```
lostella: `self.quantile` (invoked below) accepts `Union[float, str]` so this appears to be a (breaking) restriction.

lostella (changes requested):
Hey @eugeneteoh thanks! See my comment in the code

src/gluonts/model/forecast.py
eugeneteoh: Turns out `QuantileForecast.__init__` executes `Quantile.from_str` (see L685) which only accepts str, hence this will fail with float type.

lostella: I see! Makes sense then

src/gluonts/model/forecast.py
```diff
@@ -637,10 +637,10 @@ def __repr__(self):
...
     ) -> "QuantileForecast":
         return QuantileForecast(
-            forecast_arrays=np.array([self.quantile(q) for q in quantiles]),
+            forecast_arrays=np.array([self.quantile(q) if q != "mean" else self.mean() for q in quantiles]),
```
lostella: This line is probably too long and `black` is complaining. You can run `black src` and commit the changes it applies, to fix it.

lostella (approved):

lostella (changes requested):
Just run `black src` and commit the changes, then all tests should be green 🟢

lostella (approved):

PR #2226, opened 2022-08-19
Allow DeepVAR model to use (global) dynamic features

rshyamsundar:
*Issue #, if available:*

*Description of changes:*

`DeepVAREstimator` and hence the hierarchical model currently do not utilize external dynamic features. This PR allows these models to use any external features that are common to all the targets in the multivariate data.


src/gluonts/dataset/stat.py
```diff
@@ -176,6 +176,10 @@ def calculate_dataset_statistics(ts_dataset: Any) -> DatasetStatistics:
 
             # TARGET
             target = ts[FieldName.TARGET]
+            if target.ndim == 1:
+                target_length = len(target)
+            else:
+                target_length = len(target[0])
```
lostella, line 182: You could just unconditionally use

```python
target_length = target.shape[-1]
```

mbohlkeschneider (approved):
Nice, thank you!

lostella (commented):
Looks like I forgot to publish the review…

PR #2227, opened 2022-08-22
DeepNPTS: Save and load the model using torch API

rshyamsundar:
*Issue #, if available:*

*Description of changes:*

Checking if loading and saving the model via torch API still creates strange issue during pre-commit checks.


lostella:
Looks like so. I suppose it works locally? If so, something really strange is happening in the test environment…

kashif:
also can you kindly check if `torch.load` potentially needs a `map_location`? else a model saved on the GPU will not load on CPU etc.

Issue #2221, opened 2022-08-18
TreePredictor not compatible with real dynamic features


axeng:
## Description
When using a TreePredictor estimator with a dataset containing real dynamic features, the training crashes.

## To Reproduce

```python
import pandas as pd

from gluonts.model.rotbaum import TreeEstimator
from gluonts.dataset.pandas import PandasDataset

# Generate dataset (see: https://ts.gluon.ai/stable/tutorials/data_manipulation/pandasdataframes.html#Include-static-and-dynamic-features)
def generate_single_ts(date_range, item_id=None) -> pd.DataFrame:
    """create sum of `n_f` sin/cos curves with random scale and phase."""
    n_f = 2
    period = np.array([24 / (i + 1) for i in range(n_f)]).reshape(1, n_f)
    scale = np.random.normal(1, 0.3, size=(1, n_f))
    phase = 2 * np.pi * np.random.uniform(size=(1, n_f))
    periodic_f = lambda x: scale * np.sin(np.pi * x / period + phase)

    t = np.arange(0, len(date_range)).reshape(-1, 1)
    target = periodic_f(t).sum(axis=1) + np.random.normal(0, 0.1, size=len(t))
    ts = pd.DataFrame({"target": target}, index=date_range)
    if item_id is not None:
        ts["item_id"] = item_id
    return ts

def generate_single_ts_with_features(date_range, item_id) -> pd.DataFrame:
    ts = generate_single_ts(date_range, item_id)
    T = ts.shape[0]

    ts["dynamic_real_1"] = np.random.normal(size=T)
    
    return ts

ts = generate_single_ts_with_features(pd.date_range(start="2000-01-01", freq="D", periods=50), item_id=0)
ts_dataset = PandasDataset(
    ts,
    feat_dynamic_real=["dynamic_real_1"],
)

estimator = TreeEstimator(
    freq=ts_dataset.freq,
    prediction_length=10,
    use_feat_dynamic_cat=False,
    use_feat_dynamic_real=True,
    use_feat_static_real=False,
    method="QRX",
)

predictor = estimator.train(ts_dataset)
```

## Error message or code output

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
----> 1 predictor = estimator.train(ts_dataset)

~/.venv/lib/python3.7/site-packages/gluonts/model/rotbaum/_estimator.py in train(self, training_data, validation_dataset)
     40         self, training_data: Dataset, validation_dataset=None
[...]

PR #2230, opened 2022-08-22
Rotbaum: Add support for `nan` values and long dynamic features; add "explain" function.

zoolhasson:
*Issue #, if available:*

*Description of changes:*
Added support for nan values and support for long dynamic features, and added an "explain" function for the method "QuantileRegression", which aggregates lightgbm's feature importances based on which features are associated with target/short or long dynamic features/static features. Tests are added as appropriate.


src/gluonts/model/rotbaum/_preprocess.py
```diff
@@ -429,9 +487,13 @@ def make_features(self, time_series: Dict, starting_index: int) -> List:
...
                     *[
-                        list(ent[0]) + list(ent[1].values())
+                        prefix + list(ent[0]) + list(ent[1].values())
                         for ent in [
-                            self._pre_transform(ts[starting_index:end_index])
+                            self._pre_transform(
+                                ts[starting_index:end_index],
+                                self.subtract_mean,
+                                self.count_nans,
+                            )
```
lostella: @zoolhasson this change (and the next one below) appear to be fixing #2221 (please confirm). Do you think you could split this into a separate bugfix PR? This way we can backport the change to release branches. Thanks!

PR #2235, opened 2022-08-24
testing pydantic v1.10.0a2

samuelcolvin:
### TO NOT MERGE

This is pull request is just to test pydantic v1.10, see https://github.com/pydantic/pydantic/discussions/4359


samuelcolvin:
Hi, please can you kick off tests so I can check if v1.10 will break anything.

samuelcolvin:
pydantic v1.10 drops support for python3.6 which was EOL at the end of last year, so those failures are to be expected.

samuelcolvin:
Great, this is passing except for 3.6 which is expected.

jaheba:
> pydantic v1.10 drops support for python3.6 which was EOL at the end of last year, so those failures are to be expected.

We've looked at our download-stats and realised that people still use Python 3.6 a lot, so we re-added support for it.

samuelcolvin:
:disappointed: sorry to hear that.

I'm afraid it's definitely not possible to re-add it for pydantic v1.10.

jaheba:
Not that I want to ask for it! Luckily, for now the pain is manageable.

lostella:
I guess we can still support pydantic v1.10 conditioned on python being 3.7 or newer? Not sure whether that will require ad hoc fixes here and there and cause us headache, but we can try.

samuelcolvin:
Or you can release a new version that support pydantic v1.10 and drops python 3.6 and tell users to use the old version if they're still (for some strange reason) using python 3.6.

But python 3.6 is EOL, no longer supported, toxic, run away.

lostella:
> But python 3.6 is EOL, no longer supported, toxic, run away.

You're knocking at an open door 😄

Issue #2244, opened 2022-08-25
Incremental training


lostella:
Related to #272, #557

Once a model is trained, and new data is available, users may want to continue training the model on the new data without discarding the previously trained model (that is, without randomly re-initializing the network). This can make the new training converge significantly faster.

Since training is controlled by `Estimator` classes, which are stateless, while the trained model is hold into `Predictor` objects, we need to allow 

```python
predictor = estimator.train(dataset)

[...]

new_predictor = estimator.train(dataset, network=predictor.network)
```

where `network` is a `@property` of the `predictor` object. Things to note:
* In the above, we need to make sure that the training loop does not initialize the network the second time around.
* The `predictor.network` needs to be compatible with the estimator that will be used for re-training, which may be annoying to track.
* Should the `predictor` then hold information on what `Estimator` object originated it?
* What if one wants to do the second training with different training settings? Like number of epochs, optimizer, and batch size? Then everything needs to be "compatible" in the estimator, except for these details.

jaheba:
Looking at some estimators, I would assume that most parameters should be fixed, but only the trainer would change.

I guess we could also haven an `IncrementalTrainer`?

lostella:
> I guess we could also haven an IncrementalTrainer?

What would that do?

jaheba:
Not sure, probably just making sure that the network is the same.

kashif:
on the PyTorch lightning side of things, one can train from a checkpoint model... although with more max_epochs as the checkpoint records the epochs and will not train beyond the trainer's max epoch

lostella:
> on the PyTorch lightning side of things, one can train from a checkpoint model...

Yes, good point, checkpoints can also be exploited. I still think it makes sense, in GluonTS, to have a general mechanism to start off with a given `Predictor` that was serialized somewhere. A couple of reasons:
* in case training checkpoints are gone
* to retrieve the whole data-preprocessing pipeline

That still requires the `Predictor` to know what estimator generated it, or otherwise to allow setting up a new training loop to continue training.

lostella:
Could require improvements, but done for now

mraapshockwavemedical:
@lostella could you please show me how to do this? I tried

`new_predictor = estimator.train(dataset, network=predictor.network)`

as suggested in an earlier comment. But that didn't work. Thank you!

PR #2249, opened 2022-08-25
Add incremental training

lostella:
*Issue #, if available:* related to #2244 

*Description of changes:* POC to support re-training of already-trained models. Note that for how this works internally, the structure of the network will be validated at parameter-copy-time. 

Summary of the changes:
- estimators can opt-in for the feature by implementing `train_from(predictor, ...)`
- network initialization moved out of the `Trainer` class; now it’s done by copying the parameters of the given model (if any, otherwise same initialization as before)
- `PyTorchPredictor` now gets the full `LightningModule` (so that some parameter naming headache is solved), which in turn now must implement `forward` an be `@validated()`
- add access to the input model channel in `gluonts.shell` (work in progress) as described in https://docs.aws.amazon.com/sagemaker/latest/dg/incremental-training.html

Example usage:

```python
from gluonts.mx import DeepAREstimator, Trainer

estimator = DeepAREstimator(
    freq="1H",
    prediction_length=24,
    trainer=Trainer(epochs=5),
)

print("Training the first time...")

# get some new data...
# ok, not really

print("Training the second time...")
other_predictor = estimator.train_from(predictor, dataset.train)
```


src/gluonts/model/estimator.py
```diff
@@ -40,7 +40,10 @@ def __init__(self, lead_time: int = 0, **kwargs) -> None:
         self.lead_time = lead_time
 
     def train(
-        self, training_data: Dataset, validation_data: Optional[Dataset] = None
+        self,
+        training_data: Dataset,
+        validation_data: Optional[Dataset] = None,
+        model_init: Optional[Predictor] = None,
```
jaheba: I'm not sure what the right name would be. If we pass a `Predictor`, then `model` seems "wrong".

jaheba: Also, can/should we mark this somehow as experimental?

src/gluonts/mx/model/n_beats/_ensemble.py
```diff
@@ -439,14 +439,33 @@ def from_hyperparameters(
...
+        if model_init is None:
+            for index, estimator in enumerate(self.estimators):
+                logging.info(
+                    f"Training estimator {index + 1}/{len(self.estimators)}."
+                )
+                predictors.append(
+                    estimator.train(training_data, validation_data)
+                )
+        else:
+            assert isinstance(model_init, NBEATSEnsemblePredictor)
```
jaheba: I think this is weird type-wise. Each `Estimator` class has exactly one companion predictor-class. I guess we would want something like Rust's [Associated types](https://doc.rust-lang.org/rust-by-example/generics/assoc_items/types.html#associated-types)

src/gluonts/shell/env.py
```diff
@@ -29,18 +30,27 @@ def __init__(self, *args, **kwargs):
...
         if "metadata" in self.channels:
             path = self.channels.pop("metadata")
             self.hyperparameters["freq"] = MetaData.parse_file(
                 path / "metadata.json"
             ).freq
 
+        model: Optional[Predictor] = None
+        if "model" in self.channels:
+            path = self.channels.pop("model")
+            with tarfile.open(path / "model.tar.gz") as targz:
```
jaheba, line 43: Have you tried running this?

But I guess we need to extract this ourselves, since SageMaker doesn't do it for us this time?

lostella: I did, needed a couple of attempts :-) but this works

src/gluonts/model/estimator.py
lostella: You mean in the documentation, or somehow raise a warning?

lostella: > If we pass a Predictor, then model seems "wrong"

Yes, this could be `initial_predictor` maybe

jaheba: How about something like `from_predictor`, or just `predictor`?

src/gluonts/mx/model/n_beats/_ensemble.py
lostella: This would just be a class attribute in this context, right?

src/gluonts/model/estimator.py
```diff
@@ -40,7 +40,10 @@ def __init__(self, lead_time: int = 0, **kwargs) -> None:
         self.lead_time = lead_time
 
     def train(
-        self, training_data: Dataset, validation_data: Optional[Dataset] = None
+        self,
+        training_data: Dataset,
+        validation_data: Optional[Dataset] = None,
+        from_predictor: Optional[Predictor] = None,
```
jaheba: Is this a breaking feature, because I have to accept the `from_predictor` argument in subclasses?

Maybe a second method would be better than, so models can opt in to implement it?

src/gluonts/model/estimator.py
```diff
@@ -59,6 +59,34 @@ def train(
         """
         raise NotImplementedError
 
+    def train_from(
```
jaheba, line 83: Should we make this a `Protocol`?

lostella:
This results from removing the trailing `**kwargs` from `train`. However, it's unused: should we just update the tests and mark this as breaking? Maybe in a separate PR

```
2022-09-13T11:22:58.7320002Z >       predictor = estimator.train(dataset_train, num_workers=None)
2022-09-13T11:22:58.7320252Z E       TypeError: train() got an unexpected keyword argument 'num_workers'
```

jaheba:
> This results from removing the trailing `**kwargs` from `train`. However, it's unused: should we just update the tests and mark this as breaking? Maybe in a separate PR
> 
> ```
> 2022-09-13T11:22:58.7320002Z >       predictor = estimator.train(dataset_train, num_workers=None)
> 2022-09-13T11:22:58.7320252Z E       TypeError: train() got an unexpected keyword argument 'num_workers'
> ```

I think so. Probably also shows that we should disentangle data loading more.

src/gluonts/mx/model/estimator.py
```diff
@@ -39,7 +40,7 @@ class TrainOutput(NamedTuple):
     predictor: Predictor
 
 
-class GluonEstimator(Estimator):
+class GluonEstimator(Estimator, IncrementallyTrainable):
```
jaheba: I don't think we want to make this dependency explicit. I remember we had issues with the `PandasDataset` deriving from `Dataset`.

src/gluonts/model/estimator.py
```diff
@@ -77,6 +78,36 @@ def from_inputs(cls, train_iter, **params):
         return cls.from_hyperparameters(**params)
 
 
+class IncrementallyTrainable(Protocol):
```
jaheba, line 82: We could derive from `Estimator` here.

src/gluonts/shell/train.py
```diff
@@ -99,6 +101,17 @@ def run_train(
         else None
     )
 
+    if from_predictor is not None:
```
jaheba, line 104: We would need to check that `isinstance(forecaster, IncrementallyTrainable)`

src/gluonts/model/estimator.py
jaheba, line 82: I think we also want to mark this as `@runtime_checkable`.

src/gluonts/mx/model/n_beats/_ensemble.py
```diff
@@ -237,7 +236,7 @@ def __eq__(self, that):
         return True
 
 
-class NBEATSEnsembleEstimator(Estimator):
+class NBEATSEnsembleEstimator(Estimator, IncrementallyTrainable):
```
jaheba:
```suggestion
class NBEATSEnsembleEstimator(Estimator):
```

jaheba (changes requested):

jaheba (dismissed):

jaheba (approved):

PR #2250, opened 2022-08-26
update SplicedBinnedPareto demos from nursery version to gluonts version

elenaehrlich:
*Issue #, if available:* 1410

*Description of changes:* Updated the SBP example notebooks from the nursery version to the gluonts version. Also, included a small fix to ensure the tail_percentile_gen_pareto to be on same device as logits. 


elenaehrlich:
Demo notebook for gluonts-compatible SplicedBinnedPareto now added

lostella (changes requested):
@elenaehrlich looks good to me! Only one thing: maybe it is better to commit the notebooks without the cells outputs (so clearing all outputs before committing), would that be ok?

elenaehrlich:
> 

Thank you Lorenzo, I've cleared the notebooks' outputs and re-pushed

lostella (approved):
thanks @elenaehrlich! 🚀

PR #2259, opened 2022-08-30
Optimize `TimeSeriesSlice` performance

lostella:
*Description of changes:* the `TimeSeriesSlice` class used in `gluonts.dataset.split` tools adds a lot of overhead due to pandas conversions. This PR gets rid of most of the overhead by operating directly on `numpy` arrays.

To test the improvement one can use the following snippet, iterating `N` times over the `m4_hourly` dataset and splits obtained from it:

```python
import time
from contextlib import AbstractContextManager

dataset = get_dataset("m4_hourly").train
# dataset = list(get_dataset("m4_hourly").train)

class Timer(AbstractContextManager):
    def __init__(self, name: str):
        self.name = name

    def __enter__(self):
        self.start = time.time()

    def __exit__(self, *args):
        end = time.time()
        print(f"{self.name}: {end - self.start} seconds")

N = 100

with Timer("dataset    "):
    for k in range(N):
        for entry in dataset:
            pass

training_data, test_template = split(dataset, offset=-100)

with Timer("training   "):
    for k in range(N):
        for entry in training_data:
            pass

test_pairs = test_template.generate_instances(prediction_length=24)

with Timer("test pairs "):
    for k in range(N):
        for entry in test_pairs:
            pass
```

Before the fix:

```
dataset    : 3.6620118618011475 seconds
training   : 13.0590181350708 seconds
test pairs : 22.129828214645386 seconds
```

After the fix:

```
dataset    : 3.6478400230407715 seconds
training   : 3.8759350776672363 seconds
test pairs : 4.291091203689575 seconds
```

### DIFF
--- a/src/gluonts/dataset/split.py
+++ b/src/gluonts/dataset/split.py
@@ -29,14 +29,14 @@ This module defines strategies to split a whole dataset into train and test
 subsets. The :func:`split` function can also be used to trigger their logic.
 
 For uniform datasets, where all time-series start and end at the same point in
-time :class:`OffsetSplitter` can be used::
+time :class:`OffsetSplitter` can be used:
 
 .. testcode::
 
     splitter = OffsetSplitter(offset=7)
     train, test_template = splitter.split(whole_dataset)
 
-For all other datasets, the more flexible :class:`DateSplitter` can be used::
+For all other datasets, the more flexible :class:`DateSplitter` can be used:
 
 .. testcode::
 
@@ -47,7 +47,7 @@ For all other datasets, the more flexible :class:`DateSplitter` can be used::
 
 In the above examples, the ``train`` output is a regular ``Dataset`` that can
 be used for training purposes; ``test_template`` can generate test instances
-as follows::
+as follows:
 
 .. testcode::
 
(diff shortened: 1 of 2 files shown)


src/gluonts/dataset/split.py
```diff
@@ -73,130 +73,87 @@
...
-    class Config:
-        arbitrary_types_allowed = True
-
-    target: pd.Series
-    item_id: Optional[str] = None
+def slice_data_entry(
+    entry: DataEntry, slc: slice, prediction_length: int = 0
+) -> DataEntry:
+    if slc.stop is not None:
+        slc_extended = slice(slc.start, slc.stop + prediction_length, slc.step)
```
lostella: The sum here could have different sign compared to `slc.stop` (if `slc.stop` is negative); need to make this more robust

src/gluonts/dataset/split.py
```diff
@@ -73,130 +73,122 @@
...
 from gluonts.dataset import Dataset, DataEntry
 from gluonts.dataset.field_names import FieldName
 
 
-class TimeSeriesSlice(pydantic.BaseModel):
+def to_positive_slice(slc: slice, length: int) -> slice:
     """
-    Like DataEntry, but all time-related fields are of type pd.Series and is
-    indexable, e.g `ts_slice['2018':]`.
+    Returns an equivalent slice with positive bounds, given the
```
jaheba: A bit pedantic, but this should be `Return` instead of `Returns`.

src/gluonts/dataset/split.py
```diff
@@ -73,130 +73,122 @@
...
-import numpy as np
 import pandas as pd
-import pydantic
 
 from gluonts.dataset import Dataset, DataEntry
 from gluonts.dataset.field_names import FieldName
 
 
-class TimeSeriesSlice(pydantic.BaseModel):
+def to_positive_slice(slc: slice, length: int) -> slice:
```
jaheba: Not sure I like `slc`, can't we do `slice_`?

src/gluonts/dataset/split.py
```diff
@@ -73,130 +73,122 @@
...
 
-    feat_dynamic_cat: List[pd.Series] = []
-    feat_dynamic_real: List[pd.Series] = []
+    if isinstance(slc.start, pd.Period):
+        start_offset = (slc.start - start).n
+        assert start_offset >= 0
+    elif start_is_int:
+        start_offset = slc.start
+    else:
+        raise ValueError()
```
jaheba: Can we get some error message?

src/gluonts/dataset/split.py
```diff
@@ -73,130 +73,122 @@
...
+    )
 
-        feat_static_cat = list(item.get("feat_static_cat", []))
+    if slc.stop is not None:
+        slc_extended = slice(slc.start, slc.stop + prediction_length, slc.step)
+    else:
+        slc_extended = slc
 
-        feat_static_real = list(item.get("feat_static_real", []))
+    sliced_entry = entry.copy()
```
jaheba: I like this notion better

```suggestion
    sliced_entry = dict(entry)
```

jaheba (dismissed):

huibinshen (dismissed):

huibinshen (approved):

PR #2262, opened 2022-08-30
Max idle transform fix

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/mx/model/estimator.py
+++ b/src/gluonts/mx/model/estimator.py
@@ -175,7 +175,7 @@ class GluonEstimator(Estimator):
 
         transformation = self.create_transformation()
 
-        with env._let(max(max_idle_transforms=len(training_data), 100)):
+        with env._let(max_idle_transforms=max(len(training_data), 100)):
             transformed_training_data = transformation.apply(training_data)
             if cache_data:
                 transformed_training_data = Cached(transformed_training_data)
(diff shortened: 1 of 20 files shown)


jaheba:
Looking at the code, there is quite a lot of duplication.

lostella (approved):
LGTM, thanks! 🚀

kashif:
do we need to update the pytorch estimators too?

lostella:
> do we need to update the pytorch estimators too?

Probably yes 👍 not the same fix as here, but definitely setting the max idle transform in the base estimator class

src/gluonts/mx/model/estimator.py
```diff
@@ -170,27 +172,34 @@ def train_model(
...
         cache_data: bool = False,
     ) -> TrainOutput:
+
         transformation = self.create_transformation()
 
-        transformed_training_data = transformation.apply(training_data)
+        with env._let(max_idle_transforms=max(len(training_data), 100)):
+            transformed_training_data = transformation.apply(training_data)
+            if cache_data:
+                transformed_training_data = Cached(transformed_training_data)
```
lostella, line 181: Here we are setting the `max_idle_transforms` when applying the transformation to the data.

src/gluonts/transform/_base.py
```diff
@@ -169,7 +169,7 @@ class FlatMapTransformation(Transformation):
 
     @validated()
     def __init__(self):
-        self.max_idle_transforms = max(env.max_idle_transforms, 100)
+        self.max_idle_transforms = env.max_idle_transforms
```
lostella, line 172: However, here we are reading the value at construction time.

lostella (commented):
~~@jaheba are we sure this is correct?~~ nevermind, it's correct: the instance splitter transformation gets constructed with the data loader

PR #2263, opened 2022-08-31
Fix ignore hidden files when generating datasets

kashif:
the data sets contain ._ json mac osx specific hidden files which get read when regenrating datasets e.g. "taxi_30min"

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/dataset/common.py
+++ b/src/gluonts/dataset/common.py
@@ -156,7 +156,7 @@ def FileDataset(
     use_timestamp: bool = False,
     loader_class=None,
     ignore=False,
-    pattern="[!._]*",
+    pattern="*",
     levels=1,
 ) -> Dataset:
     path = Path(path)
--- a/src/gluonts/dataset/common.py
+++ b/src/gluonts/dataset/common.py
@@ -156,7 +156,7 @@ def FileDataset(
     use_timestamp: bool = False,
     loader_class=None,
     ignore=False,
-    pattern="*",
+    pattern="[!._]*",
     levels=1,
 ) -> Dataset:
     path = Path(path)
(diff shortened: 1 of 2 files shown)


src/gluonts/dataset/repository/_gp_copula_2019.py
```diff
@@ -161,4 +161,4 @@ def get_data(dataset_path: Path, ds_info: GPCopulaDataset):
 
 def clean_up_dataset(dataset_path: Path, ds_info: GPCopulaDataset):
     os.remove(dataset_path.parent / f"{ds_info.name}.tar.gz")
-    shutil.rmtree(dataset_path / "metadata")
+    shutil.rmtree(dataset_path / "metadata", ignore_errors=True)
```
lostella, line 164: Is this also needed?

kashif: yes was getting an error with get_dataset("taxi_30min", regenerate=True) without it... somehow the metadata/ folder gets removed before... I didn't figure it out...

src/gluonts/dataset/common.py
```diff
@@ -156,7 +156,7 @@ def FileDataset(
     use_timestamp: bool = False,
     loader_class=None,
     ignore=False,
-    pattern="*",
+    pattern="[!._]*",
```
lostella: Instead of making this the global default, we could set this option in line 157 of `_gp_copula_2019.py` maybe?

lostella (changes requested):

src/gluonts/dataset/common.py
kashif: ok yes sure

lostella (approved):
Thanks @kashif!

Issue #2268, opened 2022-09-04
QuantileForecast.plot() is broken


abdulfatir:
## Description
The specialized `plot()` function in `QuantileForecast` doesn't work anymore, possibly since the type of `Forecast.index` was changed to `pd.PeriodIndex` from `pd.DateTimeIndex`.

## To Reproduce

```python
import numpy as np
import pandas as pd
from gluonts.model.forecast import QuantileForecast

qf = QuantileForecast(
    forecast_arrays=np.random.randn(3, 10),
    forecast_keys=["0.1", "0.5", "0.9"],
    start_date=pd.Period("01-01-2019", freq="5min"),
    item_id=None,
)
qf.plot()
```

## Error message or code output

```
TypeError                                 Traceback (most recent call last)
<ipython-input-1-0861b3a04566> in <module>
      9     item_id=None,
     10 )
---> 11 qf.plot()

~/miniconda3/envs/rcf/lib/python3.6/site-packages/gluonts/model/forecast.py in plot(self, label, output_file, keys, *args, **kwargs)
    782                 label=f"{label_prefix}q{k}",
    783                 *args,
--> 784                 **kwargs,
    785             )
    786             plt.legend()

~/miniconda3/envs/rcf/lib/python3.6/site-packages/matplotlib/pyplot.py in plot(scalex, scaley, data, *args, **kwargs)
   2840     return gca().plot(
   2841         *args, scalex=scalex, scaley=scaley,
-> 2842         **({"data": data} if data is not None else {}), **kwargs)
   2843 
   2844 

~/miniconda3/envs/rcf/lib/python3.6/site-packages/matplotlib/axes/_axes.py in plot(self, scalex, scaley, data, *args, **kwargs)
   1743         lines = [*self._get_lines(*args, data=data, **kwargs)]
   1744         for line in lines:
-> 1745             self.add_line(line)
...
---> 83     return array(a, dtype, copy=False, order=order)
     84 
     85 

TypeError: float() argument must be a string or a number, not 'Period'
```

## Environment
- Operating system: MacOS 12.5.1
- Python version: 3.6.10
- GluonTS version: 0.10.3
- MXNet version: 1.9.1

lostella:
@abdulfatir does it also happen in the `dev` branch? Also, does turning the start date to `Timestamp` fix it?

lostella:
[...]

PR #2269, opened 2022-09-05
Fix `QuantileForecast.plot()` to use `DateTimeIndex`

abdulfatir:
*Issue #, if available:* #2268

*Description of changes:* Changed plotting function from `plt.plot()` to `Series.plot()` to be consistent with the superclass and index to `DateTimeIndex`.

### DIFF
--- a/requirements/requirements-test.txt
+++ b/requirements/requirements-test.txt
@@ -7,3 +7,4 @@ ujson
 orjson
 requests
 holidays~=0.9
+matplotlib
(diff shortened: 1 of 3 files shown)


lostella:
Thanks @abdulfatir! Would it make sense to add a test for this? There are some tests in https://github.com/awslabs/gluon-ts/blob/dev/test/model/test_forecast.py#L30-L40, I'm assuming that trying to plot the `QuantileForecast` given there would fail prior to this fix?

lostella (dismissed):
🚀

lostella (approved):
🚀 🚀

PR #2286, opened 2022-09-09
ForecastBatch POC

lostella:
*Description of changes:*
The main idea is to keep forecast data (samples, distribution parameters, quantiles) in batches, and expose all relevant methods for evaluation at the batch level. So we can compute statistics (mean, quantiles) over batches and evaluation metrics also over batches, speeding up evaluation.

This way the `ForecastGenerator` classes become obsolete: the predictor just uses the `network.forecast` over inference batches, and "flattening" a sequence of batches is handled by the `ForecastBatch`. Predictors could have a `predict_batches` method, useful in case forecast batches can be processed more efficiently downstream, and the usual `predict` method that yields single forecasts, for plotting or anything.

Pros:
* The model (intended as the network) is explicit at declaring what it expects as input batch, and how its output is to be interpreted (wrapped into some structured forecast object). This removes some focus from the estimator, which is currently overly complex, and make the network usable on its own (once one has batches).
* `ForecastGenerator` classes will become obsolete, in favor of the arguably simpler `ForecastBatch`
* Batches of forecasts are not deconstructed until strictly required

Cons:
* Some models don't operate on batches internally, so in those cases the `predict_batch` method would either be absent, or require that we can stack together multiple forecasts in a single batch (not done in this PR)


src/gluonts/model/forecast_generator.py
```diff
@@ -241,3 +262,114 @@ def __call__(
                     info=batch["info"][i] if "info" in batch else None,
                 )
             assert i + 1 == len(batch[FieldName.FORECAST_START])
+
+
+class ForecastBatch:
+    @property
+    def batch_size(self) -> int:
+        raise NotImplementedError
```
jaheba: I'm not sure, but maybe declaring `batch_size: int` is enough here.

src/gluonts/model/forecast_generator.py
```diff
@@ -241,3 +262,114 @@ def __call__(
...
+@dataclass
+class DistributionForecastBatch(ForecastBatch):
+    start_date: list
+    item_id: Optional[list]
+    info: Optional[list]
+    distr_output: Any  # TODO fix
+    distr_args: list
+
+    def __post_init__(self):
+        self.distr = self.distr_output.distribution(*self.distr_args)
```
jaheba, line 101: You would normally include `distr` in the declarations, but use `distr: Type = field(init=False)`

src/gluonts/model/forecast_generator.py
```diff
@@ -241,3 +262,114 @@ def __call__(
...
+
+@dataclass
+class SampleForecastBatch(ForecastBatch):
+    start_date: list
+    item_id: Optional[list]
+    info: Optional[list]
+    sample_batch: np.ndarray
+
+    def __post_init__(self):
+        self._sorted_samples_value = None
```
jaheba: Why not just directly assign sorted samples here? Do we require it to be lazy?

src/gluonts/model/forecast_generator.py
```diff
@@ -241,3 +262,114 @@ def __call__(
...
+    @property
+    def mean(self) -> np.ndarray:
+        if self._mean is not None:
+            return self._mean
+        else:
+            return np.mean(self.samples, axis=1)
+
+    def quantile(self, q: Union[float, str]) -> np.ndarray:
+        q = Quantile.parse(q).value
+        sample_idx = int(np.round((self.num_samples - 1) * q))
```
jaheba: No interpolation?

src/gluonts/model/forecast_generator.py
lostella: It may be premature optimization, but as long as you don't need to access quantiles, there's no need for sorting

src/gluonts/model/forecast_generator.py
lostella: I’ll make this feature-complete eventually :-)

src/gluonts/model/forecast_generator.py
```diff
@@ -241,3 +262,114 @@ def __call__(
...
+
+@dataclass
+class QuantileForecastBatch(ForecastBatch):
+    start_date: list
+    item_id: Optional[list]
+    info: Optional[list]
+    quantile_batch: np.ndarray
+    quantile_levels: List[str]
+
+    def __iter__(self) -> Iterator[SampleForecast]:
```
codingWhale13: The return type should be`Iterator[QuantileForecast]`, right?

src/gluonts/model/forecast_generator.py
```diff
@@ -102,142 +120,113 @@ def make_distribution_forecast(distr, *args, **kwargs) -> Forecast:
...
+    def mean(self) -> np.ndarray:
+        return to_numpy(self.distr.mean())
+
+    def quantile(self, q: Union[float, str]) -> np.ndarray:
+        return to_numpy(self.distr.quantile(q))
+
+
+@dataclass
+class SampleForecastBatch(ForecastBatch):
+    start_date: list
```
jaheba:
```suggestion
    start: list
```

or maybe `starts`?

And what is the type -- `pd.Period`?

src/gluonts/model/forecast_generator.py
```diff
@@ -102,142 +120,113 @@ def make_distribution_forecast(distr, *args, **kwargs) -> Forecast:
...
+    item_id: Optional[list]
+    info: Optional[list]
```
jaheba: This is weird, can we merge this into one field, which stores dicts?

(I know not related to this PR)

src/gluonts/model/forecast_generator.py
```diff
@@ -102,142 +120,113 @@ def make_distribution_forecast(distr, *args, **kwargs) -> Forecast:
...
+    sample_batch: np.ndarray
```
jaheba: Maybe `samples` or just `data`?

src/gluonts/model/forecast_generator.py
```diff
@@ -102,142 +120,113 @@ def make_distribution_forecast(distr, *args, **kwargs) -> Forecast:
...
+        self._sorted_samples_value = None
+
+    @property
+    def _sorted_samples(self) -> np.ndarray:
+        if self._sorted_samples_value is None:
+            self._sorted_samples_value = np.sort(self.sample_batch, axis=1)
+        return self._sorted_samples_value
+
+    def __iter__(self) -> Iterator[SampleForecast]:
+        for i, sample in enumerate(self.sample_batch):
```
jaheba: This looks like a use for `zip`? It's col-to-row again?

src/gluonts/model/forecast_generator.py
```diff
@@ -52,6 +55,21 @@ def predict_to_numpy(prediction_net, args) -> np.ndarray:
     raise NotImplementedError
 
 
+@singledispatch
+def to_numpy(object):
```
jaheba: Is this actually doing something?

lostella: It was, yes

jaheba: I don't see what

lostella:
There's some fundamental issue with this change, related to `SymbolBlockPredictor`, which breaks things there.

I think I will revert many of the changes to MXNet-based models, and keep them for PyTorch-based ones only.

lostella:
The initial proposal for this change was to have a `forecast` method added to the model classes. However, extending the API surface of these objects is tricky (pretty much regardless of the framework), because when models are compiled in any way you won't get the added methods back when deserializing. This was the reason in the first place to have some external class (`Predictor`) handle all of the pre-post processing around the model.

I still like the change because it simplifies over the `ForecastGenerator` approach, but this will need to go into the `Predictor` anyway I think.

PR #2287, opened 2022-09-09
Added LR scheduler to DeepAR Pytorch

shubhamkapoor:
*Issue #, if available:*

*Description of changes:*


src/gluonts/torch/model/estimator.py
```diff
@@ -23,6 +23,7 @@
 from gluonts.env import env
 from gluonts.itertools import Cached
 from gluonts.model.estimator import Estimator
+from gluonts.nursery.utils import matching_parameters
```
lostella: Better not to have imports from the nursery, maybe you could have this util elsewhere in the package?

src/gluonts/nursery/utils.py
```diff
@@ -0,0 +1,6 @@
+from inspect import signature
+from typing import Callable, Dict, Any
+
+
+def matching_parameters(f: Callable, hps: Dict[str, Any]) -> Dict[str, Any]:
```
lostella: Maybe “arguments” is better than “parameters”

lostella: Could you add a docstring here?

shubhamkapoor: Done.

src/gluonts/torch/model/estimator.py
shubhamkapoor: Done.

src/gluonts/core/utils.py
```diff
@@ -0,0 +1,35 @@
...
+from typing import Callable, Dict, Any
+
+
+def matching_arguments(f: Callable, args: Dict[str, Any]) -> Dict[str, Any]:
+    """
+    Extract valid keyword arguments for a Callable from a dictionary.
+
+    Parameters
+    ----------
+    f
```
jaheba: I generally prefer `fn` as name for a function.

src/gluonts/torch/model/estimator.py
```diff
@@ -197,6 +198,8 @@ def train_model(
         custom_callbacks = self.trainer_kwargs.get("callbacks", [])
         callbacks = [checkpoint] + custom_callbacks
         trainer_kwargs = {**self.trainer_kwargs, "callbacks": callbacks}
+
+        trainer_kwargs = matching_arguments(pl.Trainer, trainer_kwargs)
```
lostella: I'm wondering: is this desirable? Someone providing the wrong `kwargs` may not realize they are filtered out, while they should error

lostella: oh I see what's going on: `patience` needs to be filtered out... maybe then `patience` doesn't really belong to `trainer_kwargs`

src/gluonts/core/utils.py
shubhamkapoor: Done

src/gluonts/torch/model/estimator.py
shubhamkapoor: Since LR scheduling is a part of model training, ideally, it should be an option for `Trainer` function. The interface to interact with training functionality in Pytorch estimators is via `trainer_kwargs`, so I put the option of setting `patience` here.

shubhamkapoor: Also, I just realized that Pytorch `DeepAREstimator` doesn't provide any interface to set learning rate explicitly by the user. So, maybe we can expose `lr` in `trainer_kwargs` as well or something else. Any suggestions are welcome.

lostella: Sure, however this contrasts a bit with what the `DeepAREstimator` docstring says:

```
trainer_kwargs
        Additional arguments to provide to ``pl.Trainer`` for construction.
```

In fact, there we put other training-related arguments like batch size as separate. I think the docstring above should really say `Arguments to pl.Trainer for construction` and be in this class. And then the role of `trainer_kwargs` is clear and there should be no special cases.

So, I vote for making `patience` an explicit argument to the constructor :-)

src/gluonts/torch/model/deepar/lightning_module.py
```diff
@@ -47,13 +49,15 @@ def __init__(
         loss: DistributionLoss = NegativeLogLikelihood(),
         lr: float = 1e-3,
         weight_decay: float = 1e-8,
+        patience: Optional[int] = None,
```
lostella: Maybe you can set the same good default as in the estimator here?

lostella (changes requested):

src/gluonts/torch/model/estimator.py
lostella: > So, maybe we can expose `lr` in `trainer_kwargs` as well or something else. Any suggestions are welcome.

Good spot! Same as for patience, I would keep it a separate argument, and let trainer_kwargs feed directly into pl.Trainer.

However I would address the `lr` issue in a separate PR since its technically a bug fix that we would want to backport.

src/gluonts/torch/model/deepar/lightning_module.py
shubhamkapoor: 👍

shubhamkapoor:
I removed the `matching_arguments` util as we don't need it for this PR.

lostella (dismissed):

lostella (approved):
🚀

shubhamkapoor:
The metrics in the table below are **mean_wQuantileLoss** for different DeepAR configurations calculated over 5 runs. 

Dataset | DeepAR_mxnet default args | DeepAR_torch default args | DeepAR_torch w/ LR scheduler
-- | -- | -- | --
  |   |   |  
electricity | 0.055±0.003 | 0.051±0.003 | 0.051±0.004
exchange_rate | 0.008±0.001 | 0.016±0.004 | 0.01±0.002
m4_daily | 0.027±0.005 | 0.026±0.004 | 0.025±0.001
m4_hourly | 0.037±0.004 | 0.053±0.019 | 0.046±0.021
m4_monthly | 0.114±0.007 | 0.117±0.008 | 0.11±0.005
m4_quarterly | 0.085±0.001 | 0.083±0.003 | 0.084±0.003
m4_weekly | 0.048±0.003 | 0.05±0.004 | 0.047±0.001
m4_yearly | 0.113±0.002 | 0.114±0.003 | 0.124±0.006
solar-energy | 0.365±0.006 | 0.379±0.005 | 0.384±0.004
traffic | 0.122±0.002 | 0.12±0.002 | 0.119±0.001

PR #2289, opened 2022-09-09
Add learning rate and weight decay arguments to PyTorch estimators

### LINKED ISSUE #2288: Learning rate setting not available for Pytorch models
Hello everyone,

I hesitated to tag this issue as a "bug" because I think it's a simple design problem.

Indeed, unlike the **MXNet** equivalent, in the **Pytorch** estimators, the trainer does not carry the `learning_rate` information. 
It's defined in LightningModule, but this one is not instantiated before training: 

```
def train_model(
    self,
    training_data: Dataset,
    num_workers: int = 0,
    shuffle_buffer_length: Optional[int] = None,
    cache_data: bool = False,
    ckpt_path: Optional[str] = None,
    **kwargs,
) -> TrainOutput:transformation = self.create_transformation()

    transformed_training_data = transformation.apply(
        training_data, is_train=True
    )
    
    training_network = self.create_lightning_module()  <== here
    
    ...
    
    trainer.fit(
            model=training_network,
            train_dataloaders=training_data_loader,
            val_dataloaders=validation_data_loader,
            ckpt_path=ckpt_path,
        )
    
    logger.info(f"Loading best model from {checkpoint.best_model_path}")
    best_model = training_network.load_from_checkpoint(
        checkpoint.best_model_path
    )
    
    return TrainOutput(
        transformation=transformation,
        trained_net=best_model,
        trainer=trainer,
        predictor=self.create_predictor(transformation, best_model),
    )
```

So I have the impression that it's not possible to set the learning rate at the estimator initialization, nor to specify it later during the `train` call.

Am I missing something?

Thanks in advance !

shubhamkapoor:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/torch/model/deepar/estimator.py
+++ b/src/gluonts/torch/model/deepar/estimator.py
@@ -95,6 +95,10 @@ class DeepAREstimator(PyTorchLightningEstimator):
         Number of RNN layers (default: 2).
     hidden_size
         Number of RNN cells for each layer (default: 40).
+    lr
+        Learning rate (default: ``1e-3``).
+    weight_decay
+        Weight decay regularization parameter (default: ``1e-8``).
     dropout_rate
         Dropout regularization parameter (default: 0.1).
     num_feat_dynamic_real
@@ -149,6 +153,8 @@ class DeepAREstimator(PyTorchLightningEstimator):
         context_length: Optional[int] = None,
         num_layers: int = 2,
         hidden_size: int = 40,
+        lr: float = 1e-3,
+        weight_decay: float = 1e-8,
         dropout_rate: float = 0.1,
         num_feat_dynamic_real: int = 0,
         num_feat_static_cat: int = 0,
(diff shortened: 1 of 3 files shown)


lostella (approved):
Looks good to me, thanks @shubhamkapoor 👍

src/gluonts/torch/model/deepar/estimator.py
```diff
@@ -95,6 +95,10 @@ class DeepAREstimator(PyTorchLightningEstimator):
         Number of RNN layers (default: 2).
     hidden_size
         Number of RNN cells for each layer (default: 40).
+    lr
```
jaheba, line 98: Can we use `learning_rate` to be consistent with our `mx` version?

jaheba (changes requested):
Let's check consistency before merging.

lostella:
> Let's check consistency before merging.

The thing is, `lr` is consistent with the PyTorch ecosystem (see https://pytorch.org/docs/stable/optim.html) which would also be desirable in my opinion

src/gluonts/torch/model/deepar/estimator.py
shubhamkapoor: `mx` version doesn't use `learning_rate` directly in estimator, but rather in `mx` trainer. So as @lostella mentioned, I think `lr` is consistent with `torch` trainer/optimizer. And to make `mx` and `torch` interfaces consistent, we can use `lr` in `mx` trainer as well.

jaheba (approved):
I guess we have to die at least one death :)

PR #2294, opened 2022-09-12
Rework dev-setup.

jaheba:
* `.devtools` becomes now part of `gluonts.meta.qa`
* Remove doc devsetup it, since it is outdated
* Remove checks from setup.py


docs/community/devsetup.md
```diff
@@ -1,121 +0,0 @@
-# Development Setup
```
lostella, line 1: CONTRIBUTING.md points to this: https://github.com/awslabs/gluonts/blob/dev/CONTRIBUTING.md

maybe we leave this here and update it?

jaheba: Do you mean also remove it from there?

lostella: Either that, or we just keep the page around and update it with current instructions

PR #2301, opened 2022-09-14
Initial TorchServe support for PyTorch-based models.

jgasthaus:
*Description of changes:*

Add support for exporting `PyTorchPredictor`s into TorchScript and enable serving such models using TorchServe. 

This is rudimentary at this point, and we may want to move the handler directory to the nursery for the time being.

### DIFF
--- a/src/gluonts/torch/model/predictor.py
+++ b/src/gluonts/torch/model/predictor.py
@@ -102,15 +102,20 @@ class PyTorchPredictor(Predictor):
             that.prediction_net.state_dict(),
         )
 
-    def serialize(self, path: Path) -> None:
+    def serialize(self, path: Path, use_torchscript=False) -> None:
         super().serialize(path)
 
         # serialize network
-        with (path / "prediction_net.json").open("w") as fp:
-            print(dump_json(self.prediction_net), file=fp)
-        torch.save(
-            self.prediction_net.state_dict(), path / "prediction_net_state"
-        )
+        if use_torchscript:
+            self.prediction_net.to_torchscript(
+                file_path=path / "prediction_net.pt", method="trace"
+            )
+        else:
+            with (path / "prediction_net.json").open("w") as fp:
+                print(dump_json(self.prediction_net), file=fp)
+            torch.save(
+                self.prediction_net.state_dict(), path / "prediction_net_state"
+            )
 
         # serialize transformation chain
         with (path / "input_transform.json").open("w") as fp:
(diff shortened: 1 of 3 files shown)


src/gluonts/evaluation/metrics.py
```diff
@@ -87,7 +87,7 @@ def coverage(target: np.ndarray, forecast: np.ndarray) -> float:
 
         coverage = mean(Y < \hat{Y})
     """
-    return np.mean(target < forecast)
+    return float(np.mean(target < forecast))
```
lostella: Just wondering: is this related?

jgasthaus: This was somehow necessary to make the mypy check happy. No idea why -- probably some change in numpy?

Let me take it out of this PR for now.

src/gluonts/torch/model/predictor.py
```diff
@@ -102,15 +102,37 @@ def __eq__(self, that):
             that.prediction_net.state_dict(),
         )
 
-    def serialize(self, path: Path) -> None:
+    def trace(self):
+        dummy_dataset = ListDataset(
```
lostella: A slightly better way (I’m not a fan of `ListDataset` and would like to get rid of it at some point) is to use a list directly:
```
dummy_dataset = [dict(start=pd.Period(…, freq=…), target=np.zeros(1))]
```
The frequency being arbitrary here looks weird, but should not be an issue, since we’re only tracing the network and it should only use frequency information it’s configured with. (Also I don’t see how else to do it.)

src/gluonts/torch/serve/handler.py
```diff
@@ -0,0 +1,56 @@
...
+
+        model_dir = properties.get("model_dir")
+        self.predictor = PyTorchPredictor.deserialize(Path(model_dir))
+        self.model = self.predictor.prediction_net
+        self.model.to(self.device)
+        self.model.eval()
+        self.initialized = True
+
+    def inference(self, data, *args, **kwargs):
+        # FIXME: Get freq from model
```
lostella, line 47: Frequency is not there (in the predictor) anymore, see #2017, so we should find another way. I think the reasoning behind the change is that a model may be “agnostic” to data frequency, so requiring it being an attribute of the model is weird.

It’s really a property of the data, so maybe we should find a way to come from there. Or couldn’t the handler be constructed with it? I’m not sure how the handler is used though…

src/gluonts/torch/model/predictor.py
jgasthaus: I have removed this entire dummy_input code now. The PR #2307 allows us to construct the required example input much more cleanly, so that we can directly rely on the `to_torchscript` provided by PytorchLightning.

Issue #2239, opened 2022-08-25
Schema Translation


jaheba:
Models and input data may have different schemas which we want to reconcile providing some abstractions.

## Summary

Models expect input data to accord to a certain schema. However, input data may have its own incompatible data schema, requiring a translation.

It may further be the case that a model's schema has optional columns, meaning that for a given case the column may or may not be available. Thus, a model may want to lookup available fields in the data schema to enable features. For this document we use the term negotiation for this process.

We want to make schema translation and negotiation part of GluonTS's dataset abstraction.

## Background

Currently, there is a default schema that most models use. Users need to convert their datasets to be in accordance with that schema for things to generally work.

There is benefit in having a unified schema, making it easy to run different models on the same dataset. Alas, when having data which is not in compliance with the schema, one needs to translate the data in a preceding step. Also, it is not straightforward to add a new field to the default schema. Further, there are models which don't fit the default schema at all.

To enable more flexibility, we want to allow each model to define its own schema. Further, we want to enable users to define a translation of their data schema to the model's schema.

## Implementation

Users can define a mapping to translate input data on the fly:

```py
model_schema = {"target", "feat_dynamic_real"}
data_schema = {"sales", "price"}

mapping = {"target": "sales", "feat_dynamic_real": ["price"]}
```

Using that mapping, we can then check, whether we can translate one schema into the other:

```py
def check(model_schema: set, data_schema: set, mapping: dict):
    # by default, we use all fields of `data_schema` for usage in the model
    assert model_schema <= data_schema | mapping.keys()

    # make sure that all selected data-columns are actually there
    assert get_values(mapping) <= data_schema
```

```py
@dataclass
class Get:
    column: str

    def apply(self, entry):
        return entry[column]


@dataclass
class Stack:
    columns: list[str]

    def apply(self, entry):
        return np.stack([entry[column] for column in self.columns])


def get_translation(model_schema: set, data_schema: set, mapping: dict):
    translation = {}

    for column in model_schema:
        if column in mapping:
            select = mapping[column]
            if isinstance(select, str):
                translation[column] = Get(select)
            elif isinstance(select, list):
                translation[column] = Stack(select)
        else:
            translation[column] = Get(column)

    return translation

def apply_translation(translation, dataset):
    for entry in dataset:
        yield {column: fn(entry) for column, fn in translation.items()}
```

## Discussion

The described approach assumes that datatypes are always compatible. This may not be the case.

[...]

PR #2304, opened 2022-09-15
Add dataset.schema.translate.

jaheba:
This implements a solution for #2239

The idea is to define translation steps as strings, which resemble Python expressions:

```py
tl = Translator.new(target="price")
```

What is currently supported is to use some indexing, attribute access and invocation of methods.

Further, passing a list will stack the values.

---

### DIFF
--- a/src/gluonts/dataset/schema/translate.py
+++ b/src/gluonts/dataset/schema/translate.py
@@ -150,10 +150,6 @@ def check_type(token, ty, val):
     return False
 
 
-def is_boundary(stream):
-    return not stream or stream.peek("DOT") or stream.peek("PARAN_OPEN")
-
-
 @dataclass
 class Parser:
     stream: TokenStream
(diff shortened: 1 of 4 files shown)


src/gluonts/dataset/schema/translate.py
```diff
@@ -0,0 +1,284 @@
...
+
+    def parse_dot(self, obj):
+        self.stream.pop("DOT")
+
+        name = self.stream.pop("NAME").value
+
+        return GetAttr(obj, name)
+
+    def parse_invoke(self, obj):
+        self.stream.pop("PARAN_OPEN", "(")
```
lostella: PAREN?

lostella (commented):
Looks good to me. Maybe before merging, it would be good to have module documentation, and docstrings at least for the main user-facing components (`Translator`?)

src/gluonts/dataset/schema/translate.py
jaheba: Short for parentheses.

lostella: Yeah but you have PARAN

lostella (approved):

PR #2305, opened 2022-09-15
Estimator schema

jaheba:
*Issue #, if available:*

*Description of changes:*


jaheba:
Here is what I'm unsure about: When do we apply the schema. Should be a general thing imposed by `Estimator`, or do we want each instance to use it manually?

The most manual way would be something like:

```py
class MyEstimator:
    def train(
        self,
        training_data: Dataset,
    ) -> Predictor:
        schema = self.get_schema()

        training_data = Map(schema, training_data)

        if validation_data is not None:
            validation_data = Map(schema, validation_data)

        ...
```

We could simplify it by using a decorator, which would apply the schema for us:


```py
class MyEstimator:
    @with_schema
    def train(
        self,
        training_data: Dataset,
    ) -> Predictor:
        ...
```

Or we apply the schema in the base `Estimator` and do something like this:

```py
class Estimator:
    def train(
        self,
        training_data: Dataset,
    ) -> Predictor:
        schema = self.get_schema()

        training_data = Map(schema, training_data)

        if validation_data is not None:
            validation_data = Map(schema, validation_data)

        self._train(training_data, validation_data)


class MyEstimator:
    def _train(
        self,
        training_data: Dataset,
    ) -> Predictor:
        ...
```

Looking at these examples, I currently prefer the decorator approach, as it looks like a good mix. But maybe we should also consider the broader context of data-loading here. Applying the schema is just one part of it, we also sample, stack, etc the data without having interfaces for them as well.

/cc @lostella

lostella:
Where would you apply the decorator? To `GluonEstimator.train` for example? Then it’s not too different from doing it manually there, since it’s just a couple of lines (also not a huge fan of metaprogramming)

lostella:
Otherwise this appears to be making the input transform significantly simpler to define.

Applying this to all estimators will be a good testbed for this proposal.

jaheba:
> Where would you apply the decorator? To `GluonEstimator.train` for example? Then it’s not too different from doing it manually there, since it’s just a couple of lines (also not a huge fan of metaprogramming)

I like the opt-in behaviour by using a decorator. It is also more explicit looking at the code.

But I think we should also apply it to `GluonEstimator` as it is "our" own class which impose a certain interface on the subclasses.

Maybe we can add some connection on how to get the schema to the decorator.

```py
@with_schema(method="get_schema")
def train(...):
    ...
```

lostella:
> I like the opt-in behaviour by using a decorator. It is also more explicit looking at the code.

Eh, not sure it’s more explicit than

```python
training_data = Map(schema, training_data)
```

More compact for sure

jaheba:
It's more explicit in the sense that it clarifies that this `train` method uses a schema.



> ```python
> training_data = Map(schema, training_data)
> ```

This is something which would need to happen outside of the `Estimator`.

jaheba:
I'm a bit unsure on how explicit the schema interface should be. It would be nice to have a standard way to get a schema, which would allow to compare two schemas (model and data schemas) directly.

PR #2307, opened 2022-09-15
Add input size and type information to DeepARModel

jgasthaus:
*Description of changes:*

Add input size and type information to DeepARModel, in order to (among other uses) support the creation of an `example_input_array`, which is used by PytorchLightning when tracing a model for conversion to TorchScript.

### DIFF
--- a/src/gluonts/torch/model/deepar/lightning_module.py
+++ b/src/gluonts/torch/model/deepar/lightning_module.py
@@ -61,9 +61,15 @@ class DeepARLightningModule(pl.LightningModule):
         self.lr = lr
         self.weight_decay = weight_decay
         self.patience = patience
+        self.example_input_array = tuple(
+            [
+                torch.zeros(shape, dtype=self.model.input_types()[name])
+                for (name, shape) in self.model.input_shapes().items()
+            ]
+        )
 
     def forward(self, *args, **kwargs):
-        return self.model.forward(*args, **kwargs)
+        return self.model(*args, **kwargs)
 
     def _compute_loss(self, batch):
         feat_static_cat = batch["feat_static_cat"]
--- a/src/gluonts/torch/model/deepar/module.py
+++ b/src/gluonts/torch/model/deepar/module.py
@@ -11,7 +11,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-from typing import List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple
 
 import torch
 import torch.nn as nn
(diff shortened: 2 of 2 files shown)


src/gluonts/torch/model/deepar/lightning_module.py
```diff
@@ -61,9 +61,15 @@ def __init__(
...
+        self.example_input_array = tuple(
+            [
+                torch.zeros(shape, dtype=self.model.input_types()[name])
+                for (name, shape) in self.model.input_shapes().items()
+            ]
+        )
 
     def forward(self, *args, **kwargs):
-        return self.model.forward(*args, **kwargs)
+        return self.model(*args, **kwargs)
```
lostella, line 72: Is there a reason for this? I think `forward` should just call `forward`: `__call__` does some additional things like triggering hooks, but I think here the idea is just to delegate the `forward` computation to the wrapped `model`.

This would be automatic were the LightningModule to inherit from the model, but we went back then for the wrapper solution, so it needs to be explicit.

src/gluonts/torch/model/deepar/module.py
```diff
@@ -143,6 +143,34 @@ def _number_of_features(self) -> int:
     def _past_length(self) -> int:
         return self.context_length + max(self.lags_seq)
 
+    def input_shapes(self, batch_size=1) -> Dict[str, Tuple[int, ...]]:
```
lostella, line 146: Since #2249, the model actually wraps the `LightningModule`: this was necessary so that, when re-training a model, the `LightningModule` parameters used by the estimator could be recovered easily from the predictor, without any strange name-prefix-logic.

Now I guess it would make sense to have this information there on the LightningModule? Once again just delegating to the wrapped module there should be fine.

src/gluonts/torch/model/deepar/lightning_module.py
jgasthaus: Yeah, the reason is exactly what you mentioned: `__call__()` triggers the hooks, and PytorchLightning uses a hook on `forward()` to compute the input sizes when getting a `model_summary()`. Without this, the input size to the `DeepARModel` will show up as `?`.

src/gluonts/torch/model/deepar/module.py
jgasthaus: Not sure we also need to have this information on the LightningModule. These are essentially just helpers to set the `example_input_array` on the LightningModule. The added benefit of having them in `DeepARModel` is that it makes the sizes of the inputs explicit for readers of the source code.

src/gluonts/torch/model/deepar/lightning_module.py
lostella: But this somehow contrasts with https://github.com/Lightning-AI/lightning/blob/51ab51a731fb56010113df5a304eaf28dc3a35f7/src/pytorch_lightning/core/module.py#L614

lostella (approved):
🚀

PR #2314, opened 2022-09-20
Fix `gluonts.dataset.split` for multivariate case

lostella:
*Description of changes:* Fixes to `gluonts.dataset.split` in the multivariate target case, and added test cases

### DIFF
--- a/src/gluonts/dataset/rolling_dataset.py
+++ b/src/gluonts/dataset/rolling_dataset.py
@@ -111,7 +111,7 @@ def truncate_features(timeseries: dict, max_len: int) -> dict:
     ):
         if key not in timeseries:
             continue
-        timeseries[key] = [feature[:max_len] for feature in timeseries[key]]
+        timeseries[key] = timeseries[key][:, :max_len]
 
     return timeseries
 
--- a/src/gluonts/dataset/split.py
+++ b/src/gluonts/dataset/split.py
@@ -137,7 +137,7 @@ def slice_data_entry(
 ) -> DataEntry:
     slice_ = to_positive_slice(
         to_integer_slice(slice_, entry[FieldName.START]),
-        len(entry[FieldName.TARGET]),
+        entry[FieldName.TARGET].shape[-1],
     )
 
     if slice_.stop is not None:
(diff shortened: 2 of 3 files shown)


src/gluonts/dataset/rolling_dataset.py
```diff
@@ -111,7 +111,7 @@ def truncate_features(timeseries: dict, max_len: int) -> dict:
     ):
         if key not in timeseries:
             continue
-        timeseries[key] = [feature[:max_len] for feature in timeseries[key]]
+        timeseries[key] = timeseries[key][:, :max_len]
```
jaheba, line 114: How is this related to the split?

src/gluonts/dataset/split.py
```diff
@@ -137,7 +137,7 @@ def slice_data_entry(
 ) -> DataEntry:
     slice_ = to_positive_slice(
         to_integer_slice(slice_, entry[FieldName.START]),
-        len(entry[FieldName.TARGET]),
+        entry[FieldName.TARGET].shape[-1],
```
jaheba, line 140: I think the named axis approach could be handy here.

src/gluonts/dataset/split.py
```diff
@@ -152,10 +152,16 @@ def slice_data_entry(
     if slice_.start is not None:
         offset = slice_.start
         if offset < 0:
-            offset += entry["target"].shape[0]
+            offset += entry["target"].shape[-1]
         sliced_entry[FieldName.START] += offset
 
-    sliced_entry[FieldName.TARGET] = sliced_entry[FieldName.TARGET][slice_]
+    # TODO fix
```
jaheba, line 158: What would the fix be?

jaheba (approved):
I think this is another proof that our schema is too fixed.

src/gluonts/dataset/rolling_dataset.py
lostella: I think it just ended up there. In any way, it's a fix :-)

src/gluonts/dataset/split.py
lostella: I was thinking of getting rid of the branching in favor of constructing a better `slice_` that would work in all cases. But it's OK for now

PR #2315, opened 2022-09-20
Improve `TestData` class in `gluonts.dataset.split`

lostella:
*Description of changes:* This makes the `TestData` (formerly `TestDataset`) output proper iterable objects (effectively, `Dataset`) when asked for the `input` and `label` portions of the test data.

### DIFF
--- a/src/gluonts/dataset/split.py
+++ b/src/gluonts/dataset/split.py
@@ -408,7 +408,7 @@ class LabelDataset:
         return len(self.test_data)
 
     def __iter__(self):
-        for input_, label in self.test_data:
+        for _input, label in self.test_data:
             yield label
 
 
--- a/src/gluonts/dataset/split.py
+++ b/src/gluonts/dataset/split.py
@@ -253,11 +253,9 @@ class AbstractBaseSplitter(ABC):
                 )
 
                 if max_history is not None:
-                    input = TimeSeriesSlice(test[0])[-max_history:]
+                    yield TimeSeriesSlice(test[0])[-max_history:], test[1]
                 else:
-                    input = test[0]
-
-                yield input, test[1]
+                    yield test[0], test[1]
 
 
 @dataclass
@@ -398,8 +396,8 @@ class InputDataset:
         return len(self.test_data)
 
     def __iter__(self):
-        for input, _ in self.test_data:
-            yield input
+        for input_, label in self.test_data:
+            yield input_
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


src/gluonts/dataset/split.py
```diff
@@ -328,11 +328,11 @@ def test_pair(
 
 
 @dataclass
-class TestDataset:
+class TestData:
```
jaheba, line 335: Should we mark this as breaking?

src/gluonts/dataset/split.py
```diff
@@ -372,20 +372,39 @@ def __iter__(self) -> Generator[Tuple[DataEntry, DataEntry], None, None]:
...
+
+@dataclass
+class InputDataset:
+    test_data: TestData
+
+    def __len__(self):
+        return len(self.test_data)
+
+    def __iter__(self):
+        for input, _ in self.test_data:
```
jaheba: I guess there are so many other ways to do this, but this should be fine.

However, can we maybe also deconstruct the second argument to clarify what we are discarding?

src/gluonts/dataset/split.py
lostella: it's only breaking wrt `dev` itself, since this code was merged after the `v0.10.0` release

src/gluonts/dataset/split.py
lostella: addressed, plus got rid of some `input` variables to avoid shadowing the builtin `input` (not that it's a big issue I think)

src/gluonts/dataset/split.py
```diff
@@ -378,20 +376,39 @@ def __iter__(self) -> Generator[Tuple[DataEntry, DataEntry], None, None]:
...
+
+@dataclass
+class LabelDataset:
+    test_data: TestData
+
+    def __len__(self):
+        return len(self.test_data)
+
+    def __iter__(self):
+        for input_, label in self.test_data:
```
jaheba:
```suggestion
        for _input, label in self.test_data:
```

src/gluonts/dataset/split.py
```diff
@@ -378,20 +376,39 @@ def __iter__(self) -> Generator[Tuple[DataEntry, DataEntry], None, None]:
...
+@dataclass
+class InputDataset:
+    test_data: TestData
+
+    def __len__(self):
+        return len(self.test_data)
+
+    def __iter__(self):
+        for input_, label in self.test_data:
+            yield input_
```
jaheba: I think it's fine to overwrite `input`. Also, I would use `_` as a prefix to indicate that the value is not used.

```suggestion
        for input, _label in self.test_data:
            yield input
```

jaheba (approved):

Issue #2333, opened 2022-09-27
Plotting utilities


jaheba:
We should add some plotting helpers, which make it easy to display time-series, predictions, etc.

For example, [these](https://ts.gluon.ai/stable/tutorials/forecasting/quick_start_tutorial.html) [tutorials](https://ts.gluon.ai/dev/tutorials/data_manipulation/dataset_splitting_example.html) spend a lot of lines on setting up plotting.

Ideally, we add something like `gluonts.lab.plot`, which makes it easy to visualize things.

ivanvoid:
Tried to go through [Tutorial](https://ts.gluon.ai/stable/tutorials/forecasting/quick_start_tutorial.html#Quick-Start-Tutorial), just copy/pasted code, plot function for SampleForecast entity  doesn't seems to work the way it should. 

![image](https://user-images.githubusercontent.com/59028817/230828945-7dab5775-7fa4-4f70-a663-2466a4eb7b6a.png)

PR #2338, opened 2022-10-04
Plotting

codingWhale13:
*Description of changes:*
 This PR addresses #2333 and proposes a `plot` function that is given a forecast and optionally a time series for reference as well as a bunch of options about how to plot things.

The most interesting part is plotting the different prediction intervals. This now works pretty much as in `Forecast.plot`. The only major difference is that the lower and upper parts of each prediction interval are created separately to keep the colors true (in Forecast.plot, when plotting a lot of prediction intervals, the colors in the middle get very dark because the `alpha` of the color gets less with each overlap).

The `plot_prob_forecasts` used in [this tutorial](https://ts.gluon.ai/stable/tutorials/forecasting/quick_start_tutorial.html) can be replaced by a call to `lab.plot`. Also, the vertical red line in the tutorial can be created using the `train_test_separator` parameter.

It might make sense to return the  Matplotlib `fig` and `ax` so that the user can draw more custom things on top of it (like highlighting the train vs. test data in different colors as is done [here](https://ts.gluon.ai/dev/tutorials/data_manipulation/dataset_splitting_example.html)).

This proposal would pretty much replace the `plot` method of `Forecast`. I think this makes sense because plotting is something separate and the Forecasts would be more lightweight this way.

Biggest caveat at the moment: 150 lines is a lot of code to just plot a forecast. Just providing all the different knobs for the user to choose (like color, legend position etc.) makes things a bit messy but not providing them would render the `plot` function pretty useless, I think. So, maybe there's an easier way to do this.


src/gluonts/lab/plot.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def plot(
+    forecast: Optional[Forecast],
+    timeseries: Optional[pd.Series] = None,
+    prediction_intervals: Collection[float] = (50.0, 90.0),
+    show_mean: bool = False,
+    color: str = "g",
+    label: Optional[str] = None,
+    xlabel: Optional[str] = None,
```
lostella: x and y labels could default to “time” and “value” maybe?

src/gluonts/lab/plot.py
```diff
@@ -0,0 +1,165 @@
...
+                alpha=alphas[i],
+                interpolate=True,
+                label=info["label"],
+                *args,
+                **kwargs,
+            )
+
+
+def plot(
+    forecast: Optional[Forecast],
```
lostella: Should this be optional?

src/gluonts/lab/plot.py
```diff
@@ -0,0 +1,165 @@
...
+    timeseries: Optional[pd.Series] = None,
+    prediction_intervals: Collection[float] = (50.0, 90.0),
+    show_mean: bool = False,
+    color: str = "g",
+    label: Optional[str] = None,
```
lostella: Maybe “label_prefix” or “prefix”?

src/gluonts/lab/plot.py
codingWhale13: No, this should be the only required parameter. Good catch, thanks!

kashif:
another potential use case is plotting some of the variates in the multivariate forecasting case... can we get a helper for that too?

codingWhale13:
Thanks for the input, @kashif. 👍 The `plot` function now supports plotting multivariate forecasts as well. You can choose which variates to plot using the `variates_to_plot` parameter.

src/gluonts/lab/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+from .plot import plot
```
jaheba: I think it's weird when module and function have the same name.

We should probably rename the module to something else. Something like `viz` could work.

jaheba (commented):
My feeling is that we should have many different functions for different tasks. Say I want to compare the output of two models on the same data, then we probably want to plot it differently than just plotting the data for one model.

kashif:
Can you check if this will also work for Quantile forecasts?

codingWhale13:
> Can you check if this will also work for Quantile forecasts?

It does not. 😄 PR #2352 should fix this.

PR #2341, opened 2022-10-05
Speed up smoke tests

lostella:
*Description of changes:* This reduces run time for this test to 1:20 minutes, down from 3:30 minutes, on my machine.


lostella:
Saved a couple tenths of seconds here and there as well

jaheba (approved):
Does this only affect MQCNN?

Is there a reason to run for at least two epochs, to make sure that looping works?

lostella:
> Does this only affect MQCNN?

No, fixed the title

> Is there a reason to run for at least two epochs, to make sure that looping works?

I don't think so, there are other tests that do training on multiple batches and epochs

PR #2344, opened 2022-10-06
Fix numerical bug in `BinnedUniforms`

moudheus:
Computing prob.cumsum() - prob is a numerically unsafe way to shift prob.cumsum() to the right. The numerical error will create small gaps or overlaps in the windows defined by the lower and upper tensors. If a quantile falls in one of these overlap or gap the one hot bin indicator will have 0 or 2 hot bins and generate a shape mismatch error. 

*Issue #, if available:*

*Description of changes:*

The fix consists of doing a shift by copy.

---

#bug fix


lostella:
@moudheus good spot! Thanks for the fix!

lostella (approved):

PR #2351, opened 2022-10-10
Add predict_one to `Predictor`.

jaheba:
*Issue #, if available:*

*Description of changes:*


lostella:
I think this makes sense to have

jaheba:
In general, there are two kinds of predictors, those who can predict one item, and those who predict batches. Currently we have lumped everything together.

What we could do is to introduce `BatchPredictor` and `OnePredictor` (or something along those lines) and then derive from either of these.

But maybe we start with this for now and think about generalising later?

PR #2352, opened 2022-10-10
Add `copy_dim` to `QuantileForecast`, change `dim` method for multivariate data

codingWhale13:
*Issue #, if available:*

*Description of changes:*
This PR adds `copy_dim` to the `QuantileForecast` class, similar to how it is done in `SampleForecast`


jaheba:
What is the purpose of `copy_dim`?

codingWhale13:
> What is the purpose of `copy_dim`?

The purpose is stated in the base class `Forecast`: "Returns a new Forecast object with only the selected sub-dimension."

jaheba:
That's some weird naming.

Should we add a test for this?

codingWhale13:
While trying to write a test, I found that in the multivariate case, the 3d NumPy arrays for `samples`/`forecast_array` (for `SampleForecast`/`QuantileForecast` respectively) seem inconsistent:

`SampleForecast`: [(num_samples, prediction_length, target_dim)](https://github.com/awslabs/gluonts/blob/2225d9b84eca5154f9f3392da7f2ac40dd9aebe4/src/gluonts/model/forecast.py#L617-L619)
`QuantileForecast`: [(num_samples, target_dim, prediction_length)](https://github.com/awslabs/gluonts/blob/2225d9b84eca5154f9f3392da7f2ac40dd9aebe4/src/gluonts/model/forecast.py#L758-L760).

Before [this commit](https://github.com/awslabs/gluonts/commit/36630666b97208bd3089e2fea008775be9bad00e), it was both (num_samples, target_dim, prediction_length) and then it was changed in `SampleForecast`.

It seems to me like the change for `QuantileForecast`s slipped through and multivariate `QuantileForecast`s weren't much used since then.

Please correct me if I'm wrong and if changing the meaning of `QuantileForecast.forecast_array` in this way breaks anything.

codingWhale13:
With N = batch axis, T = time axis, C = channel axis (in the context of multivariate forecasts: `target_dim`), here are some examples of what's used in a few random places of the code:

- [InstanceSplitter](https://github.com/awslabs/gluonts/blob/04ceba2c271a806e205427fd3e3a709466291182/src/gluonts/transform/split.py#L37-L39) and [CanonicalInstanceSplitter](https://github.com/awslabs/gluonts/blob/04ceba2c271a806e205427fd3e3a709466291182/src/gluonts/transform/split.py#L179-L182): (N, T) - this coincides with the univariate case of `Forecast`s, everything good here.
- [MXNet Seq2SeqEnc2Dec](https://github.com/awslabs/gluonts/blob/04ceba2c271a806e205427fd3e3a709466291182/src/gluonts/mx/block/enc2dec.py#L47-L49): (N, T, C)
- [MXNet MeanScaler](https://github.com/awslabs/gluonts/blob/d5d897cd02b3aaa499f3c3632d3a18863b1a24e7/src/gluonts/mx/block/scaler.py#L136): both (N, T, C) and (N, C, T) are possible.

Even though comparing these different contexts only makes limited sense, it looks like T usually precedes C, to which things have been updated in the `SampleForecast` a while ago, as mentioned in an above comment.

Long story short, in my view this confirms that [changing](https://github.com/awslabs/gluonts/pull/2352/commits/80969697fecdaa8bb39dc60ee6ab3c53aba44f25) the dimensions of the multivariate `QuantileForecast` makes sense.

lostella:
We changed to the `(N, T, C)` layout because that's closer to the distribution abstraction: for multivariate distributions, it's over the trailing dimension (`C`) that data is sampled jointly and density is evaluated. This is the case for example in PyTorch distributions, but also in the MXNet based ones we have in GluonTS.

Indeed, the change you point out just missed the `QuantileForecast` case. I think it makes sense to change that as you proposed: it will be a breaking change, but hopefully a minor one.

lostella (approved):

PR #2357, opened 2022-10-10
Remove `DummyEstimator`.

jaheba:
*Issue #, if available:*

*Description of changes:*


lostella:
Does this have any consequence in terms of `gluonts.shell`-based containers? Or can predictor types directly specified there?

jaheba:
> Does this have any consequence in terms of `gluonts.shell`-based containers? Or can predictor types directly specified there?

No, `DummyEstimator` is not used in shell.

We could add `train` to `Predictor`, which would just returns `self`. But I don't think I like it.

lostella (dismissed):

lostella (approved):

PR #2360, opened 2022-10-11
Add wrapper for `statsforecast` models

lostella:
*Description of changes:* This adds a predictor type that wraps around models from the `statsforecast` package https://github.com/Nixtla/statsforecast

TODOs:
- [x] ~~add usage of exogenous features~~ (leaving as TODO)
- [x] return `QuantileForecast` with multiple quantiles, instead of `SampleForecast` containing only the mean (the way `statsforecast` outputs probabilistic forecasts currently complicates this)
- [x] add direct constructors for specific models
- [x] add tests

cc @abdulfatir 

Demo:

```python
import os
from pathlib import Path

from gluonts.model.predictor import Predictor
from gluonts.model.statsforecast import (
    AutoARIMAPredictor,
    ETSPredictor,
    AutoCESPredictor,
)

def plot(entry, forecast, context_length):
    series = to_pandas(entry)
    series[-context_length:].plot()
    forecast.plot()

airpassengers = get_dataset("airpassengers")

predictors = {
    str(constructor): constructor(
        prediction_length=12,
        **kwargs,
    )
    for constructor, kwargs in [
        (AutoARIMAPredictor, dict(season_length=12, quantile_levels=[0.5, 0.10, 0.30, 0.5, 0.7, 0.9, 0.99])),
        (ETSPredictor, dict(season_length=12)),
        (AutoCESPredictor, dict(season_length=12)),
    ]
}

for k, (name, predictor) in enumerate(predictors.items()):
    model_path = Path(os.path.realpath(__file__)).parent / str(k)
    model_path.mkdir(exist_ok=True)
    predictor.serialize(model_path)

    predictor_copy = Predictor.deserialize(model_path)
    
    for entry, forecast in zip(airpassengers.train, predictor.predict(airpassengers.train)):
        plot(entry, forecast, context_length=36)
        plt.title(name)
        plt.show()
```


src/gluonts/model/statsforecast/predictor.py
```diff
@@ -0,0 +1,301 @@
...
+from gluonts.core.component import validated
+from gluonts.dataset import DataEntry
+from gluonts.model.predictor import RepresentablePredictor
+from gluonts.model.forecast import QuantileForecast
+
+
+# currently it's tricky to get quantiles out of statsforecast
+# see https://github.com/Nixtla/statsforecast/issues/256
+# therefore we need the following function
+def quantiles_to_intervals(quantile_levels: Optional[List[float]] = None):
```
jaheba: What's the return type?

src/gluonts/model/statsforecast/predictor.py
```diff
@@ -0,0 +1,301 @@
...
+    if quantile_levels is None:
+        return None, dict()
```
jaheba: `dict()` should just be `{}`

src/gluonts/model/statsforecast/predictor.py
```diff
@@ -0,0 +1,301 @@
...
+    def __init__(
+        self,
+        model_constructor,
+        kwargs: dict,
+        prediction_length: int,
+        quantile_levels: Optional[List[float]] = None,
+    ) -> None:
+        super().__init__(prediction_length=prediction_length)
+        self.model = model_constructor(**kwargs)
+        self.intervals, self.keys = quantiles_to_intervals(quantile_levels)
```
jaheba: I hate this. I don't think we should assign like this to self.

src/gluonts/model/statsforecast/predictor.py
```diff
@@ -0,0 +1,301 @@
...
+        self.statsforecast_keys = ["mean"] + list(self.keys.values())
+        self.forecast_keys = ["mean"] + list(self.keys.keys())
+
```
jaheba: This looks quite weird, can we find a better name for it?

.github/workflows/tests-statsforecast.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install other dependencies
+      run: |
+        python -m pip install -U pip
+        pip install .
+        pip install -r requirements/requirements-test.txt
+        pip install statsforecast
```
jaheba: Should we have a requirements file?

src/gluonts/model/statsforecast/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+from .predictor import (
```
jaheba: If this is basically just importing stuff from `predictor.py` we could simplify this module to `statsforecast.py`.

test/model/statsforecast/test_statsforecast.py
```diff
@@ -0,0 +1,65 @@
...
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+import pandas as pd
+import pytest
+import numpy as np
+
+from gluonts.dataset import Dataset
+from gluonts.model import Predictor
+from gluonts.model.forecast import QuantileForecast
```
jaheba: Should we also expose `QuantileForecast` directly in `model`?

src/gluonts/model/statsforecast/predictor.py
```diff
@@ -0,0 +1,301 @@
...
+# see https://github.com/Nixtla/statsforecast/issues/256
+# therefore we need the following function
+def quantiles_to_intervals(quantile_levels: Optional[List[float]] = None):
+    if quantile_levels is None:
+        return None, dict()
+
+    intervals = set()
+    keys = dict()
+
+    for ql in quantile_levels:
```
jaheba: I don't like abbrevations!
```suggestion
    for quantile_level in quantile_levels:
```

src/gluonts/model/statsforecast/predictor.py
```diff
@@ -0,0 +1,301 @@
...
+
+        return QuantileForecast(
+            forecast_arrays=np.stack(arrays, axis=0),
+            forecast_keys=self.forecast_keys,
+            start_date=entry["start"] + len(entry["target"]),
+            item_id=entry.get("item_id"),
+        )
+
+
+def ADIDAPredictor(
```
jaheba: What's the advantage of using functions over sub-classing?

src/gluonts/model/statsforecast/predictor.py
```diff
@@ -0,0 +1,301 @@
...
+    prediction_length: int, quantile_levels=None, **kwargs
+) -> StatsForecastPredictor:
+    """
+    A predictor based on the ``ADIDA`` model from `statsforecast`_.
```
jaheba: I think "wrapping" might be the more accurate term?

src/gluonts/model/statsforecast/predictor.py
```diff
@@ -0,0 +1,301 @@
...
+            h=self.prediction_length,
+            **kwargs,
+        )
+
+        arrays = [pred[k] for k in self.statsforecast_keys]
+
+        return QuantileForecast(
+            forecast_arrays=np.stack(arrays, axis=0),
+            forecast_keys=self.forecast_keys,
+            start_date=entry["start"] + len(entry["target"]),
```
jaheba: There is `forecast_start` as a util somewhere.

src/gluonts/model/statsforecast/predictor.py
```diff
@@ -0,0 +1,301 @@
...
+        if self.intervals is not None:
+            kwargs["level"] = self.intervals
+
+        pred = self.model.forecast(
+            y=entry["target"],
+            h=self.prediction_length,
+            **kwargs,
+        )
+
+        arrays = [pred[k] for k in self.statsforecast_keys]
```
jaheba: I think this is a bad name. Can we be more specific?

src/gluonts/model/statsforecast/predictor.py
lostella: Why? I can make it a method if that's better

src/gluonts/model/statsforecast/predictor.py
lostella: I started with subclassing, I can go back to it, to me it doesn't make much of a difference, this was just more compact

src/gluonts/model/statsforecast/predictor.py
lostella: Yeah, it's a bit messy, I'll streamline it a bit

src/gluonts/model/statsforecast/predictor.py
jaheba: Can't we get the same with subclassing? We could also look into using `serde.dataclass` instead of `validated`.

requirements/requirements-extras-statsforecast.txt
```diff
@@ -0,0 +1 @@
+statsforecast
```
jaheba: What is their versioning like?

src/gluonts/model/__init__.py
```diff
@@ -17,8 +17,15 @@
...
+    "Predictor",
+    "Forecast",
+    "SampleForecast",
+    "QuantileForecast",
+]
 
 from .estimator import Estimator, IncrementallyTrainable
 from .predictor import Predictor
-from .forecast import Forecast
+from .forecast import Forecast, SampleForecast, QuantileForecast
```
jaheba: Can we do this in another PR?

src/gluonts/model/statsforecast.py
```diff
@@ -0,0 +1,329 @@
...
+        for quantile_level in self.quantile_levels:
+            interval = round(
+                200 * (max(quantile_level, 1 - quantile_level) - 0.5)
+            )
+            intervals.add(interval)
+            side = "hi" if quantile_level > 0.5 else "lo"
+            self.forecast_keys.append(str(quantile_level))
+            self.statsforecast_keys.append(f"{side}-{interval}")
+
+        self.intervals = sorted(list(intervals))
```
jaheba:
```suggestion
        self.intervals = sorted(intervals)
```

Sorted returns a list already.

src/gluonts/model/statsforecast.py
```diff
@@ -0,0 +1,329 @@
...
+
+    .. _statsforecast: https://github.com/Nixtla/statsforecast
+    """
+
+    @validated()
+    def __init__(
+        self, prediction_length: int, quantile_levels=None, **kwargs
+    ) -> None:
+        super().__init__(
+            model_constructor=ADIDA,
```
jaheba: Can't we just make this a class variable? Then we don't need to have this `__init__` at all.

src/gluonts/model/statsforecast.py
```diff
@@ -0,0 +1,329 @@
...
+        self.model = model_constructor(**kwargs)
+        self.config = ModelConfig(quantile_levels=quantile_levels)
+
+    def predict_item(self, entry: DataEntry) -> QuantileForecast:
+        # TODO use also exogenous features
+        kwargs = {}
+        if self.config.intervals is not None:
+            kwargs["level"] = self.config.intervals
+
+        pred = self.model.forecast(
```
jaheba:
```suggestion
        prediction = self.model.forecast(
```

src/gluonts/model/__init__.py
jaheba: https://github.com/awslabs/gluonts/pull/2366

requirements/requirements-extras-statsforecast.txt
lostella: appears to be semver

src/gluonts/model/statsforecast.py
```diff
@@ -0,0 +1,232 @@
...
+    """
+
+    ModelType: Type
+
+    @validated()
+    def __init__(
+        self,
+        prediction_length: int,
+        quantile_levels: Optional[List[float]] = None,
+        **kwargs,
```
jaheba: It might be better to not use `**kwargs` but something like `model_params`. That way the two sets of parameters are separated more clearly.

It's also what we do for prophet, where it's called `prophet_params`.

src/gluonts/model/statsforecast.py
```diff
@@ -0,0 +1,232 @@
...
+    ) -> None:
+        super().__init__(prediction_length=prediction_length)
+        self.model = self.ModelType(**kwargs)
+        self.config = ModelConfig(quantile_levels=quantile_levels)
+
+    def predict_item(self, entry: DataEntry) -> QuantileForecast:
+        # TODO use also exogenous features
+        kwargs = {}
+        if self.config.intervals is not None:
+            kwargs["level"] = self.config.intervals
```
jaheba, line 108: Is there a neutral value for `level`?

jaheba: This suggest we could just use `None`.

https://github.com/Nixtla/statsforecast/blob/fecbfc31c6cba3d615147565196f7be1459d5a90/statsforecast/models.py#L278

lostella: the thing is, only arima accepts this `level` argument, so for all other models one should just not set the quantiles and `level` not be passed

jaheba: But if only arima takes it, why do we expose it for all methods?

lostella: Mainly https://github.com/Nixtla/statsforecast/issues/248#issuecomment-1256539527

jaheba (approved):

PR #1834, opened 2021-12-16
Adding torch.deepvar

KelvinKan:
Adding PyTorch implementation of DeepVAR (Vec-LSTM)

Modules added:
1. deepvar.estimator and deepvar.module
2. test_deepvar
3. distribution.lowrank_multivariate_normal (LowRankMultivariateNormalOutput)
4. fixed a bug in torch.deepar.lightning_module

Due to time constraint on my side (internship ending), I left the following additional features as TODO:
1. Use the `target_dimension_indicator` feature in training and prediction
2. Use Gaussian copula

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Please tag this pr with at least one of these labels to make our release process faster:** BREAKING, new feature, bug fix, other change, dev setup

PR #2365, opened 2022-10-12
Add target_dim to deepAR and to distributions

kashif:
*Issue #, if available:*

*Description of changes:*

We can just abstract the input size / target dim in deepAR model and obtain the deepVAR model.


kashif:
this PR make a few refactorings to deepAR to allow for training independent multivariate models. I believe if we add the other multivariate distributions output heads then there is no need to add another exact copy of deepAR as in #1834

src/gluonts/torch/distributions/affine_transformed.py
```diff
@@ -35,13 +35,20 @@ class AffineTransformed(TransformedDistribution):
         Scaling parameter of the affine transformation.
     """
 
-    def __init__(self, base_distribution: Distribution, loc=None, scale=None):
+    def __init__(
+        self,
+        base_distribution: Distribution,
+        loc=None,
+        scale=None,
+        event_dim: int = 0,
```
lostella, line 46: Could you add this to the docstring as well? So it's clear what the meaning of it is

src/gluonts/torch/distributions/distribution_output.py
```diff
@@ -106,13 +107,17 @@ class DistributionOutput(Output):
...
     @validated()
     def __init__(self) -> None:
         pass
 
     def _base_distribution(self, distr_args):
-        return self.distr_cls(*distr_args)
+        if self.dim == 1:
+            return self.distr_cls(*distr_args)
+        else:
+            return Independent(self.distr_cls(*distr_args), 1)
```
lostella: Also here `dim` is unclear: as far as I understand, it controls how many batch dimensions we want to turn into event dimensions. So maybe the following is better?

```python
dim: int = 0
[...]
distr = self.distr_cls(*distr_args)
if self.dim > 0:
    return Independent(distr, self.dim)
return distr
```

lostella: Also I'm unsure about this mechanism: what if `self.distr_cls` returns a distribution that already has the correct `event_dim`, like a multivariate normal?

kashif: sorry for the confusion... yes the dim here is the target_dim of the distribution... which is different to the 1 in the Independent which we always want to be 1 i believe

kashif: ok i belieive i fixed that usecase now too

kashif:
also note with this refactoring the beta-nll doesn't work either...

PR #2376, opened 2022-10-17
Fix SmoothSeasonality Sine Wave Generation.

mvanness354:
*Issue #, if available:*
No associated issue.

*Description of changes:*
There was a small error in the code for generating synthetic seasonal data using `rcp.SmoothSeasonality`. In short, the phase parameter was being using incorrectly, not following the standard formula for a sine curve (see [here](https://en.wikipedia.org/wiki/Sine_wave)). The code has been updated to match the correct formula.

### DIFF
--- a/src/gluonts/dataset/artificial/recipe.py
+++ b/src/gluonts/dataset/artificial/recipe.py
@@ -803,7 +803,7 @@ class SmoothSeasonality(Lifted):
         period = resolve(self.period, x, length, *args, **kwargs)
         phase = resolve(self.phase, x, length, *args, **kwargs)
         return (
-            np.sin(2.0 / period * np.pi * (np.arange(length) + phase)) + 1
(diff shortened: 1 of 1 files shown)


jgasthaus:
This doesn't strike me as a bug, but rather a (admittedly somewhat unusual) parametrization of the phase parameter in the same units as the period. I'm ok with making this change to radians, but it will be breaking change. Maybe providing a docstring that calls out the somewhat unusual units might be a viable alternative?

If we make this change, the synthetic data generation tutorial also needs to be updated: https://github.com/awslabs/gluonts/blob/6148459b6c91944b4f3a28f208864974f45965e4/docs/tutorials/data_manipulation/synthetic_data_generation.md.template#L199

huibinshen:
Is there some reference we can cite for this parametrization? If so, putting a reference in the doc string makes sense to me and it will avoid the breaking change. Otherwise, updating the code and tutorial may avoid future confusion. What do you think?

Issue #2363, opened 2022-10-11
Best way to generate large `Dataset`s


caseytomlin:
I have a long pandas DataFrame with a very large number of `item_id`s (~3 million) - is there a recommended way to process it into a `Dataset` that can be passed to models?

`PandasDataset.from_long_dataframe` works, but is rather compute-intensive (~60 minutes + ~70gb RAM). It can be sped up quite a bit with multiprocessing, but with the new arrow functionality (and issues like [this one](https://github.com/awslabs/gluonts/issues/2236)) I feel I am missing a simpler solution that is perhaps just not yet extensively documented.

jaheba:
I've done some trying around.

What should work is to sort the long-dataframe by `item_id` and `index`.

Then we can do something like this:

```py
idx = np.concatenate([[0], np.where(df["item_id"][:-1].values != df["item_id"][1:].values)[0] + 1, [len(df)]])

for start, stop in zip(idx, idx[1:]):
    slice_ = long.iloc[start: stop]
```

@caseytomlin Can you share a simplified example of your dataframe?

caseytomlin:
Thanks @jaheba - I don't immediately understand how to use your suggestion but I will think about it. Please find a sample below.

I was rather thinking one might parallelize the creation of multiple `PandasDataset`s, save to arrow/parquet, then load again as a single dataset using the new arrow support, but it seems a bit convoluted and again like I'm missing something. 

[long_df_sample.zip](https://github.com/awslabs/gluonts/files/9771502/long_df_sample.zip)

jaheba:
Thanks @caseytomlin for sharing the sample.

What actually takes a lot of time is checking that the timestamp column/index is correct (delta between each two consecutive timestamp is equal to the frequency). In the sample that you've shared it appears to be the case. Thus, removing these checks would speed up the process by a lot.

Maybe we can add an `unchecked` flag or something similar, which would allow to skip these checks.

lostella:
> Maybe we can add an `unchecked` flag or something similar, which would allow to skip these checks.

I think it makes sense. What kind of speedup are we talking about?

jaheba:
Running on the provided sample data:

![Screenshot 2022-10-14 at 10 40 42](https://user-images.githubusercontent.com/133732/195802939-0e285161-15d3-484e-9a4b-adb2179affc6.png)

jaheba:
I don't think that this would change the problem with the used memory.

[...]

PR #2377, opened 2022-10-17
Add LongDataset.

jaheba:
*Issue #2363 , if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/dataset/pandas.py
+++ b/src/gluonts/dataset/pandas.py
@@ -256,10 +256,12 @@ def prepare_prediction_data(
     Remove ``ignore_last_n_targets`` values from ``target`` and
     ``past_feat_dynamic_real``.  Works in univariate and multivariate case.
 
-    >>> prepare_prediction_data(
-    >>>    {"target": np.array([1., 2., 3., 4.])}, ignore_last_n_targets=2
-    >>> )
-    {'target': array([1., 2.])}
+        >>> import numpy as np
+        >>> prepare_prediction_data(
+        ...    {"target": np.array([1., 2., 3., 4.])}, ignore_last_n_targets=2
+        ... )
+        {'target': array([1., 2.])}
+
     """
     entry = deepcopy(dataentry)
     for fname in [FieldName.TARGET, FieldName.PAST_FEAT_DYNAMIC_REAL]:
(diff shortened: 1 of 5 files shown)


src/gluonts/dataset/pandas.py
```diff
@@ -307,3 +311,82 @@ def is_uniform(index: pd.PeriodIndex) -> bool:
...
+                    column_as_start,
+                    column=self.timestamp,
+                    freq=self.freq,
+                    unchecked=self.unchecked,
+                ),
+                dataset,
+            )
+
+        yield from dataset
+        self.unchecked = True
```
lostella, line 410: That's smart

src/gluonts/dataset/pandas.py
```diff
@@ -307,3 +311,82 @@ def is_uniform(index: pd.PeriodIndex) -> bool:
...
+            sort_by = [self.item_id]
+            if self.timestamp is not None:
+                sort_by.append(self.timestamp)
+                dataset = Map(methodcaller("sort_values", by=sort_by), dataset)
+
+        dataset = Map(methodcaller("to_dict", orient="list"), dataset)
+        dataset = Map(self._pop_item_id, dataset)
+
+        if self.translator is not None:
+            dataset = Map(self.translator, dataset)
```
lostella, line 395: I'm wondering whether this class is taking care of too much now: applying `translator` here interferes with applying `column_as_start` below.

lostella (commented):
Should this PR have `PandasDataset` return a `LongDataset` then?

src/gluonts/dataset/pandas.py
jaheba: Right, but the problem is that dataframe only have 1D columns, so we need a way to stack columns. One could of course do it themselves afterwards, but since this is a more lab-focused setup I think there is an argument for having it all in one place.

src/gluonts/dataset/pandas.py
```diff
@@ -298,12 +275,139 @@ def is_uniform(index: pd.PeriodIndex) -> bool:
...
+    indicate that the timestamp column is uniform (see ``is_uniform``).
+    """
+
+    df: pd.DataFrame
+    item_id: Union[str, List[str]]
+    timestamp: Optional[str] = None
+    freq: Optional[str] = None
+
+    assume_sorted: bool = False
+    translate: InitVar[Optional[dict]] = None
```
lostella, line 337: There’s both `translate` and `translator`?

PR #2387, opened 2022-10-21
Enable dropping of columns in dataset.schema.translate.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/dataset/schema/translate.py
+++ b/src/gluonts/dataset/schema/translate.py
@@ -27,6 +27,7 @@ more flexibility in the input data::
 
 import re
 from dataclasses import dataclass, InitVar
+from itertools import chain
 from typing import Any, Dict, List, Union, Optional, ClassVar
 
 import numpy as np
@@ -35,6 +36,9 @@ from toolz import valfilter, valmap
 
 class Op:
     def __call__(self, item):
+        raise NotImplementedError
+
+    def fields(self):
         raise NotImplementedError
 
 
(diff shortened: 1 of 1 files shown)


src/gluonts/dataset/schema/translate.py
```diff
@@ -307,21 +326,32 @@ class Translator:
...
+        return Translator(valmap(parse, fields_), drop)
 
     def __call__(self, item):
-        result = dict(item)
+        if self.drop:
+            keys = item.keys() - self.get_fields()
+            result = {key: item[key] for key in keys}
+        else:
+            result = dict(item)
+
```
codingWhale13, line 349: What is supposed to happen when we call the `Translator` with `drop=True`? I could imagine two things:
- Dropping `item` from `fields` or
- returning the `fields` that are not present in `item`

It seems to me like `item` would be a subset of `fields` so `item.keys() - self.get_fields()` would always be empty...

jaheba: `item` is the input, so not sure what you mean with dropping item from fields.

The idea is that if `drop` is enabled, only those fields are copied to the result which are not selected by some translation.

This is what it does:

```py
t = Translator.parse(a="b", drop=True)
t(
    {
        "b": 1, # will be removed if `drop` because selected by a
        "c": 42, # will remain always
    }
)
````

codingWhale13: I see, that makes sense. 👍
Could you add a similar explanation to the docstring, explaining what `drop` does?

codingWhale13 (approved):

PR #2391, opened 2022-10-24
Add `dataset.schema.Schema` + types.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/dataset/schema/types.py
+++ b/src/gluonts/dataset/schema/types.py
@@ -49,13 +49,10 @@ class Array(GenericType[T]):
     dimension.
 
     This class ensures that the handled output data, will have `ndim` number of
-    dimensions, expanding the input if necessary.
-
-    If specified, `dtype` will be applied to the input to force a consistent
-    type, e.g. ``np.float32``.
-
-    `time_dim` is just a marker, indicating which axis notes the time-axis,
-    useful for splitting. If `time_dim` is none, the array is time invariant.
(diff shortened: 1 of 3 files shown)


src/gluonts/dataset/schema/types.py
```diff
@@ -0,0 +1,83 @@
...
+    dtype: typing.Optional[typing.Type[T]] = None
+    time_dim: typing.Optional[int] = None
+
+    def apply(self, data):
+        arr = np.asarray(data, dtype=self.dtype)
+
+        if arr.ndim < self.ndim:
+            to_expand = self.ndim - arr.ndim
+            new_shape = (1,) * to_expand + arr.shape
+            arr = arr.reshape(new_shape)
```
lostella: It looks like `time_dim` ends up referring to the reshaped array. Then if `to_expand > 0` (and reshaping kicks in) one needs to track what `time_dim` will refer to. Not sure how to fix this though

lostella: Maybe reshaping should not happen at all? And simply, a badly-shaped list not be convertible to the desired array type?

jaheba: Maybe that's easiest. I don't think we allow 1d arrays for `dynamic_feat`, or do we?

The other thing one can do is to use translate, to add dimensions on the fly.

lostella (approved):

PR #2395, opened 2022-10-25
Updating mx.trainer's internal epoch number counting to align with pr…

sighellan:
…intout and batch number counting.

*Issue #, if available:*

*Description of changes:*
This updates the internal epoch counting in mx.trainer to align with the epoch numbers displayed to the user. 


jaheba:
I think there is a reference to `epoch_no` in `test/mx/trainer/test_model_averaging.py` which we also might want to update.

sighellan:
On it!

There's a reference in this test as well: [https://github.com/awslabs/gluonts/blob/dev/test/core/test_serde.py](https://github.com/awslabs/gluonts/blob/dev/test/core/test_serde.py) but I don't think it needs changing.

src/gluonts/mx/model/deepvar_hierarchical/_network.py
```diff
@@ -216,7 +216,7 @@ def get_samples_for_loss(self, distr: Distribution) -> Tensor:
 
         # Determine which epoch we are currently in.
         self.batch_no += 1
-        epoch_no = self.batch_no // self.num_batches_per_epoch + 1
+        epoch_no = (self.batch_no - 1) // self.num_batches_per_epoch + 1
```
jaheba, line 219: Not sure we want to update this one.

sighellan:
I'm not sure how much needs updating in this file: https://github.com/awslabs/gluonts/blob/dev/src/gluonts/mx/trainer/model_iteration_averaging.py 
Updating the default epoch to 1 makes sense, but does this `epoch_no + 1` need attention? https://github.com/awslabs/gluonts/blob/dev/src/gluonts/mx/trainer/model_iteration_averaging.py#L353

lostella:
> I'm not sure how much needs updating in this file: https://github.com/awslabs/gluonts/blob/dev/src/gluonts/mx/trainer/model_iteration_averaging.py Updating the default epoch to 1 makes sense, but does this `epoch_no + 1` need attention? https://github.com/awslabs/gluonts/blob/dev/src/gluonts/mx/trainer/model_iteration_averaging.py#L353

The entry-point for that module is just the `ModelIterationAveraging` class, the others are used to configure its behaviour. It seems like the `epoch_no + 1` there is adjusting for the epoch counting being zero-based, so changing it to `epoch` should maintain the desired behaviour within your overall change.

All of this would be of course easier to reason on, had the `update_average_trigger` method had a docstring explaining precisely what it does and what it expects :-)

src/gluonts/mx/model/deepvar_hierarchical/_network.py
sighellan: I think we're currently undercounting the first epoch. 

Assuming batches_per_epoch=10, I think it  goes
Batch no, epoch no
1, 1
2, 1
3, 1
4, 1
5, 1
6, 1
7, 1
8, 1
9, 1
10, 2
11, 2
...

PR #2406, opened 2022-10-31
Add hierarchical time series reconciliation method from R/hts: MinT

melopeo:
*Description of changes:*

Addition of MinT reconciliation method for hierarchical time series, as done in R/hts package.

**Please tag this pr with at least one of these labels to make our release process faster:**  new feature

### DIFF
--- a/src/gluonts/ext/r_forecast/_hierarchical_predictor.py
+++ b/src/gluonts/ext/r_forecast/_hierarchical_predictor.py
@@ -139,6 +139,7 @@ class RHierarchicalForecastPredictor(RBasePredictor):
             "frequency": self.period,
             "fmethod": fmethod,
             "nonnegative": nonnegative,
+            "level": level
         }
 
         if level:
--- a/src/gluonts/ext/r_forecast/_hierarchical_predictor.py
+++ b/src/gluonts/ext/r_forecast/_hierarchical_predictor.py
@@ -106,7 +106,7 @@ class RHierarchicalForecastPredictor(RBasePredictor):
         fmethod: str,
         period: Optional[int] = None,
         trunc_length: Optional[int] = None,
-        nonnegative: Optional[bool] = False,
+        nonnegative: bool = False,
         level: Optional[int] = None,
         algorithm: Optional[str] = "cg",
         covariance: Optional[str] = "shr",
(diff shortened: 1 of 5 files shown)


src/gluonts/ext/r_forecast/_hierarchical_predictor.py
```diff
@@ -102,13 +102,15 @@ def __init__(
...
         num_bottom_ts: int,
         nodes: List,
-        nonnegative: bool,
         method_name: str,
         fmethod: str,
-        period: int = None,
+        period: Optional[int] = None,
         trunc_length: Optional[int] = None,
-        params: Optional[Dict] = None,
+        nonnegative: Optional[bool] = False,
```
lostella: Why `Optional`? Shouldn’t this just be either `False` or `True`?

test/conftest.py
```diff
@@ -52,69 +52,79 @@ class HierarchicalTrainDatasets(NamedTuple):
...
-            [1, 1, 0, 0],
-            [0, 0, 1, 1],
-            [1, 0, 0, 0],
-            [0, 1, 0, 0],
-            [0, 0, 1, 0],
-            [0, 0, 0, 1],
-        ],
-    )
+def sine7():
+    def _sine7(
```
lostella, line 56: I think this can be avoided? With this change you call `sine7` to construct a function, you can have `sine7` just be that function, or am I missing something?

lostella (changes requested):

test/conftest.py
melopeo: The reason for this is that now I am adding tests for the parameter `nonnegative` to be either `True` or `False`. Previously the default value for this parameter was taken, i.e. `nonnegative = False`.

Hence, I am interested in having the `sine7` data with only positive values when `nonnegative = True`, and I thought that the proposed modification was a quick way to get this done, in particular considering that this function belongs to `conftest.py`.

Is there any suggestion or hint on how to better handle this? I will be happy to investigate it and update the code accordingly.

lostella: I see, fixtures are a bit weird; nevermind! We can revisit this later if necessary

src/gluonts/ext/r_forecast/_hierarchical_predictor.py
melopeo: Fixed!

test/conftest.py
melopeo: Thanks!

lostella (approved):

PR #2411, opened 2022-11-02
`TrainDatasets` vs. `TrainingDataset` and `TestData`

codingWhale13:
*Description of changes:*

The datasets provided by GluonTS come in the form of `TrainDatasets`. Recently (starting with #2223), `TrainingDataset` and `TestData` were introduced.

Bridging the gap is a bit tedious as we have to do something like this:

```
dataset = get_dataset("electricity")

prediction_length = dataset.metadata.prediction_length

training_dataset = split(dataset.train, offset=-prediction_length)[0]
test_dataset = split(dataset.test, offset=-prediction_length)[1].generate_instances(prediction_length=prediction_length)
```

To at least make the translation from `TrainDatasets` to `TestData` easier, maybe something like `from_TrainDatasets` (see commit) can be useful because `TrainDatasets` always split by an offset of `prediction_length`.

I'm not super convinced of this myself as it entangles the two approaches more rather than leading to clarity. But maybe this can start a discussion of how training and test data should be dealt with in the future.

Is the plan to switch entirely to the new way eventually?

### DIFF
--- a/src/gluonts/dataset/split.py
+++ b/src/gluonts/dataset/split.py
@@ -209,11 +209,11 @@ class TimeSeriesSlice:
 
 class AbstractBaseSplitter(ABC):
     """
-    Base class for all other splitter.
+    Base class for all other splitters.
     """
 
     @abstractmethod
-    def training_entry(self, entry: DataEntry) -> DataEntry:
+    def split_entry(self, entry: DataEntry) -> DataEntry:
         pass
 
     @abstractmethod
@@ -224,16 +224,16 @@ class AbstractBaseSplitter(ABC):
 
     def split(
         self, dataset: Dataset
-    ) -> Tuple["TrainingDataset", "TestTemplate"]:
+    ) -> Tuple["DatasetSplit", "TestTemplate"]:
         return (
-            TrainingDataset(dataset=dataset, splitter=self),
+            DatasetSplit(dataset=dataset, splitter=self),
             TestTemplate(dataset=dataset, splitter=self),
         )
 
-    def generate_training_entries(
+    def generate_split_entries(
         self, dataset: Dataset
     ) -> Generator[DataEntry, None, None]:
-        yield from map(self.training_entry, dataset)
+        yield from map(self.split_entry, dataset)
 
     def generate_test_pairs(
         self,
@@ -277,7 +277,7 @@ class OffsetSplitter(AbstractBaseSplitter):
 
     offset: int
 
-    def training_entry(self, entry: DataEntry) -> DataEntry:
+    def split_entry(self, entry: DataEntry) -> DataEntry:
         return TimeSeriesSlice(entry)[: self.offset]
 
     def test_pair(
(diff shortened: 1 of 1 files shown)


lostella:
I'm thinking, the class introduced in #2223 (currently `TrainingDataset`) could be called something more general since it could contain data to be used e.g. for validation rather than training. It's really a `SlicedDataset` or a `DatasetSplit` or something like that?

codingWhale13:
That's a good point because in the end, it's nothing more than another `Dataset` (the Protocol) that has the last few values omitted. I like `DatasetSplit`. This would lead to the following changes:

- `TrainigDataset` -> `DatasetSplit`
- `generate_training_entries` -> `generate_entry_splits`
- `training_entry` -> `split_entry` (This is especially nice as it implies an action.)

The "test" in `TestData`, `test_pair`should be okay to stay.

codingWhale13:
Here's another thought: Currently, the (Offset/Date)`Splitter` does two things at once:
- dividing the datasets into two parts ("training part" and "testing part")
- dividing the "testing part" entries into "input" and "label"

Maybe it makes sense to separate these two actions to give the user more control?
Then, we would allow for usage like this:
1. Split the dataset into "training" and "other" 
2. Split "other" into "validation" and "testing"
3. Split entries in "testing" into "input" and "label" parts

EDIT: I believe I have confused what the `Splitter` actually does. It doesn't separate the dataset into different groups of data entries but rather splits each data entry of a dataset. I initially thought, there were two different kinds of splitting going on ("horizontally" and "vertically", if you will) but that's not the case.

lostella:
> * `generate_training_entries` -> `generate_entry_splits`
> * `training_entry` -> `split_entry` (This is especially nice as it implies an action.)

I think the former could be `generate_split_entries` which would align with the latter

jaheba:
Reworking the provided datasets is something I also want see happening!

In my view we should probably store most of these datasets without splitting them into train and test. Instead, we can then create "scenarios", concrete splitting of the data into train/test, which we can materialise if we want to.

I guess some of these datasets have some scenarios attached to them, that is there is a canonical way to use them for benchmarking.

codingWhale13:
I looked a bit into how the datasets included in GluonTS are created. While doing that, I was not able to download any dataset that uses the `generate_forecasting_dataset` function in the [dataset_recipes](https://github.com/awslabs/gluonts/blob/d5d897cd02b3aaa499f3c3632d3a18863b1a24e7/src/gluonts/dataset/repository/datasets.py#L55).
Also, “taxi_30min” strangely sometimes could be downloaded and sometimes not (because metadata.json was saved in its own metadata folder, for some reason). Probably some user error on my side but it didn’t work out of the box...

Apart from that, I found that for some datasets, the number of training entries is equal to the number of test entries because the *training entries are the same as the testing entries*, just truncated to exclude the prediction length. These are all datasets generated by the following functions:

* generate_artificial_dataset
* generate_m3/m4_dataset
* generate_uber_dataset

This makes it pretty easy to put everything into a dataset and applying a "scenario" on it.

For most other datasets, the amount of test entries is usually larger than the amount of training entries. It's always doable to hide away the values to be predicted to get training data out of the "test data". Using the entries that are meant for training for testing however, might not be generally a good idea...

jaheba:
> For most other datasets, the amount of test entries is usually larger than the amount of training entries.

Could it be that we just slice multiple examples out of a given input time series?

Issue #2410, opened 2022-11-02
Add exogenous features to `statsforecast` model wrappers


lostella:
## Description
The wrapper for `statsforecast` models added in #2360 does not use any exogenous features (only the past target is passed on to the model). It would be good to allow other standard fields from data entries, like at least `feat_dynamic_real` (or the `past_` one) and `feat_static_real`, to be used.

## Reference
See `statsforecast`'s documentation on exogenous features: https://nixtla.github.io/statsforecast/examples/exogenous.html

PR #2423, opened 2022-11-06
initial support for statsforecast covariates

kashif:
fixes issue #2410

*Issue #, if available:*

*Description of changes:*


lostella (commented):
Hi @kashif I was actually thinking of doing this simpler: just forward any additional feature in the data to the model, without engineering anything there. In the end, these are supposed to be very thin wrappers, so not much should happen

kashif:
True... I was thinking the same but then i was wondering if one can use the date-time features like all the other deep learning models (apart from static covariates) and turns out it works as well... I don't know happy to close this...

lostella:
> I don't know happy to close this...

You could also adjust it to only take in features and passed them on to the model

lostella:
> I was thinking the same but then i was wondering if one can use the date-time features like all the other deep learning models (apart from static covariates) and turns out it works as well

True, but that can be done outside by the user. I think it makes sense to have minimal wrappers for external models. For example I don't think the wrapped R models do any feature engineering

kashif:
ok i'll adjust it!

kashif:
@lostella for some reason a few models are failing...

src/gluonts/ext/statsforecast.py
```diff
@@ -102,13 +103,42 @@ def __init__(
...
+            static_real_cov = entry[FieldName.FEAT_STATIC_REAL]
+
+            repeat_static_real_cov = static_real_cov[None, :].repeat(
+                length, axis=0
+            )
+
+            if covariates is not None:
+                covariates = np.hstack([covariates, repeat_static_real_cov])
+            else:
+                covariates = repeat_static_real_cov
```
lostella, line 145: You could append arrays to a list and then stack them in the end without the `if`

lostella (changes requested):
@kashif looks like tests are failing for predictor 1 (`AutoARIMAPredictor`) and datasets 2, 3, 5: what these have in common is the fact that features extend only as long as the target, in which case `X_future = None`. Probably this is not supported by `AutoARIMA` under the hood, so I would suggest sticking to the case where features extend to the whole prediction length as well. (This is anyway the unwritten assumption for `feat_dynamic_real`.)

src/gluonts/ext/statsforecast.py
kashif: thanks! let me fix!

PR #2435, opened 2022-11-11
Speed up `PandasDataset` for long dataframes

lostella:
*Issue #, if available:* #2363

*Description of changes:* This PR is similar in spirit to #2377, but without adding a dedicated class. Also strengthened tests a bit.

### DIFF
--- a/src/gluonts/dataset/pandas.py
+++ b/src/gluonts/dataset/pandas.py
@@ -239,28 +239,6 @@ def to_df(series):
     return series.to_frame(name="target")
 
 
-def series_to_dataframe(
-    series: Union[pd.Series, List[pd.Series], Dict[str, pd.Series]]
-) -> Union[pd.DataFrame, List[pd.DataFrame], Dict[str, pd.DataFrame]]:
-    if isinstance(series, list):
-        return list(map(to_df, series))
-    elif isinstance(series, dict):
-        return valmap(to_df, series)
-    return to_df(series)
-
-
-def is_series(series: Any) -> bool:
-    """
-    return True if ``series`` is ``pd.Series`` or a collection of
-    ``pd.Series``.
-    """
-    if isinstance(series, list):
-        return is_series(series[0])
-    elif isinstance(series, dict):
-        return is_series(list(series.values()))
-    return isinstance(series, pd.Series)
-
-
 def as_dataentry(
     data: pd.DataFrame,
     target: Union[str, List[str]],
--- a/src/gluonts/dataset/pandas.py
+++ b/src/gluonts/dataset/pandas.py
@@ -102,7 +102,9 @@ class PandasDataset:
     ignore_last_n_targets: int = 0
     unchecked: bool = False
     assume_sorted: bool = False
-    _pairs: Any = None
+    _pairs: Iterable[Tuple[Any, Union[pd.Series, pd.DataFrame]]] = field(
+        init=False
+    )
 
     def __post_init__(self) -> None:
         if isinstance(self.target, list) and len(self.target) == 1:
@@ -134,7 +136,7 @@ class PandasDataset:
         item_id, df = pair
 
         if isinstance(df, pd.Series):
-            df = df.to_frame(name="target")
+            df = df.to_frame(name=self.target)
 
         if self.timestamp:
             df.index = pd.PeriodIndex(df[self.timestamp], freq=self.freq)
@@ -212,7 +214,7 @@ class PandasDataset:
         return cls(dataframes=dataframe.groupby(item_id), **kwargs)
 
 
-def pair_with_item_id(obj: Union[Tuple, pd.DataFrame, pd.Series]) -> Tuple:
+def pair_with_item_id(obj: Union[Tuple, pd.DataFrame, pd.Series]):
     if isinstance(obj, tuple) and len(obj) == 2:
         return obj
     if isinstance(obj, (pd.DataFrame, pd.Series)):
--- a/src/gluonts/dataset/pandas.py
+++ b/src/gluonts/dataset/pandas.py
@@ -13,14 +13,26 @@
 
 from copy import deepcopy
 from dataclasses import dataclass, field
-from typing import Any, cast, Dict, Iterator, List, Optional, Union
+from typing import (
+    Any,
+    cast,
+    Collection,
+    Dict,
+    Iterator,
+    Iterable,
+    List,
+    Optional,
+    Union,
+    Tuple,
+)
 
 import pandas as pd
 from pandas.core.indexes.datetimelike import DatetimeIndexOpsMixin
-from toolz import valmap
+from toolz import first, valmap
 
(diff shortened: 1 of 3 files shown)


lostella:
I'm using the following example to evaluate speed (iterating twice since index checks are disabled after the first iteration)

```python
from pathlib import Path
from time import time
import pandas as pd
from tqdm import tqdm

df = pd.read_parquet(Path(__file__).resolve().parent / "long_df_sample.parquet")

t0 = time()
ds = PandasDataset.from_long_dataframe(
    dataframe=df,
    item_id="item_id",
    timestamp="timestamp",
    freq="M",
)
t1 = time()
print(f"construction time: {t1 - t0}")

N = 2

t0 = time()
for _ in range(N):
    for entry in tqdm(ds):
        pass
t1 = time()
print(f"average iteration time: {(t1 - t0)/N}")
```

Before this PR:

```
construction time: 32.1102340221405
100%|███████████████████████████████████████████████████████████████████████████| 25000/25000 [00:04<00:00, 6237.54it/s]
100%|██████████████████████████████████████████████████████████████████████████| 25000/25000 [00:01<00:00, 17872.74it/s]
```

After this PR:

```
construction time: 0.44616174697875977
100%|███████████████████████████████████████████████████████████████████████████| 25000/25000 [00:11<00:00, 2164.93it/s]
100%|███████████████████████████████████████████████████████████████████████████| 25000/25000 [00:08<00:00, 2803.81it/s]
```

So construction is much faster (not as fast as in #2377, since there the `groupby` operation is delayed), but iteration is slower. This is because the original code essentially pre-processes and caches all dataframes at construction time: with this PR, one can get the same by simply caching the dataset explicitly using `list`, in case dataset iteration is a bottleneck.

lostella:
If I cache the dataset at construction time with `ds = list(ds)`, with the PR code I get

```
construction + caching time: 11.770990133285522
100%|████████████████████████████████████████████████████████████████████████| 25000/25000 [00:00<00:00, 2887525.47it/s]
100%|████████████████████████████████████████████████████████████████████████| 25000/25000 [00:00<00:00, 2800833.38it/s]
```

which is anyway much faster than the original code.

src/gluonts/dataset/pandas.py
```diff
@@ -179,18 +200,48 @@ def from_long_dataframe(
...
+        return len(self.data)
+
+
+def to_pairs(data):
+    if isinstance(data, (pd.DataFrame, pd.Series)):
+        return [(None, data)]
+    if isinstance(data, dict):
+        return data.items()
+    if isinstance(data, (Iterable, Collection)):
+        first_element = first(data)
```
jaheba: You might want to use `toolz.peek` here.

src/gluonts/dataset/pandas.py
```diff
@@ -81,53 +100,59 @@ class PandasDataset:
...
+
+        assert isinstance(self.dataframes, SizedIterable)
+
         if isinstance(self.dataframes, dict):
-            self._dataframes = list(self.dataframes.items())
-        elif isinstance(self.dataframes, list):
-            self._dataframes = [(None, df) for df in self.dataframes]
-        else:  # case single dataframe
-            self._dataframes = [(None, self.dataframes)]
+            self._pairs: Any = self.dataframes.items()
```
jaheba: Anything that is assigned to should be listed as a definition.

src/gluonts/dataset/pandas.py
```diff
@@ -179,35 +207,22 @@ def from_long_dataframe(
...
 
-def series_to_dataframe(
-    series: Union[pd.Series, List[pd.Series], Dict[str, pd.Series]]
-) -> Union[pd.DataFrame, List[pd.DataFrame], Dict[str, pd.DataFrame]]:
-    def to_df(series):
-        assert isinstance(
-            series.index, DatetimeIndexOpsMixin
-        ), "series index has to be a DatetimeIndex."
-        return series.to_frame(name="target")
+def pair_with_item_id(obj):
```
jaheba: Type annotation is missing.

I think this should use a `NamedTuple`. I think it's generally bad style to rely on some implied meaning through ordering.

src/gluonts/dataset/pandas.py
lostella: isn't it pretty much the same for collections or iterables?

src/gluonts/dataset/pandas.py
lostella: not sure I get it

src/gluonts/dataset/pandas.py
jaheba: I guess we rely on the re-iterable aspect?

src/gluonts/dataset/pandas.py
jaheba: There should be a type annotation on the class level

src/gluonts/dataset/pandas.py
```diff
@@ -81,53 +100,60 @@ class PandasDataset:
...
-        elif isinstance(self.dataframes, list):
-            self._dataframes = [(None, df) for df in self.dataframes]
-        else:  # case single dataframe
-            self._dataframes = [(None, self.dataframes)]
-
-        for i, (item_id, df) in enumerate(self._dataframes):
-            if self.timestamp:
-                df = df.set_index(keys=self.timestamp)
+        if isinstance(self.dataframes, (pd.Series, pd.DataFrame)):
+            self._dataframes = [self.dataframes]
```
jaheba: Why have one with and one without underscore?

src/gluonts/dataset/pandas.py
lostella: Yes

src/gluonts/dataset/pandas.py
lostella: Long story short: some test (`test_hts_to_dataset`) is accessing the `dataframes` attribute and was breaking with my change, and I'm trying to avoid touching tests at all (in the sense of modifying existing assertions). But I can also do that.

src/gluonts/dataset/pandas.py
```diff
@@ -62,13 +73,21 @@ class PandasDataset:
...
+        (Default: ``False``)
+    assume_sorted
+        Whether to assume that indexes are sorted by time, and skip sorting.
+        (Default: ``False``)
     """
 
     dataframes: Union[
         pd.DataFrame,
         pd.Series,
-        List[pd.DataFrame],
```
jaheba, line 70: I think we can make use of TypeVar here: https://stackoverflow.com/questions/59933946/difference-between-typevart-a-b-and-typevart-bound-uniona-b

jaheba (approved):

jaheba:
Do we know what impact on memory consumption this change has?

lostella:
> Do we know what impact on memory consumption this change has?

I used [scalene](https://github.com/plasma-umass/scalene) with Python 3.8.13, on `dev` vs this PR, using the data from #2363. Max memory consumption is as follows:
* `dev`: 575.1 MB
* this PR: 315.0 MB

PR #2439, opened 2022-11-15
Add cdf and icdf methods for StudentT distribution

shchur:
*Description of changes:*
- Implement `cdf` and `icdf` methods for the StudentT distribution. This enables quantile computation for the StudentT `DistributionForecast` (e.g., for `SimpleFeedForwardEstimator`).
- The new `cdf` and `icdf` methods use `scipy.stats.t` under the hood, so  require moving data to CPU and are not differentiable.
- Move `StudentTDistribtuionOutput` to `gluonts.torch.distributions.studentT`.
- Add a test that ensures that torch and scipy versions have the same parametrization.

**Please tag this pr with at least one of these labels to make our release process faster:** enhancement

### DIFF
--- a/src/gluonts/torch/distributions/studentT.py
+++ b/src/gluonts/torch/distributions/studentT.py
@@ -22,8 +22,8 @@ from .distribution_output import DistributionOutput
 
 
 class StudentT(TorchStudentT):
-    """Student's t-distribution parametrized by degree of freedom `df`, mean `loc`
-    and scale `scale`.
+    """Student's t-distribution parametrized by degree of freedom `df`,
+    mean `loc` and scale `scale`.
 
     Based on torch.distributions.StudentT, with added `cdf` and `icdf` methods.
     """
--- a/src/gluonts/torch/distributions/studentT.py
+++ b/src/gluonts/torch/distributions/studentT.py
@@ -11,7 +11,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-from typing import Dict, Tuple
+from typing import Dict, Tuple, Union
 
 import torch
 import torch.nn.functional as F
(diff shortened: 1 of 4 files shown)


kashif:
I could try to add the appropriate methods to pytorch?

kashif:
ah damn no `betainc` in pytorch i believe...

shchur:
> ah damn no `betainc` in pytorch i believe...

Yeah, pytorch doesn't have the necessary special functions, unfortunately

lostella:
> ah damn no `betainc` in pytorch i believe...

Sounds like that needs to be added too 🙃

kashif:
tensorflow uses the implementation from `eigen` i forget if that is available on the pytorch side...

src/gluonts/torch/distributions/studentT.py
```diff
@@ -0,0 +1,71 @@
...
+        df = self.df.detach().cpu().numpy()
+        loc = self.loc.detach().cpu().numpy()
+        scale = self.scale.detach().cpu().numpy()
+        self.scipy_student_t = ScipyStudentT(df=df, loc=loc, scale=scale)
+
+    def cdf(self, value: torch.Tensor) -> torch.Tensor:
+        if self._validate_args:
+            self._validate_sample(value)
+        result = self.scipy_student_t.cdf(value.detach().cpu().numpy())
+        return torch.tensor(result, device=value.device, dtype=value.dtype)
```
kashif, line 50: shall we also `require_grads=False` this tensor and the one below?

shchur: I thought this was already the [default](https://pytorch.org/docs/stable/generated/torch.tensor.html), are there some cases where the output ends up having `requires_grad=True`?

kashif: you are right! my bad

lostella (approved):
Awesome 👏🏻 thanks @shchur!

PR #2441, opened 2022-11-16
Speed up `PandasDataset` further

lostella:
*Description of changes:* More speedup following #2435. Testing on Python 3.8.13 with the data from #2363, using the following code:

```python
from pathlib import Path
from time import time
import pandas as pd
from tqdm import tqdm

df = pd.read_parquet(Path(__file__).resolve().parent / "long_df_sample.parquet")

t0 = time()
ds = PandasDataset.from_long_dataframe(
    dataframe=df,
    item_id="item_id",
    timestamp="timestamp",
    freq="M",
)
t1 = time()
print(f"construction time: {t1 - t0}")

N = 3

t0 = time()
for _ in range(N):
    for entry in tqdm(ds):
        pass
t1 = time()
print(f"average iteration time: {(t1 - t0)/N}")
```

Before the PR:

```
construction time: 0.3739509582519531
100%|█████████████████████████████████████████████| 25000/25000 [00:10<00:00, 2294.39it/s]
100%|█████████████████████████████████████████████| 25000/25000 [00:06<00:00, 3660.79it/s]
100%|█████████████████████████████████████████████| 25000/25000 [00:06<00:00, 3618.71it/s]
average iteration time: 8.413320620854696
```

After the PR:

```
construction time: 0.3727140426635742
100%|█████████████████████████████████████████████| 25000/25000 [00:08<00:00, 2825.26it/s]
100%|█████████████████████████████████████████████| 25000/25000 [00:05<00:00, 4383.24it/s]
100%|█████████████████████████████████████████████| 25000/25000 [00:05<00:00, 4380.71it/s]
average iteration time: 6.958552281061809
```

### DIFF
--- a/src/gluonts/dataset/pandas.py
+++ b/src/gluonts/dataset/pandas.py
@@ -27,6 +27,7 @@ from typing import (
 
 import pandas as pd
 import numpy as np
+from pandas.core.indexes.datetimelike import DatetimeIndexOpsMixin
 from toolz import first
 
 from gluonts.dataset.common import DataEntry, _as_period
@@ -183,13 +184,7 @@ class PandasDataset:
 
     @classmethod
     def from_long_dataframe(
-        cls,
-        dataframe: pd.DataFrame,
-        item_id: str,
-        *,
-        timestamp: str,
-        freq: str,
-        **kwargs
+        cls, dataframe: pd.DataFrame, item_id: str, **kwargs
     ) -> "PandasDataset":
         """
         Construct ``PandasDataset`` out of a long dataframe.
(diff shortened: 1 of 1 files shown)


test/dataset/test_pandas.py
```diff
@@ -228,16 +228,3 @@ def test_long_csv_3M():
         assert entry["start"] == expected_entry["start"]
         assert entry["item_id"] == expected_entry["item_id"]
         assert np.allclose(entry["target"], expected_entry["target"])
-
-    with io.StringIO(data) as fp:
-        ds = pandas.PandasDataset.from_long_dataframe(
-            pd.read_csv(fp, index_col="timestamp"),
```
lostella: I removed this part of the test since I don't think it's realistic to load a long dataframe this way, since timestamps can be repeated: it only makes sense to set this column as index after the `groupby` in general.

lostella:
I think this requires improving tests before it can be merged: for each use case for `PandasDataset`, we should assert that each dataset element is exactly what we expect (field names, types, shapes).

test/dataset/test_pandas.py
jaheba: An index in pandas does not need to be unique.

jaheba (approved):

PR #2443, opened 2022-11-17
Change schema.Type to behave like invokable types.

jaheba:
With this change, the types behave like other types in Python, e.g one uses `int(x)` instead of `int.apply(x)`.


lostella:
breaking?

lostella (approved):

PR #2462, opened 2022-11-25
Fix pandas issue with inferring start of X frequency.

### LINKED ISSUE #2461: GluonTS example airpassengers dataset not found
am trying to run the GluonTS example code, going through some struggle to install the libraries, now I get the following error:

`FileNotFoundError: C:\Users\abcde\.mxnet\gluon-ts\datasets\airpassengers\test`

The `C:\Users\abcde\.mxnet\gluon-ts\datasets\airpassengers\` does exist but contains only `train` folder. Have tried reinstalling but to no avail. Any ideas how to fix this and run the example, even if finding the dataset in correct format elsewhere?

jaheba:
Closes: #2461

*Description of changes:*

### DIFF
--- a/src/gluonts/dataset/pandas.py
+++ b/src/gluonts/dataset/pandas.py
@@ -222,10 +222,18 @@ def pair_with_item_id(obj: Union[Tuple, pd.DataFrame, pd.Series]):
     raise ValueError("input must be a pair, or a pandas Series or DataFrame.")
 
 
-def infer_freq(index: pd.Index):
+def infer_freq(index: pd.Index) -> str:
     if isinstance(index, pd.PeriodIndex):
         return index.freqstr
-    return pd.infer_freq(index)
+
+    freq = pd.infer_freq(index)
+    # pandas likes to infer the `start of x` frequency, however when doing
+    # df.to_period("<x>S"), it fails, so we avoid using it. It's enough to
+    # remove the trailing S, e.g `MS` -> `M
+    if len(freq) > 1 and freq.endswith("S"):
+        return freq[:-1]
+
+    return freq
 
 
 def extract_dynamic_array(


src/gluonts/dataset/pandas.py
```diff
@@ -222,10 +222,18 @@ def pair_with_item_id(obj: Union[Tuple, pd.DataFrame, pd.Series]):
...
-    return pd.infer_freq(index)
+
+    freq = pd.infer_freq(index)
+    # pandas likes to infer the `start of x` frequency, however when doing
+    # df.to_period("<x>S"), it fails, so we avoid using it. It's enough to
+    # remove the trailing S, e.g `MS` -> `M
+    if len(freq) > 1 and freq.endswith("S"):
+        return freq[:-1]
+
+    return freq
```
melopeo, line 237: Reading in about [offset-aliases](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases), it seems that `MS` and `M` have different connotations, meaning:

- M : month end frequency
- MS : month start frequency

Do we know if we want to keep these differences or just oversee them? 

Nevertheless, both of them are monthly frequency. I just wonder how critical this difference can be.

jaheba: For periods, only `M` makes sense, since there is no day information unlike for timestamps. Pandas throws an error if you try to create a `Period` with an `xS` frequency.

melopeo: I see, I missed the fact that this PR was having `pd.Period` in mind!

melopeo (approved):

src/gluonts/dataset/pandas.py
jaheba: We no longer use `pd.Timestamp`.

Issue #2215, opened 2022-08-15
Weighted average for custom evaluation function


eugeneteoh:
## Description

`gluonts.evaluation.Evaluator` provides a parameter, `custom_eval_fn` which allows defining custom evaluation functions. However, the global (across multiple TS) and local (within TS) aggregations are limited to mean & sum, and mean & median, respectively. My custom metric involves a weighted average-based aggregation, which the current implementation does not provide. It would be useful to have a way to define more complex custom evaluation functions.

eugeneteoh:
Seems to be relevant to https://github.com/awslabs/gluon-ts/issues/2048

lostella:
@eugeneteoh exactly, thanks for the issue: flexibility of the evaluator is something we're currently working on, I'll keep you posted when there are updates

PR #2470, opened 2022-11-29
Add `weights` option to `ev`'s Sum and Mean

codingWhale13:
*Description of changes:*
#2215 mentions that a weighted average might be a desirable feature for the evaluation.

Supporting this in the new `ev` module is rather simple if we just multiply by a `weights` array before applying aggregations (see first commit of this PR).

On top of this, calculating **normalized** weighted averages might also be useful. This however would in my opinion add unreasonable complexity to the aggregation code and should not be implemented natively by the `ev` module.


melopeo (commented):
Can we add a test for this new feature?

PR #2476, opened 2022-12-06
Include loss computation in torch DeepAR module, decouple MQF2

lostella:
*Description of changes:* This moves the loss computation to the `DeepARModel` class. This way the model is more usable in its own right, and can be used for density estimation directly. A few changes are added that clean up the handling of shapes in the class. In particular, the model is now forced to deal with univariate data (`distr_output.event_dim == 1`), as it was anyway broken for other cases (for example, arrays are sliced along the `-1` axis irrespectively of `distr_output.event_dim`, which was wrong).

### DIFF
--- a/src/gluonts/torch/util.py
+++ b/src/gluonts/torch/util.py
@@ -131,7 +131,7 @@ def lagged_sequence_values(
         end_index = -lag_index if lag_index > 0 else None
         lags_values.append(
             slice_along_dim(
-                full_sequence, dim=dim, slc=slice(begin_index, end_index)
+                full_sequence, dim=dim, slice_=slice(begin_index, end_index)
             )
         )
 
--- a/src/gluonts/torch/util.py
+++ b/src/gluonts/torch/util.py
@@ -115,8 +115,7 @@ def lagged_sequence_values(
     Returns
     -------
     Tensor
-        A tensor of shape ``(N, T, L)``: if ``I = len(indices)``,
-        and ``sequence.shape = (N, T, C)``, then ``L = C * I``.
+        A tensor of shape (*sequence.shape, len(indices)).
     """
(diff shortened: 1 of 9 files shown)


lostella:
Marking this as breaking since it slightly changes the interface of `lagged_sequence_values`: add the mandatory `dim` argument, avoid the reshaping of the output.

There are other questionable reshapings going on here and there (see the newly added test) that we could consider getting rid of.

src/gluonts/torch/util.py
```diff
@@ -104,33 +105,38 @@ def lagged_sequence_values(
...
     prior_sequence
         Tensor containing the input sequence prior to the time range for
-        which the output is required (shape: ``(N, H, C)``).
+        which the output is required.
     sequence
         Tensor containing the input sequence in the time range where the
-        output is required (shape: ``(N, T, C)``).
+        output is required.
+    dim
+        Time dimension.
```
lostella, line 113: Need to fix these

src/gluonts/torch/util.py
```diff
@@ -139,3 +145,24 @@ def __init__(self, iterable):
...
+    idx[dim] = slc
+    return a[idx]
+
+
+def repeat_along_dim(a, dim, repeats):
+    if repeats == 1:
+        return a
+    r = [1] * len(a.shape)
+    r[dim] = repeats
+    return a.repeat(*r)
```
lostella, line 172: This is allocating, see https://pytorch.org/docs/stable/generated/torch.Tensor.repeat.html

Can probably get the same without allocations using `unsqueeze`, `expand`, and `view`.

jaheba (dismissed):

jaheba (approved):

Issue #2477, opened 2022-12-07
DeepVARHierarchicalNetwork does not respect context of trainer


ziqin8:
## Description
DeepVARHierarchicalNetwork and DeepVARHierarchicalEstimator has 2 `mx.nd.array` fields `A` and `M`. They are created on `mx.current_context()` instead of `estimator.trainer.ctx`, which leads to errors in `estimator.train` and `estimator.predict` (samples on trainer.ctx, but A and M on different ctx)

## To Reproduce
(Please provide minimal example of code snippet that reproduces the error. For existing examples, please provide link.)

Code below is same as in [Hierarchical Model Tutorial](https://ts.gluon.ai/dev/tutorials/forecasting/hierarchical_model_tutorial.html), but the context of trainer is changed to `mx.gpu()`
```python
import pandas as pd
from gluonts.dataset.hierarchical import HierarchicalTimeSeries

# Load (only!) the time series at the bottom level of the hierarchy.
ts_at_bottom_level_csv = (
    "https://gist.githubusercontent.com/rshyamsundar/39e57075743537c4100a716a7b7ec047/"
    "raw/f02f5aeadad73e3f3e9cf54478606d3507826492/example_bottom_ts.csv"
)

# Make sure the dataframe has `PeriodIndex` by explicitly casting it to `PeriodIndex`.
ts_at_bottom_level = pd.read_csv(
    ts_at_bottom_level_csv,
    index_col=0,
    parse_dates=True,
).to_period()

# Load the aggregation matrix `S`.
S_csv = (
    "https://gist.githubusercontent.com/rshyamsundar/17084fd1f28021867bcf6f2d69d9b73a/raw/"
    "32780ca43f57a78f2d521a75e73b136b17f34a02/example_agg_mat.csv"
)
S = pd.read_csv(S_csv).values

hts = HierarchicalTimeSeries(
    ts_at_bottom_level=ts_at_bottom_level,
    S=S,
)

dataset = hts.to_dataset()

from gluonts.mx.model.deepvar_hierarchical import DeepVARHierarchicalEstimator
from gluonts.mx.trainer import Trainer
import mxnet as mx

prediction_length = 24
estimator = DeepVARHierarchicalEstimator(
    freq=hts.freq,
    prediction_length=prediction_length,
    # ONLY CHANGE TO ORIGINAL TUTORIAL
    trainer=Trainer(epochs=2, ctx=mx.gpu()),
    target_dim=hts.num_ts,
    S=S,
[...]

PR #2478, opened 2022-12-07
Fix: use right context in DeepVARHierarchicalEstimator

ziqin8:
*Issue #, if available:*
#2477 

*Description of changes:*
create ndarray A and M with context same as trainer.ctx


lostella (approved):
Looks good @c3-ziqin, thanks for catching this and fixing it! ;-)

PR #2482, opened 2022-12-09
Remove serde dump_code/load_code.

jaheba:
In addition to `serde.encode` which encodes an object into a json-serializable form, `serde.dump_code` returns a string which is something like the `repr` of the object.

I don't think there is much value in this, since it's error prone and basically never used. Thus, this PR removes `dump_code`.


test/transform/test_transform.py
```diff
@@ -1046,13 +1046,10 @@ def make_dataset(N, train_length):
...
     t = fqname_for(x.__class__)
     y = load_json(dump_json(x))
-    z = load_code(dump_code(x))
+
     assert dump_json(x) == dump_json(
         y
     ), f"Code serialization for transformer {t} does not work"
-    assert dump_code(x) == dump_code(
-        z
-    ), f"JSON serialization for transformer {t} does not work"
```
lostella, line 1055: Assertion messages here appeared to be swapped?

lostella (approved):

test/transform/test_transform.py
jaheba: Not sure I understand

PR #2492, opened 2022-12-13
Fix dataclass handling of member inheritance.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/test/core/test_serde_dataclass.py
+++ b/test/core/test_serde_dataclass.py
@@ -60,3 +60,6 @@ def test_dataclass_subtype():
     x = X(a=B())
 
     assert isinstance(x.a, B)
+
+    x2 = serde.decode(serde.encode(x))
+    assert isinstance(x2.a, B)
--- a/test/core/test_serde_dataclass.py
+++ b/test/core/test_serde_dataclass.py
@@ -13,6 +13,8 @@
 
 from typing import List
 
+from pydantic import BaseModel
+
 from gluonts.core import serde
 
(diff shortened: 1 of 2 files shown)


test/core/test_serde_dataclass.py
```diff
@@ -33,9 +35,31 @@ def __eventually__(self, cardinality):
...
     assert est.cardinality == [1]
+
+
+def test_dataclass_subtype():
+    x = X(a=B())
+
+    assert isinstance(x.a, B)
+
+    x2 = serde.decode(serde.encode(x))
+    assert isinstance(x2.a, B)
```
lostella, line 65: Interesting, `decode` was instantiating this back to an `A` object?

lostella (approved):

test/core/test_serde_dataclass.py
jaheba: Yes, the problem is this:

```py
In [10]: from pydantic import BaseModel
    ...: 
    ...: 
    ...: class A(BaseModel):
    ...:     pass
    ...: 
    ...: class B(A):
    ...:     pass
    ...: 
    ...: 
    ...: class M(BaseModel):
    ...:     a: A
    ...: 
    ...: m = M(a=B())
    ...: 
    ...: m.dict()
Out[10]: {'a': {}}
```

When deserialising, we want to reconstruct some `A` from `{}`, which then returns an `A` instead of the `B`.

PR #2493, opened 2022-12-14
Add `gluonts.util`.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/util.py
+++ b/src/gluonts/util.py
@@ -13,6 +13,7 @@
 
 from dataclasses import dataclass
 from typing import Tuple, Type, Optional
+from typing_extensions import Literal
 
 import numpy as np
 
@@ -54,8 +55,8 @@ def pad_and_slice(
     *,
     axis=0,
     pad_value=0,
-    pad_to: str = "left",
-    take_from: str = "left",
+    pad_to: Literal["left", "right"],
+    take_from: Literal["left", "right"],
 ) -> Tuple[np.ndarray, np.ndarray]:
     """Ensure `data` has correct `size` along `axis`. Returns a tuple of
     resized data, as well as a padding indicator, where `1` marks a padded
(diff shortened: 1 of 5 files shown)


src/gluonts/util.py
```diff
@@ -0,0 +1,102 @@
...
+@dataclass
+class AxisView:
+    data: np.ndarray
+    axis: int
+
+    def __getitem__(self, index):
+        slices = [slice(None)] * self.data.ndim
+        slices[self.axis] = index
+
+        return self.data[tuple(slices)]
```
lostella, line 49: Will this work? For example what if you want to turn this into a `torch.Tensor`? I'm wondering if a function returning `data[tuple(slices)]` would be better

src/gluonts/util.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def pad_and_slice(
+    data: np.ndarray,
+    size: int,
+    *,
+    axis=0,
+    pad_value=0,
+    pad_to: str = "left",
+    take_from: str = "left",
```
lostella: I'm wondering whether both options are needed: wouldn't one want to slice from one side and pad to the other one?

lostella (commented):
Maybe tests would be good. Especially `pad_and_slice`

src/gluonts/util.py
jaheba: No idea. For the instance splitting we indeed want to pad from the left and take from the right.

Having two parameters make it very explicit though. Maybe we change the default for `take_from` to `right`?

src/gluonts/util.py
jaheba: Can you expand on what you mean here?

src/gluonts/util.py
lostella: Maybe we leave both `pad_to` and `take_from` without default? My worry is that if one changes only one of them they may get unexpected results, maybe they should be required arguments

src/gluonts/util.py
lostella: ~What happens if you do `torch.Tensor(v)` where `v: AxisView`?~

lostella: Nevermind, I got it now

PR #2500, opened 2022-12-16
Fix `DateSplitter` for multiples of base frequencies

### LINKED ISSUE #2499: Incorrect test split start point and length when using frequency with unit >1
## Description
(A clear and concise description of what the bug is.)

The behaviour of ```split``` is not as expected when using a dataset and split point with frequency that has larger unit than 1, e.g. 30T, 6M, etc. This causes the training set length and test set lengths to be incorrect.

This likely stems from the ```to_integer_slice``` function not behaving as expected when sliced with ```pd.Period``` with frequency that is larger than 1 unit, e.g. 30T, 6M, etc. Related issue: #1833 but code seems to have changed significantly since.

Issue seems to be due to this. Could consider doing something like ```start_offset = (slice_.start - start).n // pd.tseries.frequencies.to_offset(start.freq).n```.
https://github.com/awslabs/gluonts/blob/cac5e6b91cbd545c0811b20849290e9215c1b822/src/gluonts/dataset/split.py#L111
https://github.com/awslabs/gluonts/blob/cac5e6b91cbd545c0811b20849290e9215c1b822/src/gluonts/dataset/split.py#L122

## To Reproduce

```python
import pandas as pd
import numpy as np

n = 100
split_n = 20
freq = '30T'
date = "2015-04-07 00:00:00"

df = pd.DataFrame(np.random.randn(n), index=pd.period_range(date, periods=n, freq=freq))
dataset = PandasDataset(df, target=0)
split_date = pd.Period(date, freq=freq) + pd.to_timedelta(freq) * (split_n - 1)
training_dataset, test_template = split(
    dataset, date=split_date
)
print(df.index[0])
print(split_date)
print(df.index[-1])
print(next(iter(training_dataset))['target'].shape)
print(next(iter(test_template.generate_instances(prediction_length=6)))[0]['target'].shape)
print(next(iter(test_template.generate_instances(prediction_length=6)))[1]['target'].shape)
```

```
2015-04-07 00:00
2015-04-07 09:30
2015-04-09 01:30
(100,)
(100,)
(0,)
```

## Expected message
```
2015-04-07 00:00
2015-04-07 09:30
2015-04-09 01:30
(20,)
(20,)
(6,)
```

## Environment
- GluonTS version: 0.11.4

lostella:
*Issue #, if available:* Fixes #2499 

*Description of changes:*
- Simplify the logic of date splitting, by using the newly introduced `periods_between`
- Render `to_integer_slice` pointless (can be removed in a separate breaking change)
- Also render `TimeSeriesSlice` pointless (`slice_data_entry` is sufficient)
- Extend and consolidate tests

### DIFF
--- a/src/gluonts/dataset/split.py
+++ b/src/gluonts/dataset/split.py
@@ -76,9 +76,23 @@ from dataclasses import dataclass
 from typing import Generator, Optional, Tuple
 
 import pandas as pd
+from pandas.tseries.frequencies import to_offset
 
 from gluonts.dataset import Dataset, DataEntry
 from gluonts.dataset.field_names import FieldName
+
+
+def periods_between(
+    start: pd.Period,
+    end: pd.Period,
+) -> int:
+    """
+    Counts how many periods fit between ``start`` and ``end``
+    (inclusive).
+
+    The frequency is taken from ``start``.
+    """
+    return ((end - start).n // to_offset(start.freq).n) + 1
 
 
 def to_positive_slice(slice_: slice, length: int) -> slice:
@@ -277,7 +291,7 @@ class OffsetSplitter(AbstractBaseSplitter):
     offset: int
 
     def training_entry(self, entry: DataEntry) -> DataEntry:
-        return TimeSeriesSlice(entry)[: self.offset]
+        return slice_data_entry(entry, slice(None, self.offset))
 
     def test_pair(
         self, entry: DataEntry, prediction_length: int, offset: int = 0
@@ -286,17 +300,19 @@ class OffsetSplitter(AbstractBaseSplitter):
         if self.offset < 0 and offset_ >= 0:
             offset_ += len(entry)
         if offset_ + prediction_length:
-            return (
-                TimeSeriesSlice(entry, prediction_length)[:offset_],
-                TimeSeriesSlice(entry, prediction_length)[
-                    offset_ : offset_ + prediction_length
-                ],
-            )
+            input_slice = slice(None, offset_)
+            label_slice = slice(offset_, offset_ + prediction_length)
         else:
-            return (
-                TimeSeriesSlice(entry, prediction_length)[:offset_],
-                TimeSeriesSlice(entry, prediction_length)[offset_:],
-            )
+            input_slice = slice(None, offset_)
+            label_slice = slice(offset_, None)
+        return (
+            slice_data_entry(
+                entry, input_slice, prediction_length=prediction_length
+            ),
+            slice_data_entry(
+                entry, label_slice, prediction_length=prediction_length
+            ),
+        )
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


src/gluonts/dataset/split.py
```diff
@@ -81,6 +81,13 @@
 from gluonts.dataset.field_names import FieldName
 
 
+def ceildiv(a, b):
+    """
+    Ceiling division between ``a`` and ``b``.
+    """
+    return -(a // -b)
```
jaheba: What sorcery is this?

jaheba: I guess it's the same as

```py
sum(divmod(a, b))
```

test/dataset/test_split.py
```diff
@@ -238,3 +303,137 @@ def test_split(dataset, date, offset, windows, distance, max_history):
...
+#         {
+#             "start": pd.Period("2015-04-07 00:00:00", freq="30T"),
+#             "target": np.random.randn(100),
+#         },
+#         pd.Period("2015-04-07 10:00:00", freq="30T"),
+#         (21,),
+#         6,
+#         (21,),
+#         (6,),
+#     )
```
jaheba:
```suggestion
# if __name__ == "__main__":
#     test_split_date_multiple_base_freq(
#         {
#             "start": pd.Period("2015-04-07 00:00:00", freq="30T"),
#             "target": np.random.randn(100),
#         },
#         pd.Period("2015-04-07 10:00:00", freq="30T"),
#         (21,),
#         6,
#         (21,),
#         (6,),
#     )
```

src/gluonts/dataset/split.py
jaheba: Ah, no, of course not.

src/gluonts/dataset/split.py
```diff
@@ -108,7 +115,10 @@ def to_integer_slice(slice_: slice, start: pd.Period) -> slice:
         return slice_
 
     if isinstance(slice_.start, pd.Period):
-        start_offset = (slice_.start - start).n
+        start_offset = ceildiv(
+            (slice_.start - start).n,
+            pd.tseries.frequencies.to_offset(start.freq).n,
+        )
```
jaheba: I think importing it like `from .. import to_offset` makes this easier to read.

src/gluonts/dataset/split.py
lostella: Python's `//` is the floor division, that's the usual integer division where `7 // 2 == 3` for example.

This is the ceiling division, where `7 // 2 == 4`, that is the quotient is rounded up (if it's not a whole number).

src/gluonts/dataset/split.py
```diff
@@ -322,11 +334,12 @@ def training_entry(self, entry: DataEntry) -> DataEntry:
     def test_pair(
         self, entry: DataEntry, prediction_length: int, offset: int = 0
     ) -> Tuple[DataEntry, DataEntry]:
-        date = self.date.asfreq(entry[FieldName.START].freq)
         return (
-            TimeSeriesSlice(entry, prediction_length)[: date + offset],
+            TimeSeriesSlice(entry, prediction_length)[: self.date + offset],
             TimeSeriesSlice(entry, prediction_length)[
```
lostella: I would actually get rid of the `TimeSeriesSlice` class, but maybe in a separate PR (so this one is not breaking)

src/gluonts/dataset/split.py
lostella: Addressed

src/gluonts/dataset/split.py
jaheba: +1

jaheba (approved):

src/gluonts/dataset/split.py
```diff
@@ -76,11 +76,25 @@
...
+    start: pd.Period,
+    end: pd.Period,
+) -> int:
+    """
+    Counts how many periods fit between ``start`` and ``end``
+    (inclusive).
+
+    The frequency is taken from ``start``.
+    """
+    return ((end - start).n // to_offset(start.freq).n) + 1
```
jaheba, line 95: Why do you need to call to_offset here?

src/gluonts/dataset/split.py
```diff
@@ -76,11 +76,25 @@
...
 from gluonts.dataset import Dataset, DataEntry
 from gluonts.dataset.field_names import FieldName
 
 
+def periods_between(
+    start: pd.Period,
+    end: pd.Period,
+) -> int:
+    """
+    Counts how many periods fit between ``start`` and ``end``
```
jaheba, line 90: Should be "count ...". 

I think an example would be great

src/gluonts/dataset/split.py
lostella: Yeah I think this can be removed

PR #2513, opened 2022-12-20
Fix `QuantileForecast.quantile` in case only mean is stored

### LINKED ISSUE #2498: AssertionError while using ets from wrapper.
## Description
ets from statforecast model producing error.

## To Reproduce

```python
from gluonts.ext.statsforecast import ETSPredictor
from gluonts.evaluation.backtest import backtest_metrics
dataset = get_dataset("pedestrian_counts")
predictor = ETSPredictor(prediction_length=dataset.metadata.prediction_length)
agg_metrics, agg_perite = backtest_metrics(dataset.test, predictor)

```

```
RemoteTraceback                           Traceback (most recent call last)
RemoteTraceback: 
"""
  File "/usr/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/usr/lib/python3.7/multiprocessing/pool.py", line 44, in mapstar
    return list(map(*args))
  File "/home/bhaskar/Desktop/gluonts/src/gluonts/evaluation/_base.py", line 56, in worker_function
    return evaluator.get_metrics_per_ts(ts, forecast)
  File "/home/bhaskar/Desktop/gluonts/src/gluonts/evaluation/_base.py", line 343, in get_metrics_per_ts
    median_fcst = forecast.quantile(0.5)
  File "/home/bhaskar/Desktop/gluonts/src/gluonts/model/forecast.py", line 726, in quantile
    quantiles, quantile_predictions
  File "/home/bhaskar/Desktop/gluonts/src/gluonts/model/forecast.py", line 52, in __init__
    ), "Need at least two points for linear interpolation."
AssertionError: Need at least two points for linear interpolation.
"""
```

## Environment
- Operating system:Ubuntu 18.04
- Python version:3.7
- GluonTS version:dev
- MXNet version:

lostella:
*Issue #, if available:* Fixes #2498

*Description of changes:* There was a bug into the logic that decided whether to do linear interpolation or not, in the edge case where only mean is stored in the forecast object. This fixes it and adds a test.

### DIFF
--- a/test/model/test_forecast.py
+++ b/test/model/test_forecast.py
@@ -53,7 +53,7 @@ MULTIVARIATE_FORECASTS = {
 
 
 @pytest.mark.parametrize("name", FORECASTS.keys())
-def test_Forecast(name):
+def test_forecast(name):
     forecast = FORECASTS[name]
 
     def percentile(value):
@@ -73,6 +73,23 @@ def test_Forecast(name):
     assert forecast.index[0] == START_DATE
 
     forecast.plot()
+
+
+def test_mean_only_forecast():
+    forecast = QuantileForecast(
+        forecast_arrays=np.ones(shape=(1, 12)),
+        start_date=pd.Period("2022-03-04 00", freq="H"),
+        forecast_keys=["mean"],
+    )
+
+    assert forecast.prediction_length == 12
+    assert len(forecast.index) == 12
+    assert forecast.index[0] == pd.Period("2022-03-04 00", freq="H")
+
+    for level in [0.1, 0.5, 0.7]:
+        assert np.isnan(forecast.quantile(level)).all()
+    
+    assert np.equal(forecast.mean, 1).all()
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


test/model/test_forecast.py
```diff
@@ -75,6 +75,23 @@ def percentile(value):
...
+        forecast_arrays=np.ones(shape=(1, 12)),
+        start_date=pd.Period("2022-03-04 00", freq="H"),
+        forecast_keys=["mean"],
+    )
+
+    assert forecast.prediction_length == 12
+    assert len(forecast.index) == 12
+    assert forecast.index[0] == pd.Period("2022-03-04 00", freq="H")
+
+    for level in [0.1, 0.5, 0.7]:
```
melopeo, line 89: Ok. It seems that the main message here is that, if we only have the field `mean` in the `QuantileForecast` then we will return `np.nan` if any quantile is requested. LGTM.

melopeo (approved):

test/model/test_forecast.py
lostella: Exactly

jaheba:
What looks weird to me is that we treat `mean` as a quantile. Shouldn't we just keep them separate?

lostella:
> What looks weird to me is that we treat `mean` as a quantile. Shouldn't we just keep them separate?

Yes, I agree. Either that or `QuantileForecast` gets a different name.

PR #2536, opened 2023-01-04
Add TemporalFusionTransformer implementation in PyTorch

shchur:
*Description of changes:*

Unlike the MXNet implementation of TFT, the PyTorch versions of model/estimator are compatible with the dataset schema used by DeepAR. For example, if we construct the estimator with as follows
```python
estimator = TemporalFusionTransformerEstimator(
   ...,
   dynamic_dims=[2, 5, 7],
   dynamic_cardinalities=[6, 5, 5, 4, 2],
   past_dynamic_dims=[3, 3],
   past_dynamic_cardinalities=[5, 2, 4],
)
```
it will expect to receive a dataset, where each time series has keys
- `"feat_dynamic_real"`: shape `[..., sum(dynamic_dims)]` (inside the network this will feature get partitioned into 3 chunks of dim 2, 5, 7)
- `"feat_dynamic_cat"`: shape `[..., len(dynamic_cardinalities)]`
- `"past_feat_dynamic_real"`: shape `[..., sum(past_dynamic_dims)]`
- `"past_feat_dynamic_cat"`: shape `[..., len(past_dynamic_cardinalities)]`

*To do:*
- [x] Move `QuantileOutput` and `TFTInstanceSplitter` to a another folder?
- [x] Benchmarking
- [x] Add tests

### DIFF
--- a/src/gluonts/model/forecast_generator.py
+++ b/src/gluonts/model/forecast_generator.py
@@ -116,7 +116,7 @@ class QuantileForecastGenerator(ForecastGenerator):
         for batch in inference_data_loader:
             # TODO: Change inputs to dict to avoid assuming matching order in input_names / model signature?
             inputs = [batch.get(k) for k in input_names]
-            outputs = prediction_net(*inputs).cpu().detach().numpy()
+            outputs = predict_to_numpy(prediction_net, inputs)
             if output_transform is not None:
                 outputs = output_transform(batch, outputs)
 
--- a/src/gluonts/model/forecast_generator.py
+++ b/src/gluonts/model/forecast_generator.py
@@ -114,8 +114,9 @@ class QuantileForecastGenerator(ForecastGenerator):
         **kwargs
     ) -> Iterator[Forecast]:
         for batch in inference_data_loader:
-            inputs = [batch[k] for k in input_names]
-            outputs = predict_to_numpy(prediction_net, inputs)
+            # TODO: Change inputs to dict to avoid assuming matching order in input_names / model signature?
+            inputs = [batch.get(k) for k in input_names]
+            outputs = prediction_net(*inputs).cpu().detach().numpy()
             if output_transform is not None:
                 outputs = output_transform(batch, outputs)
 
--- a/src/gluonts/torch/model/tft/estimator.py
+++ b/src/gluonts/torch/model/tft/estimator.py
@@ -54,6 +54,24 @@ from .lightning_module import TemporalFusionTransformerLightningModule
 from .transformation import TFTInstanceSplitter
 
 
+PREDICTION_INPUT_NAMES = [
+    "feat_static_cat",
+    "feat_static_real",
+    "feat_dynamic_cat",
+    "feat_dynamic_real",
+    "past_feat_dynamic_cat",
+    "past_feat_dynamic_real",
+    "past_target",
+    "past_observed_values",
+    "past_is_pad",
+]
+
+TRAINING_INPUT_NAMES = PREDICTION_INPUT_NAMES + [
+    "future_target",
+    "future_observed_values",
+]
+
+
 class TemporalFusionTransformerEstimator(PyTorchLightningEstimator):
     """
     Estimator class to train a Temporal Fusion Transformer model, as described in [SFG17]_.
@@ -95,14 +113,14 @@ class TemporalFusionTransformerEstimator(PyTorchLightningEstimator):
         context_length: Optional[int] = None,
         quantiles: Optional[List[float]] = None,
         num_heads: int = 4,
-        hidden_size: int = 40,
-        num_feat_static_real: int = 0,
-        num_feat_dynamic_real: int = 0,
-        num_past_feat_dynamic_real: int = 0,
-        cardinalities_static: Optional[List[int]] = None,
-        cardinalities_dynamic: Optional[List[int]] = None,
-        cardinalities_past_dynamic: Optional[List[int]] = None,
-        embedding_dimension: Optional[List[int]] = None,
+        hidden_dim: int = 32,
+        variable_dim: int = 32,
+        static_dims: Optional[List[int]] = None,
+        dynamic_dims: Optional[List[int]] = None,
+        past_dynamic_dims: Optional[List[int]] = None,
+        static_cardinalities: Optional[List[int]] = None,
+        dynamic_cardinalities: Optional[List[int]] = None,
+        past_dynamic_cardinalities: Optional[List[int]] = None,
         time_features: Optional[List[TimeFeature]] = None,
         lr: float = 1e-3,
         dropout_rate: float = 0.1,
@@ -130,14 +148,16 @@ class TemporalFusionTransformerEstimator(PyTorchLightningEstimator):
         # Model architecture
         self.quantiles = quantiles
         self.num_heads = num_heads
-        self.hidden_size = hidden_size
-        self.num_feat_dynamic_real = num_feat_dynamic_real
-        self.num_feat_static_real = num_feat_static_real
-        self.num_past_feat_dynamic_real = num_past_feat_dynamic_real
-        self.cardinalities_static = cardinalities_static
-        self.cardinalities_dynamic = cardinalities_dynamic
-        self.cardinalities_past_dynamic = cardinalities_past_dynamic
-        self.embedding_dimension = embedding_dimension
+        self.hidden_dim = hidden_dim
+        self.variable_dim = variable_dim
+
+        self.static_dims = static_dims
+        self.dynamic_dims = dynamic_dims
+        self.past_dynamic_dims = past_dynamic_dims
+        self.static_cardinalities = static_cardinalities
+        self.dynamic_cardinalities = dynamic_cardinalities
+        self.past_dynamic_cardinalities = past_dynamic_cardinalities
+
         if time_features is None:
             time_features = time_features_from_frequency_str(self.freq)
         self.time_features = time_features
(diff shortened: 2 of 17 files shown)


kashif:
very cool! I too have a version of TFT which is auto-regressive and the output can be a prob distrb class, rather than quantiles.... kindly have a look here: https://github.com/kashif/pytorch-transformer-ts/tree/main/tft

shchur:
Thank you, @kashif! Some parts of this PR are based on your code (there is an acknowledgement in the docstring), thanks a lot for making it available! My main focus here was to add support for all types of dynamic features, plus we want to stick to quantile output, since faster inference is critical for some downstream applications in AutoGluon-TS.

src/gluonts/model/forecast_generator.py
```diff
@@ -114,8 +114,9 @@ def __call__(
         **kwargs
     ) -> Iterator[Forecast]:
         for batch in inference_data_loader:
-            inputs = [batch[k] for k in input_names]
-            outputs = predict_to_numpy(prediction_net, inputs)
+            # TODO: Change inputs to dict to avoid assuming matching order in input_names / model signature?
+            inputs = [batch.get(k) for k in input_names]
+            outputs = prediction_net(*inputs).cpu().detach().numpy()
```
lostella: I think this works only with some type of tensors, right? `predict_to_numpy` should take care of this (although it's missing `.detach()`?)

src/gluonts/torch/model/tft/transformation.py
```diff
@@ -0,0 +1 @@
+from gluonts.mx.model.tft._transform import TFTInstanceSplitter
```
lostella: As you point out in the PR, it would probably be better to move this somewhere else, or otherwise just copy the code here. Otherwise this may put a dependency on `mxnet`.

lostella: I'm also wondering which the model needs such a dedicated component? We could probably simplify how this step works across the board. But maybe let's revisit this separately and just have the code copied here for now?

src/gluonts/torch/modules/scaler.py
```diff
@@ -111,3 +111,52 @@ def forward(
...
+        self.keepdim = keepdim
+        self.register_buffer("minimum_scale", torch.tensor(minimum_scale))
+
+    def forward(
+        self, data: torch.Tensor, weights: torch.Tensor
+    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        assert (
+            data.shape == weights.shape
+        ), "data and weights must have same shape"
+        with torch.no_grad():
```
lostella, line 163: Also to be addressed separately, but: all these scalers don't really have learnable parameters, so they could really be functions or very simple `__call__`able dataclasses. Thoughts?

kashif: I agree

src/gluonts/model/forecast_generator.py
shchur: My bad, this change is actually unnecessary. 

There is still the question of whether `inputs` should be a `dict` instead of list to remove the implicit assumption that the of arguments in `input_names` matches the order expected by `forward` method of the model.

src/gluonts/torch/model/tft/transformation.py
shchur: I've moved `TFTInstanceSplitter` to `gluonts.transform.split` & added a docstring.

src/gluonts/torch/modules/scaler.py
shchur: I agree too. We could also modify `MeanScaler` and `NOPScaler` to also return `data, loc, scale` for consistency (currently they only return `data, scale`).

kashif: I can do that and change it for every model... and also add log1p of loc as a feature together with log scale?

timoschowski:
any chance that this is going in before the Sphere course starts on Jan 17? That would be rather awesome!

shchur:
Here are some benchmarking results for PyTorch vs. MXNet implementations of TFT on randomly chosen 12 datasets. Winrate is computed as % of metrics where the PyTorch version is better (assuming higher-is-better for coverage/related metrics, and lower-is-better for the rest). 

The PyTorch version seems to consistently be 2-4x faster, but the performance is rather different between the two models - with PyTorch being better in 8/12 cases. Is this an acceptable difference @lostella?

<details>

- max_epochs = 50
- batch_size = 64
- context_length = 64
- quantile_levels = [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9]
- no validation set - just train on the train set until max_epochs is reached
- both models trained on 1xV100 GPU

</details>

| dataset                       |   train_time_torch |  train_time_mxnet |   torch_winrate |
|:------------------------------|--------:|--------:|----------------:|
| australian_electricity_demand |     555 |    1007 |            70   |
| bitcoin                       |     178 |     598 |            77.5 |
| fred_md                       |     150 |     488 |            85   |
| kdd_2018                      |     206 |     694 |            30   |
| m3_monthly                    |     149 |     494 |            10   |
| m4_hourly                     |     194 |     627 |            95   |
| m4_weekly                     |     172 |     483 |            27.5 |
| restaurant                    |     177 |     611 |            77.5 |
| rideshare                     |     205 |     672 |            57.5 |
| rossmann                      |     177 |     642 |            22.5 |
| tourism_monthly               |     153 |     543 |            67.5 |
| walmart                       |     169 |     595 |            57.5 |

Raw metrics for each dataset - [benchmark_tft.zip](https://github.com/awslabs/gluonts/files/10411229/benchmark_tft.zip)

src/gluonts/torch/model/tft/estimator.py
```diff
@@ -0,0 +1,390 @@
...
+    num_heads
+        Number of attention heads in self-attention layer in the decoder.
+    hidden_dim
+        Size of the LSTM & transformer hidden states.
+    variable_dim
+        Size of the feature embeddings.
+    static_dims
+        Sizes of the real-valued static features.
+    dynamic_dims
+        Sizes of the real-valued dynamic features that are known in the future.
```
lostella, line 107: I’m a bit confused: why does one need to specify multiple sizes?

lostella (commented):
I think it would be nice to also add this to this smoke test: https://github.com/awslabs/gluonts/blob/9e5c4e90693085a4067ba77e43128cbbc4eccbf5/test/torch/model/test_modules.py#L88

src/gluonts/torch/model/tft/module.py
```diff
@@ -0,0 +1,371 @@
...
+
+    Partially based on the implementation in github.com/kashif/pytorch-transformer-ts.
+    """
+
+    @validated()
+    def __init__(
+        self,
+        context_length: int,
+        prediction_length: int,
+        d_past_feat_dynamic_real: List[int],
```
lostella: I’m wondering whether it would make sense to have defaults for most of these. For example, in case one has no additional features (only target series) what would these be set to? I’m assuming some minimal values and then constant dummy tensors would be given to `forward` in place of features?

shchur: Added defaults here + modified the docstring

src/gluonts/torch/model/tft/estimator.py
shchur: I've added an explanation in the docstring for `TemporalFusionTransformerEstimator`. Does this make sense?

```
    TFT internally performs feature selection when making forecasts. For this
    reason, the dimensions of real-valued features can be grouped together if
    they correspond to the same variable (e.g., treat weather features as a
    one feature and holiday indicators as another feature).

    For example, if the dataset contains key "feat_static_real" with shape
    [batch_size, 3], we can, e.g.,
    - set ``static_dims = [3]`` to treat all five dimensions as a single feature
    - set ``static_dims = [1, 1, 1]`` to treat each dimension as a separate feature
    - set ``static_dims = [2, 1]`` to treat the first two dims as a single feature

    See ``gluonts.torch.model.tft.TemporalFusionTransformerModel.input_shapes``
    for more details on how the model configuration corresponds to the expected
    input shapes.
```

This way the TFT estimator / model is compatible with the DeepAR schema if we set
```python
d_feat_dynamic_real = [num_feat_dynamic_real],
```

src/gluonts/torch/model/tft/module.py
```diff
@@ -0,0 +1,374 @@
...
+    def __init__(
+        self,
+        context_length: int,
+        prediction_length: int,
+        d_feat_static_real: List[int] = None,  #  Defaults to [1]
+        c_feat_static_cat: List[int] = None,  #  Defaults to [1]
+        d_feat_dynamic_real: List[int] = None,  # Defaults to [1]
+        c_feat_dynamic_cat: List[int] = None,  # Defaults to []
+        d_past_feat_dynamic_real: List[int] = None,  # Defaults to []
+        c_past_feat_dynamic_cat: List[int] = None,  # Defaults to []
```
lostella: These should be `Optional[List[int]]`

src/gluonts/torch/model/tft/estimator.py
```diff
@@ -0,0 +1,406 @@
...
+    Estimator class to train a Temporal Fusion Transformer (TFT) model, as described in [LAL+21]_.
+
+    TFT internally performs feature selection when making forecasts. For this
+    reason, the dimensions of real-valued features can be grouped together if
+    they correspond to the same variable (e.g., treat weather features as a
+    one feature and holiday indicators as another feature).
+
+    For example, if the dataset contains key "feat_static_real" with shape
+    [batch_size, 3], we can, e.g.,
+    - set ``static_dims = [3]`` to treat all five dimensions as a single feature
```
lostella:
```suggestion
    - set ``static_dims = [3]`` to treat all three dimensions as a single feature
```

lostella (changes requested):

shchur:
Resolved 👍

src/gluonts/torch/modules/scaler.py
lostella: I guess we can make incremental changes: first make these not-`torch.nn.Module` objects, then possibly change their interface. The `loc` change will be breaking, so let's discuss that separately maybe.

lostella:
@shchur did you get those results with default settings? I'm interested in `context_length` in particular, which appears to default to `prediction_length` when not specified (in the estimator).

shchur:
These were obtained using `context_length = 64` (other hyperparameters also mentioned in the dropdown "Details" in the original message). The [reference implementation](https://github.com/google-research/google-research/blob/master/tft/data_formatters/base.py) uses a different `context_length` for each dataset, so I just set it to be the same as in the MXNet implementation of TFT in GluonTS.

Though, most likely, the choice of `context_length = prediction_length` is very poor, do you think we should change it to something else?

lostella:
> Though, most likely, the choice of context_length = prediction_length is very poor, do you think we should change it to something else?

Exactly, it should probably be much longer. Maybe let's do that separately from this PR, we could think of either a larger multiple of `prediction_length` or some heuristics based on the frequency of the data, or a combination of both.

lostella (approved):

kashif:
awesome!

Issue #2483, opened 2022-12-09
Random `NotFittedError`s with `rotbaum.TreeEstimator`


caseytomlin:
## Description
Executing the code below _sometimes_ succeeds, _sometimes_ fails, with a `NotFittedError`. I am running it in a notebook cell, and it does not seem to change whether or not I restart the kernel.

## To Reproduce
As the title suggests, no guarantee....

Mostly lifted from [this passing test](https://github.com/awslabs/gluonts/blob/dev/test/ext/rotbaum/test_rotbaum_smoke.py):
```python
from gluonts.testutil.dummy_datasets import make_dummy_datasets_with_features
from gluonts.model.rotbaum import TreeEstimator

datasets = make_dummy_datasets_with_features(
    cardinality=[3, 10, 42], num_feat_dynamic_real=3
)

dataset_train, dataset_test = datasets
hps = {
    "freq": "D",
    "prediction_length": 3,
    "context_length": 3,
}
estimator = TreeEstimator.from_inputs(dataset_train, **hps)
predictor = estimator.train(dataset_train)

forecasts = list(predictor.predict(dataset_test))
forecasts[0].median
```

## Error message or code output

```
---------------------------------------------------------------------------
NotFittedError                            Traceback (most recent call last)
Cell In[5], line 27
     19 # forecast_it, ts_it = make_evaluation_predictions(
     20 #     dataset=dataset_test,
     21 #     predictor=predictor
     22 # )
     23 # fcs = list(forecast_it)
     24 # fcs[0].quantile_ts(q=0.5)
     26 forecasts = list(predictor.predict(dataset_test))
---> 27 forecasts[0].median

File ~/.venv/lib/python3.9/site-packages/gluonts/model/forecast.py:149, in Forecast.median(self)
    147 @property
    148 def median(self) -> np.ndarray:
--> 149     return self.quantile(0.5)

File ~/.venv/lib/python3.9/site-packages/gluonts/model/rotbaum/_predictor.py:68, in RotbaumForecast.quantile(self, q)
     61 """
     62 Returns np.array, where the i^th entry is the estimate of the q
     63 quantile of the conditional distribution of the value of the i^th
     64 step in the forecast horizon.
     65 """
     66 assert 0 <= q <= 1
     67 return np.array(
---> 68     list(
     69         chain.from_iterable(
     70             model.predict(self.featurized_data, q)
     71             for model in self.models
     72         )
     73     )
     74 )

[...]

PR #2537, opened 2023-01-04
nan values and explainability support for rotbaum

zoolhasson:
Issue #, if available: #2483

Description of changes: 

1. Remade preprocessing with multiprocessing, leading to faster training.
2. For each timestamp in the future, for the corresponding tabular model/models I remove the datapoints that have missing targets. This fixes a bug that occurred occasionally when missing values were present, where because of the missing values in the target of the tabularized data, individual models corresponding to some timestamps in the future did not train.
3. Added an "explain" function to get model explainability in the case that the model is QuantileRegression.
4. Some more assertions.


lostella (approved):
Thanks @zoolhasson!

PR #2541, opened 2023-01-09
Loader cache

jaheba:
Adding the option to cache the dataset, post transformation.

Currently, this uses `pickle` to dump and load the data, however, we could also look at `arrow`, which could speed things up even more. Another option is to use an in-memory cache.

This can significantly speed up training, especially when using slow datasources, such as `.json.gz`.

### DIFF
--- a/src/gluonts/dataset/loader.py
+++ b/src/gluonts/dataset/loader.py
@@ -16,7 +16,7 @@ import io
 import pickle
 from tempfile import TemporaryFile
 from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


lostella:
Would this replace the `cache_data` option in estimators? And `gluonts.itertools.Cached`?

jaheba:
Right, this just adds caching add a different place.

Two things:

* We should allow caching for both in-memory and on disk using something like pickle/arrow.
* I think it makes sense to move the responsibility for caching to the loader, such that the estimator has to do less.

PR #2546, opened 2023-01-10
Simplify linear interpolation in forecast.py

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -24,67 +24,27 @@ from pydantic.dataclasses import dataclass
 from gluonts.core.component import validated
 
 
-class LinearInterpolation:
-    """
-    Linear interpolation based on datapoints (x_coord, y_coord)
+def _linear_interpolation(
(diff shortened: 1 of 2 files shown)


src/gluonts/model/forecast.py
```diff
@@ -24,67 +24,27 @@
 from gluonts.core.component import validated
 
 
-class LinearInterpolation:
-    """
-    Linear interpolation based on datapoints (x_coord, y_coord)
+def _linear_interpolation(
+    xs: np.ndarray, ys: np.ndarray, x: float
+) -> np.ndarray:
+    assert sorted(xs) == xs
```
melopeo, line 30: Can we complement the asserts with a text message?

jaheba: We could, but I don't see much value in it. The traceback should already contain enough information. Also, this function is only used internally.

melopeo: Sounds good.

melopeo (approved):
LGTM

PR #2553, opened 2023-01-12
Add default_scale to MeanScaler and enable the option in DeepAR-PyTorch

shubhamkapoor:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/torch/model/deepar/estimator.py
+++ b/src/gluonts/torch/model/deepar/estimator.py
@@ -126,6 +126,10 @@ class DeepAREstimator(PyTorchLightningEstimator):
         (default: ``NegativeLogLikelihood()``).
     scaling
         Whether to automatically scale the target values (default: true).
+    default_scale
+        Default scale that is applied if the context length window is
+        completely unobserved. If not set, the scale in this case will be
+        the mean scale in the batch.
     lags_seq
         Indices of the lagged target values to use as inputs of the RNN
         (default: None, in which case these are automatically determined
--- a/src/gluonts/torch/model/deepar/estimator.py
+++ b/src/gluonts/torch/model/deepar/estimator.py
@@ -170,6 +170,7 @@ class DeepAREstimator(PyTorchLightningEstimator):
         distr_output: DistributionOutput = StudentTOutput(),
         loss: DistributionLoss = NegativeLogLikelihood(),
         scaling: bool = True,
+        default_scale: Optional[float] = None,
         lags_seq: Optional[List[int]] = None,
         time_features: Optional[List[TimeFeature]] = None,
         num_parallel_samples: int = 100,
@@ -209,6 +210,7 @@ class DeepAREstimator(PyTorchLightningEstimator):
         )
         self.embedding_dimension = embedding_dimension
         self.scaling = scaling
+        self.default_scale = default_scale
         self.lags_seq = lags_seq
         self.time_features = (
             time_features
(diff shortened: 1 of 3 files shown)


src/gluonts/torch/modules/scaler.py
```diff
@@ -38,12 +38,20 @@ class MeanScaler(nn.Module):
 
     @validated()
     def __init__(
-        self, dim: int, keepdim: bool = False, minimum_scale: float = 1e-10
+        self,
+        dim: int,
+        keepdim: bool = False,
+        default_scale: Optional[float] = None,
```
lostella, line 45: Maybe this could be added to the docstring above?

shubhamkapoor: Done! I think the docstring of minimum_scale was wrong. Updated it as well. Please take a look.

src/gluonts/torch/model/deepar/estimator.py
```diff
@@ -126,6 +126,10 @@ class DeepAREstimator(PyTorchLightningEstimator):
         (default: ``NegativeLogLikelihood()``).
     scaling
         Whether to automatically scale the target values (default: true).
+    default_scale
+        Default scale that is applied if the context length window is
+        completely unobserved. If not set, the scale in this case will be
+        the mean scale in the batch.
```
lostella, line 132: I think we probably want to fix this: having the model behaviour on one batch element depend on what the rest of the batch looks like is probably not desirable, right?

src/gluonts/torch/modules/scaler.py
```diff
@@ -31,19 +31,29 @@ class MeanScaler(nn.Module):
     keepdim
         controls whether to retain dimension ``dim`` (of length 1) in the
         scale tensor, or suppress it.
-    minimum_scale
+    default_scale
         default scale that is used for elements that are constantly zero
+    minimum_scale
+        minimum possible scale that is used for any item.
         along dimension ``dim``.
```
lostella: Forgot to remove a line?

src/gluonts/torch/modules/scaler.py
```diff
@@ -57,7 +67,10 @@ def forward(
         denominator = torch.max(
             total_observed, torch.ones_like(total_observed)
         )
-        default_scale = weighted_sum.sum(dim=0) / denominator
```
lostella, line 60: I think this was wrong in the first place, mixing things across dimension 0 should not occur probably

lostella (commented):
I'm wondering whether we should just remove the previous behaviour to determine the `default_scale`, and default the new `default_scale` to `1.0` (too many "default" here, I know)

src/gluonts/torch/modules/scaler.py
shubhamkapoor: Done!

src/gluonts/torch/model/deepar/estimator.py
shubhamkapoor: I think the reason for using batch statistics for default_scale makes sense from the point of the computation of the average batch loss. A single time series on a very high scale, if not scaled with appropriate a batch statistic, can mess up the loss, no?

lostella (approved):

lostella (approved):
Thanks! Adding the default_scale option makes total sense.

If we want to challenge the “batch-wise” default_scale computation that we already had, we can do that separately

PR #2554, opened 2023-01-13
Fix: add missing params in rotbaum

zoolhasson:
Bug fix: self.subtract_mean, self.count_nans were missing from calls to _pre_transform; causing errors if related time series were used.


lostella (approved):
Thanks @zoolhasson!

PR #2558, opened 2023-01-13
Compact PandasDataset string representation

lostella:
*Description of changes:* Makes the string representation of PandasDataset more compact. It otherwise freezes the terminal when there's lots of data in the dataset.

### DIFF
--- a/src/gluonts/dataset/pandas.py
+++ b/src/gluonts/dataset/pandas.py
@@ -183,7 +183,7 @@ class PandasDataset:
         return len(self._data_entries)
 
     def __str__(self) -> str:
-        return f"PandasDataset with {len(self)} series, freq={self.freq}"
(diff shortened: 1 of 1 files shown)


src/gluonts/dataset/pandas.py
```diff
@@ -182,6 +182,9 @@ def __iter__(self) -> Iterator[DataEntry]:
     def __len__(self) -> int:
         return len(self._data_entries)
 
+    def __str__(self) -> str:
+        return f"PandasDataset with {len(self)} series, freq={self.freq}"
```
jaheba: I would prefer something like
```suggestion
        return f"PandasDataset<size={len(self)}, freq={self.freq}>"
```

jaheba (approved):

PR #2561, opened 2023-01-13
Add statsforecasts models

melopeo:
*Description of changes:*
We add more forecasting models from Nixtla/StatsForecast package.


src/gluonts/ext/statsforecast.py
```diff
@@ -25,8 +25,26 @@
     CrostonClassic,
     CrostonOptimized,
     CrostonSBA,
+    DynamicOptimizedTheta,
+    DynamicTheta,
+    ETS,
```
lostella: I think this was there and we removed it, since it’s just an alias for AutoETS, see https://github.com/Nixtla/statsforecast/blob/20be88335ba8c5872fce53407e2d550017139985/statsforecast/models.py#L719

melopeo: Good point! I updated the PR by removing ETS accordingly. Thanks!

lostella (approved):

Issue #2560, opened 2023-01-13
Handling NaNs in predictions


abdulfatir:
## Description
Currently, the default behavior in the evaluator is to skip NaN values when aggregating metrics. This results in the unexpected behavior where a model that produces NaNs as forecasts for a subset of time series may seem to perform better (particularly when looking at the `ND` and `mean_wQuantileLoss`) metrics. In the following snippet, I show how injecting NaNs into forecast "improves" the model performance. Here's what I do in the snippet:

- Create a test dataset with two time series (sine and cosine wave).
- Evaluate using `MeanPredictor` (the predictor doesn't really matter here).
- Inject NaNs in all forecast samples for the first time series.
- The `ND` and `mean_wQuantileLoss` improves (almost by half).
- Inject NaNs in all forecast samples for the second time series too.
- The `ND` and `mean_wQuantileLoss` is now `0.0` and other metrics are NaN. This is because the aggregation function for MSE-like metrics is `mean` whereas for the components of `ND` and `mean_wQuantileLoss` it is `sum` and:

```
pd.Series([float("nan")]).agg("sum", skipna=True) # equals 0.0
pd.Series([float("nan")]).agg("mean", skipna=True) # equals NaN
```

I'm not exactly sure what should the behavior be in the case when forecast contains NaNs; however, my suggestion is as follows.
- Add a `strict` parameter to the `Evaluator`.
- If `strict` is `True`, raise an error when the forecast contains invalid values.
- If `strict` is `False`, raise a warning when the forecast contains invalid values. 

Thanks to @marcelk1102 for initially spotting this issue. 

cc @lostella @dcmaddix 

## To Reproduce
(Please provide minimal example of code snippet that reproduces the error. For existing examples, please provide link.)

```python
In [1]: import numpy as np
   ...: import pandas as pd
   ...: from gluonts.model.trivial.mean import MeanPredictor
   ...: from gluonts.evaluation import make_evaluation_predictions, Evaluator

In [2]: x = np.arange(7 * 24)
   ...: y = np.vstack([np.sin(x * 2 * np.pi / 24), np.cos(x * 2 * np.pi / 24)])
   ...: y = y + np.random.randn(*y.shape) * 0.1
   ...: test_dataset = [{"target": yi, "start": pd.Period("2023-01-01", freq="H")} for yi in y]

In [3]: predictor = MeanPredictor(prediction_length=24, context_length=24)
[...]

PR #2568, opened 2023-01-16
Add NaN validation to Evaluator

abdulfatir:
*Issue #, if available:* #2560

*Description of changes:*
This PR adds NaN validation to the evaluator via a `strict` parameter which defaults to True. When `strict` is `True`, an error is raised if the forecasts contain NaN values, otherwise a warning is logged. 

### DIFF
--- a/src/gluonts/evaluation/_base.py
+++ b/src/gluonts/evaluation/_base.py
@@ -104,7 +104,7 @@ def aggregate_valid(
 
 
 def validate_forecast(
-    forecast: Forecast, quantiles: Iterable[Union[float, str]]
+    forecast: Forecast, quantiles: Iterable[Quantile]
 ) -> bool:
     """Validates a Forecast object by checking it for `NaN` values.
     The supplied quantiles and mean (if available) are checked.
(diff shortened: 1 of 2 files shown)


jaheba:
As I said in #2560, I'm against checking this in the evaluator.

src/gluonts/evaluation/_base.py
```diff
@@ -150,11 +182,14 @@ class Evaluator:
     ignore_invalid_values
         Ignore `NaN` and `inf` values in the timeseries when calculating
         metrics.
-    aggregation_strategy:
+    aggregation_strategy
         Function for aggregating per timeseries metrics.
         Available options are:
         aggregate_valid | aggregate_all | aggregate_no_nan
         The default function is aggregate_no_nan.
+    strict
```
jaheba: Maybe something like `allow_nan_forecast` is more explicit?

abdulfatir: Changed.

lostella:
image

lostella:
In all seriousness: I believe this is the desired behaviour, and the previous one was incorrect in that numbers coming out of evaluation can be misleading if a model happens to output a `nan`.

lostella (approved):

PR #2569, opened 2023-01-16
Add missing value imputation to Seasonal Naive

abdulfatir:
*Issue #, if available:* #2560

*Description of changes:*
This PR adds missing value imputation to `SeasonalNaivePredictor`. Previously, the predictor would return `nan` as forecast if the value at previous season was `nan`. I have set the default imputation strategy to `LastValueImputation` which makes sense in the case of `SeasonalNaivePredictor` but we can change this to something else.

### DIFF
--- a/src/gluonts/model/seasonal_naive/_predictor.py
+++ b/src/gluonts/model/seasonal_naive/_predictor.py
@@ -18,8 +18,13 @@ import numpy as np
 from gluonts.core.component import validated
 from gluonts.dataset.common import DataEntry
 from gluonts.dataset.util import forecast_start
+from gluonts.dataset.field_names import FieldName
 from gluonts.model.forecast import Forecast, SampleForecast
 from gluonts.model.predictor import RepresentablePredictor
+from gluonts.transform.feature import (
+    LastValueImputation,
+    MissingValueImputation,
+)
 from gluonts.time_feature import get_seasonality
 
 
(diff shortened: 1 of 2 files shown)


lostella:
@abdulfatir I think it makes sense to add some tests for `SeasonalNaivePredictor`. You could add a dedicated test module in `test/model` in the same spirit as https://github.com/awslabs/gluonts/blob/dev/test/model/trivial/test_moving_average.py, or merge the two into a single test module for baseline models.

abdulfatir:
@lostella Sure, I will add the tests.

abdulfatir:
@lostella It looks like that tests that I added are not running on github. Do I need to change something else to make them run?

lostella:
is it not running? where do you see that?

abdulfatir:
[In the logs](https://github.com/awslabs/gluonts/actions/runs/3938533195/jobs/6737337368). I searched for `test_seasonal_naive`.

lostella:
I think that's because we use the `-n2` option from `pytest-xdist` to run tests in parallel over multiple CPUs. In that case the tests are not listed individually by module, but rather displayed in a more compact format like

```
........................................................................ [  2%]
.......................................X................................ [  4%]
ssssssssssssssssssssssss................................................ [  7%]
...................x.......X..................................s.ssssssss [  9%]
sssssssssssssss......................................................... [ 11%]
........................................................................ [ 14%]
.......................X.x..X.....ssss........sss....................... [ 16%]
............................x.X........X..sss..........ssss............. [ 18%]
........................................................................ [ 21%]
```

lostella (approved):
🚢 it!

PR #2573, opened 2023-01-18
Improve `PandasDataset`

lostella:
*Description of changes:* This PR proposes an improved `PandasDataset` class where:
- Static features are provided through a separate `DataFrame`, whose index contains the `item_id`'s for the dataset, and whose columns dtypes are used to distinguish numerical/categorical features. In particular, categorical features are encoded as integers automatically using pandas
- Overall the code is simplified

In particular, now it is easy to use the class for loading the data from the M5 competition, with a bit of initial (and I believe necessary, and rather simple) pandas manipulation:

```python
# load dates information and sales data
cal = pd.read_csv(
    Path("m5-forecasting-accuracy") / "calendar.csv"
)
df = pd.read_csv(
    Path("m5-forecasting-accuracy")
    / "sales_train_validation.csv",
    index_col=0,
)

# slice out categorical columns
df_feat = df.iloc[:, 1:5].astype("category")

# slice out columns with numerical series data, transpose it
df_series = df.iloc[:, 5:].transpose()
df_series.index = pd.to_datetime(cal.date[: len(df_series.index)])

dataset = PandasDataset(
    dict(df_series),
    static_features=df_feat,
)

print(dataset)
```

gives

```
PandasDataset<size=30490, freq=D, num_dynamic_real=0, num_past_dynamic_real=0, num_static_real=0, num_static_cat=4, cardinalities=[7, 3, 10, 3]>
```

With the current implementation, one would instead need to add static features as “constant dynamic” ones, for the class to then consider only the first value, and this preprocessing step takes quite long since it’s tens of thousands of series.

Still left to do:
- [x] fix other tests
- [x] fix the `past_feat_dynamic_real` story
- [ ] add common dynamic features (M5 has "special events" of a few types, and some other numerical indicators)
- [ ] add proper tests for features
- [x] update documentation
- [x] benchmark performance against previous implementation
- [x] define deprecation path for previous implementation, if we believe it's necessary

### DIFF
--- a/docs/tutorials/data_manipulation/pandasdataframes.md.template
+++ b/docs/tutorials/data_manipulation/pandasdataframes.md.template
@@ -328,10 +328,11 @@ estimator_with_features = DeepAREstimator(
 )
 ```
 
-Let's now generate single time series and multiple time series
-(as list and as long-dataframe).
+Let's now generate multiple time series, both as a dictionary of data frames,
+and as a single long data frame. We also generate a dedicated data frame of
+static features.
 Note we don't provide `freq` and `timestamp`, because they are automatically
-inferred from the index. We are also not passing the `target` argument, since
+inferred from the time index. We are also not passing the `target` argument, since
 it's default value is "target", which is the target column in our dataframes.
 
 
--- a/docs/tutorials/data_manipulation/pandasdataframes.md.template
+++ b/docs/tutorials/data_manipulation/pandasdataframes.md.template
@@ -298,7 +298,6 @@ def generate_single_ts_with_features(date_range, item_id) -> pd.DataFrame:
     ts = generate_single_ts(date_range, item_id)
     T = ts.shape[0]
     # static features are constant for each series
-    ts["static_cat_1"] = np.random.randint(low=0, high=2) # high is exclusive
     ts["dynamic_real_1"] = np.random.normal(size=T)
     ts["dynamic_real_2"] = np.random.normal(size=T)
     # ... we can have as many static or dynamic features as we like
@@ -322,8 +321,9 @@ estimator_with_features = DeepAREstimator(
     freq=ds.freq,
     prediction_length=prediction_length,
     use_feat_dynamic_real=True,
+    use_feat_static_real=True,
     use_feat_static_cat=True,
-    cardinality=[2,],
+    cardinality=[3,],
     trainer=Trainer(epochs=1),
 )
 ```
@@ -336,48 +336,35 @@ it's default value is "target", which is the target column in our dataframes.
 
 
 ```python
-single_ts = generate_single_ts_with_features(date_range, item_id=0)
-multiple_ts = [generate_single_ts_with_features(date_range, item_id=i) for i in range(N)]
-multiple_ts_long = pd.concat(multiple_ts)
-
-single_ts_dataset = PandasDataset(
-    single_ts,
-    feat_dynamic_real=["dynamic_real_1", "dynamic_real_2"],
-    feat_static_cat=["static_cat_1"],
+multiple_ts = {i: generate_single_ts_with_features(date_range, item_id=i) for i in range(N)}
+static_features = pd.DataFrame(
+    {
+        "color": pd.Categorical(np.random.choice(["red", "green", "blue"], size=N)),
+        "height": np.random.normal(loc=100, scale=15, size=N),
+    },
+    index=list(multiple_ts.keys())
 )
+multiple_ts_long = pd.concat(multiple_ts.values())
+
 multiple_ts_dataset = PandasDataset(
     multiple_ts,
     feat_dynamic_real=["dynamic_real_1", "dynamic_real_2"],
-    feat_static_cat=["static_cat_1"],
-)
-multiple_ts_dataset_dict = PandasDataset(
-    {i: _ts for i, _ts in enumerate(multiple_ts)},
-    feat_dynamic_real=["dynamic_real_1", "dynamic_real_2"],
-    feat_static_cat=["static_cat_1"],
+    static_features=static_features,
 )
 # for long-dataset we use a different constructor and need a `item_id` column
 multiple_ts_long_dataset = PandasDataset.from_long_dataframe(
     multiple_ts_long,
     item_id="item_id",
     feat_dynamic_real=["dynamic_real_1", "dynamic_real_2"],
-    feat_static_cat=["static_cat_1"],
+    static_features=static_features,
 )
 ```
 
 That's it! We can now call `train_and_predict` with all
 the datasets.
 
-
-```python
-train_and_predict(single_ts_dataset, estimator_with_features)
-```
-
 ```python
 train_and_predict(multiple_ts_dataset, estimator_with_features)
-```
-
-```python
-train_and_predict(multiple_ts_dataset_dict, estimator_with_features)
 ```
 
 ```python
@@ -399,13 +386,17 @@ a bunch of metrics for us
 
 
 ```python
-to_pandasdataset = lambda data: PandasDataset(
-    data,
+train = PandasDataset(
+    {item_id: df[:-prediction_length] for item_id, df in multiple_ts.items()},
     feat_dynamic_real=["dynamic_real_1", "dynamic_real_2"],
-    feat_static_cat=["static_cat_1"],
+    static_features=static_features,
 )
-train = to_pandasdataset([ts.iloc[:-prediction_length, :] for ts in multiple_ts])
-test = to_pandasdataset(multiple_ts)
+
+test = PandasDataset(
+    multiple_ts,
+    feat_dynamic_real=["dynamic_real_1", "dynamic_real_2"],
+    static_features=static_features,
+)
 ```
 
 
--- a/src/gluonts/dataset/pandas.py
+++ b/src/gluonts/dataset/pandas.py
@@ -29,7 +29,6 @@ from gluonts.itertools import SizedIterable
 # - feat_dynamic_cat -> gone, too painful to do now, let's think about it later
 # - TODO have dtype somewhere?
 # - TODO have assertions for stuff being the right dtype?
-# - TODO add the "ignore_last_n_targets" in some way (I don't like it)
 # - TODO benchmark
 # - TODO adjust tests, especially add tests for static features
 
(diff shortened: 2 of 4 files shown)


lostella:
One question is how to make sure that categorical features are encoded the same in two datasets, when these are supposed to feed the same downstream model.

For example if one dataset only has a subset of the categories, these may end up encoded differently from the other one.

lostella:
> One question is how to make sure that categorical features are encoded the same in two datasets, when these are supposed to feed the same downstream model.
> 
> For example if one dataset only has a subset of the categories, these may end up encoded differently from the other one.

I think this can be overcome by assuming that the same static features dataframe is passed around to construct datasets that need to be "homogeneous" with respect to what encoding is used. Of course this needs to be clear from the documentation.

melopeo (commented):
The challenge of having homogenous categorical features is an interesting one. One should keep in mind that these need to be (at least if we follow the sagemaker requirements): 
1) non-negative,
2) the smallest categorical value has to be equal to zero,
3) sequential, i.e. there should be no gaps between consecutive integer values.

lostella:
I think that's required in general whenever the model uses embedding layers that assume values to span the [0, N-1] range, with N the number of categories. This should be guaranteed here, since I'm using [pandas.Categorical.codes](https://pandas.pydata.org/docs/reference/api/pandas.Categorical.codes.html#pandas.Categorical.codes) to encode categorical values into integers.

src/gluonts/dataset/pandas.py
```diff
@@ -127,11 +111,53 @@ def __post_init__(self) -> None:
...
+    def num_feat_static_real(self):
+        return 0 if self._static_reals is None else self._static_reals.shape[1]
+
+    @property
+    def num_feat_dynamic_real(self):
+        return (
+            0
+            if self.feat_dynamic_real is None
+            else len(self.feat_dynamic_real)
+        )
```
jaheba: I much prefer the statements over the expression.

test/dataset/test_hierarchical.py
```diff
@@ -136,38 +137,19 @@ def test_hts_to_dataset(mode: str):
             ds = hts.to_dataset(feat_dynamic_real=features_df)
     else:
         ds = hts.to_dataset(feat_dynamic_real=features_df)
+        entry = next(iter(ds))
```
jaheba: https://toolz.readthedocs.io/en/latest/_modules/toolz/itertoolz.html#first

src/gluonts/dataset/pandas.py
```diff
@@ -11,39 +11,30 @@
...
-    A pandas.DataFrame-based dataset type.
+    A dataset type based on ``pandas.DataFrame``.
 
-    This class is constructed with a collection of pandas.DataFrame-objects
-    where each ``DataFrame`` is representing one time series.
-    A ``target`` and a ``timestamp`` columns are essential. Furthermore,
-    static/dynamic real/categorical features can be specified.
+    This class is constructed with a collection of ``pandas.DataFrame``
+    objects where each ``DataFrame`` is representing one time series.
+    A ``target`` and a ``timestamp`` columns are essential. Dynamic features
```
jaheba:
```suggestion
    Both ``target`` and ``timestamp`` columns are essential. Dynamic features
```

src/gluonts/dataset/pandas.py
```diff
@@ -11,39 +11,30 @@
...
+    of a series can be specified with together with the series' ``DataFrame``,
+    while static features can be specified in a separate ``DataFrame`` object
```
jaheba, line 35: Can I have static features in a column?

src/gluonts/dataset/pandas.py
```diff
@@ -93,24 +80,21 @@ class PandasDataset:
...
-    past_feat_dynamic_real: List[str] = field(default_factory=list)
+    static_features: Optional[pd.DataFrame] = None
     ignore_last_n_targets: int = 0
     unchecked: bool = False
     assume_sorted: bool = False
     _pairs: Iterable[Tuple[Any, Union[pd.Series, pd.DataFrame]]] = field(
         init=False
     )
+    _static_reals: Optional[pd.DataFrame] = None
+    _static_cats: Optional[pd.DataFrame] = None
```
jaheba: If they are not meant to be part of the arguments, let's use `= field(init=False)`.

src/gluonts/dataset/pandas.py
```diff
@@ -238,109 +282,56 @@ def infer_freq(index: pd.Index) -> str:
...
-    Returns
-    -------
-    DataEntry
-        A dictionary with at least ``target`` and ``start`` field.
+    Return a new array with last ``n`` elements removed from the
+    trailing axis, if ``n`` is positive, and the array itself otherwise.
     """
-    start = data.loc[:, timestamp].iloc[0] if timestamp else data.index[0]
-    dataentry = {FieldName.START: start}
+    if n <= 0:
```
jaheba, line 311: When would that be the case?

src/gluonts/dataset/pandas.py
lostella: No

src/gluonts/dataset/pandas.py
lostella: ah! I somehow forgot to do that

test/dataset/test_hierarchical.py
lostella: I can do that of course, but this is literally what `first` does

src/gluonts/dataset/pandas.py
jaheba: But what if I do?

lostella: To expand a bit on this -- of course, one could have static features as columns to begin with: when that's the case, they can be extracted into a separate dataframe pretty easily. For example from a long dataframe

```python
import pandas as pd

df = pd.DataFrame(
    data={
        'item_id': [1, 1, 2, 2],
        'static_feature': ['a', 'a', 'b', 'b'],
        'target': [1.0, 2.0, 3.0, 4.0]
    }
)

df[["item_id", "static_feature"]].drop_duplicates().set_index("item_id")
```

gives 

```
        static_feature
item_id
1                    a
2                    b
```

On the other hand, forcing one to specify static features as columns is inefficient when datasets are big (see M5 example above). Also, I think that offering this single entry point for static features is better since it allows us to do integer encoding automatically, in the case of categorical features, while this would be difficult in case each input dataframe had its own columns of static features (need to go through the whole dataset and find all values and so on).

lostella: I guess the above ☝️ could be automated in `from_long_dataframe` actually

lostella: (Done)

src/gluonts/dataset/pandas.py
```diff
@@ -127,11 +111,52 @@ def __post_init__(self) -> None:
             ), "You need to provide `freq` along with `timestamp`"
             self.freq = infer_freq(first(self._pairs)[1].index)
 
+        (
+            self._static_reals,
+            self._static_cats,
+        ) = split_numerical_categorical(self.static_features)
+
+        if self._static_reals is not None:
+            self._static_reals = self._static_reals.astype(np.float32)
```
jaheba: Not sure I like this. Should we enforce a single dtype for all values instead?

src/gluonts/dataset/pandas.py
```diff
@@ -238,109 +300,59 @@ def infer_freq(index: pd.Index) -> str:
...
-    data: pd.DataFrame,
-    target: Union[str, List[str]],
-    timestamp: Optional[str] = None,
-    feat_dynamic_real: List[str] = [],
-    feat_dynamic_cat: List[str] = [],
-    feat_static_real: List[str] = [],
-    feat_static_cat: List[str] = [],
-    past_feat_dynamic_real: List[str] = [],
-) -> DataEntry:
+def split_numerical_categorical(
```
jaheba: I think I would remove this function together and just do this inline:

```py
self._static_real = self.static_features[
    [col_name for col_name, col in df.items() if col.dtype == "category"]
]
```

And probably even better:

```py
self._static_real = static_features.select_dtypes("number")
self._static_cats = static_features.select_dtypes("category")
```

src/gluonts/dataset/pandas.py
```diff
@@ -127,11 +111,52 @@ def __post_init__(self) -> None:
...
+        (
+            self._static_reals,
+            self._static_cats,
+        ) = split_numerical_categorical(self.static_features)
+
+        if self._static_reals is not None:
+            self._static_reals = self._static_reals.astype(np.float32)
+
+        if self._static_cats is not None:
+            self._static_cats = category_to_int(self._static_cats)
```
jaheba: This should just be:

```
.apply(lambda col: col.cat.codes)
```

src/gluonts/dataset/pandas.py
lostella: Yes, `dtype` should be an attribute of the dataset and used all over the place

jaheba (approved):

jaheba (changes requested):
Need to fix the docs as well.

jaheba:
Let's merge this and continue to improve `PandasDataset`?

Should this go into v0.12?

docs/tutorials/data_manipulation/pandasdataframes.md.template
```diff
@@ -322,64 +321,53 @@ estimator_with_features = DeepAREstimator(
...
-```
-
 ```python
 train_and_predict(multiple_ts_dataset, estimator_with_features)
 ```
 
-```python
-train_and_predict(multiple_ts_dataset_dict, estimator_with_features)
-```
-
```
lostella, line 382: These example need to be removed since the separate `static_features` data frame requires that `item_id` is given, so only the dictionary case or the long data frame case make sense.

Moreover, if one needs to use static features it better be because the model is being trained on multiple series, otherwise static features are of no use.

jaheba (approved):

src/gluonts/dataset/pandas.py
```diff
@@ -61,19 +51,15 @@ class PandasDataset:
         frequency.
     feat_dynamic_real
         List of column names that contain dynamic real features.
-    feat_dynamic_cat
```
jschemm, line 64: What is the reasoning behind no longer allowing dynamic categorical features?

PR #2576, opened 2023-01-20
Fix: avoid automatic device detection via serialized tensors when deserializing `PyTorchPredictor`.

shubhamkapoor:
*Issue #, if available:*

*Description of changes:*
When `map_location=None` is passed to `torch.load`, it automatically loads the tensors to the device they were serialized on. It results in runtime errors when the model is trained using a GPU and serialized, but inference is done on CPU when predictor is constructed by `deserialize` method of PytorchPredictor with `device=None`.


lostella (approved):
Thanks! We can then consolidate this with the same logic happening in other places, maybe having a `default_device` function to do this would be good

jaheba (approved):

PR #2578, opened 2023-01-23
Clean up `gluonts.time_feature.holiday`

lostella:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/time_feature/holiday.py
+++ b/src/gluonts/time_feature/holiday.py
@@ -78,44 +78,25 @@ CyberMonday = Holiday(
     offset=[pd.DateOffset(weekday=TH(4)), Day(4)],
 )
 
-NEW_YEARS_DAY = "new_years_day"
-MARTIN_LUTHER_KING_DAY = "martin_luther_king_day"
-SUPERBOWL = "superbowl"
-PRESIDENTS_DAY = "presidents_day"
-GOOD_FRIDAY = "good_friday"
-EASTER_SUNDAY = "easter_sunday"
-EASTER_MONDAY = "easter_monday"
-MOTHERS_DAY = "mothers_day"
-INDEPENDENCE_DAY = "independence_day"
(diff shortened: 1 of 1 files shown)


src/gluonts/time_feature/holiday.py
```diff
@@ -78,44 +78,25 @@ def distance_to_day(index):
...
-    LABOR_DAY: distance_to_holiday(USLaborDay),
-    MEMORIAL_DAY: distance_to_holiday(USMemorialDay),
-    COLUMBUS_DAY: distance_to_holiday(USColumbusDay),
-    THANKSGIVING: distance_to_holiday(USThanksgivingDay),
-    CHRISTMAS_EVE: distance_to_holiday(ChristmasEve),
-    CHRISTMAS_DAY: distance_to_holiday(ChristmasDay),
-    NEW_YEARS_EVE: distance_to_holiday(NewYearsEve),
-    BLACK_FRIDAY: distance_to_holiday(BlackFriday),
-    CYBER_MONDAY: distance_to_holiday(CyberMonday),
+    "new_years_day": distance_to_holiday(NewYearsDay),
```
jaheba, line 82: Some of them are referenced in the docstings below.

lostella: I think we can remove that class entirely, it's only used in tests and I don't see any way the user can use it with estimators.

Instead, holidays should probably work just like any other `TimeFeature`?

kashif: so if they are time features we just need to make sure they are not called on the fly as these operations are slow and in theory are just needed once and then every time series in a dataset can use them...

lostella: Good point @kashif, some caching would help there I guess. In general I believe we should re-think how these features are specified to estimators, since now there seems to be too much going on.

PR #2583, opened 2023-01-24
Simplify PandasDataset.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/dataset/pandas.py
+++ b/src/gluonts/dataset/pandas.py
@@ -13,6 +13,7 @@
 
 from __future__ import annotations
 
+from dataclasses import dataclass, field, InitVar
 from typing import Any, Iterable, Optional, Type, Union
 
 import numpy as np
@@ -25,6 +26,7 @@ from gluonts.dataset.common import DataEntry
 from gluonts.itertools import Map, StarMap, SizedIterable
 
 
+@dataclass
 class PandasDataset:
     """
     A dataset type based on ``pandas.DataFrame``.
@@ -69,9 +71,8 @@ class PandasDataset:
         (Default: ``False``)
     """
 
-    def __init__(
-        self,
-        dataframes: Union[
+    dataframes: InitVar[
+        Union[
             pd.DataFrame,
             pd.Series,
             Iterable[pd.DataFrame],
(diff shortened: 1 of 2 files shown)


src/gluonts/dataset/pandas.py
```diff
@@ -57,8 +57,8 @@ class PandasDataset:
     static_features
         ``pd.DataFrame`` containing static features for the series. The index
         should contain the key of the series in the ``dataframes`` argument.
-    ignore_last_n_targets
-        For target and past dynamic features last ``ignore_last_n_targets``
+    future_length
```
lostella, line 63: 👍

lostella: Let's mark the PR as breaking then

src/gluonts/dataset/pandas.py
```diff
@@ -84,67 +84,66 @@ class PandasDataset:
...
 
-        assert isinstance(self._pairs, SizedIterable)
+        assert isinstance(pairs, SizedIterable)
+        self._data_entries = Map(self._pair_to_dataentry, pairs)
+
+        if self.feat_dynamic_real is None:
+            self.feat_dynamic_real = pd.DataFrame()
+
+        if self.past_feat_dynamic_real is None:
+            self.past_feat_dynamic_real = pd.DataFrame()
```
lostella: Does the typing of these attributes need to be adjusted?

jaheba: The problem is that one can pass a `None`, but internally it is guaranteed to be not None.

Maybe it's easier to not make this a dataclass.

lostella: But then I don't understand the change: this is supposed to be a list of column names, here it gets turned into a dataframe if it was None, what happens in the other case?

src/gluonts/dataset/pandas.py
```diff
@@ -84,71 +84,63 @@ class PandasDataset:
...
     def num_feat_dynamic_real(self):
-        if self.feat_dynamic_real is None:
-            return 0
         return len(self.feat_dynamic_real)
 
     @property
     def num_past_feat_dynamic_real(self):
-        if self.past_feat_dynamic_real is None:
-            return 0
         return len(self.past_feat_dynamic_real)
```
lostella: These changes work as well if `self.feat_dynamic_real` is forced to be `[]` when found `None` (same for the other)

src/gluonts/dataset/pandas.py
```diff
@@ -84,71 +84,63 @@ class PandasDataset:
...
 
-    @property
-    def cardinalities(self):
-        return [
-            max(self._static_cats[c]) + 1 for c in self._static_cats.columns
-        ]
```
lostella: Why remove this? I think this is valuable information

src/gluonts/dataset/pandas.py
lostella: I think `dataclass` makes sense in this case, I'm not yet ready to give that up

src/gluonts/dataset/pandas.py
```diff
@@ -50,15 +53,15 @@ class PandasDataset:
         Frequency of observations in the time series. Must be a valid pandas
         frequency.
     feat_dynamic_real
-        List of column names that contain dynamic real features.
+        list of column names that contain dynamic real features.
     past_feat_dynamic_real
-        List of column names that contain dynamic real features only available
+        list of column names that contain dynamic real features only available
```
lostella: not sure I agree, it's a word (not a code expression) starting a sentence

jaheba: I think I accidentally changed this as well.

src/gluonts/dataset/pandas.py
```diff
@@ -171,31 +169,30 @@ def _pair_to_dataentry(self, pair) -> DataEntry:
...
+        target = df[self.target].values
+        target = target[: len(target) - self.future_length]
+        entry["target"] = target.T
+
         if item_id is not None:
             entry["item_id"] = item_id
 
-        if self.num_feat_static_cat > 0:
-            entry["feat_static_cat"] = self._static_cats.loc[item_id].values
+        if self.num_feat_static_cat:
```
lostella: I prefer `self.num_feat_static_cat > 0` (not a fan of implicit conversions to `bool`)

jaheba: But that's the pythonic way to do it.

src/gluonts/dataset/pandas.py
```diff
@@ -207,23 +204,24 @@ def __len__(self) -> int:
...
     def __str__(self) -> str:
-        return (
-            f"PandasDataset<"
-            f"size={len(self)}, "
-            f"freq={self.freq}, "
-            f"num_dynamic_real={self.num_feat_dynamic_real}, "
-            f"num_past_dynamic_real={self.num_past_feat_dynamic_real}, "
-            f"num_static_real={self.num_feat_static_real}, "
-            f"num_static_cat={self.num_feat_static_cat}, "
-            f"cardinalities={self.cardinalities}>"
```
lostella, line 218: Why not printing the cardinalities? I think it can be useful for quick validation

lostella (approved):

Issue #2577, opened 2023-01-20
Empty `target` when series is `split` too close to the end


lostella:
## Description
The `split` function returns an object that generates instances with empty ground truth when the given `prediction_length` is longer than the available data.

## To Reproduce

```python
from gluonts.dataset.split import split
import pandas as pd
import numpy as np

dataset = [{"start": pd.Period("2022-03-04", freq="D"), "target": np.arange(20)}]

train, valid = split(dataset, offset=-3)
list(valid.generate_instances(prediction_length=4, windows=1))  # generates empty target in the ground truth
```

## Error message or code output
```
[({'start': Period('2022-03-04', 'D'),
   'target': array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15, 16])},
  {'start': Period('2022-03-21', 'D'), 'target': array([], dtype=int64)})]
```

PR #2587, opened 2023-01-24
Add assertion to split function ensuring valid windows

marcelkollovieh:
*Issue:* #2577

*Description of changes:*
This PR addresses issue #2577 and added assertions for `OffsetSplitter` and `DateSplitter` to ensure that the targets of the label have valid target lengths, i.e. match `prediction_length.`

### DIFF
--- a/src/gluonts/dataset/split.py
+++ b/src/gluonts/dataset/split.py
@@ -355,9 +355,8 @@ class DateSplitter(AbstractBaseSplitter):
         input_slice = slice(None, base + offset)
         label_slice = slice(base + offset, base + offset + prediction_length)
         assert (
-            base + offset + prediction_length
-            <= entry[FieldName.TARGET].shape[-1]
-        ), "Offset too short to generate windows"
+            label_slice.stop <= entry[FieldName.TARGET].shape[-1]
+        ), "Date is too early to generate windows"
         return (
             slice_data_entry(
                 entry, input_slice, prediction_length=prediction_length
--- a/src/gluonts/dataset/split.py
+++ b/src/gluonts/dataset/split.py
@@ -356,7 +356,7 @@ class DateSplitter(AbstractBaseSplitter):
         label_slice = slice(base + offset, base + offset + prediction_length)
         assert (
             label_slice.stop <= entry[FieldName.TARGET].shape[-1]
-        ), "Date is too early to generate windows"
+        ), "Date is too late to generate windows"
         return (
             slice_data_entry(
                 entry, input_slice, prediction_length=prediction_length
--- a/src/gluonts/dataset/split.py
+++ b/src/gluonts/dataset/split.py
@@ -306,8 +306,12 @@ class OffsetSplitter(AbstractBaseSplitter):
         self, entry: DataEntry, prediction_length: int, offset: int = 0
     ) -> Tuple[DataEntry, DataEntry]:
         offset_ = self.offset + offset
-        if self.offset < 0 and offset_ >= 0:
-            offset_ += len(entry)
+
+        assert (-offset_ >= prediction_length and self.offset < 0) or (
+            offset_ + prediction_length <= entry[FieldName.TARGET].shape[-1]
+            and self.offset >= 0
+        ), "Offset too short to generate windows"
+
         if offset_ + prediction_length:
             input_slice = slice(None, offset_)
             label_slice = slice(offset_, offset_ + prediction_length)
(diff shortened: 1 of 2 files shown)


src/gluonts/dataset/split.py
```diff
@@ -350,6 +354,10 @@ def test_pair(
         base = periods_between(entry["start"], self.date)
         input_slice = slice(None, base + offset)
         label_slice = slice(base + offset, base + offset + prediction_length)
+        assert (
+            base + offset + prediction_length
+            <= entry[FieldName.TARGET].shape[-1]
+        ), "Offset too short to generate windows"
```
lostella: rhs of the inequality is just `label_slice.stop`, you could use that I guess. Also maybe in this case the message should be `"Date is too early to generate windows"`?

src/gluonts/dataset/split.py
```diff
@@ -306,8 +306,12 @@ def test_pair(
         self, entry: DataEntry, prediction_length: int, offset: int = 0
     ) -> Tuple[DataEntry, DataEntry]:
         offset_ = self.offset + offset
-        if self.offset < 0 and offset_ >= 0:
-            offset_ += len(entry)
```
lostella, line 310: This was clearly a bug, but I'm wondering whether `offset_ += entry[FieldName.TARGET].shape[-1]` is still needed here: if `self.offset == -offset` then `offset_ == 0` then the `input_slice` below may be incorrect? It would be `slice(None, 0)` which would be empty I think.

lostella (commented):
I think we probably miss a few tests here @MarcelK1102 do you want to add them? This module is covered by https://github.com/awslabs/gluonts/blob/dev/test/dataset/test_split.py

src/gluonts/dataset/split.py
marcelkollovieh: I think it is not needed. If `self.offset == -offset` and `offset_ == 0` then the assertion fails since `-offset_ < prediction_length`. If we would add `offset_ += entry[FieldName.TARGET].shape[-1]` then the assertion would fail as well since `-offset_ < prediction_length` again. I can't think of any scenario where it would change the outcome.

src/gluonts/dataset/split.py
marcelkollovieh: Done

lostella: My bad: `"Date is too late to generate windows"` :-)

marcelkollovieh:
> I think we probably miss a few tests here @MarcelK1102 do you want to add them? This module is covered by https://github.com/awslabs/gluonts/blob/dev/test/dataset/test_split.py

@lostella Sure, I will do that!

src/gluonts/dataset/split.py
lostella: Right. I'm wondering though whether ensuring `offset_` is always positive would render the logic simpler to read.
So,

```python
if self.offset < 0:
    offset_ += entry[FieldName.TARGET].shape[-1]

assert offset_ + prediction_length <= entry[FieldName.TARGET].shape[-1], "Offset too short"
```

(didn't check this ☝️ thoroughly but makes sense to me)

src/gluonts/dataset/split.py
marcelkollovieh: Corrected :)

src/gluonts/dataset/split.py
marcelkollovieh: The two inequalities are equal then. I tested it, it works. I updated it accordingly.

src/gluonts/dataset/split.py
```diff
@@ -306,8 +306,12 @@ def test_pair(
...
+        if self.offset < 0:
+            offset_ += entry[FieldName.TARGET].shape[-1]
+        assert (
+            offset_ + prediction_length <= entry[FieldName.TARGET].shape[-1]
+        ), "Offset too short to generate windows"
```
lostella: Also here "short" sounds strange. To give proper context maybe a message like this would work better

```
"Not enough data to generate some of the windows; try splitting data at an earlier offset"
"Not enough data to generate some of the windows; try splitting data at an earlier date"
```

lostella (changes requested):

src/gluonts/dataset/split.py
marcelkollovieh: I agree with you. I adapted the assertions.

lostella:
Looks good now! Let's add the tests and then this can be merged!

lostella (approved):

PR #2588, opened 2023-01-24
Fix maybe map_or/map_or_else return types.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/maybe.py
+++ b/src/gluonts/maybe.py
@@ -75,9 +75,7 @@ def map(self: Optional[T], fn: Callable[[T], U]) -> Optional[U]:
     return map_or(self, None, fn)
 
 
-def map_or(
-    self: Optional[T], default: Optional[U], fn: Callable[[T], U]
-) -> Optional[U]:
+def map_or(self: Optional[T], default: Optional[U], fn: Callable[[T], U]) -> U:
     """
     ::
         >>> map_or(["x"], 0, len)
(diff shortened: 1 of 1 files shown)


src/gluonts/maybe.py
```diff
@@ -75,9 +75,7 @@ def map(self: Optional[T], fn: Callable[[T], U]) -> Optional[U]:
     return map_or(self, None, fn)
 
 
-def map_or(
-    self: Optional[T], default: Optional[U], fn: Callable[[T], U]
-) -> Optional[U]:
+def map_or(self: Optional[T], default: Optional[U], fn: Callable[[T], U]) -> U:
```
lostella: `default: U`?

lostella:
Maybe you could also add docstrings? I find it a bit hard to read through what these do, maybe because of the ordering of the arguments (reading `map` I would expect a function first)

lostella:
Also, did `mypy` not catch these?

src/gluonts/maybe.py
jaheba: Thanks!

jaheba:
> Also, did `mypy` not catch these?

It does if you run it directly. But I think our current setup doesn't, because of the marker files.

jaheba:
> Maybe you could also add docstrings? I find it a bit hard to read through what these do, maybe because of the ordering of the arguments (reading `map` I would expect a function first)

Granted, it is a bit weird. But I essentially followed Rust ordering of arguments, putting the `self` as the first one. I think the reason for putting the `fn` last is that you can then have longish inline functions, which are not easily doable with Python's lamdas.

lostella (approved):

PR #2591, opened 2023-01-24
Add wrapper for Nixtla/hierarchicalforecast

melopeo:
*Description of changes:*
Add wrapper to call Nixtla/hierarchicalforecast package.

### DIFF
--- a/src/gluonts/ext/hierarchicalforecast/_predictor.py
+++ b/src/gluonts/ext/hierarchicalforecast/_predictor.py
@@ -130,7 +130,6 @@ def format_reconciled_forecasts(
 def prune_fcst_df(
     df: pd.DataFrame, base_reconciliation_model_name: str
 ) -> pd.DataFrame:
-
     # keep certain columns
     columns_to_keep = ["ds"]
     columns_to_keep.extend(
--- a/src/gluonts/ext/hierarchicalforecast/_predictor.py
+++ b/src/gluonts/ext/hierarchicalforecast/_predictor.py
@@ -244,7 +244,7 @@ class HierarchicalForecastPredictor(RepresentablePredictor):
             df=Y_df,
             models=self.models,
             freq=entry["start"].freq,
-            n_jobs=self.config.n_jobs,
+            n_jobs=self.n_jobs,
         )
 
         # compute base forecasts
--- a/src/gluonts/ext/hierarchicalforecast/_predictor.py
+++ b/src/gluonts/ext/hierarchicalforecast/_predictor.py
@@ -218,7 +218,7 @@ class HierarchicalForecastPredictor(RepresentablePredictor):
             if ts_names is not None
             else [str(x) for x in range(len(S))]
         )
-        self.S = get_formatted_S(S, ts_names)
+        self.S = get_formatted_S(S, self.ts_names)
         self.tags = {key: np.array(val) for key, val in tags.items()}
         self.intervals_method = intervals_method
         self.config = ModelConfig(
--- a/src/gluonts/ext/hierarchicalforecast/_predictor.py
+++ b/src/gluonts/ext/hierarchicalforecast/_predictor.py
@@ -67,6 +67,7 @@ def format_data_entry(entry: DataEntry, S: pd.DataFrame) -> pd.DataFrame:
     1) the index corresponds to the name of the time series,
     2) the columns 'ds'/'y' correspond to timestamps/actuals, respectively.
     """
+
     df = pd.DataFrame(entry["target"]).T
     df.columns = S.index.tolist()
     df.index = pd.date_range(
@@ -102,17 +103,6 @@ def unpivot(df: pd.DataFrame) -> pd.DataFrame:
             "y": df.to_numpy().ravel("F"),
         }
     )
-
-
-# def get_model_name(df: pd.DataFrame) -> Union[str, List[str]]:
-#     df_columns = df.columns.to_list()
-#     drop_cols = ["ds", "y"] if "y" in df_columns else ["ds"]
-#     model_names = [x for x in df_columns if x not in drop_cols]
-#
-#     if len(model_names) == 1:
-#         return model_names[0]
-#
-#     return model_names
 
 
 def format_reconciled_forecasts(
--- a/src/gluonts/ext/hierarchicalforecast/_predictor.py
+++ b/src/gluonts/ext/hierarchicalforecast/_predictor.py
@@ -41,7 +41,7 @@ models_without_fitted_capability = [
 
 def get_formatted_S(
     _S: Union[List[List[int]], np.ndarray],
-    ts_names: List[Union[str, Any]],
+    ts_names: List[str],
 ) -> pd.DataFrame:
     S = np.array(_S)
 
--- a/src/gluonts/ext/hierarchicalforecast/_predictor.py
+++ b/src/gluonts/ext/hierarchicalforecast/_predictor.py
@@ -12,7 +12,7 @@
 # permissions and limitations under the License.
 
 from dataclasses import dataclass
-from typing import List, Optional, Union, Iterable, Any
+from typing import List, Optional, Union, Any, Dict
 
 import numpy as np
 import pandas as pd
@@ -28,7 +28,7 @@ from gluonts.model.predictor import RepresentablePredictor
 from gluonts.model.forecast import QuantileForecast
 from gluonts.ext.statsforecast import ModelConfig as _ModelConfig
 
-models_with_fitted_capability = [
+models_without_fitted_capability = [
     "ADIDA",
     "CrostonClassic",
     "CrostonOptimized",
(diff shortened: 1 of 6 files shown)


src/gluonts/ext/hierarchicalforecast/_predictor.py
```diff
@@ -0,0 +1,309 @@
...
+    "IMAPA",
+    "SeasWA",
+    "TSB",
+    "WindowAverage",
+]
+
+
+def get_formatted_S(
+    _S: Union[List[List[int]], np.ndarray],
+    ts_names: List[Union[str, Any]],
```
lostella: `Union[str, Any]` is equivalent to `Any`

src/gluonts/ext/hierarchicalforecast/_predictor.py
```diff
@@ -0,0 +1,309 @@
...
+) -> pd.DataFrame:
+    S = np.array(_S)
+
+    return pd.DataFrame(
+        S, index=ts_names, columns=ts_names[S.shape[0] - S.shape[1] :]
```
lostella, line 55: So it looks like we're just discarding the first few rows and never use them again (if I see through the code correctly), why then are the first rows required at all?

src/gluonts/ext/hierarchicalforecast/_predictor.py
```diff
@@ -0,0 +1,309 @@
...
+    )
+
+
+def format_data_entry(entry: DataEntry, S: pd.DataFrame) -> pd.DataFrame:
```
lostella, line 59: I think docstrings here would help understanding, even simple ones quickly describing what is it that this functions return

src/gluonts/ext/hierarchicalforecast/_predictor.py
```diff
@@ -0,0 +1,309 @@
...
+
+    return df.rename(columns=mapper)
+
+
+def prune_fcst_df(
+    df: pd.DataFrame, base_reconciliation_model_name: str
+) -> pd.DataFrame:
+
+    df = keep_model_columns(df, base_reconciliation_model_name)
+    df = rename_model_columns(df, base_reconciliation_model_name)
```
lostella: I think these two function are simple enough, that just having the entire behaviour here in `prune_fcst_df` would be easier to follow, maybe

src/gluonts/ext/hierarchicalforecast/_predictor.py
```diff
@@ -0,0 +1,309 @@
...
+
+    return df
+
+
+@dataclass
+class ModelConfig(_ModelConfig):
+    n_jobs: Optional[int] = 1
```
lostella: Maybe `n_jobs` does not really belong together with the rest of `ModelConfig`: it seems like `ModelConfig` is really concerned with quantile levels and input/output interface of models. For example, why `n_jobs` goes there, but `S` does not?

src/gluonts/ext/hierarchicalforecast/_predictor.py
```diff
@@ -0,0 +1,309 @@
...
+
+        Y_df = format_data_entry(entry, self.S)
+
+        # set up forecaster
+        forecaster = StatsForecast(
+            df=Y_df,
+            models=self.models,
+            freq=entry["start"].freq,
+            n_jobs=self.config.n_jobs,
+        )
```
lostella: How does this `StatsForecast` class work? It fits different models independently? And its output is what?

I see in the constructor that `self.models` is instantiated to a singleton model, so I'm wondering whether it is necessary to use this class?

melopeo: This is an interesting point. In short: calling `self.models` is a bit more convolved in this case as we have multiple time series - one per element in the given hierarchy.

When using the class StatsForecast, it automatically handles these multiple time series via the class [GroupedArray](https://github.com/Nixtla/statsforecast/blob/main/statsforecast/core.py#L32). This class is done in a way that it generates more class attributes that do not make it clear how to get out of the logics of StatsForecast. For instance, forecasts, in our case, are computed with [self.ga.forecast](https://github.com/Nixtla/statsforecast/blob/cb80fffc33655f5c7f6cbf44bb3baa4347e431d1/statsforecast/core.py#L750).

With the class GroupedArray, formatting of forecasts is done automatically, and makes it easier for us by avoiding to handle this ourselves. Of course we could explore this further.

src/gluonts/ext/hierarchicalforecast/_predictor.py
melopeo: All the rows are taken. What we do here is to take the matrix `S` and convert it to a suitable format for `hierarchicalforecast`, namely: make it a dataframe, set the index to `ts_names`, and provide names to the columns of `S`, which correspond to the first entries of `ts_names`. I will add a docstring to make it more accessible.

src/gluonts/ext/hierarchicalforecast/_predictor.py
melopeo: Thank you. I updated it to `List[str]`

src/gluonts/ext/hierarchicalforecast/_predictor.py
melopeo: Sounds good. I added the docstring.

src/gluonts/ext/hierarchicalforecast/_predictor.py
melopeo: Yes, I updated the code accordingly.

src/gluonts/ext/hierarchicalforecast/_predictor.py
melopeo: I agree. The new version follows your suggestion.

src/gluonts/ext/hierarchicalforecast/_predictor.py
```diff
@@ -0,0 +1,316 @@
...
+def format_data_entry(entry: DataEntry, S: pd.DataFrame) -> pd.DataFrame:
+    """
+    formats data entry as required by hierarchicalforecast.
+
+    entry is a dictionary with keys: 'start', 'item_id', 'target'.
+    entry['target'] is a np.narray with shape (num_ts, num_timestamps),
+    and each row corresponds to one time series of the hierarchy.
+    The goal is to reshape this DataEntry as a dataframe where:
+    1) the index corresponds to the name of the time series,
+    2) the columns 'ds'/'y' correspond to timestamps/actuals, respectively.
```
lostella:
```suggestion
    Format data entry as required by hierarchicalforecast.

    ``entry`` is a dictionary with keys: ``"start"``, ``"item_id"``, ``"target"``.
    ``entry["target"]`` is a ``np.ndarray`` with shape ``(num_ts, num_timestamps)``,
    and each row corresponds to one time series of the hierarchy.
    The goal is to reshape this DataEntry as a dataframe where:
    1) the index corresponds to the name of the time series,
    2) the columns ``"ds"`` and ``"y"`` correspond to timestamps and actuals, respectively.
```

src/gluonts/ext/hierarchicalforecast/_predictor.py
```diff
@@ -0,0 +1,316 @@
...
+    """
+    unpivots data frame
+
+    The input dataframe has as index the time stamps
+    and columns per time series of the hierarchy.
+    We unpivot this so that the final dataframe has
+    three columns, i.e. "unique_id", "ds", and "y", where
+    1) "unique_id" has the name of the corresponding time series,
+    2) "ds" has the corresponding time stamps,
+    3) "y" has the actuals.
```
lostella:
```suggestion
    Unpivot data frame.

    The input dataframe has as index the time stamps,
    and one column per each time series of the hierarchy.
    We unpivot this so that the final dataframe has
    three columns, i.e. ``"unique_id"``, ``"ds"``, and ``"y"``, where
    1) ``"unique_id"`` has the name of the corresponding time series,
    2) ``"ds"`` has the corresponding time stamps,
    3) ``"y"`` has the actuals.
```

lostella (changes requested):
@melopeo just suggested some changes to the docstrings, then I think this can be merged 🚀

src/gluonts/ext/hierarchicalforecast/_predictor.py
melopeo: Thank you! I updated the PR with it.

src/gluonts/ext/hierarchicalforecast/_predictor.py
melopeo: Thank you! I updated the PR with it.

lostella (approved):

jaheba:
Should we add a github workflow to test these changes?

lostella:
This could be bundled with the statsforecast tests, since the packages are very closely related

melopeo:
Good idea. I will open a PR on this.

PR #2593, opened 2023-01-25
Rework `maybe`.

jaheba:
* Swap operands for `map_or` and `map_or_else` such that the function argument comes before the default argument.
* Add `Maybe` class so one can do `Maybe(x).unwrap()` in addition to `unwrap(x)`, etc.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/maybe.py
+++ b/src/gluonts/maybe.py
@@ -12,17 +12,30 @@
 # permissions and limitations under the License.
 
 """
-This module contains functions that work on `Optional` values. In contrast to
+This module contains functions that work on ``Optional`` values. In contrast to
 other approaches, this does not wrap values into a dedicated type, but works
-on normal Python values, which are of type `Optional[T]`.
+on normal Python values, which are of type ``Optional[T]``.
 
 Thus, some functions are implemented identically but have different type
-signatures. For example, both `map` and `and_then` both just apply a function
-to a value if it is not `None`, but the result of `map` is `T` and the result
-of `and_then` is `Optional[T]`.
+signatures. For example, both ``map`` and ``and_then`` both just apply a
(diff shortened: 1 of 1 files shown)


melopeo (changes requested):
Can we add tests?

src/gluonts/maybe.py
```diff
@@ -43,69 +43,71 @@ def expect(self: Optional[T], msg: str) -> T:
...
-    if self is None:
+    if val is None:
         raise ValueError(msg)
 
-    return self
+    return val
 
 
-def do(self: Optional[T], fn: Callable[[T], U]) -> Optional[T]:
+def do(val: Optional[T], fn: Callable[[T], U]) -> Optional[T]:
```
melopeo, line 68: Why do we have three inputs here, i.e. `val, fn, U`? I don't see how `U` is used here. Most likely I am missing something here ;)

melopeo: No I see, I was indeed missing something.

src/gluonts/maybe.py
```diff
@@ -115,69 +117,71 @@ def unwrap(self: Optional[T]) -> T:
...
         42
         >>> and_then(None, lambda xs: xs[0] if xs else None)
     """
-    return map(self, fn)  # type: ignore
+    return map(val, fn)  # type: ignore
 
 
-def or_(self: Optional[T], default: Optional[T]) -> Optional[T]:
-    if self is None:
+def or_(val: Optional[T], default: Optional[T]) -> Optional[T]:
```
melopeo, line 217: Can we add a small docstring to this one as it is done for the other functions?

jaheba:
> Can we add tests?

The docstrings are the tests.

melopeo (dismissed):

PR #2599, opened 2023-01-27
Fix bug with static cardinalities in `PandasDataset`

lostella:
*Issue #, if available:*

*Description of changes:*


jaheba (approved):
Thanks!

One thing I've seen is that `.to_numpy()` should be used in favour of `.values`.

jaheba (approved):

lostella:
> Thanks!
> 
> One thing I've seen is that `.to_numpy()` should be used in favour of `.values`.

Yes, we can do that all over the place separately

PR #2604, opened 2023-01-29
Fix requirements following breaking change in setuptools

lostella:
Recent [breaking changes](https://github.com/pypa/setuptools/blob/main/CHANGES.rst#breaking-changes) in setuptools appear to have broken our setup.py, this PR fixes it.


jaheba:
Thanks for fixing it. How did you know which was was the bad one?

jaheba (approved):

lostella:
> Thanks for fixing it. How did you know which was was the bad one?

I just excluded the extras one by one and did `pip install -e.` until it was working

PR #2609, opened 2023-01-31
Zebras

jaheba:
Your favourite black and white animal.

*Issue #, if available:*

*Description of changes:*

With `TimeFrame` we can simplify some transformations:

```py
@dataclass
class AddAgeFeature:
    output_field: str
    log_scale: bool = True
    dtype: Type = np.float32

    def __call__(self, tf):
        age = np.arange(len(tf), dtype=self.dtype)

        if self.log_scale:
            age = np.log10(2 + age)

        return tf.set(self.output_field, age)
```

```py
@dataclass
class AddObservedValuesIndicator:
    target: str = "target"
    output_field = "observed_values"
    dtype: Type = np.float32

    def __call__(self, tf):
        target = tf.c(self.target)
        nan_entries = np.isnan(target)

        target[nan_entries] = 0.0

        return tf.set_like(
            self.target,
            self.target,
            target,
            replace=True,
        ).set_like(
            self.target,
            self.output_field,
            np.invert(nan_entries).astype(self.dtype),
        )
```

```py
@dataclass
class AddTimeFeatures:
    output_field: str
    time_features: List[TimeFeature]
    dtype: Type = np.float32

    def __call__(self, tf):
        return tf.set(
            self.output_field,
            np.vstack(
                [feat(tf.index) for feat in self.time_features],
            ).astype(self.dtype),
        )
```

---


src/gluonts/time_feature/_base.py
```diff
@@ -27,109 +27,109 @@ def second_of_minute(index: pd.PeriodIndex) -> np.ndarray:
     """
     Second of minute encoded as value between [-0.5, 0.5]
     """
-    return index.second.values / 59.0 - 0.5
+    return index.second / 59.0 - 0.5
 
 
 def second_of_minute_index(index: pd.PeriodIndex) -> np.ndarray:
```
lostella: So here the annotation does not apply anymore I guess

lostella (commented):
To clarify: is performance the goal here? Is this intended to be as general as `pd.Period`? If so, would it make sense to contribute to `pandas` as well?

jaheba:
The goal is mainly independence from pandas. Performance is a nice additional benefit.

Also, this should be more consistent and more ergonomic to use.

src/gluonts/time_feature/_base.py
```diff
@@ -192,25 +192,22 @@ def time_features_from_frequency_str(freq_str: str) -> List[TimeFeature]:
...
-        offsets.BusinessDay: [day_of_week, day_of_month, day_of_year],
-        offsets.Hour: [hour_of_day, day_of_week, day_of_month, day_of_year],
-        offsets.Minute: [
+    features_by_offsets: Dict[str, List[TimeFeature]] = {
+        "Y": [],
+        "Q": [month_of_year],
+        "M": [month_of_year],
+        "W": [day_of_month, week_of_year],
+        "D": [day_of_week, day_of_month, day_of_year],
+        # offsets.BusinessDay: [day_of_week, day_of_month, day_of_year],
```
jaheba, line 191: Need to look into this. Do we have a use-case for `BusinessDay`?

src/gluonts/dataset/arrow/enc.py
```diff
@@ -102,6 +103,11 @@ def _encode_py_pd_preiod(val: pd.Period):
     return val.to_timestamp()
 
 
+@_encode_py_to_arrow.register
+def _encode_py_pd_preiod(val: zb.Period):
```
lostella:
```suggestion
def _encode_py_pd_period(val: zb.Period):
```

src/gluonts/zebras/_period.py
```diff
@@ -0,0 +1,291 @@
...
+
+class Periods(_BasePeriod):
+    def _astype(self, ty):
+        return self.data.astype(ty)
+
+    def first(self, count: int) -> Periods:
+        return self[:count]
+
+    def last(self, count: int) -> Periods:
+        return self[-count:]
```
lostella: There should be one, and preferably only one, obvious way to do it!

jaheba: Do you prefer the slicing syntax? Because I don't.

Issue #2594, opened 2023-01-26
num_batches_per_epoch not well accounted for by gluonts.torch.model.deepar.DeepAREstimator


pbruneau:
## Description
During each epoch, validation starts at the `num_batches_per_epoch` specified in the `DeepAREstimator` constructor call, but after this, training exhausts the whole training dataset, while validation runs concurrently.

Maybe I am not using the API correctly, or something is wrong with the gluonts version I'm using?

## To Reproduce
Please check out [this minimal example](https://gist.github.com/pbruneau/bef11c78662a2a2dbfcf6381be4f2c04)

## Error message or code output
Before `num_batches_per_epoch` is reached, training proceeds as expected:
```
141 K     Trainable params
0         Non-trainable params
141 K     Total params
0.564     Total estimated model params size (MB)
Epoch 0: : 24it [00:01, 23.54it/s, loss=7.04, v_num=35]
```

After it is reached (50 in that case), training and validation run concurrently:
```
141 K     Trainable params
0         Non-trainable params
141 K     Total params
0.564     Total estimated model params size (MB)
Epoch 0: : 169it [00:08, 19.96it/s, loss=6.34, v_num=37]                                                                                                                        
Validation DataLoader 0: : 119it [00:05, 20.99it/s]
```

## Environment
- Operating system: Linux
- Python version: 3.7.10
- GluonTS version: v0.11.7
- MXNet version: 1.9.1

lostella:
@pbruneau I simplified your example by removing `num_workers`:

```python
from gluonts.dataset.repository.datasets import get_dataset
from gluonts.torch.model.deepar import DeepAREstimator

dataset = get_dataset("electricity")
context_length = 2 * 7 * 24
prediction_length = dataset.metadata.prediction_length

model = DeepAREstimator(
    prediction_length=prediction_length,
    context_length=context_length,
    freq='1H',
    batch_size=32,
    num_batches_per_epoch=2,
)

predictor = model.train(
    dataset.train,
    dataset.test,
)
```

The validation data consists of 2247 series in this case, which are 71 batches of size 32; after the `num_batches_per_epoch=2` batches are consumed for training, 71 are accounted for in the validation logging line, adding up to 73. I think the first logging line is just counting and timing the total number of batches processed.

It's left to see whether
* logging is working as expected (this is handled by PyTorch Lightning, but the estimator class may be using it wrong)
* `num_workers` is working as expected

pbruneau:
[...]

PR #2615, opened 2023-02-01
Improvement of parallel work

Alex2251:
Improvement of duplicates in parallel work. Previous realization did duplicates in parallel work.

*Issue #, if available:*
In parallel work function duplicates output if you use DataLoader with num_workers > 1

*Description of changes:*
Added functionality to check on parallel work. If solution works in parallel than we take number of workers and each worker returns unique sequence.

### DIFF
--- a/src/gluonts/torch/util.py
+++ b/src/gluonts/torch/util.py
@@ -151,7 +151,9 @@ class IterableDataset(torch.utils.data.IterableDataset):
         worker_total_num = torch.utils.data.get_worker_info().num_workers
         worker_id = torch.utils.data.get_worker_info().id
 
-        yield from itertools.islice(self.iterable, worker_id, None, worker_total_num)
+        yield from itertools.islice(
+            self.iterable, worker_id, None, worker_total_num
+        )
 
 
 def repeat_along_dim(a: torch.Tensor, dim: int, repeats: int) -> torch.Tensor:
--- a/src/gluonts/torch/util.py
+++ b/src/gluonts/torch/util.py
@@ -151,9 +151,7 @@ class IterableDataset(torch.utils.data.IterableDataset):
         worker_total_num = torch.utils.data.get_worker_info().num_workers
         worker_id = torch.utils.data.get_worker_info().id
 
-        yield from itertools.islice(
-            self.iterable, worker_id, None, worker_total_num
-        )
+        yield from itertools.islice(self.iterable, worker_id, None, worker_total_num)
 
 
 def repeat_along_dim(a: torch.Tensor, dim: int, repeats: int) -> torch.Tensor:
--- a/src/gluonts/torch/util.py
+++ b/src/gluonts/torch/util.py
@@ -13,6 +13,7 @@
 
 from typing import List, Optional, Type
 import inspect
+import itertools
 
 import torch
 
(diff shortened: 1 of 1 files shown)


lostella:
Hi @Alex2251, thanks for this! By the looks of it, this might be related to #2594?

To fix the style checks failures:

```
pip install -U black
black src test
```

and commit the changes

Alex2251:
Hi @lostella! Thanks for comment and this commit will solve #2594.

lostella:
@Alex2251 while this fixes #2594, each worker will still process each entry it needs to skip, and will slow down because of it. The downstream network will not be invoked (and there may be some time saving because of it), but any lazy transformation will still be applied to each element of the dataset. I think this was also pointed out by @gorold in [this comment](https://github.com/awslabs/gluonts/issues/2594#issuecomment-1418637840)

Alex2251:
@lostella Yes it is. I tried to explaine my opinion in #2594, I think it is critical bug because you have a lot of the same batches.

jaheba:
I wonder if we should use that ``DataLoader`` at all.

As I understand it, it does three things:

* collect inputs into batches
* stack those batches
* turns things into ``torch.tensor``

That's something we can easily do ourselves and in fact already do elsewhere.

Alex2251:
Additionaly this DataLoader prevent duplicates, today there are duplicates. It shuld be prevented to use such dataloader because many people and me first time thought that it works from a box in a right way in parallel mode.

lostella:
I think @jaheba means torch’s `DataLoader`: whatever we use, as long as the data is accessible only as an iterable stream, if multiple processes are to be used then there must be some logic to handle it. I’m not sure that avoiding torch’s `DataLoader` buys us anything.

The solution in this PR solves an issue, but it’s not ideal. The ideal solution, as pointed out, is either to rely on a “map” (“indexable”) dataset, or at least on a dataset type that can be sharded so that each process iterates its own shard: we used to have something like the latter, but it was arguably a bit clunky.

lostella:
@jaheba what torch’s `DataLoader` buys us, I think, is to be able to use PyTorch Lightning: whether one likes it or not, I think it’s better than having to maintain our own training loop.

The best code is no code!

Alex2251:
I agree.

jaheba:
> @jaheba what torch’s `DataLoader` buys us, I think, is to be able to use PyTorch Lightning: whether one likes it or not, I think it’s better than having to maintain our own training loop.
> 
> The best code is no code!

That's just not true.

gorold:
torch's ```DataLoader``` helps handle the multiprocessing for datasets! It's really important for larger batch sizes and can speed up training by 10x (assuming the underlying dataset is sharded properly). Implementing custom dataloading tooling which supports multiprocessing and interops with pytorch lightning (and all the accelerators, e.g. deepspeed, ddp, etc.) may be quite complicated.

jaheba:
We have had bad experiences when using `multiprocessing` for loading data in the past and have therefore removed it. Our current strategy is to make the data-loading pipeline as fast as possible, and looking at parallelising it in the future.

Something we should do is to have some simple benchmarks to look at pipeline throughput, so that we actually have some numbers that we can talk about.

jaheba:
We've removed the `DataLoader` in #2628.

I am going to create an issue to track the re-introduction of parallelised data loaders.

PR #2618, opened 2023-02-06
Fix off-by-one in torch DeepAR

### LINKED ISSUE #2616: Off-by-one bug in PyTorch DeepAR implementation
## Description
(A clear and concise description of what the bug is.)

Please correct me if I'm wrong, but there might be an off by 1 error in the below code snippet. What should happen is that ```time_feat``` at time step **t** should be concatenated with ```target``` at time step **t-1.** But because lags is taken as the input without concatenating the target itself, what is happening is that ```time_feat``` **t+1** is concatenated with ```target``` **t-1**.

https://github.com/awslabs/gluonts/blob/60e1767d80e1d5b932d5e94cc2b51439a970dca9/src/gluonts/torch/model/deepar/module.py#L267-L283

I think the below change should be made? 

```python
 time_feat = ( 
     torch.cat( 
         ( 
             past_time_feat[..., -self.context_length :, :], 
             future_time_feat, 
         ), 
         dim=-2, 
     ) 
     if future_time_feat is not None 
     else past_time_feat[..., -self.context_length :, :] 
 ) 
```

lostella:
*Issue #, if available:* Fixes #2616

*Description of changes:* The RNN in torch DeepAR was getting dynamic features shifted by one as input.
- Split `prepare_rnn_input` out of `unroll_lagged_rnn`
- Add test to verify alignment of RNN inputs (target values vs dynamic features, test is failing prior to fix)
- Change is breaking in that previously serialized models will not work as before (network changes)

Todo:
- [x] Take the opportunity to improve the code
- [x] Test effect of bug/fix (e.g. simple indicator that zeroes-out a series, model should not get it right prior to fix)

### DIFF
--- a/src/gluonts/torch/model/deepar/module.py
+++ b/src/gluonts/torch/model/deepar/module.py
@@ -132,6 +132,7 @@ class DeepARModel(nn.Module):
             else [min(50, (cat + 1) // 2) for cat in cardinality]
         )
         self.lags_seq = lags_seq or get_lags_for_frequency(freq_str=freq)
+        self.lags_seq = [l - 1 for l in self.lags_seq]
         self.num_parallel_samples = num_parallel_samples
         self.past_length = self.context_length + max(self.lags_seq)
         self.embedder = FeatureEmbedder(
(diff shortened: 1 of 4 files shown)


src/gluonts/torch/model/deepar/module.py
```diff
@@ -191,14 +192,62 @@ def input_types(self) -> Dict[str, torch.dtype]:
...
+            )
+
+        lags = lagged_sequence_values(
+            self.lags_seq, prior_input, input, dim=-1
+        )
+
+        time_feat = torch.cat(
+            (
+                past_time_feat[..., -self.context_length + 1 :, :],
+                future_time_feat,
```
kashif, line 239: is there a need to append this when there is no `future_target`

src/gluonts/torch/model/deepar/module.py
```diff
@@ -191,14 +192,62 @@ def input_types(self) -> Dict[str, torch.dtype]:
...
+        past_observed_values: torch.Tensor,
+        future_time_feat: torch.Tensor,
+        future_target: Optional[torch.Tensor] = None,
+    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor,]:
+        context = past_target[..., -self.context_length :]
+        observed_context = past_observed_values[..., -self.context_length :]
+        _, scale = self.scaler(context, observed_context)
+
+        prior_input = past_target[..., : -self.context_length] / scale
+        input = context / scale
```
kashif:
```suggestion
        input, scale = self.scaler(context, observed_context)

        prior_input = past_target[..., : -self.context_length] / scale
```

src/gluonts/torch/model/deepar/module.py
lostella: Yes, `future_time_feat` is not an `Optional` argument anymore (it never should have been, probably). The assumption is that at least one "future output" will be produced, for which the corresponding dynamic features are needed. In fact, it is assumed that there's one less `future_target` than `future_time_feat`, if any, see lines 211-215 above. Let me improve this a bit

kashif:
BTW @lostella do you think the same issue also affects the transformer model?  https://github.com/awslabs/gluonts/blob/dev/src/gluonts/mx/model/transformer/_network.py#L172-L178

lostella:
> BTW @lostella do you think the same issue also affects the transformer model? https://github.com/awslabs/gluonts/blob/dev/src/gluonts/mx/model/transformer/_network.py#L172-L178

It's hard to say. The easiest way to verify this, I think, is to set up a small unit test what the input to whatever layer will be: in this case the tensor is split into encoder + decoder input, so one could have a method prepare those two directly and then test that lagged target values, dynamic features, and what not, are aligned as expected.

test/torch/model/test_deepar_modules.py
```diff
@@ -145,3 +145,81 @@ def test_deepar_modules(
...
+        scaling=False,
+        num_parallel_samples=num_samples,
+    )
+
+    batch = {
+        "feat_static_cat": torch.tensor([[0]], dtype=torch.int64),
+        "feat_static_real": torch.tensor([[0.0]], dtype=torch.float32),
+        "past_time_feat": torch.arange(history_length, dtype=torch.float32)
+        .unsqueeze(0)
+        .unsqueeze(-1),
```
jaheba: I think I prefer `reshape(1, history_length, 1)` instead of `unsqueeze`. Makes it more clear what the output shapes is in my view.

kashif: or even `.view()`

test/torch/model/test_estimators.py
```diff
@@ -67,14 +67,14 @@
...
-            epochs=2,
-        ),
+        # lambda dataset: DeepNPTSEstimator(
+        #     freq=dataset.metadata.freq,
+        #     prediction_length=dataset.metadata.prediction_length,
+        #     context_length=2 * dataset.metadata.prediction_length,
+        #     batch_size=4,
+        #     num_batches_per_epoch=3,
+        #     epochs=2,
+        # ),
```
lostella: For some reason this test started failing again with

```python
E       RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn

../.pyenv/versions/3.8.13/envs/gluonts/lib/python3.8/site-packages/torch/autograd/__init__.py:197: RuntimeError
```

I think we need to understand this better and fix it separately.

lostella: This kind of error kept #2496 on hold for a while, then it disappeared on its own. But something must be off.

src/gluonts/torch/model/deepar/module.py
```diff
@@ -132,6 +132,7 @@ def __init__(
             else [min(50, (cat + 1) // 2) for cat in cardinality]
         )
         self.lags_seq = lags_seq or get_lags_for_frequency(freq_str=freq)
+        self.lags_seq = [l - 1 for l in self.lags_seq]
```
mbohlkeschneider, line 135: Maybe I'm completely not getting the issue here, but if we do `l-1` for the lags, don't we feed in the current time stamp during training when `l=1` thereby introducing a data leak? Maybe I'm wrong, though.

lostella: It depends on what time index the lag is applied: for how the RNN input preparation is written, the lag index gets subtracted from `t`, when we have data up to time `t` and want a prediction for time `t+1`. The `test_rnn_input` test function aims at verifying precisely this. Let me try to see whether the test can be made more apparent.

kashif:
@lostella one issue i found with the `lagged_sequence_values` helper is that with multivariate inputs this helper doesn't return the stacked tensors... e.g.

```py
# prior_input.shape torch.Size([1, 167, 137])
# input.shape torch.Size([1, 24, 137])

lags = lagged_sequence_values([0, 167], prior_input, input, dim=1)
# lags.shape torch.Size([1, 24, 137, 2])
```

where i would have expected `torch.Size([1, 24, 137*2])`

jaheba (approved):

PR #2627, opened 2023-02-09
make Pytorch scaler's forward API consistent

kashif:
*Description of changes:*

Made the API of the scalers consistent... this will allow us to use the different scalers interchangeably (which is not possible for the StdScaler)


lostella:
Thanks @kashif! I think we could even have a follow up that makes these classes simple `__call__`-able dataclasses?

kashif:
indeed!

lostella (approved):

PR #2628, opened 2023-02-09
Remove torch specific Dataloader, remove `num_workers` from torch models.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/torch/model/simple_feedforward/estimator.py
+++ b/src/gluonts/torch/model/simple_feedforward/estimator.py
@@ -19,7 +19,8 @@ import pytorch_lightning as pl
 from gluonts.core.component import validated
 from gluonts.dataset.common import Dataset
 from gluonts.dataset.field_names import FieldName
-from gluonts.dataset.loader import data_loader
+from gluonts.dataset.loader import as_stacked_batches
+from gluonts.itertools import Cyclic
 from gluonts.model.forecast_generator import DistributionForecastGenerator
 from gluonts.torch.modules.loss import DistributionLoss, NegativeLogLikelihood
 from gluonts.transform import (
@@ -203,12 +204,12 @@ class SimpleFeedForwardEstimator(PyTorchLightningEstimator):
         shuffle_buffer_length: Optional[int] = None,
         **kwargs,
     ) -> Iterable:
+        data = Cyclic(data).stream()
         instances = self._create_instance_splitter(module, "training").apply(
             data, is_train=True
         )
-        return data_loader(
+        return as_stacked_batches(
             instances,
-            cycle=True,
             batch_size=self.batch_size,
             shuffle_buffer_length=shuffle_buffer_length,
             field_names=TRAINING_INPUT_NAMES,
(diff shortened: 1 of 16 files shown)


lostella:
When setting `num_batches_per_epoch`, and iterating multiple times, this goes over the same data again and again. I think `iter` is missing somewhere. Example:

```python
from gluonts.dataset.loader import data_loader
import numpy as np

dataset = [{"a": np.array([k])} for k in range(100)]

batches = data_loader(dataset, batch_size=5, num_batches_per_epoch=3, output_type=np.array)
```

Iterate twice with

```python
for batch in batches:
    print(batch)
```

Outputs both times

```python
{'a': array([[0],
       [1],
       [2],
       [3],
       [4]])}
{'a': array([[5],
       [6],
       [7],
       [8],
       [9]])}
{'a': array([[10],
       [11],
       [12],
       [13],
       [14]])}
```

src/gluonts/torch/model/simple_feedforward/estimator.py
```diff
@@ -207,27 +203,17 @@ def create_training_data_loader(
...
-            ),
-            self.num_batches_per_epoch,
+        return data_loader(
+            instances,
+            cycle=True,
+            batch_size=self.batch_size,
+            shuffle_buffer_length=shuffle_buffer_length,
+            field_names=TRAINING_INPUT_NAMES,
+            output_type=torch.tensor,
+            num_batches_per_epoch=self.num_batches_per_epoch,
```
lostella: Here, `Cyclic` is applied after instance splitting, while prior to the change it was applied before

jaheba:
I think you need to use cycle in this case

lostella:
> I think you need to use cycle in this case

That doesn't do it, `cycle` only makes sure the data is wrapped over when finished, but not that it takes off from where it left

src/gluonts/torch/model/simple_feedforward/estimator.py
jaheba: I think it shouldn't matter, since we are not doing any caching.

lostella: There might be weird edge cases though: if the dataset has just one series, and the splitter happens to sample zero instances, then `Cyclic` will exhaust, see https://github.com/awslabs/gluonts/blob/c6be56264fefe5d879a1fe47ba57798097327533/src/gluonts/itertools.py#L77-L81

jaheba:
> > I think you need to use cycle in this case
> 
> That doesn't do it, `cycle` only makes sure the data is wrapped over when finished, but not that it takes off from where it left

I see, you are right. The arg to `IterableSlice` needs to be wrapped in an `iter`.

src/gluonts/torch/model/simple_feedforward/estimator.py
jaheba: Should we remove `cycle` from this then and ensure that the passed dataset is already cycled?

lostella:
So maybe we need some good test here 😇

lostella (dismissed):

lostella (approved):

PR #2632, opened 2023-02-09
Simplify scalers, move to `gluonts.torch.scaler`

lostella:
*Description of changes:* There's no need for scalers to be `torch.nn.Module` since they don't really hold parameters. Also fixes the default `keepdim` of `MeanScaler` for consistency.

cc @kashif 

### DIFF
--- a/src/gluonts/torch/scaler.py
+++ b/src/gluonts/torch/scaler.py
@@ -16,7 +16,7 @@ from typing import Optional
 
 import torch
 
-from gluonts.core.serde import dataclass
+from gluonts.core.component import validated
 
 
 class Scaler:
@@ -26,7 +26,6 @@ class Scaler:
         raise NotImplementedError
 
 
-@dataclass
 class MeanScaler(Scaler):
     """
     Computes a scaling factor as the weighted average absolute value along
@@ -45,10 +44,18 @@ class MeanScaler(Scaler):
         minimum possible scale that is used for any item.
     """
 
-    dim: int = -1
-    keepdim: bool = True
-    default_scale: Optional[float] = None
-    minimum_scale: float = 1e-10
+    @validated()
+    def __init__(
+        self,
+        dim: int = -1,
+        keepdim: bool = False,
+        default_scale: Optional[float] = None,
+        minimum_scale: float = 1e-10,
+    ) -> None:
+        self.dim = dim
+        self.keepdim = keepdim
+        self.default_scale = default_scale
+        self.minimum_scale = minimum_scale
 
     def __call__(
         self, data: torch.Tensor, observed_indicator: torch.Tensor
@@ -88,7 +95,6 @@ class MeanScaler(Scaler):
         return scaled_data, loc, scale
 
 
-@dataclass
 class NOPScaler(Scaler):
     """
     Assigns a scaling factor equal to 1 along dimension ``dim``, and therefore
@@ -103,8 +109,14 @@ class NOPScaler(Scaler):
         scale tensor, or suppress it.
     """
 
-    dim: int = -1
-    keepdim: bool = False
+    @validated()
+    def __init__(
+        self,
+        dim: int = -1,
+        keepdim: bool = False,
+    ) -> None:
+        self.dim = dim
+        self.keepdim = keepdim
 
     def __call__(
         self, data: torch.Tensor, observed_indicator: torch.Tensor
@@ -117,7 +129,6 @@ class NOPScaler(Scaler):
         return data, loc, scale
 
 
-@dataclass
 class StdScaler(Scaler):
     """
     Computes a std scaling  value along dimension ``dim``, and scales the data accordingly.
@@ -134,9 +145,16 @@ class StdScaler(Scaler):
         along dimension ``dim``.
     """
 
-    dim: int = -1
-    keepdim: bool = False
-    minimum_scale: float = 1e-5
+    @validated()
+    def __init__(
+        self,
+        dim: int = -1,
+        keepdim: bool = False,
+        minimum_scale: float = 1e-5,
+    ) -> None:
+        self.dim = dim
+        self.keepdim = keepdim
+        self.minimum_scale = minimum_scale
 
     def __call__(
         self, data: torch.Tensor, weights: torch.Tensor
--- a/src/gluonts/torch/scaler.py
+++ b/src/gluonts/torch/scaler.py
@@ -60,12 +60,12 @@ class MeanScaler(Scaler):
 
         # If `default_scale` is provided, we use it, otherwise we use the scale
         # of the batch.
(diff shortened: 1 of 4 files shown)


kashif:
very cool! LGTM!

src/gluonts/torch/scaler.py
```diff
@@ -37,21 +36,12 @@ class MeanScaler(nn.Module):
         minimum possible scale that is used for any item.
     """
 
-    @validated()
```
jaheba: We loose the validation property?

src/gluonts/torch/scaler.py
```diff
@@ -37,21 +36,12 @@ class MeanScaler(nn.Module):
...
-        self.dim = dim
-        self.keepdim = keepdim
-        self.register_buffer("minimum_scale", torch.tensor(minimum_scale))
-        self.register_buffer("default_scale", torch.tensor(default_scale))
-
-    def forward(
+    dim: int = -1
+    keepdim: bool = True
+    default_scale: float = 0.0
+    minimum_scale: float = 1e-10
```
jaheba: I think these need to be `torch.tensor`

src/gluonts/torch/scaler.py
```diff
@@ -141,20 +130,11 @@ class StdScaler(nn.Module):
...
-        assert dim > 0, (
-            "Cannot compute scale along dim = 0 (batch dimension), please"
-            " provide dim > 0"
-        )
-        self.dim = dim
-        self.keepdim = keepdim
-        self.register_buffer("minimum_scale", torch.tensor(minimum_scale))
+    dim: int = -1
+    keepdim: bool = False
+    minimum_scale: float = 1e-5
```
shubhamkapoor: Also, `minimum_scale` should be a `torch.Tensor`.

src/gluonts/torch/scaler.py
lostella: Should I use our own `dataclass`?

src/gluonts/torch/scaler.py
lostella: Why? I think we can add a `torch.Tensor` to a number

src/gluonts/torch/scaler.py
lostella: `torch.clamp` also deals with numbers; for the `default_scale`, isn't this some kind of "immutable" (mind the quotes) property of the object, so that replacing `torch.where` with `if` is not really harmful?

lostella:
~~It's funny that all tests pass on macOS~~

src/gluonts/torch/scaler.py
jaheba: Yes, let's give it a try

src/gluonts/torch/scaler.py
```diff
@@ -60,21 +58,14 @@ def forward(
...
-        batch_sum = ts_sum.sum(dim=0)
-        batch_observations = torch.clamp(num_observed.sum(0), min=1)
-        batch_scale = torch.squeeze(batch_sum / batch_observations)
-
-        default_scale = torch.where(
-            self.default_scale > 0.0,
-            self.default_scale,
-            batch_scale,
-        )
+        if self.default_scale is not None:
```
jaheba: What is the effect if this change on tracing?

lostella: I think that, as long as `self.default_scale` stays constant (which it should) then tracing should be fine with it and produce the right code for the model. But I need to verify

lostella: Also according to the warning here https://pytorch.org/docs/stable/generated/torch.jit.trace.html#torch-jit-trace it should be fine as long as the control flow is not affected by the value in input tensors (and I would add: if it’s not otherwise changed throughout the model execution, like the `forward` changing the value of `self.default_scale` here for some reason)

src/gluonts/torch/scaler.py
lostella: I'll just keep them validated for now

jaheba (approved):

abdulfatir:
Sorry for the late comment but any takers for Scalers being of `torch.distributions.transforms.Transform` type, `AffineTransform` in particular. This provides us with the inverse function that you can just apply back to your scaled output and also provides `log_abs_det_jacobian`, if needed for loss computation.

kashif:
@abdulfatir on the output side the distribution is an AffineTransform which takes the loc and scale from the scalers... but you mean more general scalers? And I believe currently it's an appropriate place to have the scalers (in the model) since then the model can use the loc and scale as input as well (instead of just on the emission side).

abdulfatir:
Currently, I don't have an example of a general scaler in mind, but yes, using `torch.distributions.transforms.Transform` can provide that flexibility. Regarding access of scaler params (loc, scale) in the models, they can also be accessed from the scaler/transform object's properties. 

The primary benefit IMO is clarity. You have a scaler (Transform) that normalizes the data and then use its inverse (and `log_abs_det_jacobian`) at the output side. 
@kashif

abdulfatir:
Also, inside models I think we should be able to provide a `Scaler()` object instead of a boolean `scaling` as currently implemented.

lostella:
@abdulfatir I think it makes sense to consider this. The (log) scale should be accessible via `log_abs_det_jacobian` if one needs that as input to the model, but not the location. I’ll give it a try and share how that goes

abdulfatir:
@lostella If the transform is of `AffineTransform` type, then we can access the `.loc` and `.scale` attributes. How a model handles/uses transforms could be their internal matter. For instance, for DeepAR like models we can restrict the scale to be of `AffineTransform` type and freely access these properties.

abdulfatir:
For completeness, I would like to add that the inverse transform (which we would need at the output side) is available via the `.inv` attribute.

```python
import torch
from torch.distributions.transforms import AffineTransform

tr = AffineTransform(10., 1.)
inv_tr = tr.inv

x = torch.rand(2, 3, 4)
y = tr(x)

torch.allclose(x, inv_tr(y), atol=1e-5) # True
```

lostella:
> @lostella If the transform is of `AffineTransform` type, then we can access the `.loc` and `.scale` attributes. How a model handles/uses transforms could be their internal matter. For instance, for DeepAR like models we can restrict the scale to be of `AffineTransform` type and freely access these properties.

The catch is that these scaling operations are not really affine transformations of the data: for some array `x`, `x` and `2*x` get scaled down to the same vector. They are once you fix the `loc` and `scale`, so you can access them as properties of the transformation, but `scale` (and possibly `loc`) depend on the input data here.

lostella:
> Also, inside models I think we should be able to provide a Scaler() object instead of a boolean scaling as currently implemented.

@abdulfatir agreed! We can do that in a separate PR

PR #2633, opened 2023-02-09
Fix bugs in `MeanScaler`

lostella:
*Description of changes:* An assignment was missing, and `keepdim=False` was not working. Ported tests from the mxnet version, with minor differences due to a difference in how min_scale vs default_scale is handled cc @shubhamkapoor 


jaheba:
I guess we can adapt the tests from mx?

shubhamkapoor (approved):
Thanks! 🚀

PR #2638, opened 2023-02-10
Add `dataset.loader.as_stacked_batches`.

jaheba:
*Description of changes:*

Add a new function to do collecting and stacking data batches. The function can be used to simplify our data loaders in the future.

### DIFF
--- a/src/gluonts/dataset/loader.py
+++ b/src/gluonts/dataset/loader.py
@@ -14,11 +14,24 @@
 import logging
 from typing import Callable, Iterable, Optional
 
+import numpy as np
 from pydantic import BaseModel
 
 from gluonts.dataset import DataBatch, Dataset
-from gluonts.itertools import Cyclic, IterableSlice, PseudoShuffled, batcher
-from gluonts.transform import AdhocTransform, Identity, Transformation
+from gluonts.itertools import (
+    Cyclic,
+    IterableSlice,
+    PseudoShuffled,
+    batcher,
+    rows_to_columns,
+)
+from gluonts.transform import (
+    AdhocTransform,
+    Identity,
+    SelectFields,
+    Transformation,
+    Valmap,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


src/gluonts/dataset/loader.py
```diff
@@ -37,6 +50,63 @@ def __call__(self, data, is_train):
...
+    def __call__(self, data, is_train):
+        for batch in data:
+            yield rows_to_columns(batch, np.array)
+
+
+def as_stacked_batches(
+    dataset: Dataset,
+    *,
+    batch_size: int,
+    output_type: Callable,
```
jaheba: Should we make this `Optional`?

src/gluonts/dataset/loader.py
```diff
@@ -37,6 +50,63 @@ def __call__(self, data, is_train):
...
+    if shuffle_buffer_length:
+        dataset = PseudoShuffled(dataset, shuffle_buffer_length)
+
+    transform: Transformation = Identity()
+
+    if field_names is not None:
+        transform += SelectFields(field_names)
+
+    transform += Batch(batch_size=batch_size)
+    transform += Stack()
```
jaheba, line 95: This could also be an optional parameter, and only stack if required.

On the other hand, I don't know where we would want that.

src/gluonts/dataset/loader.py
```diff
@@ -37,6 +50,63 @@ def __call__(self, data, is_train):
         yield from batcher(data, self.batch_size)
 
 
+class Stack(Transformation, BaseModel):
+    def __call__(self, data, is_train):
+        for batch in data:
+            yield rows_to_columns(batch, np.array)
```
jaheba, line 56: There is an argument that we don't need to wrap this in an `np.array`.

However, `torch.tensor` and `mx.ndarray` like to work with `np.array` instances rather than lists.

src/gluonts/dataset/loader.py
lostella: yeah no need for that

src/gluonts/dataset/loader.py
lostella: Why not, or just default to identity?

src/gluonts/dataset/loader.py
lostella: You could wrap as numpy array in a separate step using `ValMap`? But then again, it's also fine like this

jaheba: I guess we can leave it as it is.

src/gluonts/dataset/loader.py
```diff
@@ -37,6 +50,61 @@ def __call__(self, data, is_train):
...
+
+    if output_type is not None:
+        transform += Valmap(output_type)
+
+    # Note: is_train needs to be provided but does not have an effect
+    transformed_dataset = transform.apply(dataset, is_train=True)
+    return IterableSlice(
+        transformed_dataset,
+        num_batches_per_epoch,
+    )
```
lostella: If this creates complications (we may or may not need to `iter` the `transformed_dataset`) we could even remove this wrapping and rely on the trainer to limit how much data is taken at once, see the `limit_train_batches` option at https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.trainer.trainer.Trainer.html#pytorch_lightning.trainer.trainer.Trainer

This way we can directly return `transformed_dataset` and downstream one `iter`s it or not depending on the context

jaheba: I don't think this causes issues. One is free to not set `num_batches_per_epoch`.

lostella: it prevents you from being able to `iter(transformed_dataset)` downstream

lostella: nvm

lostella (approved):

PR #2640, opened 2023-02-10
Ev geo mean

jaheba:
*Issue #, if available:*

*Description of changes:*


melopeo:
The geometric mean is only defined for non-negative values, i.e. `x >= 0`. Do we have a check on this condition?

lostella:
It may be ok to just let this fail at runtime in case `(values<=0).any()`

lostella (approved):

jaheba:
This shouldn't work yet.

PR #2643, opened 2023-02-12
Fix PyTorch training loop

gorold:
*Description of changes:*
Minor bug in the pytorch training loop.

### DIFF
--- a/src/gluonts/torch/model/estimator.py
+++ b/src/gluonts/torch/model/estimator.py
@@ -174,8 +174,8 @@ class PyTorchLightningEstimator(Estimator):
 
         validation_data_loader = None
 
-        with env._let(max_idle_transforms=max(len(training_data), 100)):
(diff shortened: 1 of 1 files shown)


jaheba (approved):
`ruff` removed the check that failed in the newer version, so this should be good to go.

src/gluonts/torch/model/estimator.py
```diff
@@ -174,8 +174,8 @@ def train_model(
 
         validation_data_loader = None
 
-        with env._let(max_idle_transforms=max(len(training_data), 100)):
-            if validation_data is not None:
+        if validation_data is not None:
+            with env._let(max_idle_transforms=max(len(validation_data), 100)):
```
jaheba, line 178: Something I'm wondering, shouldn't there never be idle transforms, since the instance splitter always returns one window for validation per ts?

PR #2644, opened 2023-02-13
RFC: zebras.TimeFrame.

jaheba:
**[Rendered](https://github.com/jaheba/gluon-ts/blob/zebras-rfc/docs/rfc/zebras_timeframe.md)**

Started implementation in #2609

### DIFF
--- a/docs/rfc/zebras_timeframe.md
+++ b/docs/rfc/zebras_timeframe.md
@@ -91,7 +91,8 @@ handles tracking of time internally.
 ## Proposal
 
 We introduce two new data structures called ``TimeFrame`` and ``SplitFrame`` to
-replace our current dictionary based approach.
+replace our current dictionary based approach. In addition there are batched
+variants called ``BatchTimeFrame`` and ``BatchSplitFrame``.
 
 A ``TimeFrame`` consists of a time index and a set of time series, which all
 have the same length. A ``SplitFrame`` consists of two ``TimeFrame`` instances,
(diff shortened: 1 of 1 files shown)


docs/rfc/zebras_timeframe.md
```diff
@@ -0,0 +1,160 @@
...
+when doing predictions, while all other data should not.
+
+Additionally, there is a padding indicator to denote which values have been
+padded: We want to allow to force frames to a certain length, and pad values if
+needed.
+
+Both classes have easy to use constructors in ``zebras.time_frame`` and
+``zebras.split_frame``.
+
+## Discussion
```
huibinshen, line 155: How about missing values?

docs/rfc/zebras_timeframe.md
```diff
@@ -0,0 +1,160 @@
...
+* A ``DataFrame`` can only store one dimensional columns. However, we also need
+  support for columns with multiple dimensions, of which one is the time
+  dimension.
+* In a ``TimeFrame`` all values are evenly spaced with respect to time. This
+  then allows us to just extend a ``TimeFrame`` into the future and adjust the
+  index automatically.
+* We can implement custom behaviour such as ``.split`` easily on ``TimeFrame``.
+* ``SplitFrame`` has a notion of past and future.
+
+## Obstacles
```
huibinshen, line 171: Agree, this seems a big change and gain should outweigh the effort.

docs/rfc/zebras_timeframe.md
jaheba: Missing values can be handled within `TimeFrame` by adding another column to it. I don't think we need to add a special case for this.

PR #2646, opened 2023-02-13
Pass prediction inputs as dict.

### LINKED ISSUE #2624: Support optional inputs for models
## Description
(A clear and concise description of what the feature is.)
GluonTS currently does not support optional inputs. 
The current paradigm is that if a model has optional inputs, the estimator class needs to define a transformation to include dummy values. However, it would be preferable to completely skip the processing of the unavailable inputs, rather than processing some additional dummy values:

```python
def forward(
    self, 
    past_target,  
    past_time_feat, 
    future_time_feat,
    feat_static_cat = None,  # optional
    feat_static_real = None  # optional
):
    if feat_static_cat is not None:
        ...
    if feat_static_real is not None:
        ...
```

Currently GluonTS does not really support this paradigm even if the model handles this logic. For example, because of how 
ForecastGenerators are defined, the below definition for ```create_predictor``` causes issues. 
https://github.com/awslabs/gluonts/blob/66a129ec258ad8f9ea087ec02c145cc37f5713b1/src/gluonts/model/forecast_generator.py#L155

```python
def create_predictor(
    self,
    transformation,
    module
):
    ...
    return PyTorchPredictor(
        input_names=["past_target", "past_time_feat", "future_time_feat", "feat_static_real"]
        ...
    )
```

Namely, because a list is fed into the model, ```feat_static_real``` is treated as ```feat_static_cat```. We can easily fix this by ```inputs = {k: batch[k] for k in input_names}``` and editing ```predict_to_numpy``` accordingly. However, I am not too sure whether there are any other parts of the library which would fail if this change is made.

## References
- list reference and related literature
- list known implementations

jaheba:
Fixes: #2624

We have a central function that is used to create predictions called ``predict_to_numpy``.

Previously, the prediction inputs were passed as positional arguments. Now, we pass ``kwargs`` instead. Since `mxnet` expects positional arguments, we pass values via `args`, while `torch` can handle ``kwargs``.

There are some benefits to this: In `torch`, we can make use of optional arguments. ONNX models expects the inputs as `dict`, so we can now just pass these directly.

---

### DIFF
--- a/src/gluonts/model/forecast_generator.py
+++ b/src/gluonts/model/forecast_generator.py
@@ -199,7 +199,7 @@ class DistributionForecastGenerator(ForecastGenerator):
     ) -> Iterator[Forecast]:
         for batch in inference_data_loader:
             inputs = select(input_names, batch, ignore_missing=True)
-            outputs = predict_to_numpy(prediction_net, inputs)
+            outputs = prediction_net(*inputs.values())
 
             if output_transform:
                 log_once(OUTPUT_TRANSFORM_NOT_SUPPORTED_MSG)
--- a/src/gluonts/model/forecast_generator.py
+++ b/src/gluonts/model/forecast_generator.py
@@ -199,7 +199,7 @@ class DistributionForecastGenerator(ForecastGenerator):
     ) -> Iterator[Forecast]:
         for batch in inference_data_loader:
             inputs = select(input_names, batch)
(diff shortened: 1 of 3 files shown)


jaheba:
There might be an issue with the simple feedforward implementation in that it defines inputs that it doesn't take.

lostella (approved):
Looks good to me 👍

PR #2648, opened 2023-02-13
Fix call to `extractall`

lostella:
*Description of changes:* There was yet one call that applied to tar archives as well.

### DIFF
--- a/src/gluonts/shell/sagemaker/dyn.py
+++ b/src/gluonts/shell/sagemaker/dyn.py
@@ -33,10 +33,6 @@ from typing import Optional
 from gluonts.util import safe_extractall
 
 
-def extractall(archive: zipfile.ZipFile, path: Path):
-    archive.extractall(path)
-
-
 class Installer:
     def __init__(self, packages):
         self.packages = packages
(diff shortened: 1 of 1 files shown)


src/gluonts/shell/sagemaker/dyn.py
```diff
@@ -78,10 +84,10 @@ def install_requirement(self, path: Path):
     def install(self, path):
         if path.is_file():
             if tarfile.is_tarfile(path):
-                self.handle_archive(tarfile.open, path)
+                self.handle_archive(tarfile.open, safe_extractall, path)
 
             elif zipfile.is_zipfile(path):
-                self.handle_archive(zipfile.ZipFile, path)
+                self.handle_archive(zipfile.ZipFile, extractall, path)
```
jaheba:
```suggestion
                self.handle_archive(zipfile.ZipFile, zipfile.ZipFile.extractall, path)
```

This should also work.

jaheba:
Let's check when we added `dyn`, I don't think we need to backport this that far back.

lostella:
was there in 0.9.x https://github.com/awslabs/gluonts/blob/v0.9.x/src/gluonts/shell/sagemaker/dyn.py

jaheba:
I'm getting old!

jaheba (approved):

PR #2651, opened 2023-02-14
Add zebras freq/period.

jaheba:
Add `zebras.Freq`, `zebras.Period` and `zebras.Periods` 

### DIFF
--- a/src/gluonts/zebras/_freq.py
+++ b/src/gluonts/zebras/_freq.py
@@ -15,7 +15,7 @@ from __future__ import annotations
 
 import re
 from dataclasses import dataclass
-from typing import Tuple, cast
+from typing import Tuple
 
 import numpy as np
 
--- a/src/gluonts/zebras/_freq.py
+++ b/src/gluonts/zebras/_freq.py
@@ -11,6 +11,8 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
+from __future__ import annotations
+
 import re
 from dataclasses import dataclass, asdict
 from typing import Tuple
--- a/src/gluonts/zebras/_freq.py
+++ b/src/gluonts/zebras/_freq.py
@@ -23,7 +23,7 @@ from gluonts.core import serde
 NpFreq = Tuple[str, int]
 
 
-def _canonical_freqstr(n: int, name: str):
+def _canonical_freqstr(n: int, name: str) -> str:
     """Canonical name of frequency.
 
     >>> _canonical_freqstr("X")
@@ -98,7 +98,7 @@ class Freq:
     n: int
 
     @property
-    def np_freq(self):
+    def np_freq(self) -> NpFreq:
         return _freq_pandas_to_numpy[self.name]
 
     @classmethod
@@ -107,7 +107,7 @@ class Freq:
         yield freq
 
     @classmethod
-    def from_pandas(cls, freq):
+    def from_pandas(cls, freq) -> Freq:
         if not isinstance(freq, str):
             if hasattr(freq, "freqstr"):
                 freq = freq.freqstr
@@ -125,18 +125,18 @@ class Freq:
 
         return cls(name, n)
 
-    def to_pandas(self) -> str:
+    def to_pandas(self) -> "pandas.Period":
         from pandas.tseries.frequencies import to_offset
 
         return to_offset(str(self))
 
-    def shift(self, start, count):
+    def shift(self, start: np.datetime64, count: int) -> np.datetime64:
         if self.name == "B":
             return np.busday_offset(data, self.n * count)
 
         return data + self.n * count
 
-    def range(self, start, count):
+    def range(self, start: np.datetime64, count: int) -> np.datetime64:
         if self.name == "B":
             # We first collect all days, even non business days to then filter
             # for business days, of which we then take, each n-th.
@@ -151,12 +151,12 @@ class Freq:
 
         return np.arange(start, count * step, step)
 
-    def __str__(self):
+    def __str__(self) -> str:
         return _canonical_freqstr(self.n, self.name)
 
 
 @serde.encode.register
-def _encode_freq(v: Freq):
+def _encode_freq(v: Freq) -> dict:
     return {
         "__kind__": "instance",
         "class": "gluonts.zebras.Freq.from_pandas",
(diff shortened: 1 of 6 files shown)


test/zebras/test_period.py
```diff
@@ -0,0 +1,53 @@
...
+import pandas as pd
+from pandas.tseries.frequencies import to_offset
+
+from gluonts import zebras as zb
+
+FREQS = ["S", "min", "T", "H", "D", "B", "W", "M", "QS", "Q", "AS", "A", "Y"]
+
+
+@pytest.mark.parametrize("freq", FREQS)
+def test_period_equal_pandas(freq):
```
lostella: Minor: `test_freq_…`

src/gluonts/zebras/_freq.py
```diff
@@ -0,0 +1,120 @@
...
+        "B": ("D", 1),
+        "W": ("D", 1),
+    },
+)
+
+
+@dataclass
+class Freq:
+    np_freq: NpFreq
+    pd_freq: str
```
lostella: So this class can be constructed with potentially inconsistent arguments? I see that `from_pandas` is the preferred construction method, but maybe it should be the default one?

src/gluonts/zebras/_freq.py
```diff
@@ -0,0 +1,120 @@
...
+    _multiple: int
+
+    @property
+    def multiple(self):
+        if self.pd_freq == "W":
+            return self._multiple * 7
```
lostella: What’s going on here? Is this method ever used?

src/gluonts/zebras/_freq.py
```diff
@@ -0,0 +1,120 @@
...
+
+        return self._multiple
+
+    @classmethod
+    def __get_validators__(cls):
+        yield freq
```
lostella, line 119: What’s `freq`?

src/gluonts/zebras/_period.py
```diff
@@ -0,0 +1,285 @@
...
+            return self.data - other.data
+
+        raise ValueError(other)
+
+
+@functools.total_ordering
+class Period(_BasePeriod):
+    def periods(self, count: int):
+        return Periods(
+            np.arange(
```
lostella: `Periods` is a `_BasePeriod` whose `data` is supposed to be `np.datetime64` according to type annotations?

src/gluonts/zebras/_period.py
```diff
@@ -0,0 +1,285 @@
...
+        return self.data[np.in1d(self, other)]
+
+    def index_of(self, period: Period):
+        idx = (period - self.start).astype(int)
+        assert 0 <= idx < len(self)
+
+        return idx
+
+    def cat(self, other):
+        return Periods(np.concatenate([self.data, other.data]), self.freq)
```
lostella: What if the data of the two is not “aligned”? One could try to concatenate two overlapping Periods, or that have some gap.

Maybe we can avoid this method until needed.

lostella (commented):
I guess docstrings would be helpful. Could you add `.typesafe`? (We should do it for all new sub packages I think)

src/gluonts/zebras/_freq.py
jaheba: The idea generally is to use `freq` to construct this class. Like you never invoke `np.ndarray` but use `np.array`.

I've just changed it though.

src/gluonts/zebras/_freq.py
jaheba: The function defined below. This means that if `Freq` is used in a pydantic model, it can create instances of `Freq` from strings.

src/gluonts/zebras/_period.py
jaheba: I think we can add a check.

jaheba:
Microbenchmark:

```py
# zebras
%%timeit

zb.periods("2020", "D", 365)

39.8 µs ± 2.14 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

```py
# pandas
%%timeit

pd.period_range("2020", freq="D", periods=365)

113 µs ± 7.29 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

```

---

```py
# zebras
%%timeit

periods.dayofyear

16.4 µs ± 381 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

```py
# pandas
%%timeit

pr.dayofyear

29.4 µs ± 848 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

src/gluonts/zebras/_freq.py
```diff
@@ -0,0 +1,174 @@
...
+
+from gluonts import maybe
+
+NpFreq = Tuple[str, int]
+
+
+def _canonical_freqstr(n: int, name: str) -> str:
+    """Canonical name of frequency.
+
+    >>> _canonical_freqstr("X")
```
lostella:
```suggestion
    >>> _canonical_freqstr(1, "X")
```

same below. Do these examples run as part of the docs build?

src/gluonts/zebras/_freq.py
```diff
@@ -0,0 +1,174 @@
...
+    def to_pandas(self):
+        from pandas.tseries.frequencies import to_offset
+
+        return to_offset(str(self))
+
+    def shift(self, start: np.datetime64, count: int) -> np.datetime64:
+        if self.name == "B":
+            return np.busday_offset(start, self.n * count)
+
+        return start + self.n * count
```
lostella: Below in the `range` method we multiply by 7 for weekly frequency, here we don't need that?

src/gluonts/zebras/_freq.py
```diff
@@ -0,0 +1,179 @@
...
+        return to_offset(str(self))
+
+    def shift(self, start: np.datetime64, count: int) -> np.datetime64:
+        if self.name == "B":
+            return np.busday_offset(start, self.n * count)
+
+        step = self.n
+
+        if self.name == "W":
+            step *= 7
```
lostella, line 152: I'm still not clear what's going on here?

jaheba: From the docstring:

> Weekly frequency needs to be handled specially, since numpy counts the
> number of weeks since Thu Jan 1 1970 and uses Thursday and aligns the
> timestamp to Thursday. We therefore use daily frequency internally and
> align to Monday.

jaheba: The `7` is needed to convert from `D` to `W`. We can't use `step`, because users can set `2W`, which we convert to `2 * 7 D`.

lostella (approved):

PR #2658, opened 2023-02-16
Add test for torch models tracing

lostella:
*Description of changes:* Add test to check that tracing a model results in the same computation, i.e., that the model's behaviour does not depend on input values.


test/torch/model/test_jit.py
```diff
@@ -0,0 +1,62 @@
...
+
+from gluonts.torch.model.deepar import DeepARModel
+from gluonts.torch.model.tft import TemporalFusionTransformerModel
+
+
+def get_model_and_input(model):
+    return model, tuple(
+        [
+            torch.zeros(shape, dtype=model.input_types()[name])
+            for (name, shape) in model.input_shapes().items()
```
jaheba: I think it would be nice to not have two different methods to get information on values.

lostella: OK, I guess that can be addressed separately?

jaheba: Yes!

test/torch/model/test_jit.py
```diff
@@ -0,0 +1,62 @@
...
+
+import pytest
+
+import torch
+
+from gluonts.torch.model.deepar import DeepARModel
+from gluonts.torch.model.tft import TemporalFusionTransformerModel
+
+
+def get_model_and_input(model):
```
jaheba: What's the purpose of this, why not just use `.example_input_array`?

Also, why should this be part of the parameter and not just be invoked in the test function?

jaheba (commented):
There also is `to_torchscript` which lightning provides.

test/torch/model/test_jit.py
lostella: I'm not sure that using these completely dummy arrays is the best approach. Maybe for tracing the model yes, but then we should probably test the model vs scripted model using some other input

lostella: So it's mostly a placeholder for now

lostella: > What's the purpose of this, why not just use .example_input_array?

That's not a property of `torch.nn.Module`

jaheba: Right. Still looks like something which could be consolidated.

jaheba (approved):

PR #2659, opened 2023-02-16
Add `hierarchicalforecast` to github workflows.

melopeo:
*Description of changes:*
add hierarchicalforecast to github workflow


lostella:
I vote for merging this with the statsforecast one https://github.com/awslabs/gluonts/blob/dev/.github/workflows/tests-statsforecast.yaml

melopeo:
Propose a name for it and we can do it. Nixtla, maybe? At the end these are wrappers from their repositories.

lostella:
> Propose a name for it and we can do it. Nixtla, maybe? At the end these are wrappers from their repositories.

Yes, sounds good to me

.github/workflows/tests-nixtla.yaml
```diff
@@ -1,4 +1,4 @@
-name: StatsForecast tests
+name: nixtla tests
```
lostella:
```suggestion
name: Nixtla tests
```

just for consistency with other workflows 🙂

test/ext/hierarchicalforecast/require-packages.txt
```diff
@@ -1 +1,2 @@
+statsforecast
```
lostella, line 1: Is this needed @jaheba? How are these used?

lostella:
Just the workflow name otherwise good to go! 🚀

lostella (approved):

test/ext/hierarchicalforecast/require-packages.txt
jaheba: I would assume this doesn't need it.

The idea behind `require-packages` is that tests in that folder are only run if the required packages are installed. Here, I would assume we only need `hierarchicalforecast` as a check.

PR #2660, opened 2023-02-16
Delay instantiation of ```ScipyStudentT``` object

gorold:
*Description of changes:*
Delay the instantiation of ```ScipyStudentT``` object until it is required. The ```ScipyStudentT``` is instantiated whenever ```StudentT``` is initialized, requiring additional operations of 3 tensors (```df, loc, scale```), moving them across device. However, the ```ScipyStudentT``` object is only needed for ```cdf``` and ```icdf``` which are not called when using sample forecasts. Furthermore, when training on bfloat16 mode, the ```.numpy()``` call leads to an error.


src/gluonts/torch/distributions/studentT.py
```diff
@@ -53,6 +50,16 @@ def icdf(self, value: torch.Tensor) -> torch.Tensor:
...
 
+    @property
+    def scipy_student_t(self):
+        if self._scipy_student_t is None:
+            self._scipy_student_t = ScipyStudentT(
+                df=self.df.detach().cpu().numpy(),
+                loc=self.loc.detach().cpu().numpy(),
+                scale=self.scale.detach().cpu().numpy(),
+            )
+        return self._scipy_student_t
```
lostella: You could use the `lazy_property` decorator here  https://github.com/awslabs/gluonts/blob/dev/src/gluonts/util.py#L53

gorold: Done!

lostella (approved):
Thanks @gorold!

PR #2663, opened 2023-02-16
Faster index building in PandasDataset

huibinshen:
*Faster index building for PandasDataset:*
Construct `pd.PeriodIndex` from list of time stamp strings can be quite slow:
```Python
periods = pd.period_range('2006-01-01 00:00:00', periods=7000, freq='H')
timestamps_list = [str(t.start_time) for t in period]

>>> timeit.timeit(lambda: pd.PeriodIndex(timestamps_list, freq='H'), number=5)
3.156371082000078
>>> timeit.timeit(lambda: pd.DatetimeIndex(timestamps_list).to_period(freq='H'), number=5)
0.02911860299991531
```

### DIFF
--- a/src/gluonts/dataset/pandas.py
+++ b/src/gluonts/dataset/pandas.py
@@ -151,7 +151,9 @@ class PandasDataset:
             df = df.to_frame(name=self.target)
 
         if self.timestamp:
-            df.index = pd.PeriodIndex(df[self.timestamp], freq=self.freq)
+            df.index = pd.DatetimeIndex(
+                df[self.timestamp], freq=self.freq
+            ).to_period()
 
         if not isinstance(df.index, pd.PeriodIndex):
             df = df.to_period(freq=self.freq)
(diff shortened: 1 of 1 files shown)


src/gluonts/dataset/pandas.py
```diff
@@ -151,7 +151,9 @@ def _pair_to_dataentry(self, item_id, df) -> DataEntry:
             df = df.to_frame(name=self.target)
 
         if self.timestamp:
-            df.index = pd.PeriodIndex(df[self.timestamp], freq=self.freq)
+            df.index = pd.DatetimeIndex(
+                df[self.timestamp], freq=self.freq
+            ).to_period()
```
lostella: I would try

```suggestion
                df[self.timestamp]
            ).to_period(freq=self.freq)
```

huibinshen: This works, thanks!

lostella: Does it give the same speedup?

huibinshen: Yes!

jaheba:
This shouldn't be considered a bug fix, I think.

lostella (approved):

PR #2670, opened 2023-02-18
Fix ```DateSplitter``` when split date is before start

gorold:
*Description of changes:*
DateSplitter aims to split time series such that "Training entries obtained from this class will be limited to observations up to (including) the given ``date``." However, its current behaviour is similar to OffsetSplitter, when the split date is before the start date, i.e. ```start > end```, ```periods_between``` returns a negative value, and slicing the array via negative indexing. This is not the expected behaviour, and instead should return an empty array.

### DIFF
--- a/src/gluonts/dataset/split.py
+++ b/src/gluonts/dataset/split.py
@@ -101,6 +101,8 @@ def periods_between(
         >>> periods_between(start, end)
         9
     """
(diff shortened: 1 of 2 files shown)


lostella (approved):
Looks good! Thanks @gorold

PR #2672, opened 2023-02-20
Zebras timeframe

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/zebras/_timeframe.py
+++ b/src/gluonts/zebras/_timeframe.py
@@ -273,23 +273,26 @@ class TimeFrame:
 
     def stack(
         self,
-        columns: List[str],
+        select: List[str],
         into: str,
         drop: bool = True,
     ) -> TimeFrame:
         # Ensure all tdims are the same.
         # TODO: Can we make that work for different tdims? There might be a
         # problem with what the resulting dimensions are.
-        assert len(set([self.tdims[column] for column in columns])) == 1
+        tdims = set([self.tdims[column] for column in select])
+        assert len(tdims) == 1
+        tdim = first(tdims)
 
         if drop:
-            columns = dissoc(self.columns, *columns)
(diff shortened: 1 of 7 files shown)


src/gluonts/zebras/_timeframe.py
```diff
@@ -0,0 +1,305 @@
...
+
+    return dataclasses.replace(clone, **kwargs)
+
+
+@dataclasses.dataclass
+class TimeFrame:
+    columns: dict
+    index: Optional[Periods]
+    static: dict
+    length: int
```
lostella, line 101: Could this be a `@property` instead?

src/gluonts/zebras/_timeframe.py
```diff
@@ -0,0 +1,305 @@
...
+
+
+def time_frame(
+    columns=None,
+    index=None,
+    start=None,
+    freq=None,
+    static=None,
+    tdims=None,
+    length=None,
```
lostella, line 536: I'm wondering when would one need to specify this?

src/gluonts/zebras/_timeframe.py
jaheba: No, because `columns` could be empty.

src/gluonts/zebras/_timeframe.py
jaheba: If you create an empty instance which you want to build up.

lostella: Wouldn't one in that case provide a non-empty set of columns, each one being empty?

jaheba: Not sure what you mean, but what happens if you drop all columns -- the length should stay the same.

lostella (approved):

PR #2673, opened 2023-02-20
Add interface to `gluonts.ev`

lostella:
*Description of changes:* This adds a higher-level interface to the metrics and evaluators API in `gluonts.ev`. See the example below for a feeling of how it works. In short, one calls

```python
evaluate_forecasts(forecasts, test_data=data, metrics=metrics)
```

to evaluate `metrics` to compare `forecasts` to `data`. Here `data` is a `TestData` object that contains both the "input" and "label" slices of each test series. This wraps the metric values into a `DataFrame` with appropriate index based on the aggregation dimension.

Example:

```python
from pathlib import Path
import pandas as pd
from gluonts.model import Predictor, evaluate_forecasts
from gluonts.ev import (
    SMAPE,
    NRMSE,
    MeanWeightedSumQuantileLoss,
)
import torch

torch.manual_seed(0)

# Load data from a CSV file into a PandasDataset
df = pd.read_csv(
    "https://raw.githubusercontent.com/AileenNielsen/"
    "TimeSeriesAnalysisWithPython/master/data/AirPassengers.csv",
    index_col=0,
    parse_dates=True,
)
dataset = PandasDataset(df, target="#Passengers")

# Split train/test data
training_data, test_gen = split(dataset, offset=-36)
test_data = test_gen.generate_instances(prediction_length=12, windows=3)

# Train model, or load serialized model
model_path = Path("some_deepar_model")
model_path.mkdir(exist_ok=True, parents=True)
train = True
if not train:
    model = Predictor.deserialize(model_path)
else:
    model = DeepAREstimator(
        freq="M",
        prediction_length=12,
        trainer_kwargs=dict(max_epochs=20),
    ).train(training_data=training_data)
    model.serialize(model_path)

forecasts = list(model.predict(test_data.input))

metrics = [
    SMAPE(),
    NRMSE(),
    MeanWeightedSumQuantileLoss(quantile_levels=[0.1, 0.5, 0.9]),
]
```

Then,

```python
res = evaluate_forecasts(forecasts, test_data=test_data, metrics=metrics)
print(res)

# prints
#          sMAPE     NRMSE  mean_weighted_sum_quantile_loss
# None  0.055886  0.066318                         0.035829
```

```python
res = evaluate_forecasts(forecasts, test_data=test_data, metrics=metrics, axis=1)
print(res)

# prints
#                           sMAPE     NRMSE  mean_weighted_sum_quantile_loss
# 
# NaN     1958-01        0.099674  0.109566                         0.072287
#         1959-01        0.030778  0.035876                         0.018940
#         1960-01        0.037205  0.044245                         0.021850
```

```python
res = evaluate_forecasts(forecasts, test_data=test_data, metrics=metrics, axis=0)
print(res)

# prints
#                        sMAPE     NRMSE  mean_weighted_sum_quantile_loss
# 
# 0                   0.040189  0.044771                         0.021694
# 1                   0.083287  0.082183                         0.051847
# 2                   0.056241  0.067684                         0.039326
# 3                   0.046029  0.074129                         0.037520
# 4                   0.040745  0.059825                         0.030936
# [...]
# 11                  0.096292  0.110979                         0.067913
```

### DIFF
--- a/src/gluonts/model/evaluation.py
+++ b/src/gluonts/model/evaluation.py
@@ -1,0 +1,246 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+import logging
+from collections import ChainMap
+from typing import Iterable, Optional, Union
+from dataclasses import dataclass
+from toolz import first, valmap
+
+import numpy as np
+import pandas as pd
+from tqdm import tqdm
+
+from gluonts.dataset import DataEntry
+from gluonts.dataset.split import TestData
+from gluonts.time_feature.seasonality import get_seasonality
+from gluonts.model import Forecast, Predictor
+from gluonts.ev import seasonal_error
+from gluonts.itertools import prod
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class BatchForecast:
+    """
+    Wrapper around ``Forecast`` objects, that adds a batch dimension
+    to arrays returned by ``__getitem__``, for compatibility with
+    ``gluonts.ev``.
+    """
+
+    forecast: Forecast
+    allow_nan: bool = False
+
+    def __getitem__(self, name):
+        value = self.forecast[name]
+        if np.isnan(value).any():
+            if not self.allow_nan:
+                raise ValueError("Forecast contains NaN values")
+
+            logger.warning(
+                "Forecast contains NaN values. Metrics may be incorrect."
+            )
+
+        return np.expand_dims(value.T, axis=0)
+
+
+def _get_data_batch(
+    input_: DataEntry,
+    label: DataEntry,
+    forecast: Forecast,
+    seasonality: Optional[int] = None,
+    mask_invalid_label: bool = True,
+    allow_nan_forecast: bool = False,
+) -> dict:
+    forecast_dict = BatchForecast(forecast, allow_nan=allow_nan_forecast)
+
+    freq = forecast.start_date.freqstr
+    if seasonality is None:
+        seasonality = get_seasonality(freq=freq)
+
+    label_target = label["target"]
+    if mask_invalid_label:
+        label_target = np.ma.masked_invalid(label_target)
+
+    other_data = {
+        "label": np.expand_dims(label_target, axis=0),
+        "seasonal_error": np.expand_dims(
+            seasonal_error(
+                input_["target"], seasonality=seasonality, time_axis=-1
+            ),
+            axis=0,
+        ),
+    }
+
+    return ChainMap(other_data, forecast_dict)
+
+
+def evaluate_forecasts_raw(
+    forecasts: Iterable[Forecast],
+    *,
+    test_data: TestData,
+    metrics,
+    axis: Optional[Union[int, tuple]] = None,
+    mask_invalid_label: bool = True,
+    allow_nan_forecast: bool = False,
+    seasonality: Optional[int] = None
+) -> dict:
+    """
+    Evaluate ``forecasts`` by comparing them with ``test_data``, according
+    to ``metrics``.
+
+    .. note:: This feature is experimental and may be subject to changes.
+
+    The optional ``axis`` arguments controls aggregation of the metrics:
+    - ``None`` (default) aggregates across all dimensions
+    - ``0`` aggregates across the dataset
+    - ``1`` aggregates across the first data dimension (time, in the univariate setting)
+    - ``2`` aggregates across the target dimension (time, in the multivariate setting)
+
+    Return results as a dictionary.
+    """
+    label_ndim = first(test_data.label)["target"].ndim
+
+    assert label_ndim in [1, 2]
+
+    if axis is None:
+        axis = tuple(range(label_ndim + 1))
+    if isinstance(axis, int):
+        axis = (axis,)
+
+    assert all(ax in range(3) for ax in axis)
+
+    evaluators = {}
+    for metric in metrics:
+        evaluator = metric(axis=axis)
+        evaluators[evaluator.name] = evaluator
+
+    index_data = []
+
+    for input_, label, forecast in tqdm(
+        zip(test_data.input, test_data.label, forecasts)
+    ):
+        if 0 not in axis:
+            index_data.append((forecast.item_id, forecast.start_date))
+
+        data_batch = _get_data_batch(
+            input_,
+            label,
+            forecast,
+            seasonality=seasonality,
+            mask_invalid_label=mask_invalid_label,
+            allow_nan_forecast=allow_nan_forecast,
+        )
+
+        for evaluator in evaluators.values():
+            evaluator.update(data_batch)
+
+    metrics_values = {
+        metric_name: evaluator.get()
+        for metric_name, evaluator in evaluators.items()
+    }
+
+    if index_data:
+        metrics_values["__index_0"] = index_data
+
+    return metrics_values
+
+
+def evaluate_forecasts(
+    forecasts: Iterable[Forecast],
+    *,
+    test_data: TestData,
+    metrics,
+    axis: Optional[Union[int, tuple]] = None,
+    mask_invalid_label: bool = True,
+    allow_nan_forecast: bool = False,
+    seasonality: Optional[int] = None
+) -> pd.DataFrame:
+    """
+    Evaluate ``forecasts`` by comparing them with ``test_data``, according
+    to ``metrics``.
+
+    .. note:: This feature is experimental and may be subject to changes.
+
+    The optional ``axis`` arguments controls aggregation of the metrics:
+    - ``None`` (default) aggregates across all dimensions
+    - ``0`` aggregates across the dataset
+    - ``1`` aggregates across the first data dimension (time, in the univariate setting)
+    - ``2`` aggregates across the target dimension (time, in the multivariate setting)
+
+    Return results as a Pandas ``DataFrame``.
+    """
+    metrics_values = evaluate_forecasts_raw(
+        forecasts=forecasts,
+        test_data=test_data,
+        metrics=metrics,
+        axis=axis,
+        mask_invalid_label=mask_invalid_label,
+        allow_nan_forecast=allow_nan_forecast,
+        seasonality=seasonality,
+    )
+    index0 = metrics_values.pop("__index_0", None)
+
+    metric_shape = metrics_values[first(metrics_values)].shape
+    if metric_shape == ():
+        index = [None]
+    else:
+        index_arrays = np.unravel_index(
+            range(prod(metric_shape)), metric_shape
+        )
+        if index0 is not None:
+            index0_repeated = np.take(index0, indices=index_arrays[0], axis=0)
+            index_arrays = (*zip(*index0_repeated), *index_arrays[1:])
+        index = pd.MultiIndex.from_arrays(index_arrays)
+
+    flattened_metrics = valmap(np.ravel, metrics_values)
+
+    return pd.DataFrame(flattened_metrics, index=index)
+
+
+def evaluate_model(
+    model: Predictor,
+    *,
+    test_data: TestData,
+    metrics,
+    axis: Optional[Union[int, tuple]] = None,
+    mask_invalid_label: bool = True,
+    allow_nan_forecast: bool = False,
+    seasonality: Optional[int] = None
+) -> pd.DataFrame:
+    """
+    Evaluate ``model`` when applied to ``test_data``, according
+    to ``metrics``.
+
+    .. note:: This feature is experimental and may be subject to changes.
+
+    The optional ``axis`` arguments controls aggregation of the metrics:
+    - ``None`` (default) aggregates across all dimensions
+    - ``0`` aggregates across the dataset
+    - ``1`` aggregates across the first data dimension (time, in the univariate setting)
+    - ``2`` aggregates across the target dimension (time, in the multivariate setting)
+
+    Return results as a Pandas ``DataFrame``.
+    """
+    forecasts = model.predict(test_data.input)
+
+    return evaluate_forecasts(
+        forecasts=forecasts,
+        test_data=test_data,
+        metrics=metrics,
+        axis=axis,
+        mask_invalid_label=mask_invalid_label,
+        allow_nan_forecast=allow_nan_forecast,
+        seasonality=seasonality,
+    )
--- a/src/gluonts/model/evaluation.py
+++ b/src/gluonts/model/evaluation.py
@@ -106,7 +106,7 @@ def evaluate_forecasts_raw(
     - ``None`` (default) aggregates across all dimensions
     - ``0`` aggregates across the dataset
     - ``1`` aggregates across the first data dimension (time, in the univariate setting)
-    - ``2`` aggregates across the target dimension (time, in the multivariate setting)
+    - ``2`` aggregates across the second data dimension (time, in the multivariate setting)
 
     Return results as a dictionary.
     """
@@ -177,7 +177,7 @@ def evaluate_forecasts(
     - ``None`` (default) aggregates across all dimensions
     - ``0`` aggregates across the dataset
     - ``1`` aggregates across the first data dimension (time, in the univariate setting)
-    - ``2`` aggregates across the target dimension (time, in the multivariate setting)
+    - ``2`` aggregates across the second data dimension (time, in the multivariate setting)
 
     Return results as a Pandas ``DataFrame``.
     """
@@ -229,7 +229,7 @@ def evaluate_model(
     - ``None`` (default) aggregates across all dimensions
     - ``0`` aggregates across the dataset
     - ``1`` aggregates across the first data dimension (time, in the univariate setting)
-    - ``2`` aggregates across the target dimension (time, in the multivariate setting)
+    - ``2`` aggregates across the second data dimension (time, in the multivariate setting)
 
     Return results as a Pandas ``DataFrame``.
     """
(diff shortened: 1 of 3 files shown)


src/gluonts/ev/interface.py
```diff
@@ -0,0 +1,191 @@
...
+from gluonts.ext.naive_2 import naive_2
+from gluonts.time_feature.seasonality import get_seasonality
+from gluonts.model.forecast import Forecast
+from gluonts.ev import seasonal_error
+
+
+@dataclass
+class ForecastDict:
+    """
+    Wrapper around ``Forecast`` objects, that adds ``__getitem__`` and
```
jaheba: I think `Forecast` objects already support this.

src/gluonts/ev/interface.py
```diff
@@ -0,0 +1,191 @@
...
+
+    for input_, label, forecast in zip(
+        test_data.input, test_data.label, forecasts
+    ):
+        forecast_length = len(forecast.index)
+        if axis == 1:
+            index.append((forecast.item_id, forecast.start_date))
+        data_batch = _get_data_batch(input_, label, forecast)
+        for evaluator in evaluators.values():
+            evaluator.update(data_batch)
```
jaheba: Some spacing would be nice.

src/gluonts/ev/interface.py
```diff
@@ -0,0 +1,191 @@
...
+) -> Tuple[dict, Optional[list]]:
+    """
+    Evaluate ``metrics`` by comparing ``forecasts`` to ``test_data``.
+
+    The optional ``axis`` arguments controls aggregation of the metrics:
+    - ``None`` (default) aggregates across all dimensions
+    - ``0`` aggregates across the batch dimension, giving one metric value per forecast horizon
+    - ``1`` aggregates across the time dimension, giving one metric value per batch item
+
+    Return results as a dictionary, plus an optional index (list).
```
jaheba: What is the index for?

src/gluonts/ev/interface.py
```diff
@@ -0,0 +1,191 @@
...
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+from typing import ChainMap, Tuple, Optional, Iterable, Union
```
jaheba: `ChainMap` should be imported from `collections`.

src/gluonts/ev/interface.py
```diff
@@ -0,0 +1,191 @@
...
+    - ``1`` aggregates across the time dimension, giving one metric value per batch item
+
+    Return results as a dictionary, plus an optional index (list).
+    """
+    evaluators = {}
+    for metric in metrics:
+        evaluator = metric(axis=axis)
+        evaluators[evaluator.name] = evaluator
+
+    assert axis is None or axis in [0, 1]
```
jaheba: Can we put the assert as the first statement of the function body?

src/gluonts/ev/interface.py
```diff
@@ -0,0 +1,191 @@
...
+        index = list(range(forecast_length))
+    elif axis is None:
+        index = None
+
+    metrics_dict = {
+        metric_name: evaluator.get()
+        for metric_name, evaluator in evaluators.items()
+    }
+
+    return metrics_dict, index
```
abdulfatir: I don't like the `index` being returned from this function. Can we change it to only return the metrics and return the index only when asked for it (by the `_pd` method)?

src/gluonts/ev/interface.py
lostella: It's mainly for the downstream `evaluate_pd` (see below, name is tentative). The thing is, as long as you provide metric values as arrays, it's hard to read through them if you don't know what the axes are. For example, which `item_id` does this "0.12345" MAPE value refers to?

This is all up for discussion, but I think that an "interface" at this level should provide a more readable structure than plain arrays.

src/gluonts/ev/interface.py
lostella: The thing is, it needs to be constructed, and you may need to iterate through the data to do that (which is what `evaluate` does). So this is why it's returned there.

Another option is to provide it as part of `metrics_dict` itself, let me try that out.

abdulfatir: Part of the metrics dict would be better imo.

src/gluonts/ev/interface.py
lostella: Ah right, then I just need to add the batch dimension

src/gluonts/ev/interface.py
```diff
@@ -0,0 +1,200 @@
...
+    freq = forecast.start_date.freqstr
+    seasonality = get_seasonality(freq=freq)
+
+    other_data = {
+        "label": np.array([label["target"]]),
+        "seasonal_error": np.array(
+            [seasonal_error(input_["target"], seasonality=seasonality)]
+        ),
+        "naive_2": np.array(
+            [naive_2(input_["target"], len(label["target"]), freq=freq)]
```
jaheba: These all should be arrays already. Can we use `expand_dims` instead?

src/gluonts/ev/interface.py
lostella: Done, I agree it's better. One question is what the index name should be though, I guess it makes sense to have a different name if it's predicition horizon vs item_id + forecast time

jaheba:
One goal of `ev` was to be independent from the rest of gluonts. This adds some dependencies.

Would it be an option to put it into model?

lostella:
> One goal of `ev` was to be independent from the rest of gluonts. This adds some dependencies.
> 
> Would it be an option to put it into model?

I think that would make sense. On the other hand, one would need to, say `from gluonts.model import evaluate` and then also `from gluonts.ev import SMAPE, NRMSE, ...`. Would you then re-export the metrics from `gluonts.model`?

src/gluonts/ev/interface.py
```diff
@@ -0,0 +1,145 @@
...
+def _get_data_batch(
+    input_: DataEntry, label: DataEntry, forecast: Forecast
+) -> ChainMap:
+    forecast_dict = BatchForecast(forecast)
+
+    freq = forecast.start_date.freqstr
+    seasonality = get_seasonality(freq=freq)
+
+    other_data = {
+        "label": np.expand_dims(label["target"], axis=0),
```
jaheba: By the way, is this actually necessary? Broadcasting aligns array shapes on the right and inserts dimensions as needed.

I think we can get away without using `expand_dims` here.

src/gluonts/ev/interface.py
```diff
@@ -0,0 +1,145 @@
...
+    Call ``evaluate``, wrap result in a ``pd.DataFrame``.
+    """
+    metrics_dict = evaluate(
+        metrics=metrics,
+        test_data=test_data,
+        forecasts=forecasts,
+        axis=axis,
+    )
+
+    __index = metrics_dict.pop("__index")
```
jaheba:
```suggestion
    index = metrics_dict.pop("__index")
```

Don't think we need to keep the `__` in a variable name.

src/gluonts/ev/interface.py
lostella: I would have hoped this were not necessary, but it seems like it is. There must be some assumption on the batch dimension actually being there. I will look deeper into this

gorold:
Thanks for the great work! Would be good to include the following:
1. Validate no nans in forecast
2. Mask invalids in targets
3. Add dimension wise aggregation function for multivariate
4. Manually define seasonality
5. Transpose targets if multivariate (no aggregation function)

lostella:
> 5. Transpose targets if multivariate (no aggregation function)

Hi @gorold, what do you mean by that? ☝️

gorold:
> > 9\. Transpose targets if multivariate (no aggregation function)
> 
> Hi @gorold, what do you mean by that? ☝️

Targets from the dataset are (C, T) while forecasts are (T,C). But I think the ev module makes some assumptions that targets should be (T,C) so we need to transpose them!

abdulfatir:
@lostella A couple of things.

- It's not clear to me how one would compute aggregate metrics like CRPS-sum in this setup?
- I think we should add an alias named `CRPS` for `MeanWeightedSumQuantileLoss` with `{0.1, 0.2, ..., 0.9}` quantiles and clarify in the docstring how this is an approximation of CRPS. Everyone uses this for CRPS but I don't think it's documented anywhere clearly. I can do this in a separate PR but I would like to know your opinion first. What do you think @kashif?

gorold:
I believe this axis is wrong:
https://github.com/awslabs/gluonts/blob/cbc66c90569d09739bb1f494af20de3985ab44e1/src/gluonts/ev/ts_stats.py#L22

based on this:
https://github.com/awslabs/gluonts/blob/cbc66c90569d09739bb1f494af20de3985ab44e1/src/gluonts/ev/ts_stats.py#L34-L37

lostella:
> I believe this axis is wrong:

Looks like so, I think it's always best to just be explicit on what the time axis is, so it could be an additional argument with default 0

src/gluonts/model/evaluation.py
```diff
@@ -0,0 +1,180 @@
...
+    seasonality: Optional[int] = None
+) -> dict:
+    """
+    Evaluate ``metrics`` by comparing ``forecasts`` to ``test_data``.
+
+    The optional ``axis`` arguments controls aggregation of the metrics:
+    - ``None`` (default) aggregates across all dimensions
+    - ``0`` aggregates across the batch dimension
+    - ``1`` aggregates across the time dimension
+    - ``2`` aggregates across the target dimension
```
huibinshen: Nit: I find the `target dimension` is a bit confusing. Is it for different quantiles or different time series in a multivariate setting or both? Might be worth to explain what does it mean in the docstring.

src/gluonts/model/evaluation.py
```diff
@@ -0,0 +1,180 @@
...
+        value = self.forecast[name]
+        valid = ~np.isnan(value).any()
+
+        if not valid and self.allow_nan:
+            logger.warning(
+                "Forecast contains NaN values. Metrics may be incorrect."
+            )
+
+        if not valid and not self.allow_nan:
+            raise ValueError("Forecast contains NaN values")
```
jaheba:
```suggestion
        if np.isnan(value).any():
            if not self.allow_nan:
                raise ValueError("Forecast contains NaN values")
            
            logger.warning(
                "Forecast contains NaN values. Metrics may be incorrect."
            )              
```

src/gluonts/model/evaluation.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def _get_data_batch(
+    input_: DataEntry,
+    label: DataEntry,
+    forecast: Forecast,
+    seasonality: Optional[int] = None,
+    mask_invalid_label: bool = True,
+    allow_nan_forecast: bool = False,
+) -> ChainMap:
```
jaheba: I don't think `ChainMap` is the interface that we want to expose. Can't this just be `dict`?

src/gluonts/model/evaluation.py
```diff
@@ -0,0 +1,180 @@
...
+    - ``2`` aggregates across the target dimension
+
+    Return results as a dictionary.
+    """
+    label_shape = first(test_data.label)["target"].shape
+
+    assert len(label_shape) in [1, 2]
+
+    if axis is None:
+        axis = tuple(range(len(label_shape) + 1))
```
jaheba:
```suggestion
    label_ndims = first(test_data.label)["target"].ndims

    assert label_ndims in [1, 2]

    if axis is None:
        axis = tuple(range(label_ndims + 1))
```

src/gluonts/model/evaluation.py
```diff
@@ -0,0 +1,180 @@
...
+    if isinstance(axis, int):
+        axis = (axis,)
+
+    assert axis is None or all(a in range(3) for a in axis)
```
jaheba: I HATE single letter variable names. Please, please, please don't use them.

src/gluonts/model/evaluation.py
```diff
@@ -0,0 +1,180 @@
...
+        for metric_name, evaluator in evaluators.items()
+    }
+
+    if index_data:
+        metrics_dict["__index_0"] = index_data
+
+    return metrics_dict
+
+
+def evaluate_pd(*args, **kwargs) -> pd.DataFrame:
```
jaheba: I think it's generally better to be explicit here, since this shows up in documentation, etc.

src/gluonts/model/evaluation.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+    Call ``evaluate``, flatten the results and wrap them
+    in a ``pd.DataFrame``.
+    """
+    metrics_dict = evaluate(*args, **kwargs)
```
jaheba: I'm really not a fan of Hungarian notation.

If you want to indicate that this is a dict, I think this is better:

```suggestion
    metrics: dict = evaluate(*args, **kwargs)
```

src/gluonts/model/evaluation.py
```diff
@@ -0,0 +1,180 @@
...
+    else:
+        index_arrays = np.unravel_index(
+            range(prod(metric_shape)), metric_shape
+        )
+        if index0 is not None:
+            index0_repeated = np.take(index0, indices=index_arrays[0], axis=0)
+            index_arrays = (*zip(*index0_repeated), *index_arrays[1:])
+        index = pd.MultiIndex.from_arrays(index_arrays)
+
+    flattened_metrics = {k: v.ravel() for k, v in metrics_dict.items()}
```
jaheba:
```suggestion
    flattened_metrics = valmap(pd.Series.ravel, metrics_dict)
```

jaheba (changes requested):

src/gluonts/model/evaluation.py
lostella: It’s the dimension in the multivariate setting. But I’ll try to explain better how the `axis` dimension is to be used

src/gluonts/model/evaluation.py
```diff
@@ -0,0 +1,202 @@
...
+    to arrays returned by ``__getitem__``, for compatibility with
+    ``gluonts.ev``.
+    """
+
+    forecast: Forecast
+    allow_nan: bool = False
+
+    def __getitem__(self, name):
+        value = self.forecast[name]
+        if np.isnan(value).any():
```
jaheba, line 47: Is this actually the place to check for nan?

Why don't we dot it in get_data_batch?

lostella: Because in `get_data_batch` I don't know what forecast keys will be accessed

src/gluonts/model/evaluation.py
lostella: done

src/gluonts/model/evaluation.py
lostella: done

src/gluonts/model/evaluation.py
lostella: done

src/gluonts/model/evaluation.py
lostella: done

src/gluonts/model/evaluation.py
lostella: done

src/gluonts/model/evaluation.py
lostella: done

src/gluonts/model/evaluation.py
lostella: done

jaheba:
Should we re-export these function in `gluonts.model`?

src/gluonts/model/evaluation.py
```diff
@@ -0,0 +1,245 @@
...
+    axis: Optional[Union[int, tuple]] = None,
+    mask_invalid_label: bool = True,
+    allow_nan_forecast: bool = False,
+    seasonality: Optional[int] = None
+) -> pd.DataFrame:
+    """
+    Evaluate ``forecasts`` by comparing them with ``test_data``, according
+    to ``metrics``.
+
+    Note: this feature is experimental and may be subject to changes.
```
jaheba: I think you might want to use it like it is described here: https://pradyunsg.me/furo/reference/admonitions/

lostella: Done!

test/model/test_evaluation.py
```diff
@@ -0,0 +1,222 @@
...
+def test_evaluation_shape(
+    metrics,
+    test_data: TestData,
+    axis: Union[int, Tuple[int]],
+):
+    num_samples = 31
+
+    sample_forecasts = [
+        SampleForecast(
+            samples=np.zeros((num_samples, *label["target"].shape)),
```
gorold: I think the samples shouldn't be constructed based on label shape (C, T). SampleForecast for multivariate follow (N, T, C).

src/gluonts/model/evaluation.py
```diff
@@ -0,0 +1,245 @@
...
+    label_ndim = first(test_data.label)["target"].ndim
+
+    assert label_ndim in [1, 2]
+
+    if axis is None:
+        axis = tuple(range(label_ndim + 1))
+    if isinstance(axis, int):
+        axis = (axis,)
+
+    assert axis is None or all(ax in range(3) for ax in axis)
```
gorold: axis wouldn't be None here

gorold:
Would be nice to have tqdm here as well!

src/gluonts/model/evaluation.py
lostella: true that

test/model/test_evaluation.py
lostella: And this is indeed pretty annoying

gorold:
I think now theres a mismatch between the docstring and actual computation.. docstring states:

- ``None`` (default) aggregates across all dimensions
- ``0`` aggregates across the dataset
- ``1`` aggregates across the time dimension
- ``2`` aggregates across the target dimension (multivariate setting)

but now the inputs and labels have (N, C, T) shape.

lostella:
@gorold that's what is annoying about this. The time axis should not change position in the univariate vs multivariate case, in my opinion, but that would mean changing the data layout assumption in the multivariate case.

jaheba (dismissed):

jaheba (dismissed):

jaheba (approved):

PR #2681, opened 2023-02-21
Allow `axis` to be a tuple in `ev.aggregations`

lostella:
*Description of changes:* This extends the behaviour of `gluonts.ev.aggregations.Aggregation` to the case where `axis` is a tuple, to reduce along multiple axes at once. A test is expanded by using the newly introduced `power_set` (see https://docs.python.org/3/library/itertools.html#itertools-recipes, I'm open to split this into a separate PR). This change can be useful for multivariate evaluation, where one may want to reduce across different combinations of batch/time/covariate dimensions.

### DIFF
--- a/src/gluonts/itertools.py
+++ b/src/gluonts/itertools.py
@@ -393,7 +393,7 @@ def power_set(iterable):
 
     >>> list(power_set(["a", "b"]))
     [(), ('a',), ('b',), ('a', 'b')]
-    
+
     Adapted from https://docs.python.org/3/library/itertools.html#itertools-recipes
     """
     return itertools.chain.from_iterable(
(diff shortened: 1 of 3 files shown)


src/gluonts/itertools.py
```diff
@@ -385,3 +385,10 @@ def pluck_attr(seq, name, default=_no_default):
         return [getattr(el, name) for el in seq]
 
     return [getattr(el, name, default) for el in seq]
+
+
+def power_set(iterable):
```
jaheba: Can you add a docstring?

lostella: moved this to #2682

jaheba (approved):

PR #2709, opened 2023-03-07
update hierarchicalforecast for new release

melopeo:
*Description of changes:*
Updating hierarchicalforecast predictor to work with newest release


lostella:
Thanks a lot @melopeo! Does the requirement also need to be bumped to `~=0.3`, or is it still compatible with `0.2` as well?

melopeo:
> 

The fix works for both versions. Should I better update the requirements to `~=0.3`?

lostella:
I think it's fine both ways

lostella (approved):

Issue #2662, opened 2023-02-16
DistributionForecast fails on GPU


gorold:
## Description
(A clear and concise description of what the bug is.)
As per the title, DistributionForecast fails on GPU. There are 2 sources of error, firstly

https://github.com/awslabs/gluonts/blob/c5b64b4952a89f172c9c1e1fa9f1cfc0ee684a95/src/gluonts/torch/model/forecast.py#L89

```torch.tensor([level])``` will be on a different device compared to the parameters in ```self.distribution```

Next,
https://github.com/awslabs/gluonts/blob/c5b64b4952a89f172c9c1e1fa9f1cfc0ee684a95/src/gluonts/model/forecast_generator.py#L202
```DistributionForecastGenerator``` should use ```predict_to_numpy``` in line with Sample and Quantile generators.

## To Reproduce
(Please provide minimal example of code snippet that reproduces the error. For existing examples, please provide link.)

```python
import pandas as pd
import numpy as np
from gluonts.dataset.pandas import PandasDataset
from gluonts.evaluation import make_evaluation_predictions
from gluonts.torch.model.simple_feedforward import SimpleFeedForwardEstimator
from gluonts.evaluation import Evaluator


n = 100
freq = 'D'
date = "2015-04-07 00:00:00"

df = pd.DataFrame(np.random.randn(n).astype(np.float32), index=pd.period_range(date, periods=n, freq=freq))
dataset = PandasDataset(df, target=0)

model = SimpleFeedForwardEstimator(prediction_length=5, trainer_kwargs={"max_epochs": 1, "accelerator": "gpu"})
predictor = model.train(dataset)

forecast_it, ts_it = make_evaluation_predictions(
    dataset, predictor=predictor, num_samples=100
)

# Error 1
next(forecast_it).quantile(0.5)


# Error 2
evaluator = Evaluator()
agg_metrics, item_metrics = evaluator(
    ts_it, forecast_it,
)
```

## Error message or code output
(Paste the complete error message, including stack trace, or the undesired output that the above snippet produces.)

Error 1
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Input In [17], in <cell line: 1>()
----> 1 next(forecast_it).quantile(0.5)

File /export/home/forecasting/aiops_tsf/benchmark_exp/pytorch_venv/lib/python3.9/site-packages/gluonts/torch/model/forecast.py:89, in DistributionForecast.quantile(self, level)
[...]

PR #2714, opened 2023-03-07
Fix DistributionForecast failure on GPU

huibinshen:
*Issue #, if available:*

*Description of changes:*

Fixing https://github.com/awslabs/gluonts/issues/2662

### DIFF
--- a/src/gluonts/evaluation/_base.py
+++ b/src/gluonts/evaluation/_base.py
@@ -32,7 +32,6 @@ from typing import (
 
 import numpy as np
 import pandas as pd
-import torch
 
 from gluonts.gluonts_tqdm import tqdm
 from gluonts.model.forecast import Forecast, Quantile
@@ -257,9 +256,6 @@ class Evaluator:
             desc="Running evaluation",
         ) as it, np.errstate(divide="ignore", invalid="ignore"):
             if self.num_workers and not sys.platform == "win32":
-
-                if torch.cuda.is_available():
-                    torch.multiprocessing.set_start_method('spawn', force=True)
 
                 mp_pool = multiprocessing.Pool(
                     initializer=None, processes=self.num_workers
--- a/src/gluonts/evaluation/_base.py
+++ b/src/gluonts/evaluation/_base.py
@@ -32,6 +32,7 @@ from typing import (
 
 import numpy as np
 import pandas as pd
+import torch
 
 from gluonts.gluonts_tqdm import tqdm
 from gluonts.model.forecast import Forecast, Quantile
@@ -256,6 +257,10 @@ class Evaluator:
             desc="Running evaluation",
         ) as it, np.errstate(divide="ignore", invalid="ignore"):
             if self.num_workers and not sys.platform == "win32":
+
+                if torch.cuda.is_available():
+                    torch.multiprocessing.set_start_method('spawn', force=True)
+
                 mp_pool = multiprocessing.Pool(
                     initializer=None, processes=self.num_workers
                 )
--- a/src/gluonts/evaluation/_base.py
+++ b/src/gluonts/evaluation/_base.py
@@ -256,7 +256,6 @@ class Evaluator:
             desc="Running evaluation",
         ) as it, np.errstate(divide="ignore", invalid="ignore"):
             if self.num_workers and not sys.platform == "win32":
-
                 mp_pool = multiprocessing.Pool(
                     initializer=None, processes=self.num_workers
                 )
--- a/src/gluonts/torch/__init__.py
+++ b/src/gluonts/torch/__init__.py
@@ -19,7 +19,10 @@ __all__ = [
     "SimpleFeedForwardEstimator",
     "TemporalFusionTransformerEstimator",
 ]
+import torch
 
+if torch.cuda.is_available():
+    torch.multiprocessing.set_start_method("spawn", force=True)
 
 from .model.estimator import PyTorchLightningEstimator
 from .model.predictor import PyTorchPredictor
--- a/src/gluonts/torch/__init__.py
+++ b/src/gluonts/torch/__init__.py
@@ -20,13 +20,8 @@ __all__ = [
     "TemporalFusionTransformerEstimator",
 ]
 import torch
-
-if torch.cuda.is_available():
-    torch.multiprocessing.set_start_method("spawn", force=True)
-
 from .model.estimator import PyTorchLightningEstimator
 from .model.predictor import PyTorchPredictor
-
 from .model.deep_npts import DeepNPTSEstimator
 from .model.deepar import DeepAREstimator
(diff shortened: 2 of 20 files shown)


src/gluonts/evaluation/_base.py
```diff
@@ -256,6 +257,10 @@ def __call__(
             desc="Running evaluation",
         ) as it, np.errstate(divide="ignore", invalid="ignore"):
             if self.num_workers and not sys.platform == "win32":
+
+                if torch.cuda.is_available():
+                    torch.multiprocessing.set_start_method('spawn', force=True)
+
```
lostella: I think this should not happen here, it's probably better to keep it on the caller side

test/test_on_gpu/test_torch_model.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@pytest.mark.parametrize(
+    "torch_model_cls",
+    [
+        SimpleFeedForwardEstimator,
+        DeepAREstimator
+    ]
+)
+def test_torch_model_on_gpu(
```
lostella: You can condition the execution of this test on `torch.cuda.is_available()` using `skipif` from `pytest`: https://docs.pytest.org/en/7.1.x/how-to/skipping.html#id1

test/test_on_gpu/test_torch_model.py
```diff
@@ -0,0 +1,57 @@
...
+    torch_model_cls: torch.model.PyTorchLightningEstimator
```
lostella: You're passing a class now (so, a `Type` and not a `PyTorchLightningEstimator`), but you could indeed pass the estimator object directly

src/gluonts/evaluation/_base.py
huibinshen: For now I put it into `src/gluonts/torch/__init__.py`, what do you think of it? Or is it better to put it as part of the test you mean?

test/test_on_gpu/test_torch_model.py
huibinshen: What is the preferred way in this case?

src/gluonts/evaluation/_base.py
lostella: Yes, maybe part of the test is better

test/test_on_gpu/test_torch_model.py
lostella: I think passing the estiamator object (so, creating it as part of the `pytest.mark.parametrize` decorator)

src/gluonts/torch/__init__.py
```diff
@@ -19,7 +19,10 @@
     "SimpleFeedForwardEstimator",
     "TemporalFusionTransformerEstimator",
 ]
+import torch
 
+if torch.cuda.is_available():
+    torch.multiprocessing.set_start_method("spawn", force=True)
```
jaheba: You generally want to do this after all imports, not mix them in.

src/gluonts/evaluation/_base.py
```diff
@@ -256,6 +256,7 @@ def __call__(
             desc="Running evaluation",
         ) as it, np.errstate(divide="ignore", invalid="ignore"):
             if self.num_workers and not sys.platform == "win32":
+
```
jaheba: Let's revert this change :)

src/gluonts/torch/model/forecast.py
```diff
@@ -86,7 +86,12 @@ def mean_ts(self) -> pd.Series:
 
     def quantile(self, level: Union[float, str]) -> np.ndarray:
         level = Quantile.parse(level).value
-        return self.distribution.icdf(torch.tensor([level])).cpu().numpy()
+        device = self.distribution.mean.device
+        return (
+            self.distribution.icdf(torch.tensor([level], device=device))
```
jaheba: I think I would just inline `self.distribution.mean.device` here.

test/test_on_gpu/test_torch_model.py
```diff
@@ -0,0 +1,56 @@
...
+from gluonts.dataset.pandas import PandasDataset
+from gluonts.evaluation import make_evaluation_predictions
+from gluonts.torch.model.estimator import PyTorchLightningEstimator
+from gluonts.torch.model.simple_feedforward import SimpleFeedForwardEstimator
+from gluonts.torch.model.deepar import DeepAREstimator
+from gluonts.evaluation import Evaluator
+
+
+@pytest.mark.skipif(
+    torch.cuda.is_available() is False, reason="only run the test on GPU"
```
jaheba:
```suggestion
    not torch.cuda.is_available(), reason="only run the test on GPU"
```

src/gluonts/evaluation/_base.py
huibinshen: If I leave it to the test function, the users may still face the error when using torch models.

src/gluonts/torch/__init__.py
huibinshen: Do you mean putting it at the end of the `torch/__init__.py` file?

src/gluonts/evaluation/_base.py
huibinshen: That would leave the following fix not necessary
```
if torch.cuda.is_available():
    torch.multiprocessing.set_start_method("spawn", force=True)
```
Are we sure that we want to do this? What's your input? @lostella

jaheba: No, I mean the empty line that you've introduced ;)

src/gluonts/torch/__init__.py
jaheba: Somewhere below the imports.

src/gluonts/torch/model/forecast.py
```diff
@@ -86,7 +86,13 @@ def mean_ts(self) -> pd.Series:
 
     def quantile(self, level: Union[float, str]) -> np.ndarray:
         level = Quantile.parse(level).value
-        return self.distribution.icdf(torch.tensor([level])).cpu().numpy()
+        return (
+            self.distribution.icdf(
+                torch.tensor([level], device=self.distribution.mean.device)
```
jaheba, line 91: Something we could also do is to add `.device` to the distribution class, that would this a bit cleaner in my view.

huibinshen: `Distribution` is from torch, I could add a `.device` into the `AffineTransformed` class.

jaheba:
Looks generally good to me, did you try running this on GPU and it succeeds?

jaheba:
I think my suggestion with the `device` things was actually a bad one, sorry!

huibinshen:
> Looks generally good to me, did you try running this on GPU and it succeeds?

I ran the test on a GPU instance and it was fine.

lostella (approved):

PR #2727, opened 2023-03-17
Update CP-Flow fork as extra dependency for MQF2

lostella:
*Description of changes:* relaxed `torch` and `pandas` requirements in my fork, so that this doesn't hold back dependencies in the test workflows https://github.com/lostella/CP-Flow/blob/gluonts-dependency/requirements.txt


lostella:
As expected, `torch==2.0` getting installed and not replaced when `cpflows` is installed, see https://github.com/awslabs/gluonts/actions/runs/4447126839/jobs/7808282084

melopeo (approved):

Issue #2743, opened 2023-03-21
Transformer model breaks when scaling is set to `False`


abdulfatir:
## Description

The MXNet transformer model breaks when `scaling=False` is used. The following code works as expected with `scaling=True` (default).

## To Reproduce

```python
from gluonts.mx import TransformerEstimator, Trainer
from gluonts.dataset.repository.datasets import get_dataset

dataset = get_dataset("airpassengers")

tft_default = TransformerEstimator(
    freq=dataset.metadata.freq,
    prediction_length=dataset.metadata.prediction_length,
    trainer=Trainer(epochs=10),
    scaling=False,
)
predictor_tft_default = tft_default.train(
    dataset.train,
)

```

## Error message or code output

```
  0%|          | 0/50 [00:00<?, ?it/s]/Users/ansarnd/miniconda3/envs/gluonts/lib/python3.8/site-packages/mxnet/gluon/block.py:825: UserWarning: The 3-th input to HybridBlock is not used by any computation. Is this intended?
  out = self.forward(*args)

  0%|          | 0/50 [00:00<?, ?it/s]
infer_shape error. Arguments:
  data0: (32, 1)
  data1: (32, 49, 2)
  data2: (32, 49)
  data3: (32, 49)
  data4: (32, 12, 2)
  data5: (32, 12)
Traceback (most recent call last):
  File "/Users/ansarnd/miniconda3/envs/gluonts/lib/python3.8/site-packages/mxnet/gluon/block.py", line 1102, in _build_cache
    params[name].data()
  File "/Users/ansarnd/miniconda3/envs/gluonts/lib/python3.8/site-packages/mxnet/gluon/parameter.py", line 574, in data
    return self._check_and_get(self._data, ctx)
  File "/Users/ansarnd/miniconda3/envs/gluonts/lib/python3.8/site-packages/mxnet/gluon/parameter.py", line 230, in _check_and_get
    raise DeferredInitializationError(
mxnet.gluon.parameter.DeferredInitializationError: Parameter 'dec_inputlayer0_dense0_weight' has not been initialized yet because initialization was deferred. Actual initialization happens during the first forward pass. Please pass one batch of data through the network before accessing Parameters. You can also avoid deferred initialization by specifying in_units, num_features, etc., for network layers.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #2744, opened 2023-03-21
Fix MXNet NOPScaler

abdulfatir:
*Issue #, if available:* #2743

*Description of changes:*
This PR modifies MXNet `NOPScaler` to use `observed_indicator` in the computation graph (redundantly). MXNet models break if an input is not used in the model somehow 😮 (see https://github.com/apache/mxnet/issues/14373).


lostella:
Being this related to #2743, is there no other way `TransformerEstimator` could use this indicator more productively?

src/gluonts/mx/block/scaler.py
```diff
@@ -295,4 +295,5 @@ def compute_scale(
         Tensor
             shape (N, C), identically equal to 1.
         """
-        return F.ones_like(data).mean(axis=self.axis)
+        _ones = F.ones_like(data) * F.ones_like(observed_indicator)
+        return _ones.mean(axis=self.axis)
```
lostella: All scalers use `data` anyway (see `hybrid_forward`) so maybe this could just do

```python
return F.ones_like(observed_indicator).mean(axis=self.axis)
```

abdulfatir:
It can (probably should) be used more productively in the model, I think.

src/gluonts/mx/block/scaler.py
abdulfatir: Makes sense. Changed.

lostella (approved):

PR #2748, opened 2023-03-22
Add patch-TST, D-Linear and a new lag-TST model

kashif:
*Description of changes:*

Implemented patch-TST model. https://arxiv.org/abs/2211.14730

implemented D-Linear model

added my own creation: "Lag-TST" which uses freq based lags to construct vectors


kashif:
having gone through the paper and implementation a better name for "patch-TST" might be "segment-TST" as there are no patches (which are 2D in nature?) in this method and instead, they "segment" each univariate time series into fixed-sized pieces and use the segments as vectors

lostella:
@kashif I think PatchTST is fine, if that’s what they call it in the paper…

src/gluonts/torch/model/patch_tst/module.py
```diff
@@ -0,0 +1,218 @@
...
+            past_key_values_length + seq_len,
+            dtype=torch.long,
+            device=self.weight.device,
+        )
+        return super().forward(positions)
+
+
+class PatchTSTModel(nn.Module):
+    """
+    Module implementing the PatchTST model for forecasting.
```
lostella: I think it would make sense to reference the paper here. Maybe also mention of any important differences? Like the distribution head I suppose

src/gluonts/torch/model/patch_tst/estimator.py
```diff
@@ -0,0 +1,285 @@
...
+    "future_target",
+    "future_observed_values",
+]
+
+
+class PatchTSTEstimator(PyTorchLightningEstimator):
+    """
+    An estimator training the PatchTST model for forecasting.
+
+    This class is uses the model defined in ``SimpleFeedForwardModel``,
```
lostella: I guess it’s 

```suggestion
    This class uses the model defined in ``PatchTSTModel``,
```

src/gluonts/torch/model/d_linear/lightning_module.py
```diff
@@ -0,0 +1,112 @@
...
+        assert target.shape[-1] == self.model.prediction_length
+
+        distr_args, loc, scale = self.model(
+            context, past_observed_values=past_observed_values
+        )
+        distr = self.model.distr_output.distribution(distr_args, loc, scale)
+
+        return (
+            self.loss(distr, target) * observed_target
+        ).sum() / torch.maximum(torch.tensor(1.0), observed_target.sum())
```
shchur: Nit: `observed_target.sum().clamp_min(1.0)`

src/gluonts/torch/model/d_linear/module.py
```diff
@@ -0,0 +1,155 @@
...
+import torch
+from torch import nn
+
+from gluonts.core.component import validated
+from gluonts.model import Input, InputSpec
+from gluonts.torch.distributions import StudentTOutput
+from gluonts.torch.scaler import StdScaler, MeanScaler, NOPScaler
+
+
+def make_linear_layer(dim_in, dim_out):
```
shchur: Should we reuse the same method from [`simple_feedfoward/model.py`](https://github.com/awslabs/gluonts/blob/dev/src/gluonts/torch/model/simple_feedforward/module.py#L28)?

src/gluonts/torch/model/lag_tst/module.py
```diff
@@ -0,0 +1,342 @@
...
+from gluonts.core.component import validated
+from gluonts.model import Input, InputSpec
+from gluonts.torch.distributions import StudentTOutput
+from gluonts.torch.scaler import StdScaler, MeanScaler, NOPScaler
+from gluonts.torch.util import unsqueeze_expand, lagged_sequence_values
+from gluonts.time_feature import norm_freq_str
+from gluonts.time_feature.lag import _make_lags
+
+
+def get_lags_for_frequency(
```
shchur: Should we use https://github.com/awslabs/gluonts/blob/dev/src/gluonts/time_feature/lag.py#L29 instead?

src/gluonts/torch/model/lag_tst/module.py
```diff
@@ -0,0 +1,342 @@
...
+    # flatten lags list and filter
+    lags = [
+        int(lag) for sub_list in lags for lag in sub_list if 7 < lag <= lag_ub
+    ]
+    lags = [1] + sorted(list(set(lags)))
+
+    return lags[:num_lags]
+
+
+def make_linear_layer(dim_in, dim_out):
```
shchur: Same as above - reuse method across modules?

src/gluonts/torch/model/patch_tst/lightning_module.py
```diff
@@ -0,0 +1,110 @@
...
+
+        assert context.shape[-1] == self.model.context_length
+        assert target.shape[-1] == self.model.prediction_length
+
+        distr_args, loc, scale = self.model(context, past_observed_values)
+        distr = self.model.distr_output.distribution(distr_args, loc, scale)
+
+        return (
+            self.loss(distr, target) * observed_target
+        ).sum() / torch.maximum(torch.tensor(1.0), observed_target.sum())
```
shchur: Nit: `observed_target.sum().clamp_min(1.0)`

src/gluonts/torch/model/patch_tst/module.py
```diff
@@ -0,0 +1,219 @@
...
+from torch import nn
+
+from gluonts.core.component import validated
+from gluonts.model import Input, InputSpec
+from gluonts.torch.distributions import StudentTOutput
+from gluonts.torch.scaler import StdScaler, MeanScaler, NOPScaler
+from gluonts.torch.util import unsqueeze_expand
+
+
+def make_linear_layer(dim_in, dim_out):
```
shchur: Same as above

src/gluonts/torch/model/patch_tst/module.py
```diff
@@ -0,0 +1,219 @@
...
+    lin = nn.Linear(dim_in, dim_out)
+    torch.nn.init.uniform_(lin.weight, -0.07, 0.07)
+    torch.nn.init.zeros_(lin.bias)
+    return lin
+
+
+class SinusoidalPositionalEmbedding(nn.Embedding):
```
shchur, line 28: Should we reuse the code for `SinusoidalPositionalEmbedding` across LagTST and PatchTST?

test/torch/model/test_estimators.py
```diff
@@ -75,6 +76,13 @@
             num_batches_per_epoch=3,
             epochs=2,
         ),
+        lambda dataset: PatchTSTEstimator(
+            prediction_length=dataset.metadata.prediction_length,
+            patch_len=16,
+            batch_size=4,
+            num_batches_per_epoch=3,
+            trainer_kwargs=dict(max_epochs=2),
+        ),
```
shchur, line 109: Would it be possible to add tests for `LagTST` and `DLinear` as well?

shchur (commented):
Thank you for this amazing PR, looking forward to trying out these new models! Added a few minor comments

src/gluonts/torch/model/lag_tst/module.py
kashif: we can but i wanted to avoid the default `[1, 2, 3, 4,..]` lags that are in there

test/torch/model/test_estimators.py
kashif: yes sure let me add some tests for all these models

src/gluonts/torch/model/lag_tst/module.py
shchur: We could add an argument to control this, something like
```python
def get_lags_for_frequency(
    freq_str: str, 
    lag_ub: int = 1200, 
    num_default_lags: int = 7,  # number of lags that are always included
    num_lags: Optional[int] = None,
) -> List[int]:
    ...
    # flatten lags list and filter
    lags = [
        int(lag) for sub_list in lags for lag in sub_list if num_default_lags < lag <= lag_ub
    ]
    lags = list(range(1, num_default_lags + 1)) + sorted(list(set(lags)))
    return lags
```

kashif: hehe that is exactly what i did just now haha but i had:

```py
def get_lags_for_frequency(
    freq_str: str,
    lag_ub: int = 1200,
    num_lags: Optional[int] = None,
    default_lags: List[int] = [1, 2, 3, 4, 5, 6, 7],
) -> List[int]:
```

shchur: I would prefer a tuple or an int here as the default to avoid potential problems with [mutable default arguments](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments). This won't be a problem currently, but could introduce a bug if some change is made to the function in the future.

kashif: sure sure! changing!

shchur (approved):
LGTM!

abdulfatir:
@kashif Can you please also update the list of available models [here](https://github.com/awslabs/gluonts/blob/dev/docs/getting_started/models.md)?

kashif:
opps yes sure!

abdulfatir:
Thank you!

PR #2749, opened 2023-03-23
Add cdf and icdf to torch NegativeBinomial distribution

kashif:
*Description of changes:*

For some reason, the cdf tests are failing.

### DIFF
--- a/src/gluonts/torch/distributions/negative_binomial.py
+++ b/src/gluonts/torch/distributions/negative_binomial.py
@@ -19,7 +19,7 @@ from torch.distributions import Distribution
 from torch.distributions import NegativeBinomial as TorchNegativeBinomial
 
 from gluonts.util import lazy_property
-from scipy.stats import nbinom as ScipyNegativeBinomial
+from scipy.stats import nbinom
 
 
 from .distribution_output import DistributionOutput
@@ -58,7 +58,7 @@ class NegativeBinomial(TorchNegativeBinomial):
 
     @lazy_property
     def scipy_nbinom(self):
-        return ScipyNegativeBinomial(
+        return nbinom(
             n=self.total_count.detach().cpu().numpy(),
             p=1.0 - self.probs.detach().cpu().numpy(),
         )
--- a/src/gluonts/torch/distributions/negative_binomial.py
+++ b/src/gluonts/torch/distributions/negative_binomial.py
@@ -1,0 +1,98 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


kashif:
@lostella would you kindly have a look whenever you get a chance!

src/gluonts/torch/distributions/negative_binomial.py
```diff
@@ -0,0 +1,98 @@
...
+
+from typing import Dict, Optional, Tuple, Union
+
+import torch
+import torch.nn.functional as F
+from torch.distributions import Distribution
+from torch.distributions import NegativeBinomial as TorchNegativeBinomial
+
+from gluonts.util import lazy_property
+from scipy.stats import nbinom as ScipyNegativeBinomial
```
lostella: no need to import `as` I think, either `import scipy` or `from scipy.stats import nbinom` is fine

src/gluonts/torch/distributions/negative_binomial.py
```diff
@@ -0,0 +1,98 @@
...
+        result = self.scipy_nbinom.cdf(value.detach().cpu().numpy())
+        return torch.tensor(result, device=value.device, dtype=value.dtype)
+
+    def icdf(self, value: torch.Tensor) -> torch.Tensor:
+        result = self.scipy_nbinom.ppf(value.detach().cpu().numpy())
+        return torch.tensor(result, device=value.device, dtype=value.dtype)
+
+    @lazy_property
+    def scipy_nbinom(self):
+        return ScipyNegativeBinomial(
```
lostella: For example, `return scipy.stats.nbinom(...)` here reads perfectly fine

test/torch/distribution/test_negative_binomial.py
```diff
@@ -0,0 +1,59 @@
...
+
+@pytest.mark.parametrize("probs", [0.1, 0.5, 0.8])
+@pytest.mark.parametrize("total_count", [3, 7, 100])
+@pytest.mark.parametrize("value", [11, 5, 9])
+def test_custom_neg_bin_cdf(total_count, probs, value):
+    torch_dist = NegativeBinomial(total_count=total_count, probs=probs)
+    scipy_dist = torch_dist.scipy_nbinom
+
+    torch_cdf = torch_dist.cdf(torch.as_tensor(value)).numpy()
+    scipy_cdf = scipy_dist.cdf(np.asarray(value))
```
lostella, line 44: Isn't `torch_dist.cdf` exactly defined as `torch_dist.scipy_nbinom.cdf`? I don't see how the assertion below can fail...

kashif: ah figured out the reason... its the `dtype` conversion of the resulting tensor... since for neg. bin `value` are ints...

lostella (approved):
Thanks @kashif!

PR #2750, opened 2023-03-23
Zebras: documentation and type fixes

abdulfatir:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/zebras/_period.py
+++ b/src/zebras/_period.py
@@ -65,6 +65,10 @@ class _BasePeriod:
     freq: Freq
 
     @property
+    def freqstr(self) -> str:
+        return self.freq.name
+
+    @property
     def year(self) -> np.ndarray:
         return self.data.astype("M8[Y]").astype(int) + 1970
 
@@ -340,7 +344,7 @@ def _encode_zebras_periods(v: Periods):
 
 
 def period(data: Union[Period, str], freq: Optional[str] = None) -> Period:
-    """Creates a zebras.Period object that represents a period of time.
+    """Create a `zebras.Period` object that represents a period of time.
 
     Parameters
     ----------
@@ -352,12 +356,14 @@ def period(data: Union[Period, str], freq: Optional[str] = None) -> Period:
 
     Returns
     -------
-        A zebras.Period object.
+        A `zebras.Period` object.
     """
     if hasattr(data, "freqstr") and freq is None:
         freq = Freq.from_pandas(data.freqstr)
-    else:
+    elif freq:
         freq = Freq.from_pandas(freq)
+    else:
+        raise ValueError("No frequency specified.")
 
     if isinstance(data, Period):
         data = data.data
@@ -378,7 +384,7 @@ def period(data: Union[Period, str], freq: Optional[str] = None) -> Period:
 
 
 def periods(start: Union[Period, str], freq: str, count: int) -> Period:
-    """Creates a zebras.Periods object that represents multiple consecutive
+    """Create a `zebras.Periods` object that represents multiple consecutive
     periods of time.
 
     Parameters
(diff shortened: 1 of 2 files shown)


src/zebras/_period.py
```diff
@@ -339,7 +339,21 @@ def _encode_zebras_periods(v: Periods):
     }
 
 
-def period(data, freq=None) -> Period:
+def period(data: Union[Period, str], freq: Optional[str] = None) -> Period:
```
abdulfatir: @jaheba Still not sure what to do here. Should we import a pandas? Or modify the code to allow using zebras.Period as input too.

jaheba: I think we should allow ``zebras.Period`` as well. Can we check if ``pandas.Period`` has some common base class that we could use here?

Otherwise, we just leave it undocumented for now, I think.

jaheba: I think we should also allow to pass `zebras.Freq` as `freq`.

src/zebras/_period.py
```diff
@@ -339,7 +339,21 @@ def _encode_zebras_periods(v: Periods):
...
+    """Creates a zebras.Period object that represents a period of time.
```
jaheba:
```suggestion
    """Create a zebras.Period object that represents a period of time.
```

src/zebras/_period.py
```diff
@@ -64,6 +64,10 @@ class _BasePeriod:
     data: Any
     freq: Freq
 
+    @property
+    def freqstr(self) -> str:
+        return self.freq.name
```
jaheba:
```suggestion
        return str(self.freq)
```

`.name` does not include the number part, e.g. `zb.freq("3D").name == "D"`

src/zebras/_period.py
```diff
@@ -339,10 +343,29 @@ def _encode_zebras_periods(v: Periods):
...
+    -------
+        A `zebras.Period` object.
+    """
+    if freq is None:
+        if hasattr(data, "freqstr"):
+            freq = Freq.from_pandas(data.freqstr)
+        else:
+            raise ValueError("No frequency specified.")
+    elif isinstance(freq, str):
         freq = Freq.from_pandas(freq)
```
jaheba, line 371: I think we might want to add an explicit test for `Freq` here.

src/zebras/_timeframe.py
```diff
@@ -805,17 +805,56 @@ def with_index(self, index):
...
+    index: Optional[Periods] = None,
+    start: Optional[Union[Period, str]] = None,
+    freq: Optional[str] = None,
+    static: Optional[Dict[str, Any]] = None,
+    tdims: Optional[Dict[str, int]] = None,
+    length: Optional[int] = None,
+    default_tdim: int = -1,
+    metadata: Optional[Dict] = None,
 ):
+    """Create a `zebras.TimeFrame` object that represents one
```
jaheba: We should always use double backticks for sphinx when referencing types, etc:

```suggestion
    """Create a ``zebras.TimeFrame`` object that represents one
```

jaheba (commented):
Thanks for the improvements.

src/zebras/_period.py
abdulfatir: Updated but not sure about the `ValueError`. Is this okay?

jaheba: I think `ValueError` is fine.

jaheba (approved):

abdulfatir:
@jaheba updated a few more things that need your review. Most importantly, added a `length` parameter to `TimeSeries` to be consistent with `TimeFrame` and added post init checks for the same. I will do split frame things in a separate PR.

jaheba (approved):

PR #2752, opened 2023-03-23
Zebras: Add support for weekday-based weekly frequency in Period

shubhamkapoor:
*Issue #, if available:*

*Description of changes:*
Enable support for freq like `W-SUN` where different week start days can be accommodated 

`zb.time_series([1, 2, 3], start="2023-03-20", freq="W-SAT")`
`zb.time_series([1, 2, 3], start="2023-03-20", freq="W-SUN")`

### DIFF
--- a/src/zebras/_period.py
+++ b/src/zebras/_period.py
@@ -45,6 +45,7 @@ from __future__ import annotations
 
 import datetime
 import functools
+import re
 from dataclasses import dataclass
 from typing import Any, Union, cast, overload
 
@@ -341,8 +342,10 @@ def _encode_zebras_periods(v: Periods):
 
 def period(data, freq=None) -> Period:
     if hasattr(data, "freqstr") and freq is None:
+        freqstr = data.freqstr
         freq = Freq.from_pandas(data.freqstr)
     else:
+        freqstr = freq
         freq = Freq.from_pandas(freq)
 
     if isinstance(data, Period):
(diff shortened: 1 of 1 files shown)


src/zebras/_period.py
```diff
@@ -357,11 +360,30 @@ def period(data, freq=None) -> Period:
 
     if freq.name == "W":
         period = Period(np.datetime64(data, freq.np_freq), freq)
-        period.data -= cast(int, period.dayofweek)
+        if re.match(r"W-\w{3}", freqstr):
```
jaheba, line 363: I would prefer not to use `re` for this. Generally doing it in pure python is also faster.

src/zebras/_period.py
```diff
@@ -357,11 +360,30 @@ def period(data, freq=None) -> Period:
...
+        else:
+            weekday_offset = 0
+        period.data -= (cast(int, period.dayofweek) - weekday_offset) % 7
         return period
 
     return Period(np.datetime64(data, freq.np_freq), freq)
 
 
+def get_weekday_offset(freqstr):
+    day_offsets = {
```
jaheba, line 374: I think I would move this outside of the function to increase performance.

jaheba (commented):
Thanks!

I think it would make sense to generally move some of this to `_freq.py`.

jaheba:
Also, tests!

PR #2757, opened 2023-03-24
Zebras: Add support for weekday-based weekly frequency in Period

shubhamkapoor:
*Issue #, if available:*

*Description of changes:*
Enable support for freq like `W-SUN` where different week start days can be accommodated 

`zb.time_series([1, 2, 3], start="2023-03-20", freq="W-SAT")`
`zb.time_series([1, 2, 3], start="2023-03-20", freq="W-SUN")`


src/zebras/_freq.py
```diff
@@ -126,21 +136,39 @@ def __get_validators__(cls):
...
-            else:
-                raise ValueError(f"Invalid freq {freq}: {type(freq)}")
+    def get_weekday_offset(cls, freqstr):
+        if freqstr.upper() == "W":
+            return 0
+
+        assert (
+            len(freqstr) == 5
+            and freqstr[:2] == "W-"
+            and freqstr[2:] in list(day_offsets.keys())
```
jaheba:
```suggestion
            and freqstr[2:] in day_offsets
```

Note: `list` would also make this `O(n)` instead of `O(1)`.

test/zebras/test_period.py
```diff
@@ -115,3 +115,49 @@ def test_periods_feature(freq):
...
+
+@pytest.mark.parametrize(
+    "data, freq, result",
+    [
+        (
+            "2023-03-19",
+            "W",
+            np.array(
+                ["2023-03-13", "2023-03-20", "2023-03-27", "2023-04-03"]
+            ).astype(np.datetime64),
```
jaheba: Can we delay the use of `np.array` to the function body. That would make this easier to read.

test/zebras/test_period.py
```diff
@@ -115,3 +115,49 @@ def test_periods_feature(freq):
...
+            "2023-03-19",
+            "2W-SAT",
+            np.array(
+                ["2023-03-18", "2023-04-01", "2023-04-15", "2023-04-29"]
+            ).astype(np.datetime64),
+        ),
+    ],
+)
+def test_weekly_weekday_period(data, freq, result):
+    np.testing.assert_array_equal(
```
jaheba, line 151: Can we also check that this is the same behaviour as `pandas` here?

On a second thought, probably not! `pandas` looks inconsistent again...

src/zebras/_freq.py
```diff
@@ -126,21 +136,39 @@ def __get_validators__(cls):
         yield freq
 
     @classmethod
```
jaheba, line 152: Let's not make this a classmethod, I think this can be a plain function.

src/zebras/_freq.py
```diff
@@ -126,21 +136,39 @@ def __get_validators__(cls):
...
+    @classmethod
+    def from_pandas(cls, freq) -> Freq:
+        if not isinstance(freq, str):
+            if hasattr(freq, "freqstr"):
+                freq = freq.freqstr
+            else:
+                raise ValueError(f"Invalid freq {freq}: {type(freq)}")
+
+        name, n = cls.freq_name_n(freq)
+        name = name.upper().split("-")[0]
```
jaheba, line 171: Maybe we should add the logic for `W-XXX` here?

src/zebras/_freq.py
shubhamkapoor: I'll move it out of `Freq` class.

src/zebras/_freq.py
shubhamkapoor: I think we don't want to store `W-XXX` in `Freq.name` since it is not recognized by `np`. I thought of keeping this function intact since we want `np` frequencies and just need the day offset to modify `Period.data`.

jaheba: We don't use frequencies of `numpy` but those defined by `pandas`, see:

https://github.com/awslabs/gluonts/blob/04919a894a4750c64031656689ff58b98df8757a/src/zebras/_freq.py#L85-L89

shubhamkapoor: Yes, you're right. Actually, I was referring to `numpy.datetime64` objects which are used internally for eg. in period calculation. But what exactly do you propose to do here? I didn't get it.

test/zebras/test_period.py
jaheba, line 1: Can we add some of the new freqs to line 24:

```py
FREQS = ["S", "min", "T", "H", "D", "B", "W", "M", "QS", "Q", "AS", "A", "Y"]
```

jaheba:
I think something which is missing is that currently the day info is not stored on the `Freq` object. For example:

```py
>>> zb.freq("W")
Freq(name='W', n=1)

>>> zb.freq("W-TUE")
Freq(name='W', n=1)

>>> zb.freq("W") == zb.freq("W-TUE")
True
```

PR #2758, opened 2023-03-24
Add more flexible sampler types through `Range`

lostella:
*Description of changes:* This proposes a new type of samplers, that select (not necessarily) random instances to construct training or validation batches. The main addition to the existing samplers, is that we can configure sampling ranges with timestamps (`pd.Period`) in addition to integer indices: this is obtained through the `Range` class.

This should allow composing batches as required [in this discussion](https://github.com/awslabs/gluonts/discussions/2674). In addition, it improves the code compared to existing samplers, I believe.

In summary, `Range` represents a "partially specified" `range` object, that can only be constructed once we know what sequence we intend to range over (where in time it starts, and how log it is). Once the sequence is known, a regular Python `range` object is constructed, and samplers take (not necessarily) random elements from it according to their own strategy.

Examples:

```python
# consider observations between the 30th-last and the 10th-last, going 4 by 4
rge = Range(-30, -10, step=4)

# let's just get all of them
sample = SampleAll(rge)

sample(start=pd.Period("2023-01-01", freq="D"), length=90)
# [60, 64, 68, 72, 76]
```

```python
# consider observations between these dates
rge = Range(pd.Period("2023-03-10"), pd.Period("2023-03-20"))

# sample one on average
sample = SampleOnAverage(rge, 1)

sample(start=pd.Period("2023-01-01", freq="D"), length=90)
# [68, 74]
# [76]
# [71]
# []
# [70, 75, 76]
# [76]
```

### DIFF
--- a/src/gluonts/transform/sampler.py
+++ b/src/gluonts/transform/sampler.py
@@ -11,12 +11,91 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-from typing import Tuple
+from dataclasses import dataclass
+from typing import Tuple, Optional, Union
 
 import numpy as np
+import pandas as pd
 from pydantic import BaseModel
 
 from gluonts.dataset.stat import ScaleHistogram
+
+
+def clip(value, low, high):
+    """
+    Clip ``value`` between ``low`` and ``high``, included.
+    """
+    return max(low, min(high, value))
+
+
+@dataclass
+class Range:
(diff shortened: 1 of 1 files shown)


jaheba:
I like the idea of splitting (heh) the tasks of defining windows where splits should happen and selecting the concrete instances. Still, I think splitting a dataset into train/val/test might be something different.

Say, I want to incrementally train an existing model using new data. I can define a train/test split first, and then select training windows to be only selected from a recent range.

PR #2762, opened 2023-03-27
Zebras: Split timeframe.py into multiple modules.

jaheba:
*Issue #, if available:*

*Description of changes:*


src/zebras/_split_frame.py
```diff
@@ -0,0 +1,283 @@
...
+            axis=tdim,
+        )
+
+        return _replace(
+            past=merge(self.past, {name: past}),
+            future=merge(self.future, {name: future}),
+            tdims=merge(self.tdims, {name: tdim}),
+        )
+
+    def set_like(self, ref: str, column, value, tdim=None):
```
abdulfatir, line 90: What's the use case of `set_like`?

abdulfatir (approved):

src/zebras/_split_frame.py
jaheba: The idea is to be able to do `set` on both `SplitFrame` and `TimeFrame`. Say I want to add an observed values indicator for a column. In case of `SplitFrame`, we need to know whether that value is `past`, `future` or both. So for `TimeFrame` this method is just like `set`, but for `SplitFrame` that gets translated into either `set_past`, `set_future` or `set`.

PR #2763, opened 2023-03-28
Zebras: Add batching.

jaheba:
Allows batching of `TimeSeries`, `TimeFrame` and `SplitFrame`.

The batched versions stack the underlying data into a single `array`, while other information (such as metadata) is merged into lists.

```py
ts0 = zb.time_series([1, 2, 3])
ts1 = zb.time_series([4, 5, 6])

b = zb.batch([ts0, ts1])
assert b.batch_size == 2

np.testing.assert_array_equal(
    b.values,
    [[1, 2, 3], [4, 5, 6]],
)
```

### DIFF
--- a/src/zebras/__init__.py
+++ b/src/zebras/__init__.py
@@ -18,22 +18,40 @@ __all__ = [
     "Periods",
     "period",
     "periods",
+    "BatchTimeFrame",
     "TimeFrame",
     "time_frame",
+    "BatchSplitFrame",
     "SplitFrame",
     "split_frame",
     "time_series",
+    "BatchTimeSeries",
     "TimeSeries",
 ]
 
+from typing import TypeVar
+
 from ._freq import Freq, freq
 from ._period import period, Period, periods, Periods
-from ._split_frame import split_frame, SplitFrame
-from ._timeframe import time_frame, TimeFrame
-from ._time_series import time_series, TimeSeries
+from ._split_frame import split_frame, SplitFrame, BatchSplitFrame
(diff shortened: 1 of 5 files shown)


src/zebras/__init__.py
```diff
@@ -31,3 +31,9 @@
 from ._split_frame import split_frame, SplitFrame
 from ._timeframe import time_frame, TimeFrame
 from ._time_series import time_series, TimeSeries
+
+
+def batch(xs: list):
+    assert xs
+
+    return xs[0]._batch(xs)
```
abdulfatir: This looks a bit odd. Why don't we make `_batch` a class/static method?

jaheba: This is currently a hacky way to have a `batch` function. `_batch` is in fact a `classmethod`.

jaheba: (Actually, that's what it should be, but it isn't done consistently).

Alternatively we could just have functions for each type.

src/zebras/__init__.py
```diff
@@ -18,16 +18,40 @@
...
-from ._time_series import time_series, TimeSeries
+from ._split_frame import split_frame, SplitFrame, BatchSplitFrame
+from ._timeframe import time_frame, TimeFrame, BatchTimeFrame
+from ._time_series import time_series, TimeSeries, BatchTimeSeries
+
+
+Batchable = TypeVar("Batchable", TimeSeries, TimeFrame, SplitFrame)
+
+
+def batch(xs: list):
```
abdulfatir, line 44: Looks much better now.

abdulfatir (approved):

PR #2764, opened 2023-03-28
Add Fourier option for local arima

leica2023:
*Issue #, if available:*

*Description of changes:*


src/gluonts/evaluation/_base.py
jaheba, line 1: We are planning to deprecate the `evaluation` module in favour of `ev`.

One reason is that currently too many things are fixed, including using the median prediction for `MAPE`, etc. One thing we do in `ev` is to make metrics parametric over their inputs, so you can choose either `mean` or `median`:

https://github.com/awslabs/gluonts/blob/a973a04d1681dcb3074dabb559a002ed0333d2f2/src/gluonts/ev/metrics.py#L121-L125

However, we haven't integrated this in the rest of the code base yet. We should work on that.

src/gluonts/model/forecast.py
```diff
@@ -442,6 +442,7 @@ def __init__(
         self,
         samples: np.ndarray,
         start_date: pd.Period,
+        quantiles: Optional[Dict] = None,
```
jaheba: We have a dedicated classed `QuantileForecast`. Do we want to support cases where we have both samples and quantiles?

/cc @lostella

src/gluonts/ext/r_forecast/R/univariate_forecast_methods.R
jaheba, line 1: I don't know much about these models, maybe @lovvge can take a look?

lostella:
@leica2023 could you give this a proper title and a short description? Thanks!

test/ext/r_forecast/test_benchmarking_data.py
```diff
@@ -0,0 +1,271 @@
...
+import sys
+import time
+import rpy2
+import numpy as np
+import boto3
+import yaml
+import argparse
+import pickle
+
+sys.path[0] = '/Users/linmen/Desktop/gluonts/src'
```
lostella, line 16: Is this intended?

src/gluonts/evaluation/_base.py
```diff
@@ -298,7 +298,7 @@ def __call__(
                 if col not in ["item_id", "forecast_start"]
             }
         )
-
+        # metrics_per_ts = metrics_per_ts[metrics_per_ts.seasonal_error > 0]
```
lostella, line 301: This doesn't need to be there, right?

src/gluonts/evaluation/_base.py
```diff
@@ -399,13 +399,13 @@ def get_metrics_per_ts(
...
+            "abs_error": abs_error(pred_target, mean_fcst),
             "abs_target_sum": abs_target_sum(pred_target),
             "abs_target_mean": abs_target_mean(pred_target),
             "seasonal_error": seasonal_error,
-            "MASE": mase(pred_target, median_fcst, seasonal_error),
-            "MAPE": mape(pred_target, median_fcst),
-            "sMAPE": smape(pred_target, median_fcst),
+            "MASE": mase(pred_target, mean_fcst, seasonal_error),
+            "MAPE": mape(pred_target, mean_fcst),
+            "sMAPE": smape(pred_target, mean_fcst),
```
lostella, line 408: I think we can split changes into multiple pull requests. Maybe you can keep this one about adding the option for Fourier features?

src/gluonts/model/forecast.py
```diff
@@ -493,6 +493,7 @@ def mean(self) -> np.ndarray:
             self._mean = np.mean(self.samples, axis=0)
         return self._mean
 
+
```
lostella, line 496:
```suggestion
```

src/gluonts/model/forecast.py
```diff
@@ -501,6 +502,7 @@ def mean_ts(self) -> pd.Series:
         return pd.Series(self.mean, index=self.index)
 
     def quantile(self, q: Union[float, str]) -> np.ndarray:
+
```
lostella, line 505:
```suggestion
```

src/gluonts/ext/r_forecast/_predictor.py
```diff
@@ -318,14 +324,37 @@ def predict(
...
+                xreg_full.append(np.sin(2 * np.pi * k / period * np.arange(1, len_ts + self.prediction_length + 1)))
+                xreg_full.append(np.cos(2 * np.pi * k / period * np.arange(1, len_ts + self.prediction_length + 1)))
+            xreg_full = np.transpose(np.array(xreg_full, dtype=np.float32))
+            xreg = xreg_full[:-self.prediction_length, :]
+            xreg_future = xreg_full[-self.prediction_length:, :]
+
+        if self.method_name == 'arima' and fourier:
+            self.params["seasonal"] = False
+            self.params["xreg"] = xreg
+            self.params["xreg_future"] = xreg_future
```
lostella, line 348: I think this should be a separate function, so that we can test it and benchmark it. Also, features should be computed for each dataset entry independently, since they may have different length/frequency

PR #2767, opened 2023-03-31
Zebras: Add Schema.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/zebras/_schema.py
+++ b/src/zebras/_schema.py
@@ -13,15 +13,39 @@
 
 
 from dataclasses import dataclass
-from typing import Any, Callable, Optional, Type
+from typing import Any, Callable, Optional, Union, Type, Dict
 
 import numpy as np
 from toolz import valfilter
 
 from gluonts.itertools import partition
 
-from ._timeframe import time_frame
-from ._split_frame import split_frame
+from ._freq import Freq
+from ._period import Period
+from ._timeframe import time_frame, TimeFrame
+from ._split_frame import split_frame, SplitFrame
+
+"""
+This module provides tooling to extract ``zebras.TimeFrame`` and
+``zebras.SplitFrame`` instances from Python dictionaries::
+
+    schema = zebras.Schema(...)
+    tf = schema.load_timeframe(raw_dict)
+
+    sf = schema.load_splitframe(raw_dict)
+
+The idea is to define expected types and shapes for each field (column) using
+``zebras.Field`` and then wrap them in a ``zebras.Schema``:
+
+    schema = zebras.Schema({
+        "target": Field(ndim=1, tdim=0, past_only=True),
+        "time_feat": zebras.Field(ndim=2, tdim=-1),
+        "static_feat": zebras.Field(ndim=1),
+
+    })
+
+See the documentation of ``Field`` on how to configure the columns of a schema.
+"""
 
 
 @dataclass
@@ -45,8 +69,8 @@ class Field:
         the past range when loading ``zebras.SplitFrame``. The value is ignored
         for static fields.
     required
-        When set, the field has to be in the user data. Otherwise ``default``
-        is used as a fallback value.
+        When set to true, the field has to be in the user data. Otherwise
+        ``default`` is used as a fallback value.
     internal
         Allows to ignore user provided data when set, and instead ``default``
         is always used as the value.
@@ -55,7 +79,7 @@ class Field:
         set to true.
     preprocess, optional
         This function is called on the value before validating the value. For
-        example, one can set ``preprocess = np.astleast_2d`` to also allow
+        example, one can set ``preprocess = np.atleast_2d`` to also allow
         one dimensional arrays as input even when ``ndim = 2``.
     """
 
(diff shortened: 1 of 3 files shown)


src/zebras/_schema.py
```diff
@@ -0,0 +1,151 @@
...
+    ndim: Optional[int] = None
+    tdim: Optional[int] = None
+    dtype: Type = np.float32
+    past_only: bool = False
+    required: bool = True
+    internal: bool = False
+    default: Any = None
+    preprocess: Optional[Callable] = None
+
+    def validate(self, value):
```
melopeo: Can we add the corresonding `typing` entry for `value`? Is `value` a list, a numpy array, or alike?

src/zebras/_schema.py
```diff
@@ -0,0 +1,151 @@
...
+        value = np.array(value, dtype=self.dtype)
+
+        if self.ndim is not None:
+            assert value.ndim == self.ndim
```
melopeo: Can we add a string to the assert?

src/zebras/_schema.py
```diff
@@ -0,0 +1,151 @@
...
+        dimension. When this value is ``None`` the array is classified as
+        ``"static``".
+    dtype
+        The data type, passed to ``numpy.array``.
+    past_only
+        If the value is a time series, this marks if data is only expected for
+        the past range when loading ``zebras.SplitFrame``. The value is ignored
+        for static fields.
+    required
+        When set, the field has to be in the user data. Otherwise ``default``
```
melopeo: When set to `True`, the field has...

melopeo: User data here means the dictionary provided by the user? For instance, the key (here called _field_) has to be in the dictionary provided by the user (here called _user data_)

src/zebras/_schema.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def validate(self, value):
+        value = np.array(value, dtype=self.dtype)
+
+        if self.ndim is not None:
+            assert value.ndim == self.ndim
+
+        return value
+
+    def load_from(self, dct, name):
```
melopeo: I guess `dct` is a dictionary, and `name` is a string. Is this correct? Can we add the corresponding types?

src/zebras/_schema.py
```diff
@@ -0,0 +1,151 @@
...
+    past_only: bool = False
+    required: bool = True
+    internal: bool = False
+    default: Any = None
+    preprocess: Optional[Callable] = None
+
+    def validate(self, value):
+        value = np.array(value, dtype=self.dtype)
+
+        if self.ndim is not None:
```
melopeo, line 98: If `ndim` is not provided there there is basically no validation done?

src/zebras/_schema.py
jaheba: Yes, I think I should have some more general description of this module as well :)

src/zebras/_schema.py
jaheba: Precisely, that should be explained in the docstring.

src/zebras/_schema.py
```diff
@@ -0,0 +1,151 @@
...
+        is used as a fallback value.
+    internal
+        Allows to ignore user provided data when set, and instead ``default``
+        is always used as the value.
+    default
+        The default value to use when either ``required`` or ``internal`` is
+        set to true.
+    preprocess, optional
+        This function is called on the value before validating the value. For
+        example, one can set ``preprocess = np.astleast_2d`` to also allow
```
abdulfatir: `np.atleast_2d`.

test/zebras/test_schema.py
```diff
@@ -0,0 +1,49 @@
...
+    {
+        "target": zb.Field(ndim=1, tdim=0, past_only=True),
+        "time_feat": zb.Field(
+            ndim=2, tdim=-1, past_only=True, preprocess=np.atleast_2d
+        ),
+        "static": zb.Field(ndim=1, preprocess=np.atleast_1d),
+    }
+)
+
+
```
abdulfatir, line 29: Tests for `load_splitframe`?

jaheba: Coming! Note, that this PR is still in draft mode :)

test/zebras/test_schema.py
```diff
@@ -0,0 +1,49 @@
...
+def test_schema_timeframe():
+    xs = [1, 2, 3, 4, 5]
+    row = {"target": xs, "time_feat": xs, "static": 1}
+    schema.load_timeframe(row)
+
+    row = {"target": xs, "time_feat": [xs, xs], "static": [1]}
+    schema.load_timeframe(row)
+
```
melopeo, line 37: Why do we need "static" to be a scalar or a list in each of these two cases?

melopeo: Can we add a test showing how we can add categorical/static features?

jaheba: Not sure I understand what you are asking for. The test includes static data.

melopeo: Yes, you are right. I missed it. Could it help to change `static` to `static_feat`? Maybe that way it would be more homogeneous with the name `time_feat`.

src/zebras/_schema.py
```diff
@@ -0,0 +1,202 @@
...
+
+Fields = Dict[str, Field]
+
+
+@dataclass
+class Schema:
+    fields: Fields
+
+    @property
+    def static(self) -> Fields:
```
melopeo, line 140: Would it make sense to rename this one as `return_static_fields` or something similar?

jaheba: No?

PR #2776, opened 2023-04-03
Merge zebras from proof of concept branch.

jaheba:
*Issue #, if available:*

*Description of changes:*


test/zebras/test_schema.py
```diff
@@ -0,0 +1,68 @@
...
+    row = {"target": xs, "time_feat": xs}
+
+    with pytest.raises(Exception):
+        schema.load_timeframe({})
+
+    with pytest.raises(Exception):
+        schema.load_timeframe(row)
+
+
+def test_schema_slitframe():
```
lostella:
```suggestion
def test_schema_splitframe():
```

src/gluonts/zebras/_schema.py
```diff
@@ -0,0 +1,202 @@
...
+@dataclass
+class Field:
+    """
+    Specification for user provided input data.
+
+    Parameters
+    ----------
+    ndim, optional
+        The expected number of dimensions of the input data. If provided, it is
+        ensured that the input array has the expected number of dimensions.
```
lostella, line 60: What happens if it's not provided? Is this validation just skipped, or is the argument instead assumed to be a scalar (non-array)?

test/zebras/test_schema.py
```diff
@@ -0,0 +1,68 @@
...
+)
+
+
+def test_schema_timeframe():
+    xs = [1, 2, 3, 4, 5]
+    row = {"target": xs, "time_feat": xs, "static_feat": 1}
+    schema.load_timeframe(row)
+
+    row = {"target": xs, "time_feat": [xs, xs], "static_feat": [1]}
+    schema.load_timeframe(row)
```
lostella, line 36: Any meaningful assertion that could be added here, other than just loading the data?

test/zebras/test_schema.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def test_schema_slitframe():
+    short = [1, 2, 3, 4, 5]
+    long = [1, 2, 3, 4, 5, 6, 7, 8]
+    row = {"target": short, "time_feat": long, "static_feat": 1}
+    schema.load_splitframe(row)
+
+    row = {"target": short, "time_feat": [long, long], "static_feat": [1]}
+    schema.load_splitframe(row)
```
lostella, line 57: I guess here one could check the size of past and future, and that `short` is not there in the future, or things like that.

test/zebras/test_schema.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def test_schema_splitframe_error():
```
lostella, line 60: Nitpick: this seems to be testing both `load_splitframe` and `load_timeframe`

test/zebras/test_schema.py
```diff
@@ -0,0 +1,68 @@
...
+    xs = [1, 2, 3, 4, 5]
+
+    with pytest.raises(Exception):
+        schema.load_splitframe({})
```
lostella, line 64: I guess another error condition should raise from the schema having multiple `past_only=False` and `past_only=True` fields, but they appear in the data with different sizes, so that data is not aligned.

My understanding is that `split_frame` infers past vs future length from the first column in the data, so I guess when not all columns have that size probably an error should be raised somewhere?

src/gluonts/zebras/_split_frame.py
```diff
@@ -0,0 +1,387 @@
...
+    past_length=None,
+    future_length=None,
+    metadata=None,
+    static=None,
+    start=None,
+    freq=None,
+    index=None,
+    tdims=None,
+    default_tdim=-1,
+):
```
lostella: This appears to be a fundamental entry point for the package, maybe add a docstring?

src/gluonts/zebras/_schema.py
jaheba: Just skipped.

test/zebras/test_schema.py
jaheba: The error is raised in `SplitFrame` itself, where we check for consistency.

lostella (dismissed):

huibinshen (approved):

PR #2777, opened 2023-04-03
Fix dataset file discovery.

jaheba:
*Issue #, if available:*

*Description of changes:*
This PR allows for the discovery of datasets that have a nested directory structure. Before this PR, dataset files would not be identified, and it would basically return an empty dataset.

### DIFF
--- a/src/gluonts/dataset/common.py
+++ b/src/gluonts/dataset/common.py
@@ -129,16 +129,18 @@ def infer_file_type(path):
     return None
 
 
-def _glob(path: Path, pattern="*", levels=1):
+def _glob(path: Path, pattern="*", levels=None):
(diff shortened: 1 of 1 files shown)


src/gluonts/dataset/common.py
```diff
@@ -129,16 +129,18 @@ def infer_file_type(path):
     return None
 
 
-def _glob(path: Path, pattern="*", levels=1):
+def _glob(path: Path, pattern="*", levels=None):
```
jaheba, line 132: Should we limit the number of `levels` by default?

If misused, this can be become a performance issues. Maybe we set this to a lower number and set it to `None` in `shell`?

melopeo (approved):

PR #2779, opened 2023-04-04
Fix: Loading of nested paths in FileDataset.

jaheba:
Before this change `levels` didn't do anything, since we iterated subfolders in `FileDataset` as well. Now, `_rglob` does path discovery and `FileDataset` applies the right loader.

This also removes the `ignore` flag from `FileDataset`, which wasn't really used before.

---


lostella (approved):
Looks good to me. Label as bugfix, backport?

Issue #2790, opened 2023-04-07
SimpleFeedForward crashes due to numerical errors on datasets with large fraction of zeros


shchur:
## Description
PyTorch version of SimpleFeedForward sometimes crashes because of numerical problems StudentT output. Similar to #833.

## To Reproduce

```python
from gluonts.dataset.repository.datasets import get_dataset
from gluonts.torch.model.simple_feedforward import SimpleFeedForwardEstimator
from gluonts.dataset.split import split
import pytorch_lightning as pl

pl.seed_everything(2)

data = get_dataset("covid_deaths")
train, val = split(data.train, offset=-data.metadata.prediction_length)
estimator = SimpleFeedForwardEstimator(
    prediction_length=data.metadata.prediction_length,
    trainer_kwargs={
        "accelerator": "cpu",
        "devices": 1,
    }
)
estimator.train(train)
```

## Error message or code output
```
ValueError: Expected parameter df (Tensor of shape (32, 30)) of distribution Chi2() to satisfy the constraint GreaterThan(lower_bound=0.0), but found invalid values:
tensor([[nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan],
....
```

## Environment
- Operating system: AL2
- Python version: 3.9
- GluonTS version: 0.12.5
- PyTorch version: 1.13.1

shchur:
Should be fixed by #2791.

PR #2791, opened 2023-04-07
Avoid zero scale in StudentTOutput

shchur:
*Issue #, if available:* #2790

*Description of changes:*
- Avoid zero scale in `StudentTOutput `. Similar to the fix in #1893.

### DIFF
--- a/src/gluonts/torch/distributions/studentT.py
+++ b/src/gluonts/torch/distributions/studentT.py
@@ -68,7 +68,7 @@ class StudentTOutput(DistributionOutput):
     def domain_map(
         cls, df: torch.Tensor, loc: torch.Tensor, scale: torch.Tensor
     ):
-        epsilon = np.finfo(cls._dtype).eps  # machine epsilon
+        epsilon = torch.finfo(scale.dtype).eps
         scale = F.softplus(scale).clamp_min(epsilon)
         df = 2.0 + F.softplus(df)
         return df.squeeze(-1), loc.squeeze(-1), scale.squeeze(-1)
--- a/src/gluonts/torch/distributions/studentT.py
+++ b/src/gluonts/torch/distributions/studentT.py
@@ -13,6 +13,7 @@
 
 from typing import Dict, Tuple, Union
 
+import numpy as np
 import torch
 import torch.nn.functional as F
 from scipy.stats import t as ScipyStudentT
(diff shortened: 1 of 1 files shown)


kashif:
I think we should also move from softplus to squareplus but yes this is a helpful fix!

src/gluonts/torch/distributions/studentT.py
```diff
@@ -67,7 +68,8 @@ class StudentTOutput(DistributionOutput):
     def domain_map(
         cls, df: torch.Tensor, loc: torch.Tensor, scale: torch.Tensor
     ):
-        scale = F.softplus(scale)
+        epsilon = np.finfo(cls._dtype).eps  # machine epsilon
```
kashif:
```suggestion
        epsilon = torch.finfo(scale.dtype).eps
```

shchur: Nice, thank you!

src/gluonts/torch/distributions/studentT.py
```diff
@@ -13,6 +13,7 @@
 
 from typing import Dict, Tuple, Union
 
+import numpy as np
```
jaheba: Looks unused.

jaheba:
Is there a way we can test this?

src/gluonts/torch/distributions/studentT.py
shchur: Thanks for catching this

shchur:
@jaheba The logic here looks very similar to other `DistributionOutput` classes (e.g., [Beta](https://github.com/awslabs/gluonts/blob/dev/src/gluonts/torch/distributions/distribution_output.py#L193) and [Gamma](https://github.com/awslabs/gluonts/blob/dev/src/gluonts/torch/distributions/distribution_output.py#L213)). If there are existing tests that cover these, we could extend them to StudentT.

The only thing that comes to my mind would be training a network with respective model types on data consisting of all zeros and seeing if it eventually runs into NaNs. However, even this is not guaranteed and may take a a non-deterministic number of epochs.

src/gluonts/torch/distributions/studentT.py
jaheba: `ruff` did, I was just pointing it out!

jaheba (approved):

jaheba:
I see, I guess similar to the distribution tests we have for mxnet, the test would add little value, is and just slows down the test suite.

PR #2799, opened 2023-04-13
Zebras: time length fixes.

jaheba:
* Rework `split_frame` length calculation.
* Fix missing handling of tdims.
* Add consistency checks for `SplitFrame`
* Allow integer index access on TimeSeries


src/gluonts/zebras/_split_frame.py
```diff
@@ -42,6 +43,14 @@ class SplitFrame:
     default_tdim: int = -1
     _pad: Pad = Pad()
 
+    def __post_init__(self):
+        for column in itertools.chain(self._past, self._future):
+            self.tdims.setdefault(column, self.default_tdim)
+
+        # this triggers checks for past_length and future_length
+        self.past
+        self.future
```
abdulfatir: I don't know if this is how it's generally done. Would something like this be more readable:
```py
_, _ = self.past, self.future
```

src/gluonts/zebras/_split_frame.py
```diff
@@ -357,47 +366,55 @@ def split_frame(
...
+                "Cannot determine past and future length if only "
+                "`full` is provided."
+            )
+    # No data and no lengths are passed
+    elif maybe.or_(past_length, future_length) is None:
+        past_length = 0
+        future_length = 0
+        full_length = 0
+    else:
+        full_length = None
```
abdulfatir: What is this case?

src/gluonts/zebras/_split_frame.py
jaheba: Maybe, I'm not sure. But let's do this.

src/gluonts/zebras/_split_frame.py
jaheba: Let me add a comment.

abdulfatir (approved):

Issue #2795, opened 2023-04-10
Plotting intervals do not work


abdulfatir:
Tried to go through [Tutorial](https://ts.gluon.ai/stable/tutorials/forecasting/quick_start_tutorial.html#Quick-Start-Tutorial), just copy/pasted code, plot function for SampleForecast entity  doesn't seems to work the way it should. 

![image](https://user-images.githubusercontent.com/59028817/230828945-7dab5775-7fa4-4f70-a663-2466a4eb7b6a.png)

_Originally posted by @ivanvoid in https://github.com/awslabs/gluonts/issues/2333#issuecomment-1501392375_

abdulfatir:
Hi @ivanvoid,

I created a new issue from your comment. Let's discuss this here. 

The plotting issue you mentioned happens sometimes with me as well. As far as I understand, it's due to conflicts between how pandas plotting interacts with `plt.fill_between` when using a datetime index on the x-axis. Can you share your `gluonts`, `matplotlib` (and `seaborn`, if installed), and `pandas` versions?

As a workaround, you can go to [this function](https://github.com/awslabs/gluonts/blob/160a127f64002322586359e2729c7b586acd15fb/src/gluonts/model/forecast.py#L284) in your GluonTS installation and replace all instances of `self.index.to_timestamp()` in the function to `self.index`. This has worked for me in the past and we can merge this as a bug fix.

jaheba:
In the future we ideally would not want to rely on `pandas`.

Another thing to do right with Zebras!

abdulfatir:
Fixed by #2800

Issue #2205, opened 2022-08-09
Median prediction is plotted one index shift to the left


zhichenggeng:
## Description
The plot of median prediction is shifted to the left by one index, while 90% and 50% prediction interval plots are correct. Potential reason is that the data is resampled by week, which might cause some problem on the index.

## To Reproduce

```python
import numpy as np
import pandas as pd
import matplotlib.pyplot as plt


def generate_single_ts(date_range, item_id=None) -> pd.DataFrame:
    """create sum of `n_f` sin/cos curves with random scale and phase."""
    n_f = 2
    period = np.array([24 / (i + 1) for i in range(n_f)]).reshape(1, n_f)
    scale = np.random.normal(1, 0.3, size=(1, n_f))
    phase = 2 * np.pi * np.random.uniform(size=(1, n_f))
    periodic_f = lambda x: scale * np.sin(np.pi * x / period + phase)

    t = np.arange(0, len(date_range)).reshape(-1, 1)
    target = periodic_f(t).sum(axis=1) + np.random.normal(0, 0.1, size=len(t))
    ts = pd.DataFrame({"target": target}, index=date_range)
    if item_id is not None:
        ts["item_id"] = item_id
    return ts

prediction_length, freq = 2, "1D"
T = 365 * prediction_length
date_range = pd.date_range("2021-01-01", periods=T, freq=freq)
ts = generate_single_ts(date_range)

ts = ts.resample('W', label='left').sum()

from gluonts.model.deepar import DeepAREstimator
from gluonts.mx import Trainer
from gluonts.evaluation import make_evaluation_predictions, Evaluator

prediction_length, freq = 4, "W"
estimator = DeepAREstimator(
    freq=freq, prediction_length=prediction_length, trainer=Trainer(epochs=1)
)

from gluonts.dataset.pandas import PandasDataset

train = PandasDataset(ts[:-prediction_length], target="target", freq=freq)
test = PandasDataset(ts, target="target", freq=freq)

predictor = estimator.train(train)
forecast_it, ts_it = make_evaluation_predictions(dataset=test, predictor=predictor)

forecasts = list(forecast_it)
tss = list(ts_it)
forecast_entry = forecasts[0]
ts_entry = tss[0]

fig, ax = plt.subplots(1, 1, figsize=(20, 8))
ts_entry[-prediction_length * 4:].plot(ax=ax)
forecast_entry.plot(prediction_intervals=(50, 90), color="g")
```

## Environment
[...]

PR #2800, opened 2023-04-13
Remove `.to_timestamp()` to fix interval plotting

abdulfatir:
*Issue #, if available:* #2795

*Description of changes:*
Plotting intervals sometimes do not work due to how pandas plotting interacts with `plt.fill_between`. Using a period index appears to solve the issue, in my experience. 

### DIFF
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -338,12 +338,12 @@ class Forecast:
         i_p50 = len(percentiles_sorted) // 2
 
         p50_data = ps_data[i_p50]
-        p50_series = pd.Series(data=p50_data, index=self.index.to_timestamp())
+        p50_series = pd.Series(data=p50_data, index=self.index)
         p50_series.plot(color=color, ls="-", label=f"{label_prefix}median")
 
         if show_mean:
             mean_data = np.mean(self._sorted_samples, axis=0)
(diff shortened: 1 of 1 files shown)


jaheba:
Can you check that this looks good using different frequencies? I think I had some problems using `PeriodIndex`.

abdulfatir:
Tested with different frequencies using this script. Works for all.

```py
import numpy as np
import pandas as pd

from gluonts.model.forecast import SampleForecast

freq = ["H", "2H", "T", "2T", "D", "2D", "B", "W", "M", "Y"]

for f in freq:
    print(f)
    sf = SampleForecast(
        samples=np.random.randn(100, 24), start_date=pd.Period("2023", f)
    )
    sf.plot()
    plt.show()

```

jaheba:
Should we backport these fixes? @lostella

abdulfatir:
@jaheba Do you remember the problems you faced with `PeriodIndex` previously? @rshyamsundar was wondering if we can add a test for this somehow.

jaheba:
I think the plot didn't show up right. Not sure how we would be able to test this.

jaheba (approved):

lostella:
@abdulfatir @jaheba issue #2205 is related to `PeriodIndex` vs `DatetimeIndex`, there's some context there

PR #2802, opened 2023-04-17
Update `setuptools` and `wheel` in test workflow

lostella:
*Issue #, if available:* Some test workflow was failing when installing dependencies. See eg https://github.com/awslabs/gluonts/actions/runs/4718825468/jobs/8369409495

*Description of changes:* Update `setuptools` and `wheel` as part of the workflow


lostella:
@jaheba I think it makes sense to do this to other workflows as well?

jaheba (approved):

PR #2814, opened 2023-04-20
Add test workflow for R based models

lostella:
*Issue #, if available:* The tests for R model wrappers do not run on GitHub

*Description of changes:* Add test workflow for wrapped R models. There appears to be an issue with installing the `hts` package in R, so the workflow currently excludes the tests for hierarchical models.

### DIFF
--- a/.github/workflows/tests-r.yml
+++ b/.github/workflows/tests-r.yml
@@ -1,4 +1,4 @@
-name: Tests
+name: R Tests
 
 on: [push, pull_request]
 
--- a/.github/workflows/tests-r.yml
+++ b/.github/workflows/tests-r.yml
@@ -31,7 +31,7 @@ jobs:
         pip install -r requirements/requirements-extras-r.txt
     - name: Install R dependencies
       run: |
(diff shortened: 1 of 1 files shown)


melopeo (approved):
LGTM

.github/workflows/tests-r.yml
```diff
@@ -0,0 +1,47 @@
...
+    strategy:
+      max-parallel: 4
+      fail-fast: false
+      matrix:
+        python-version: ['3.9']
+        platform: [ubuntu-latest]
+
+    runs-on: ${{ matrix.platform }}
+
+    if: contains(github.event.pull_request.labels.*.name, 'pr:test-r')
```
abdulfatir, line 16: This makes them opt-in right?

lostella: Yes, needs to be labeled as `pr:test-r` (we have a similar `pr:test-nursery` label for the nursery tests). I checked that without adding the label the workflow indeed skips

PR #2830, opened 2023-05-04
Simplify univariate R wrapper

lostella:
*Description of changes:*
- Use `QuantileForecast` for all methods, whether they give probabilistic or point predictions. In particular, `arima` and `ets` now do not rely on sampling, but on direct mean and quantile computation from the R methods.
- Simplify the R handler functions following the previous point.
- Remove additional arguments from `.predict`: options are set at model construction time, and that's it.
- Remove `_override_params` and simplify `_get_r_forecast`, `_forecast_dict_to_obj` following the previous points.
- Move `unlist` as well as interval-to-quantile level converters into the new `util` module.
- Clean up test a bit.
- Add check to test, for consistency of quantiles (no crossing).


src/gluonts/ext/r_forecast/_univariate_predictor.py
```diff
@@ -95,79 +99,59 @@ def __init__(
...
-            "output_types": ["samples"],
             "frequency": self.period,
         }
 
-        if params is not None:
-            self.params.update(params)
+        if self.method_name in UNIVARIATE_POINT_FORECAST_METHODS:
+            self.params["output_types"] = ["mean"]
+        elif self.method_name in UNIVARIATE_QUANTILE_FORECAST_METHODS:
+            self.params["output_types"] = ["quantiles", "mean"]
```
rshyamsundar: Quick question here about whether you always get "mean" forecasts: if I sent `params["output_types"] = ["quantiles"]` then `self.params` would be updated with that (line 125) and you don't get mean forecasts from the underlying method. But you are expecting the "mean" key in `forecast_dict` in the first line of function `_forecast_dict_to_obj`?

src/gluonts/ext/r_forecast/_predictor.py
```diff
@@ -291,54 +242,23 @@ def _forecast_dict_to_obj(
...
-
-        """
         self._warning_message()
 
         for data in dataset:
             data = self._preprocess_data(data=data)
-            params = self._override_params(
-                params=self.params.copy(),
-                num_samples=num_samples,
-                intervals=intervals,
```
rshyamsundar, line 326: I think the reason for using separate set of `params` for each time series instead of using `self.params` is being able to run this in parallel. Not sure if it is needed though.

lostella: That should not be an issue in case the process is forked, right @jaheba? Most importantly, if the model does not modify `self.params` on a per-series basis, then it's essentially a read-only dictionary and should be fine.

But let me verify these

src/gluonts/ext/r_forecast/_univariate_predictor.py
lostella: good point :-) let me fix it

lostella: Fixed

src/gluonts/ext/r_forecast/_univariate_predictor.py
```diff
@@ -95,79 +99,59 @@ def __init__(
...
-                    lower_quantiles + upper_quantiles[1:],
+                    lower_quantiles + upper_quantiles,
                     forecast_dict["lower_quantiles"]
-                    + forecast_dict["upper_quantiles"][1:],
+                    + forecast_dict["upper_quantiles"],
                 )
             )
 
-            # `QuantileForecast` allows "mean" as the key;
-            # we store them as well since they can differ from median.
```
rshyamsundar, line 166: The reason I had this here is to avoid returning a wrong value without warning: the user sends only `"quantiles"` in `output_types` and obtains a `QuantileForecast` which in this case returns `p50` when requested for `mean`. So if the user wants to evaluate this forecast using a point metric and requests for `mean` (because it supports mean), then `QuantileForecast` would return a wrong value without raising any warning. 

So it is better to always include mean while creating `QuantileForecast`.

lostella: You mean because of [this](https://github.com/awslabs/gluonts/blob/dev/src/gluonts/model/forecast.py#L677-L685)? I think we should also break that, if one asks for the `mean` they should either get the `mean` or an exception

lostella: Ok now I see what you mean

rshyamsundar (approved):

PR #2831, opened 2023-05-05
Move notebook compilation logic to docs workflow

lostella:
*Description of changes:* Moving the logic from the `md2ipynb.py` script to the docs workflow config for tighter control.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -39,14 +39,17 @@ jobs:
       run: |
         sudo apt-get install pandoc
 
-    - name: Compile notebooks
-      if: contains(github.event.pull_request.labels.*.name, 'pr:docs-build-notebook')
+    - name: Set build mode
       run: |
-        just compile_notebooks
+        if [ ${{ contains(github.event.pull_request.labels.*.name, 'pr:docs-build-notebook') }} ]; then
+          MODE=skip
+        else
+          MODE=release
+        fi
+      run: echo "Running in '$MODE' mode"
 
     - name: Build docs
-      run: |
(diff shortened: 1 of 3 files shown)


docs/md2ipynb.py
```diff
@@ -108,8 +89,6 @@ def convert(path, mode, kernel_name=None, timeout=40 * 60):
 @click.option("--kernel", "-k", help="Name of iPython kernel to use.")
 @click.option("--mode", "-m", default="release")
 def cli(paths, kernel, mode):
-    mode = check_github_event(mode)
```
jaheba, line 111: I think we always need to run this script, either with `--mode release` or `--mode skip`.

In the latter case we rename the `.template` files to just `.md` or `.ipynb`.

.github/workflows/docs.yml
```diff
@@ -39,9 +39,17 @@ jobs:
...
-    - name: Build docs
+    - name: Set build mode
       run: |
-        just docs
+        if [ ${{ contains(github.event.pull_request.labels.*.name, 'pr:docs-build-notebook') }} ]; then
+          DOCS_BUILD_MODE=skip
+        else
+          DOCS_BUILD_MODE=release
+        fi
+        echo "Running in '$DOCS_BUILD_MODE' mode"
```
lostella: Keeping this as a separate step so that it's easy to look into the step's output for debugging purposes: should anything change in the GH workflow logic and this step break, we can see it immediately in the logs, instead of having it hidden among the thousands of lines from the actual build step

jaheba (approved):

PR #2841, opened 2023-05-08
Fix torch DeepAREstimator in case `context_length=1`

### LINKED ISSUE #2838: `gluonts.torch.DeepAREstimator` cannot be trained with `context_length=1`
This:

```py
from gluonts.torch import DeepAREstimator

DeepAREstimator(prediction_length=1, freq="M", context_length=1).train(data)
```

fails with:

```
File ~/code/gluon-ts/src/gluonts/torch/model/deepar/module.py:255, in DeepARModel.prepare_rnn_input(self, feat_static_cat, feat_static_real, past_time_feat, past_target, past_observed_values, future_time_feat, future_target)
    249 expanded_static_feat = unsqueeze_expand(
    250     static_feat, dim=-2, size=time_feat.shape[-2]
    251 )
    253 features = torch.cat((expanded_static_feat, time_feat), dim=-1)
--> 255 return torch.cat((lags, features), dim=-1), scale, static_feat

RuntimeError: Sizes of tensors must match except in dimension 2. Expected size 1 but got size 38 for tensor number 1 in the list.
```

## Environment
- Operating system: MacOS

lostella:
Fixes: #2838

*Description of changes:* When `context_length = 1` the torch DeepAREstimator resulted in a `-0` indexing that was causing some tensor having unexpected size.

### DIFF
--- a/src/gluonts/torch/util.py
+++ b/src/gluonts/torch/util.py
@@ -188,6 +188,29 @@ def slice_along_dim(a: torch.Tensor, dim: int, slice_: slice) -> torch.Tensor:
     return a[idx]
 
 
+def take_last(a: torch.Tensor, dim: int, qty: int) -> torch.Tensor:
+    """
+    Take last elements from a given tensor along a given dimension.
+
+    Parameters
+    ----------
+    a
+        Original tensor to slice.
+    dim
+        Dimension to slice over.
+    qty
+        Quantity of trailing elements to retain (non-negative).
+
+    Returns
+    -------
+    torch.Tensor
+        A tensor with the same size as the input one, except dimension
+        ``dim`` which has length equal to ``qty``.
+    """
+    assert qty >= 0
+    return slice_along_dim(a, dim, slice(a.shape[dim] - qty, None))
+
+
 def unsqueeze_expand(a: torch.Tensor, dim: int, size: int) -> torch.Tensor:
     """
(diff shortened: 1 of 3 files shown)


src/gluonts/torch/util.py
```diff
@@ -188,6 +188,29 @@ def slice_along_dim(a: torch.Tensor, dim: int, slice_: slice) -> torch.Tensor:
     return a[idx]
 
 
+def take_last(a: torch.Tensor, dim: int, qty: int) -> torch.Tensor:
```
jaheba: `qty` looks like an unusual name.

jaheba:
Something I was thinking, `take_last` is not specific to `torch` but would work as well for `numpy`. Wouldn't it make sense to move these kind of functions elsewhere?

lostella:
> Something I was thinking, `take_last` is not specific to `torch` but would work as well for `numpy`. Wouldn't it make sense to move these kind of functions elsewhere?

It would make sense, but I guess there's something about having "framework-friendly" argument names. For example, `axis` in numpy is `dim` in torch. One could have thin wrappers around utils that have numpy-friendly names I guess.

src/gluonts/torch/util.py
lostella: changed to `num`

jaheba (approved):

PR #2843, opened 2023-05-09
Raise warning in `QuantileForecast.mean` when mean is not there

lostella:
*Issue #, if available:*

*Description of changes:* The `QuantileForecast` class falls back to the median when asked for the mean but the mean is not among the stored values. This is motivated by some downstream evaluation metrics using the mean of the predicted distribution, but some models only producing quantiles as forecasts.

To me this is pretty clearly not the right thing to do, and potential source of confusion: instead, it should be clear how to configure the evaluation to use the predicted median instead of the mean, if that's what the user wants.

### DIFF
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -12,6 +12,7 @@
 # permissions and limitations under the License.
 
 import re
+import logging
 from dataclasses import field
 from typing import Callable, Dict, List, Optional, Union, Tuple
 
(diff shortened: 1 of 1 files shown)


lostella:
Note: add warning instead, so that it's not breaking

src/gluonts/model/forecast.py
```diff
@@ -656,7 +659,10 @@ def mean(self) -> np.ndarray:
         """
         if "mean" in self._forecast_dict:
             return self._forecast_dict["mean"]
-
+        logger.warning(
+            "The mean prediction is not stored in the forecast data; "
+            "the median is being returned instead."
```
abdulfatir: "This behavior may change in the future."?

abdulfatir (approved):

PR #2855, opened 2023-05-13
Zebras: Add update to TimeSeries.

jaheba:
Allow merging of two time series.

With `update` it is possible to gradually build up a time series by updating the time series.

This works by creating a new series that spans both input series and is pre-filled with a default values. Then values are copied for the input series. This has some implications: If there is an overlap, the values of the second series overwrite the values of the first series in the overlapping span. Also, the values spanning between the two series remain the default value.

TODO:

- [x] Add consistency checks.
- [x] Merge static values
- [x] Merge metadata
- [x] Extend behaviour to other classes.

### DIFF
--- a/src/gluonts/zebras/_time_frame.py
+++ b/src/gluonts/zebras/_time_frame.py
@@ -168,6 +168,56 @@ class TimeFrame(TimeBase):
             _pad=Pad(pad_left, pad_right),
         )
 
+    def update(self, other: TimeFrame, default=np.nan) -> TimeFrame:
+        assert self.index is not None and other.index is not None
+
+        start = min(self.index.start, other.index.start)
+        end = max(self.index.end, other.index.end)
+
+        index = Periods(
+            np.arange(
+                start.to_numpy(),
+                # arange is exclusive, thus we need to add `1`
+                (end + 1).to_numpy(),
+                start.freq.step,
+            ),
+            start.freq,
+        )
+        self_idx0 = index.index_of(self.index.start)
+        other_idx0 = index.index_of(other.index.start)
+
+        columns = {}
+        for name, (self_col, other_col, tdim) in rows_to_columns(
+            (self.columns, other.columns, self.tdims)
+        ).items():
+            new_shape = list(self_col.shape)
+            new_shape[tdim] = len(index)
+
+            values = np.full(new_shape, default)
+            view = AxisView(values, tdim)
+
+            view[self_idx0 : self_idx0 + len(self)] = self_col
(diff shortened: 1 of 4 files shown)


src/gluonts/zebras/_time_frame.py
```diff
@@ -136,6 +136,76 @@ def pad(self, value, left: int = 0, right: int = 0) -> TimeFrame:
             _pad=Pad(pad_left, pad_right),
         )
 
+    def update(self, other: TimeFrame, default=np.nan) -> TimeFrame:
+        """Create a new ``TimeFrame`` which includes values of both input
+        frames.
+
+        The new frame spans both input frames and inserts default values if
+        there is a gap between the two frames. If both frames overlap, the
+        second can overwrite the values of the first frame.
```
lostella: can or does?

```suggestion
        second overwrites the values of the first frame.
```

src/gluonts/zebras/_time_series.py
```diff
@@ -88,6 +88,72 @@ def pad(self, value, left: int = 0, right: int = 0) -> TimeSeries:
             _pad=self._pad.extend(left, right),
         )
 
+    def update(self, other: TimeSeries, default=np.nan) -> TimeSeries:
+        """Create a new ``TimeSeries`` which includes values of both input
+        series.
+
+        The new series spans both input series and inserts default values if
+        there is a gap between the two series. If both series overlap, the
+        second can overwrite the values of the first series.
```
lostella:
```suggestion
        second overwrites the values of the first series.
```

src/gluonts/zebras/_time_frame.py
```diff
@@ -136,6 +136,76 @@ def pad(self, value, left: int = 0, right: int = 0) -> TimeFrame:
...
+            ),
+            start.freq,
+        )
+        self_idx0 = index.index_of(self.index.start)
+        other_idx0 = index.index_of(other.index.start)
+
+        columns = {}
+        for name, (self_col, other_col, tdim) in rows_to_columns(
+            (self.columns, other.columns, self.tdims)
+        ).items():
```
lostella: What columns are taken? Is it assumed that the two frames have the same columns? On only the ones from `self` are considered?

jaheba: Yes, the idea is that they need to be identical at the moment. I guess what we can do is to allow to update the result only partially and just fall back on the default value. That way one can even update different parts of the time frame at different times.

lostella (approved):

PR #2858, opened 2023-05-15
Zebras: Add `eq_to` to TimeFrame.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/zebras/_time_frame.py
+++ b/src/gluonts/zebras/_time_frame.py
@@ -61,7 +61,7 @@ class TimeFrame(TimeBase):
             f"Expected: {len(self)}, got {len(self.index)}."
         )
 
-    def __eq__(self, other) -> bool:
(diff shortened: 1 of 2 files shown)


jaheba:
Implementing `__eq__` like that would diverge from how pandas works. In short, pandas DataFrames behave similar to how `==` works on `numpy` arrays:

```py
>>> x = pd.DataFrame({"a": [1, 2, 3]})
>>> (x == x).to_dict("list")
{'a': [True, True, True]}
```

If the two data-frames can't be compared (have different shape) an error is thrown.

Maybe we actually want to do something similar?

As an alternative there is `np.array_equal` which tests if the both inputs are the same.

lostella (approved):

lostella:
I think `==` behaving a-la numpy would give the least amount of surprise.

PR #2859, opened 2023-05-15
Add ``join_items`` to itertools.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/itertools.py
+++ b/src/gluonts/itertools.py
@@ -452,3 +452,31 @@ def power_set(iterable):
     return itertools.chain.from_iterable(
         itertools.combinations(iterable, r) for r in range(len(iterable) + 1)
     )
+
+
+def zip_items(*dicts, default=None, strict=False):
+    """
+    Iterate over multiple dictionary items.
+
+    >>> a = {"x": "a"}
+    >>> b = {"x": "b"}
+    >>> c = {}
+    >>> for name, xa, xb, xc in zip_items(a, b, c):
+    ...     print(name, xa, xb, xc)
+    x, a, b, None
+    """
+    if not dicts:
+        return
+
+    ref = dicts[0]
+
+    if strict:
+        for dct in dicts[1:]:
+            assert ref.keys() == dct.keys()
+    else:
+        ref = set()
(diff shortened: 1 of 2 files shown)


src/gluonts/itertools.py
```diff
@@ -452,3 +452,31 @@ def power_set(iterable):
     return itertools.chain.from_iterable(
         itertools.combinations(iterable, r) for r in range(len(iterable) + 1)
     )
+
+
+def zip_items(*dicts, default=None, strict=False):
```
lostella: zip_values seems more what is happening here?

src/gluonts/itertools.py
```diff
@@ -452,3 +452,31 @@ def power_set(iterable):
...
+    """
+    Iterate over multiple dictionary items.
+
+    >>> a = {"x": "a"}
+    >>> b = {"x": "b"}
+    >>> c = {}
+    >>> for name, (xa, xb, xc) in zip_items(a, b, c):
+    ...     print(name, xa, xb, xc)
+    x a b None
```
lostella: This contrasts with what happens for lists with zip, in which case the output is limited to the common keys. So in this example the output iterator would be empty.

Maybe adding None could be optional?

src/gluonts/itertools.py
jaheba: It returns key value pairs like `items`, hence the name.

src/gluonts/itertools.py
jaheba: I guess there are generally three ways to handle this:

* be strict and require all dicts to have the same keys
* just iterate over the intersection of all keys
* iterate over the union of all keys, fall back on a default when a dict doesn't have a given key

Would we want to support all three modes?

jaheba:
I think the name is a bit confusing, maybe `join_items` or something like that is better?

lostella:
Maybe yes; then one can borrow the SQL jargon and define inner, outer, left, and right to cover all cases

jaheba:
`left` / `iight` only works for two inputs. I think that's fine?

lostella (approved):

PR #2860, opened 2023-05-16
Cache groupby result in `PandasDataset`

lostella:
*Description of changes:* On small datasets, this doesn't make much of a difference. On larger datasets such as M5, I observed a 10% slowdown at construction time, but then the same overhead is avoided when iterating the dataset over and over again.


jaheba:
I guess we don't really care about memory consumption?

jaheba (approved):

lostella:
> I guess we don't really care about memory consumption?

We can worry about that when needed

PR #2864, opened 2023-05-16
Simplify plotting of forecasts.

jaheba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -13,6 +13,7 @@
 
 import re
 from dataclasses import field
+from functools import partial
 from typing import Callable, Dict, List, Optional, Union, Tuple
 
 import numpy as np
@@ -20,6 +21,7 @@ import pandas as pd
 from pydantic.dataclasses import dataclass
 
 from gluonts.core.component import validated
+from gluonts import maybe
 
 
 def _linear_interpolation(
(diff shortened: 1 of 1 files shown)


jaheba:
Screenshot 2023-05-16 at 20 30 25

src/gluonts/model/forecast.py
```diff
@@ -283,99 +284,73 @@ def __getitem__(self, name):
...
-        label_prefix = "" if label is None else label + "-"
+        # Get current axes (gca), if not provided explicitly.
+        ax = maybe.unwrap_or_else(ax, plt.gca)
 
-        for c in prediction_intervals:
-            assert 0.0 <= c <= 100.0
+        # If no color is provided, we use matplotlib's internal color cycle.
+        # Note: This is an internal API and might change in the future.
+        color = maybe.unwrap_or_else(
+            color, lambda: next(ax._get_lines.prop_cycler)["color"]
```
abdulfatir, line 316: Perhaps we can use the CN notation for getting the default color?
https://stackoverflow.com/questions/42086276/get-default-line-colour-cycle

jaheba: We want to get the next colour in the cycle. In the above example, if `"black"` was not set explicitly, the true values would be blue, and the predictions would start with orange.

lostella (approved):

PR #2878, opened 2023-05-22
Speedup is_uniform for PandasDataset.

jaheba:
Constructing a second index is very slow. The new version is 9x faster locally when using an index of size 365.


lostella (approved):
Nice! Marking for backport

PR #2879, opened 2023-05-22
Breaking: change num_output with quantiles in TFT (mxnet)

baniasbaabe:
*Issue #, if available:*

*Description of changes:*
The TFT model with MXNet implementation, the quantiles are calculated based on `num_outputs`. This is a bit counter-intuitive, so why can I not give the model custom quantiles? Another limitation was that the automatically calculated quantiles could return negative values for high `num_outputs`. So I change the attribute to `quantiles` where you can give the model a list of floats (like MQCNN).


src/gluonts/mx/model/tft/_network.py
```diff
@@ -151,7 +151,8 @@ def __init__(
         d_var: int,
         d_hidden: int,
         n_head: int,
-        n_output: int,
+        #n_output: int,
```
lostella: Could this be removed instead of commented?

src/gluonts/mx/model/tft/_network.py
```diff
@@ -167,11 +168,8 @@ def __init__(
         self.d_var = d_var
         self.d_hidden = d_hidden
         self.n_head = n_head
-        self.n_output = n_output
-        self.quantiles = sum(
-            ([i / 10, 1.0 - i / 10] for i in range(1, (n_output + 1) // 2)),
-            [0.5],
-        )
+        # self.n_output = n_output
```
lostella: Remove?

lostella (changes requested):
Hi @baniasbaabe, thanks for the PR! I generally agree with this change, I only have a couple of comments inline.

Note that this would be a breaking change: if merged, it will not be included in a release before 0.14.0.

lostella:
@baniasbaabe just out of curiosity, is there any particular reason to prefer the MXNet implementation of TFT over the [PyTorch one](https://github.com/awslabs/gluonts/tree/dev/src/gluonts/torch/model/tft)?

src/gluonts/mx/model/tft/_network.py
baniasbaabe: Yes, it sneaked into it :)

src/gluonts/mx/model/tft/_network.py
baniasbaabe: Yes

baniasbaabe:
> @baniasbaabe just out of curiosity, is there any particular reason to prefer the MXNet implementation of TFT over the [PyTorch one](https://github.com/awslabs/gluonts/tree/dev/src/gluonts/torch/model/tft)?

I'm not a big fan of MXNet, but there are indeed some organizations that rely on MXNet and maybe build their model pipelines around it. Most of the gluonts models were MXNet based some versions ago :)

lostella (dismissed):

lostella (approved):
Thanks @baniasbaabe, and sorry it took a bit long but this can be merged 🚀  I just added minor styling changes

PR #2885, opened 2023-05-23
Fix default scale in torch DeepAR

### LINKED ISSUE #2884: Strange behaviour of `DeepAREstimator` with `NormalOutput`
## To Reproduce
From the README with minor changes

```python
import pandas as pd
from gluonts.torch.distributions import NormalOutput

df = pd.read_csv(
    "https://raw.githubusercontent.com/AileenNielsen/"
    "TimeSeriesAnalysisWithPython/master/data/AirPassengers.csv",
    index_col=0,
    parse_dates=True,
)
dataset = PandasDataset(df, target="#Passengers")

training_data, test_gen = split(dataset, offset=-36)
test_data = test_gen.generate_instances(prediction_length=12, windows=3)

model = DeepAREstimator(
    prediction_length=12,
    freq="M",
    distr_output=NormalOutput(),
    trainer_kwargs={"max_epochs": 3},
).train(training_data)
```

Loss values are way off with gluonts 0.13 compared to 0.12, suggesting a potential regression.

v0.12

```
Epoch 0: : 50it [00:02, 20.47it/s, v_num=10, train_loss=5.670]
Epoch 1: : 50it [00:02, 19.70it/s, v_num=10, train_loss=4.760]
Epoch 2: : 50it [00:02, 21.59it/s, v_num=10, train_loss=4.600]
```

v0.13

```
Epoch 0: : 50it [00:02, 19.36it/s, v_num=11, train_loss=9.3e+21]
Epoch 1: : 50it [00:02, 22.46it/s, v_num=11, train_loss=3.8e+22]
Epoch 2: : 50it [00:02, 22.23it/s, v_num=11, train_loss=4.64e+22]
```

## Environment
- Operating system: macOS 13
- Python version: 3.8
- GluonTS version: 0.13.0
- Torch: 2
- Lightning: 2

lostella:
*Issue #, if available:* Fixes #2884

*Description of changes:* With `default_scale=0`, model training becomes highly unstable as soon as an "empty" training window is encountered, in which case the `default_scale` value is used to scale the series: this is apparent especially in case short series are in the training data, such as in the minimal README example from #2884. This PR fixes the issue by setting `default_scale=None` by default, in which case the default scale value is taken from the rest of the batch (whether this is the right thing to do is questionable, but this is a separate story). The change to `default_scale=0` in DeepAR was introduced in #2600 (later modified by #2633 and #2632), so this PR partially undoes it.

### DIFF
--- a/src/gluonts/torch/model/deepar/estimator.py
+++ b/src/gluonts/torch/model/deepar/estimator.py
@@ -166,7 +166,7 @@ class DeepAREstimator(PyTorchLightningEstimator):
         distr_output: DistributionOutput = StudentTOutput(),
         loss: DistributionLoss = NegativeLogLikelihood(),
         scaling: bool = True,
-        default_scale: float = 0.0,
+        default_scale: Optional[float] = None,
         lags_seq: Optional[List[int]] = None,
         time_features: Optional[List[TimeFeature]] = None,
         num_parallel_samples: int = 100,
--- a/src/gluonts/torch/model/deepar/module.py
+++ b/src/gluonts/torch/model/deepar/module.py
@@ -105,7 +105,7 @@ class DeepARModel(nn.Module):
         distr_output: DistributionOutput = StudentTOutput(),
         lags_seq: Optional[List[int]] = None,
         scaling: bool = True,
-        default_scale: float = 0.0,
+        default_scale: Optional[float] = None,
         num_parallel_samples: int = 100,
     ) -> None:
         super().__init__()


jaheba (approved):

jaheba:
A scale of `0.0` essentially means the target values are "scaled" down to `0.0`?

timoschowski:
thanks for looking into this. Scaling indeed seems to be a problem, when I compare 0.12 and 0.13 on the M5 data set.

lostella:
> A scale of `0.0` essentially means the target values are "scaled" down to `0.0`?

Past target values are divided by the scale: since there is also a `minimum_scale=1e-10` (see [here](https://github.com/awslabs/gluonts/blob/5b73d2182b1219ac9869bc53c69083d8a9201859/src/gluonts/torch/scaler.py#L86)) then this is not really a division by zero but it blows up the model inputs.

lostella:
> thanks for looking into this. Scaling indeed seems to be a problem, when I compare 0.12 and 0.13 on the M5 data set.

Thanks for raising the issue @timoschowski, let me know in case you're able to validate that things are now better in `dev`

PR #2892, opened 2023-05-26
Use observation mask as feature in DeepAR

abdulfatir:
*Description of changes:*
This PR adds the observation mask as an input feature to DeepAR. Previously lags were fed into the model but the model had no information on whether the lag value is observed or missing/padded.

I ran a small experiment to gauge the performance with and without (previous version) the mask feature. Here are the results on the **solar** dataset averaged over 10 runs. Each model was trained for 10 epochs.

| model        |   NRMSE |     ND |   mean_wQuantileLoss |   sMAPE |   MASE |
|:-------------|--------:|-------:|---------------------:|--------:|-------:|
| with_mask (now)   |  1.0553 | 0.4896 |               0.4054 |  1.3745 | 1.1503 |
| without_mask (previous) |  1.0551 | 0.4892 |               0.4074 |  1.3741 | 1.1495 |

### DIFF
--- a/src/gluonts/torch/model/deepar/module.py
+++ b/src/gluonts/torch/model/deepar/module.py
@@ -146,7 +146,9 @@ class DeepARModel(nn.Module):
             )
         else:
             self.scaler = NOPScaler(dim=-1, keepdim=True)
-        self.rnn_input_size = len(self.lags_seq) + self._number_of_features
+        self.rnn_input_size = (
+            2 * len(self.lags_seq)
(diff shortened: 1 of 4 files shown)


lostella:
> Here are the results on the solar dataset averaged over 10 runs.

Does the dataset contain missing values?

abdulfatir:
No, it does not, so this only testing the padding aspect.

TonyHCIR (approved):

PR #2900, opened 2023-06-01
Rework `maybe.Maybe` to use a boxed container.

jaheba:
Also allow to pass `args` and `kwargs` to `map` and `and_then`.
    
Now `maybe.Maybe` return `Maybe` values:
    
```py
Some(0).map(add, 1).map(add, 2).unwrap() == 3
```

### DIFF
--- a/src/gluonts/maybe.py
+++ b/src/gluonts/maybe.py
@@ -244,7 +244,7 @@ def and_(val: OptionalOrMaybe[T], other: OptionalOrMaybe[U]) -> Optional[U]:
 
 def and_then(
     val: OptionalOrMaybe[T],
-    fn: Callable[Concatenate[T, P], U],
+    fn: Callable[Concatenate[T, P], OptionalOrMaybe[U]],
     *args: P.args,
     **kwargs: P.kwargs,
 ) -> Optional[U]:
@@ -547,7 +547,7 @@ class Maybe(ABC, Generic[T]):
     @abstractmethod
     def and_then(
         self,
-        fn: Callable[Concatenate[T, P], U],
+        fn: Callable[Concatenate[T, P], OptionalOrMaybe[U]],
         *args: P.args,
         **kwargs: P.kwargs,
     ) -> Maybe[U]:
@@ -750,7 +750,7 @@ class Some(Maybe[T]):
 
     def and_then(
         self,
-        fn: Callable[Concatenate[T, P], U],
+        fn: Callable[Concatenate[T, P], OptionalOrMaybe[U]],
         *args: P.args,
         **kwargs: P.kwargs,
     ) -> Maybe[U]:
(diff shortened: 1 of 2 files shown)


src/gluonts/maybe.py
```diff
@@ -76,25 +126,31 @@ def do(val: Optional[T], fn: Callable[[T], U]) -> Optional[T]:
...
     """
-    if val is not None:
-        fn(val)
-
-    return val
+    return box(val).do(fn).unbox()
 
 
-def map(val: Optional[T], fn: Callable[[T], U]) -> Optional[U]:
+def map(
```
lostella, line 131: I’m wondering: why is this called `map` and not for example `apply`? `map` sounds more like the built-in `map` that deals with iterables. (Same for `map_or` and `map_or_else`.)

lostella (commented):
Is `do` redundant now? (Does `map` without args do the same?)

src/gluonts/maybe.py
jaheba: The idea behind `map` is that it takes some as input `M[T]` and returns some `M[U]`. This is true for list `list[T] -> list[U]` and also for maybe `Maybe[T] -> Maybe[U]`.

Actually, it takes `fn: T -> U`, and first unwraps the value and then wraps the result again.

It's what people call functor and map.

jaheba:
> Is `do` redundant now? (Does `map` without args do the same?)

It shouldn't. `do` should return the underlying value, without changing it:

```py
_ = Some(1).do(print).map(lamda n: n + 1).do(print)

### prints
1
2
```

src/gluonts/maybe.py
```diff
@@ -426,233 +464,430 @@ def map_or_else(
...
 
         """
-        return and_(self.val, other)
+        return self.and_(other)
 
-    def and_then(self, fn: Callable[[T], Optional[U]]) -> Optional[U]:
+    @abstractmethod
+    def and_then(
+        self,
+        fn: Callable[Concatenate[T, P], U],
```
lostella: Does `fn` always return an `Optional`? The docstring says so. I guess I’m having a hard time getting the difference with `map`

lostella (commented):
So I guess the functional interface can be easier to use sometimes, but the object oriented one is better in case one needs to chain multiple transformations?

src/gluonts/maybe.py
lostella: I guess I need to refresh my category theory!

jaheba:
> So I guess the functional interface can be easier to use sometimes, but the object oriented one is better in case one needs to chain multiple transformations?

Yes! Also, we can use Maybe values on objects directly

lostella (dismissed):

src/gluonts/maybe.py
jaheba: That's the whole fun! `and_then` is `flatmap`, meaning `fn` returning an Option, makes the Monad.

lostella (approved):

lostella (approved):

PR #2905, opened 2023-06-02
Filter unused fields during inference

### LINKED ISSUE #2904: Models improperly handling additional (unused) data attributes
## Description
Unused data attributes (such as `past_feat_dynamic_real`) are not ignored by data loading, and break the predictor due to shape issues: the `past_feat_dynamic_real` field from a batch of entries is not being properly sliced, but still it's being stacked together, breaking inference when entries have different lengths.

## To Reproduce

```python

import pandas as pd
import numpy as np

estimator = DeepAREstimator(
    freq="1H",
    prediction_length=10,
    context_length=10,
    num_feat_dynamic_real=1,
    trainer_kwargs=dict(max_epochs=1),
)

data_train = [
    {
        "start": pd.Period("2020-01-01", freq="1H"),
        "target": np.ones(shape=(500,)),
        "feat_dynamic_real": np.ones(shape=(1, 500)),
        "past_feat_dynamic_real": np.ones(shape=(1, 500)),
    },
    {
        "start": pd.Period("2020-01-01", freq="1H"),
        "target": np.ones(shape=(300,)),
        "feat_dynamic_real": np.ones(shape=(1, 300)),
        "past_feat_dynamic_real": np.ones(shape=(1, 300)),
    }
]

predictor = estimator.train(data_train)

data_pred = [
    {
        "start": pd.Period("2020-01-01", freq="1H"),
        "target": np.ones(shape=(500,)),
        "feat_dynamic_real": np.ones(shape=(1, 510)),
        "past_feat_dynamic_real": np.ones(shape=(1, 500)),
    },
    {
        "start": pd.Period("2020-01-01", freq="1H"),
        "target": np.ones(shape=(300,)),
        "feat_dynamic_real": np.ones(shape=(1, 310)),
        "past_feat_dynamic_real": np.ones(shape=(1, 300)),
    }
]

forecasts = list(predictor.predict(data_pred))
```

```
  File "reproduce.py", line 68, in <module>
    forecasts = list(predictor.predict(data_pred))
  File "/Users/stellalo/gluonts/src/gluonts/torch/model/predictor.py", line 85, in predict
    yield from self.forecast_generator(
  File "/Users/stellalo/gluonts/src/gluonts/model/forecast_generator.py", line 154, in __call__
  File "/Users/stellalo/gluonts/src/gluonts/transform/_base.py", line 111, in __iter__
    yield from self.transformation(
  File "/Users/stellalo/gluonts/src/gluonts/transform/_base.py", line 136, in __call__
    raise e
  File "/Users/stellalo/gluonts/src/gluonts/transform/_base.py", line 134, in __call__
    yield self.map_transform(data_entry.copy(), is_train)
  File "/Users/stellalo/gluonts/src/gluonts/transform/_base.py", line 149, in map_transform
    return self.transform(data)
  File "/Users/stellalo/gluonts/src/gluonts/transform/_base.py", line 169, in transform
    return self.func(data.copy())
  File "/Users/stellalo/gluonts/src/gluonts/torch/model/predictor.py", line 79, in <lambda>
    stack_fn=lambda data: batchify(data, self.device),
  File "/Users/stellalo/gluonts/src/gluonts/torch/batchify.py", line 33, in batchify
    return {
  File "/Users/stellalo/gluonts/src/gluonts/torch/batchify.py", line 34, in <dictcomp>
    key: stack(data=[item[key] for item in data], device=device)
  File "/Users/stellalo/gluonts/src/gluonts/torch/batchify.py", line 24, in stack
    data = torch.tensor(np.array(data), device=device)
ValueError: could not broadcast input array from shape (500,) into shape (1,)
```

## Environment
- Operating system:
- Python version:
- GluonTS version: `dev`
- MXNet version:

abdulfatir:
*Issue #, if available:*
Fixes #2904 

*Description of changes:*

### DIFF
--- a/src/gluonts/torch/model/predictor.py
+++ b/src/gluonts/torch/model/predictor.py
@@ -30,7 +30,7 @@ from gluonts.model.forecast_generator import (
 from gluonts.model.predictor import OutputTransform, Predictor
 from gluonts.torch.batchify import batchify
 from gluonts.torch.component import equals
-from gluonts.transform import Transformation
+from gluonts.transform import Transformation, SelectFields
 
 
 @predict_to_numpy.register(nn.Module)
@@ -73,12 +73,14 @@ class PyTorchPredictor(Predictor):
     def predict(
         self, dataset: Dataset, num_samples: Optional[int] = None
     ) -> Iterator[Forecast]:
+        self.input_transform += SelectFields(
+            self.input_names + self.required_fields, allow_missing=True
+        )
         inference_data_loader = InferenceDataLoader(
             dataset=dataset,
             transform=self.input_transform,
             batch_size=self.batch_size,
             stack_fn=lambda data: batchify(data, self.device),
-            field_names=self.input_names + self.required_fields,
         )
 
         self.prediction_net.eval()
--- a/src/gluonts/torch/model/predictor.py
+++ b/src/gluonts/torch/model/predictor.py
@@ -77,7 +77,7 @@ class PyTorchPredictor(Predictor):
             self.input_names + self.required_fields, allow_missing=True
         )
         inference_data_loader = InferenceDataLoader(
-            dataset=dataset,
+            dataset,
             transform=self.input_transform,
             batch_size=self.batch_size,
             stack_fn=lambda data: batchify(data, self.device),
--- a/src/gluonts/torch/model/predictor.py
+++ b/src/gluonts/torch/model/predictor.py
@@ -59,6 +59,7 @@ class PyTorchPredictor(Predictor):
         self.forecast_generator = forecast_generator
         self.output_transform = output_transform
         self.device = device
+        self.required_fields = ["forecast_start"]
 
     def to(self, device) -> "PyTorchPredictor":
         self.prediction_net = self.prediction_net.to(device)
@@ -73,10 +74,11 @@ class PyTorchPredictor(Predictor):
         self, dataset: Dataset, num_samples: Optional[int] = None
     ) -> Iterator[Forecast]:
         inference_data_loader = InferenceDataLoader(
-            dataset,
+            dataset=dataset,
             transform=self.input_transform,
             batch_size=self.batch_size,
             stack_fn=lambda data: batchify(data, self.device),
+            field_names=self.input_names + self.required_fields,
         )
 
         self.prediction_net.eval()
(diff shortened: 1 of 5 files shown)


lostella:
Besides fixing the issue, maybe this test could be adapted to reproduce the problem: https://github.com/awslabs/gluonts/blob/d1e4889b8ab105f2e2806c966a8e9961c6ee5d58/test/torch/model/test_estimators.py#L265

For example adding a field to the data that surely no model uses (`”unused”`?)

jaheba:
I think in training we use ``as_stacked_batches``, which exposes a field_names parameter.

Should we use that for inference as well?

abdulfatir:
Yes, I tried doing that but one issue with that approach is that it tries to stack everything as `torch.tensor` which breaks when you have objects like `pd.Period`.

lostella:
The change looks good to me. Should we have it similarly in https://github.com/awslabs/gluonts/blob/8d9b45c0de76d9b5a8375358b228d4dcf8754c18/src/gluonts/mx/model/predictor.py#L156?

Also, this should cover it in tests

src/gluonts/torch/model/predictor.py
```diff
@@ -73,10 +74,11 @@ def predict(
         self, dataset: Dataset, num_samples: Optional[int] = None
     ) -> Iterator[Forecast]:
         inference_data_loader = InferenceDataLoader(
-            dataset,
+            dataset=dataset,
             transform=self.input_transform,
```
lostella, line 81: You could also just add the `SelectFields` transformation here maybe? This way `loader.py` stays untouched

abdulfatir: I am wondering if we should actually have select fields as an optional in every loader? I think you would face a similar problem for `ValidationDataLoader`.

lostella: For now, I would go for as minimal a change as possible. It seems like (some) models that use `ValidationDataLoader` explicitly add `SelectField` https://github.com/lostella/gluonts/blob/c6f90a6afeea50fdc700ee8a0f7104c8d9f683da/src/gluonts/mx/model/deepar/_estimator.py#L419-L420

lostella (dismissed):

jaheba (approved):

PR #2913, opened 2023-06-09
Add new samplers

abdulfatir:
*Issue #, if available:*

*Description of changes:*
This PR adds two new samplers: `NInstanceSampler` and `ExpectedNumInstanceWithMinSampler`.
`NInstanceSampler`: sample exactly `N` time points per time series.
`ExpectedNumInstanceWithMinSampler`: same as `ExpectedNumInstanceSampler` but ensures at a minimum number of time points are sampled per time series.

Thanks to @lostella for code and suggestions.


src/gluonts/transform/sampler.py
```diff
@@ -45,6 +45,26 @@ def __call__(self, ts: np.ndarray) -> np.ndarray:
         raise NotImplementedError()
 
 
+class NInstanceSampler(InstanceSampler):
```
lostella: I'm tempted to say that `NumInstanceSampler` would be a better name: just like `ExpectedNumInstanceSampler` samples a certain number of instances in expectations, this samples a certain number of instances (full stop).

Let's see also what others think about the naming maybe

lostella: Let's say: if all these samplers had a more "assertive" naming, I would just call this `SampleN` (and then `SampleOnAverageN`)

src/gluonts/transform/sampler.py
```diff
@@ -142,6 +162,37 @@ def __call__(self, ts: np.ndarray) -> np.ndarray:
         return indices + a
 
 
+class ExpectedNumInstanceWithMinSampler(ExpectedNumInstanceSampler):
```
lostella: I see two ways to escape the naming dilemma here:
1. Make `min_instances: int = 0` (note the default `0`) an attribute of `ExpectedNumInstanceSampler` and add the logic directly there. Then there's no name to be invented. However, strictly speaking when one sets `min_instances > 0` then this is not really sampling `num_instances` on average.
2. Make this logic a wrapper around some other sampler: `sampler = AtLeast(1, ExpectedNumInstanceSampler)`. The logic is factored a bit better maybe that way, just not as straightforward to instantiate.

Again, no strong opinions, it's just that this names is now a mouthful. @jaheba to the rescue.

src/gluonts/transform/sampler.py
abdulfatir: Makes sense. Updated.

src/gluonts/transform/__init__.py
```diff
@@ -33,6 +33,8 @@
     "ContinuousTimePredictionSampler",
     "ExpandDimArray",
     "ExpectedNumInstanceSampler",
+    "ExpectedNumInstanceWithMinSampler",
+    "NInstanceSampler",
```
lostella: These also need updating

src/gluonts/transform/sampler.py
lostella: @abdulfatir let's go for option 1

src/gluonts/transform/sampler.py
```diff
@@ -142,6 +162,37 @@ def __call__(self, ts: np.ndarray) -> np.ndarray:
...
+
+    def __call__(self, ts: np.ndarray) -> np.ndarray:
+        indices = super().__call__(ts)
+
+        a, b = self._get_bounds(ts)
+
+        if a > b:
+            return indices
+        if len(indices) < self.min_instances:
+            prefix = np.random.choice(
```
jaheba: `np.random.randint` or `.random_integer` should make this a bit more concise.

src/gluonts/transform/sampler.py
```diff
@@ -142,6 +162,37 @@ def __call__(self, ts: np.ndarray) -> np.ndarray:
...
+
+    num_instances: float
+    min_instances: int = 1
+
+    def __call__(self, ts: np.ndarray) -> np.ndarray:
+        indices = super().__call__(ts)
+
+        a, b = self._get_bounds(ts)
+
+        if a > b:
```
jaheba: Should this be `>=`?

And also, can we return `np.array([])` in that case, I think it makes it a bit more clear what this does then.

src/gluonts/transform/sampler.py
```diff
@@ -142,6 +162,37 @@ def __call__(self, ts: np.ndarray) -> np.ndarray:
...
+            return indices
+        if len(indices) < self.min_instances:
+            prefix = np.random.choice(
+                range(a, b + 1), size=self.min_instances - len(indices)
+            )
+            indices = np.concatenate([prefix, indices])
```
jaheba:
```suggestion
            return np.concatenate([prefix, indices])
```

jaheba:
I think we can just extend the current class with `min_instances=0`

TonyHCIR (approved):

src/gluonts/transform/sampler.py
abdulfatir: Done.

src/gluonts/transform/sampler.py
abdulfatir: Done

abdulfatir:
@lostella @jaheba Removed the new class and added `min_instances` option to the original class.

lostella (approved):

src/gluonts/transform/sampler.py
```diff
@@ -45,6 +45,26 @@ def __call__(self, ts: np.ndarray) -> np.ndarray:
         raise NotImplementedError()
 
 
+class NumInstanceSampler(InstanceSampler):
+    """
+    Samples N time points from each series.
+
+    Parameters
+    ----------
+    N
```
jaheba, line 54: Why is this uppercase?

PR #2914, opened 2023-06-12
Rework ev.

jaheba:
Changes:

* All metrics now have canonical names, for example `RMSE` now is `RMSE[mean]`, indicating that we compare against the mean prediction.
* Renamed `Evaluator` to `Metric`, and `Metric` to `MetricDefinition`.
* Add two classes `MetricCollection` and `MetricDefinitionCollection` to group respective objects.
* Add `evaluate` function to run evaluations on metrics end to end.

Usage:

```py
from gluonts.ev import nd, rmse, evaluate

evaluate(nd + rmse, data_batches, axis=1)
```


src/gluonts/ev/metrics.py
```diff
@@ -34,98 +35,193 @@
...
+
+
+@dataclass
+class DerivedMetric(Metric):
+    """An Metric for metrics that are derived from other metrics.
+
+    A derived metric updates multiple, simpler metrics independently and in
+    the end combines their results as defined in `post_process`."""
+
+    evaluators: Dict[str, Metric]
```
lostella: would `metrics` be a better name, to avoid confusion?

lostella:
Do we need 3 "layers" for this: `MetricDefinition`, `Metric`, and `Evaluator`. In a sense, the `Evaluator` appears to be a collection of `Metric` objects, just like `MetricCollection` is a collection ("sum") of `MetricDefinition`, so I'm wondering if everything could be more compact.

jaheba:
I think we need it.

Otherwise we loose the flexibility that we wanted in the first place.

src/gluonts/ev/metrics.py
```diff
@@ -34,115 +43,231 @@
...
+class MetricCollection:
+    metrics: List[Metric]
+
+    def update(self, data: Mapping[str, np.ndarray]) -> None:
+        for metric in self.metrics:
+            metric.update(data)
+
+    def update_all(self, stream: Iterator[Mapping[str, np.ndarray]]) -> None:
+        for element in stream:
+            self.update(element)
```
lostella: This could return `self` maybe

src/gluonts/ev/__init__.py
```diff
@@ -60,29 +72,44 @@
...
-    "Evaluator",
-    "DirectEvaluator",
-    "DerivedEvaluator",
 ]
+
+
+def evaluate(metrics, data_batches, axis=None):
+    evaluator = metrics(axis)
+    evaluator.update_all(data_batches)
+    return evaluator.get()
```
lostella: If `update_all` returned `self`, this would be `metrics(axis).update_all(data_batches).get()`

src/gluonts/ev/metrics.py
```diff
@@ -34,115 +43,231 @@
...
 
+@dataclass
+class MetricCollection:
+    metrics: List[Metric]
+
+    def update(self, data: Mapping[str, np.ndarray]) -> None:
+        for metric in self.metrics:
+            metric.update(data)
+
+    def update_all(self, stream: Iterator[Mapping[str, np.ndarray]]) -> None:
```
lostella: These methods could use a docstring I guess

src/gluonts/ev/metrics.py
```diff
@@ -34,115 +43,250 @@
 )
 
 
+@dataclass
+class MetricCollection:
```
lostella, line 47: Could this have `Metric` as base class?

lostella (approved):

PR #2916, opened 2023-06-13
Zebras: Rework Schema.

jaheba:
*Issue #, if available:*

*Description of changes:*


src/gluonts/zebras/schema.py
```diff
@@ -29,27 +29,90 @@
...
     tf = schema.load_timeframe(raw_dict)
 
     sf = schema.load_splitframe(raw_dict)
 
-The idea is to define expected types and shapes for each field (column) using
-``zebras.Field`` and then wrap them in a ``zebras.Schema``:
+The idea is to define expected types and shapes for each field (column) and
+then wrap them in a ``zebras.Schema``:
 
     schema = zebras.Schema({
```
lostella: `zebras.schema.Schema`?

src/gluonts/zebras/schema.py
```diff
@@ -29,27 +29,90 @@
...
-    schema = zebras.Schema(...)
+    schema = zebras.schema.Schema(...)
     tf = schema.load_timeframe(raw_dict)
 
     sf = schema.load_splitframe(raw_dict)
 
-The idea is to define expected types and shapes for each field (column) using
-``zebras.Field`` and then wrap them in a ``zebras.Schema``:
+The idea is to define expected types and shapes for each field (column) and
+then wrap them in a ``zebras.Schema``:
```
lostella: Why `and them write them`?

```suggestion
The idea is to define expected types and shapes for each field (column)
in a ``zebras.schema.Schema``:
```

src/gluonts/zebras/schema.py
```diff
@@ -29,27 +29,90 @@
...
 
     schema = zebras.Schema({
-        "target": Field(ndim=1, tdim=0, past_only=True),
-        "time_feat": zebras.Field(ndim=2, tdim=-1),
-        "static_feat": zebras.Field(ndim=1),
+        "target": zebras.schema.TimeSeries(),
+        "time_feat": zebras.schema.TimeSeries(ndim=2, tdim=-1, past_only=False),
+        "static_feat": zebras.schema.Array(ndim=1),
```
lostella, line 42: Could these just be keyword arguments to the constructor of `Schema`? So on can do it simpler:

```python
schema = Schema(
    target=Timeseries(),
    time_feat=Timeseries(...),
    static_feat=Array(...),
)
```

src/gluonts/zebras/schema.py
```diff
@@ -29,27 +28,90 @@
...
+
+class Array(Field):
+    ndim: Optional[int] = None
+    shape: Optional[tuple] = None
+    dtype: Type = np.float32
+
+    def load_from(self, data, name):
+        value = self._get(data, name)
+
+        value = np.array(value, dtype=self.dtype)
```
lostella: This forces copy I think. According to numpy docs:

> copy: bool, optional
> If true (default), then the object is copied. Otherwise, a copy will only be made if __array__ returns a copy, if obj is a nested sequence, or if a copy is needed to satisfy any of the other requirements ([dtype](https://numpy.org/doc/stable/reference/generated/numpy.dtype.html#numpy.dtype), order, etc.).

Maybe we should have an option?

src/gluonts/zebras/schema.py
lostella: I'm thinking to the case where we would use the schema simply for validation

src/gluonts/zebras/schema.py
jaheba: I guess we could use `asarray`?

Or indeed make it an option.

src/gluonts/zebras/schema.py
```diff
@@ -29,27 +28,90 @@
...
+    def load_from(self, data, name):
+        value = self._get(data, name)
+        return parse_obj_as(self.type, value)
+
+
+class Scalar(Field):
+    type: Type
+
+    def load_from(self, data, name):
+        value = self._get(data, name)
```
lostella: Is `load_from` always like this? That is, `self._get` followed by some logic for parsing the value? If so I'm wondering if `Field` should just define `load(self, value)`

lostella (approved):

PR #2917, opened 2023-06-13
Add `RandomYield` to `itertools`.

huibinshen:
*Issue #, if available:*

*Description of changes:*

Given a list of iterators, generate samples uniformly or according to the probabilities if given.


huibinshen:
cc @abdulfatir

src/gluonts/itertools.py
```diff
@@ -289,6 +289,43 @@ def __iter__(self):
...
+            probs = [
+                prob / sum(self.probabilities) for prob in self.probabilities
+            ]
+            idx = self.random_state.choice(range(len(self.iterators)), p=probs)
+
+            try:
+                data = next(self.iterators[idx])
+                success = True
+            except StopIteration:
+                self.probabilities[idx] = 0.0
```
jaheba: I don't think this will work, as the probabilities have to sum up to 1.

src/gluonts/itertools.py
```diff
@@ -289,6 +289,43 @@ def __iter__(self):
...
+@dataclass
+class MixIterators:
+    iterators: List[Iterator]
+    probabilities: Optional[List[float]] = None
+    random_state: np.random.RandomState = np.random.RandomState()
+
+    def __post_init__(self):
+        if self.probabilities:
+            # copy the probabilities as we may modify it
+            self.probabilities = [prob for prob in self.probabilities]
```
jaheba:
```suggestion
            self.probabilities = list(self.probabilities)
```

jaheba (commented):
There are some things which are not clear to me. The idea here is to sample from multiple underlying iterators, given some probability.

However, the question is what should happen if one source gets exhausted. I guess there are two approaches: Either stop yielding values entirely, or only yielding from the remaining sources. I can see the argument for both.

Also, should it be possible to rerun the iterator? E.g, if the input are lists then we could get different permutations each time we execute this.

src/gluonts/itertools.py
```diff
@@ -289,6 +289,43 @@ def __iter__(self):
         return filter(self.fn, self.iterable)
 
 
+@dataclass
+class MixIterators:
+    iterators: List[Iterator]
```
jaheba: I think I would make this
```suggestion
    iterators: List[Iterable]
```

src/gluonts/itertools.py
```diff
@@ -289,6 +289,43 @@ def __iter__(self):
...
+    probabilities: Optional[List[float]] = None
+    random_state: np.random.RandomState = np.random.RandomState()
```
jaheba: I think you would want to use `field` with `default_factory` here, that way each instance gets its own state.

src/gluonts/itertools.py
```diff
@@ -289,6 +289,43 @@ def __iter__(self):
...
+    def __post_init__(self):
+        if self.probabilities:
+            # copy the probabilities as we may modify it
+            self.probabilities = [prob for prob in self.probabilities]
+        else:
+            self.probabilities = [1.0 / len(self.iterators)] * len(
+                self.iterators
+            )
+
+    def __iter__(self):
```
jaheba, line 327: I think we can just do everything in `__iter__`, no need for `__next__`.

src/gluonts/itertools.py
huibinshen: It normalizes the probabilities in the beginning of the loop.

src/gluonts/itertools.py
huibinshen: Will do.

src/gluonts/itertools.py
```diff
@@ -289,6 +289,43 @@ def __iter__(self):
         return filter(self.fn, self.iterable)
 
 
+@dataclass
+class MixIterators:
```
abdulfatir: I don't like the name `MixIterators`.

src/gluonts/itertools.py
huibinshen: I talked with Lorenzo specifically on this. If we make it `Iterable`, we will need to create `Iterator`s inside the class, which will be hidden from user. To be explicit, we thought `Iterator` might be better.

jaheba:
This should work:


```py
    def __iter__(self):
        iterators = [iter(it) for it in self.iterables]
        indices = range(len(iterators))
        p = list(self.p)

        while True:
            idx = self.random_state.choice(indices, p=p)

            try:
                yield next(iterators[idx])
            except StopIteration:
                p[idx] = 0.0
                if sum(p) == 0.0:
                    return
                p /= sum(p)
```

src/gluonts/itertools.py
huibinshen: Will take a look!

src/gluonts/itertools.py
huibinshen: Any suggestion? `SampleFromIterators` or `MetaIterators`?

src/gluonts/itertools.py
huibinshen: This is related to the `Iterable` or `Iterator` question. If we stick to `Iterator`, then I think it makes sense to implement `__next__` method. So you pass in a list of `Iterator`s, you get an `Iterator` out.

src/gluonts/itertools.py
jaheba: See my proposal below, I think this should be iterators which we then call `iter` on.

src/gluonts/itertools.py
jaheba: Maybe something with "interleave"?

abdulfatir: How about `WeightedIterables` or something with probabilistic?

huibinshen:
> This should work:
> 
> ```python
>     def __iter__(self):
>         iterators = [iter(it) for it in self.iterables]
>         indices = range(len(iterators))
>         p = list(self.p)
> 
>         while True:
>             idx = self.random_state.choice(indices, p=p)
> 
>             try:
>                 yield next(iterators[idx])
>             except StopIteration:
>                 p[idx] = 0.0
>                 if sum(p) == 0.0:
>                     return
>                 p /= sum(p)
> ```

Because one creates an `Iterator` in the `__iter__`, it always give you the first element in those `Iterator`s. One way is to create these `iterators` as class fields in `__post_init__`. But if we decide to go with `Iterator`, then this would not be needed.

abdulfatir:
> Because one creates an Iterator in the __iter__, it always give you the first element in those Iterators.

Why?

huibinshen:
> > Because one creates an Iterator in the **iter**, it always give you the first element in those Iterators.
> 
> Why?

```
>>> a = list(range(10))
>>> a_it = iter(a)
>>> next(a_it)
0
>>> a_it = iter(a)
>>> next(a_it)
0
>>>
```
No? Or you mean `__iter__` will only be called once so it should be fine?

huibinshen:
> There are some things which are not clear to me. The idea here is to sample from multiple underlying iterators, given some probability.
> 
> However, the question is what should happen if one source gets exhausted. I guess there are two approaches: Either stop yielding values entirely, or only yielding from the remaining sources. I can see the argument for both.

Agree. I will make the documentation clearer.

> Also, should it be possible to rerun the iterator? E.g, if the input are lists then we could get different permutations each time we execute this.

I mean, one could create a new iterator and pass that to the class right?

abdulfatir:
> Or you mean __iter__ will only be called once so it should be fine?

Yes.

src/gluonts/itertools.py
lostella: > I think this should be iterators which we then call `iter` on

_iterables_

src/gluonts/itertools.py
jaheba: That name confuses me.

jaheba (changes requested):
Let's also add a test as docstring.

src/gluonts/itertools.py
huibinshen: Changed to `WeightedIterables`.

src/gluonts/itertools.py
```diff
@@ -289,6 +289,65 @@ def __iter__(self):
...
+        iterators = [iter(it) for it in self.iterables]
+        while True:
+            idx = self.random_state.choice(
+                range(len(iterators)), p=self.probabilities
+            )
+
+            try:
+                yield next(iterators[idx])
+            except StopIteration:
+                self.probabilities[idx] = 0
```
lostella: This makes the object stateful: if while iterating the first time some probabilities are set to zero, they will remain zero the second time you iterate. I don't think this is what we want?

I think in this iterable approach we should have a local copy of `self.probabilities` in the scope of `__iter__`

huibinshen: You are absolutely right, changed.

src/gluonts/itertools.py
```diff
@@ -289,6 +289,62 @@ def __iter__(self):
...
+
+    """
+
+    iterables: List[Iterable]
+    probabilities: Optional[List[float]] = None
+    random_state: RandomState = field(default_factory=RandomState)
+
+    def __post_init__(self):
+        if self.probabilities:
+            # copy the probabilities as we may modify it
```
jaheba: This should no longer be needed.

src/gluonts/itertools.py
```diff
@@ -289,6 +289,62 @@ def __iter__(self):
...
+    """
+    Given a list of Iterables `iterables`, generate samples from them.
+
+    When `probabilities` is given, sample iteratbles according to it.
+    When `probabilities` is not given, sample iterables uniformly.
+
+    When one iterable exhausts, the sampling probabilities for it will be set to 0.
+
+        >>> from toolz import take
+        >>> from gluonts.itertools import Cyclic
```
jaheba: No need to import things from the same module.

jaheba: Also, don't think we need it for demonstrating what this does.

src/gluonts/itertools.py
```diff
@@ -289,6 +289,62 @@ def __iter__(self):
         return filter(self.fn, self.iterable)
 
 
+@dataclass
+class WeightedIterables:
```
jaheba: Looking at just the name, I wouldn't know what this does. In some sense this is a scheduler doing resource management over multiple queues. I wish the name would make it clear that we take multiple iterators and fuse it into a single one.

src/gluonts/itertools.py
```diff
@@ -289,6 +289,62 @@ def __iter__(self):
...
+    When `probabilities` is given, sample iteratbles according to it.
+    When `probabilities` is not given, sample iterables uniformly.
+
+    When one iterable exhausts, the sampling probabilities for it will be set to 0.
+
+        >>> from toolz import take
+        >>> from gluonts.itertools import Cyclic
+        >>> a = [1, 2, 3]
+        >>> b = [4, 5, 6]
+        >>> it = iter(WeightedIterables([a, b], [1, 0]))
```
jaheba: I think it's cleaner if you provide the probabilities as a keyword parameter.

test/test_itertools.py
```diff
@@ -277,3 +279,22 @@ def test_join_items():
 
     with pytest.raises(Exception):
         oin_items(left, right, "strict")
+
+
+@pytest.mark.parametrize(
+    "iterables, probabilities, sample_size, samples",
+    [
+        ([[1, 2, 3], [4, 5, 6]], [1.0, 0.0], 5, [1, 2, 3]),
+        ([[1, 2, 3], [4, 5, 6]], [0.0, 1.0], 5, [4, 5, 6]),
```
jaheba: I don't like the test where we set the probability for an input to zero. Couldn't we initialise the random state with some seed?

jaheba (changes requested):

test/test_itertools.py
lostella: I think even if we set the seed, newer python or library versions may break the sequence generated by the same seed.

jaheba: No? The rng comes from numpy which should be fairly stable.

jaheba: No? The rng comes from numpy which should be fairly stable.

src/gluonts/itertools.py
huibinshen: I will remove the `Cyclic` import but leave the `toolz` import.

src/gluonts/itertools.py
huibinshen: `WeightedFusionIterables`?

jaheba (dismissed):

lostella (commented):
Maybe I’m late to the party, but since names are important I would consider alternatives:
- `YieldRandomlyFrom`
- `YieldRandomly`
- `RandomScheduler`

huibinshen:
> Maybe I’m late to the party, but since names are important I would consider alternatives:
> 
>     * `YieldRandomlyFrom`
> 
>     * `YieldRandomly`
> 
>     * `RandomScheduler`

Yes, you are late :-). 

From the list, I see the key world `Yield` and `Random`. Jasper was proposing `RandomYield`. I thought since we are passing probabilities to the yield, it is nice to have `Probabilistic` in the name, thus `ProbabilisticYield`. But `RandomYield` may also work for me (as it is the largest common factor).

lostella (approved):

PR #2921, opened 2023-06-15
Add scaled quantile loss

lostella:
*Description of changes:* Add scaled quantile loss, also known as scaled pinball loss, which is used for example in the M5 competition https://mofc.unic.ac.cy/m5-competition/. This just scales the quantile loss by the average naive error, similar to MASE.

$$
sQL_\alpha(\hat y, y, x) = \frac{
    QL_\alpha(\hat y, y)
}{
    \tfrac{1}{T-S} \sum_{S < t \leq T} |x_{t} - x_{t-S}|
}
$$

where $S$ is the number of seasons ($S=1$ gives the MAE of the naive forecaster in the denominator). This is aggregated via mean over the prediction length and/or dataset dimension. Reduces to MASE when the quantile level is 0.5.

I went for "quantile" instead of "pinball" for consistency with other metrics we have. Naming is not ideal, especially when we want to average over multiple quantile levels (`AverageMeanScaledQuantileLoss`), I'm open to suggestions.

### DIFF
--- a/src/gluonts/ev/__init__.py
+++ b/src/gluonts/ev/__init__.py
@@ -26,6 +26,7 @@ from .stats import (
 )
 from .metrics import (
     Metric,
+    MetricDefinition,
     MeanAbsoluteLabel,
     mean_absolute_label,
     SumAbsoluteLabel,
@@ -53,6 +54,7 @@ from .metrics import (
     WeightedSumQuantileLoss,
     MAECoverage,
     MeanScaledQuantileLoss,
+    AverageMeanScaledQuantileLoss,
     MeanSumQuantileLoss,
     MeanWeightedSumQuantileLoss,
     OWA,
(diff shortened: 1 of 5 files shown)


abdulfatir:
I'm wondering if we should use "Seasonal" or "Relative" somewhere in the loss instead of scaled. Currently, it's a bit confusing.

jaheba (approved):

shchur (approved):
@lostella Regarding the naming: I'm not very familiar with the evaluation logic in GluonTS, but I'm fine with the name if it's consistent with the rest of the metrics. 

I would also be fine with `MeanScaledQuantileLoss` corresponding to the average over all series / timesteps / quantile levels. I would guess that 99% of the users don't care about more sophisticated aggregation methods (e.g., weighted sum over quantiles, mean over series / timesteps), and if they do, they can implement these scores themselves.

@abdulfatir I prefer "scaled", as it's the more commonly used name for normalization by the seasonal difference (e.g., M4, M5 competitions, papers by [Hyndman](https://robjhyndman.com/papers/foresight.pdf))

lostella:
I like "relative" better, but I agree that consistency with MASE and how this type of (indeed relative) metric is called elsewhere.

PR #2923, opened 2023-06-19
add num_evaluations to metrics

melopeo:
*Description of changes:*

We add `num_evaluations` as a metric to know how many timestamps are evaluated. This helps to identify the amount of missing values when evaluating the given forecasts.

### DIFF
--- a/src/gluonts/ev/__init__.py
+++ b/src/gluonts/ev/__init__.py
@@ -62,6 +62,7 @@ from .metrics import (
     OWA,
     owa,
     SumNumMaskedTargetValues,
+    sum_num_masked_target_values,
 )
 from .aggregations import Aggregation, Sum, Mean
 
@@ -117,7 +118,7 @@ __all__ = [
     "Sum",
     "Mean",
     "SumNumMaskedTargetValues",
-    "num_masked_target_values"
+    "num_masked_target_values",
 ]
 
 
--- a/src/gluonts/ev/__init__.py
+++ b/src/gluonts/ev/__init__.py
@@ -24,6 +24,7 @@ from .stats import (
     scaled_interval_score,
     absolute_scaled_error,
     scaled_quantile_loss,
+    num_masked_target_values,
 )
 from .metrics import (
     Metric,
@@ -60,6 +61,7 @@ from .metrics import (
     MeanWeightedSumQuantileLoss,
     OWA,
     owa,
+    SumNumMaskedTargetValues,
 )
 from .aggregations import Aggregation, Sum, Mean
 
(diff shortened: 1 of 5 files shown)


src/gluonts/evaluation/_base.py
```diff
@@ -386,6 +386,7 @@ def get_base_metrics(
             "MASE": mase(pred_target, median_fcst, seasonal_error),
             "MAPE": mape(pred_target, median_fcst),
             "sMAPE": smape(pred_target, median_fcst),
+            "num_evaluations": pred_target.count(),
```
lostella: Not sure this is the best name. This is counting how many non-masked target values there are in a single test case: if `allow_nan_forecast` is `True`, then also those missing values will contribute to masking the evaluation, I think

lostella:
I would prioritize putting useful metrics into `gluonts.ev`, so we should at least add it there too (or only there: this evaluator class is really inflexible and we should deprecate it at some point)

melopeo:
@lostella , thanks for your comments!
Since the default evaluation framework is based on `gluonts.evaluation`, and hence this is used in production by default, I would prefer to focus on this one.

What are the plans to set the new framework as the default one?

src/gluonts/evaluation/_base.py
melopeo: I agree. I updated the code so that now we have the metric called `num_masked_target_values`, and defined it based on the case if the provided target is a masked array or not.

melopeo:
I just updated the PR so that `gluonts.ev` has the proposed metric as well.

lostella:
> What are the plans to set the new framework as the default one?

I don't think there's a default one, strictly speaking: you can use one or the other. The only place where the choice is hardcoded probably is `gluonts.shell`, which uses `gluonts.evaluation`. No plan to change that for the moment, but we probably should at some point.

src/gluonts/ev/__init__.py
```diff
@@ -114,6 +118,8 @@
     "Aggregation",
     "Sum",
     "Mean",
+    "SumNumMaskedTargetValues",
+    "sum_num_masked_target_values",
```
lostella, line 122: @jaheba I think this namespace is now becoming very confused: it's aggregating stuff from `ev.stats` and `ev.metrics`, and in this example now `num_masked_target_values` (function) is a very different thing from `sum_num_masked_target_values` (a `MetricDefinition` object).

I think we should reconsider what names are exposed: maybe just the ones from `ev.metric`?

jaheba: I guess there is an argument not to expose any of these?

This:

```py
from gluonts.ev.metrics import MASE
from gluonts.ev.stats import absolute_percentage_error
```

looks cleaner to me than:

```py
from gluonts.ev import MASE, absolute_percentage_error
```

lostella: Maybe; let's take this to a separate story I guess

lostella (approved):

PR #2924, opened 2023-06-20
Add `gluonts.itertools.Chain`

lostella:
*Description of changes:* Add a thin wrapper around `itertools.chain`. 

### DIFF
--- a/src/gluonts/itertools.py
+++ b/src/gluonts/itertools.py
@@ -143,7 +143,7 @@ class Chained:
         yield from itertools.chain(*self.iterables)
 
     def __len__(self) -> int:
-        return sum(len(iterable) for iterable in self.iterables)
+        return sum(map(len, self.iterables))
 
 
 @dataclass
--- a/src/gluonts/itertools.py
+++ b/src/gluonts/itertools.py
@@ -20,6 +20,7 @@ from dataclasses import dataclass, field
 from pathlib import Path
 from typing import (
     Callable,
+    Collection,
     Dict,
     Iterable,
     Iterator,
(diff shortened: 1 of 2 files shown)


src/gluonts/itertools.py
```diff
@@ -128,6 +129,23 @@ def get_batch():
...
+    This is a thin wrapper around ``itertools.chain``.
+    """
+
+    iterables: Collection[Iterable]
+
+    def __iter__(self):
+        yield from itertools.chain(*self.iterables)
+
+    def __len__(self) -> int:
+        return sum(len(iterable) for iterable in self.iterables)
```
jaheba:
```suggestion
        return sum(map(len, self.iterables))
```

src/gluonts/itertools.py
```diff
@@ -128,6 +129,23 @@ def get_batch():
     return iter(get_batch, [])
 
 
+@dataclass
+class Chained:
```
jaheba: I think it's nicer to use the same name as in `itertools`.

```suggestion
class Chain:
```

test/test_itertools.py
```diff
@@ -320,3 +321,18 @@ def test_random_yield(
...
+@pytest.mark.parametrize(
+    "iterables",
+    [
+        [[1, 2, 3], [4, 5, 6, 7]],
+    ],
+)
+def test_chained(iterables: List[Iterable]):
+    expected = list(itertools.chain(*iterables))
+    chained = Chained(iterables)
+    _ = list(chained)
```
jaheba: Why not just compare twice?

src/gluonts/itertools.py
lostella: @jaheba applied the change, although I had chosen `Chained` for grammatical consistency with `Cached`...

test/test_itertools.py
lostella: updated

src/gluonts/itertools.py
jaheba: Maybe `Cache` would also be better? `Cylic` should probably also be `Cycle`, but we have to start somewhere!

Also, I like `Chain` more than `Chained` regardless :).

jaheba (approved):
Thanks!

PR #2925, opened 2023-06-21
Fix JsonLinesFile slicing.

jaheba:
This also slightly reworks `start` on `JsonLinesFile`. It is now an index into `line_starts` instead of the actual offset.


lostella (approved):

lostella:
@jaheba please label as bugfix and for backport if it applies

PR #2926, opened 2023-06-21
Add splitting for Arrow based files.

jaheba:
Allow splitting of files into subsets which only yield values from respective subsets. That way one can gives different file chunks to different workers.


src/gluonts/dataset/arrow/file.py
```diff
@@ -71,6 +72,9 @@ class ArrowFile(File):
         default=None, init=False, repr=False
     )
 
+    start: int = 0
+    n: Optional[int] = None
```
lostella: Maybe a more descriptive name would be better

src/gluonts/dataset/arrow/file.py
```diff
@@ -129,6 +167,8 @@ def __getitem__(self, idx):
 class ArrowStreamFile(File):
     path: Path
     _decoder: Optional[ArrowDecoder] = field(default=None, init=False)
+    start: int = 0
+    n: Optional[int] = None
```
lostella: More descriptive name

lostella (commented):
Could you add a short description to the PR?

lostella (approved):

PR #2932, opened 2023-06-26
Remove `freq` argument in `SeasonalNaivePredictor`

lostella:
*Description of changes:* So far, `freq` was a mandatory argument in constructing `SeasonalNaivePredictor`. However, its role is not really clear this way: what is it used for? Can the predictor only treat data sampled at frequency `freq`? No, the frequency information is only used to infer the seasonality in case not directly provided.

This PR fixes this by removing `freq` altogether: now `season_length` is required.

### DIFF
--- a/src/gluonts/model/seasonal_naive/_predictor.py
+++ b/src/gluonts/model/seasonal_naive/_predictor.py
@@ -43,23 +43,25 @@ class SeasonalNaivePredictor(RepresentablePredictor):
 
     Parameters
     ----------
-    freq
-        Frequency of the input data
     prediction_length
-        Number of time points to predict
+        Number of time points to predict.
+    freq
+        Frequency of the input data, used to infer ``season_length``
+        in case it is not provided.
     season_length
-        Length of the seasonality pattern of the input data
+        Length of the seasonality pattern of the input data. If not
+        provided, it is inferred from ``freq``.
     imputation_method
         The imputation method to use in case of missing values.
-        Defaults to `LastValueImputation` which replaces each missing
+        Defaults to :py:class:`LastValueImputation` which replaces each missing
         value with the last value that was not missing.
     """
 
     @validated()
     def __init__(
         self,
-        freq: str,
         prediction_length: int,
+        freq: Optional[str] = None,
         season_length: Optional[int] = None,
         imputation_method: Optional[
             MissingValueImputation
@@ -67,6 +69,9 @@ class SeasonalNaivePredictor(RepresentablePredictor):
     ) -> None:
         super().__init__(prediction_length=prediction_length)
 
+        assert (freq is not None) or (
+            season_length is not None
+        ), "You must provide one of `freq` or `season_length`"
         assert (
             season_length is None or season_length > 0
         ), "The value of `season_length` should be > 0"
--- a/src/gluonts/model/seasonal_naive/_predictor.py
+++ b/src/gluonts/model/seasonal_naive/_predictor.py
@@ -46,11 +46,11 @@ class SeasonalNaivePredictor(RepresentablePredictor):
     prediction_length
         Number of time points to predict.
     freq
-        Frequency of the input data, used to infer ``season_length``
+        Sampling frequency of the input data, used to infer ``season_length``
         in case it is not provided.
     season_length
-        Length of the seasonality pattern of the input data. If not
-        provided, it is inferred from ``freq``.
+        Seasonality used to make predictions. If not provided, it is inferred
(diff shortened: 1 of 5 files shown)


src/gluonts/model/seasonal_naive/_predictor.py
```diff
@@ -43,30 +43,35 @@ class SeasonalNaivePredictor(RepresentablePredictor):
...
-        Frequency of the input data
     prediction_length
-        Number of time points to predict
+        Number of time points to predict.
+    freq
+        Frequency of the input data, used to infer ``season_length``
+        in case it is not provided.
     season_length
-        Length of the seasonality pattern of the input data
+        Length of the seasonality pattern of the input data. If not
```
lostella: Also, this is not accurate

melopeo (dismissed):

src/gluonts/model/seasonal_naive/_predictor.py
```diff
@@ -43,30 +43,35 @@ class SeasonalNaivePredictor(RepresentablePredictor):
...
         prediction_length: int,
+        freq: Optional[str] = None,
         season_length: Optional[int] = None,
         imputation_method: Optional[
             MissingValueImputation
         ] = LastValueImputation(),
     ) -> None:
         super().__init__(prediction_length=prediction_length)
 
+        assert (freq is not None) or (
```
jaheba: Not sure this would be nicer.

```py
maybe.or_(season_length, freq).expect("You must provide one of `freq` or `season_length`")
```

jaheba: What is the use of `freq`, if `season_length` is passed?

lostella: it's ignored (`season_length` takes precendence); another option is that not both can be specified

lostella: > Not sure this would be nicer.

not sure, I find the explicit assertion clearer

melopeo: I would keep the option of both parameters to be given, with `season_length` taking precedence.

jaheba: I think I actually like removing `freq` as a parameter and require setting `season_length` (why is it not called `seasonality`?).

That way things become much more explicit.

lostella:
Removed `freq` altogether, making `season_length` mandatory.

src/gluonts/model/seasonal_naive/_predictor.py
melopeo: Maybe I am wrong, but with this modification this model would be the only one that explicitly requires the parameter `season_length`, and I am not sure this is something we want. I agree that giving priority to `season_length` is something nice, but in many cases one computes `seasonal_naive` just as a reference without giving so much thought about the seasonality. For instance, in our metrics we use `calculate_seasonal_error` and I am not sure to what degree this would then also require an explicit value for `season_length`.

lostella: > but with this modification this model would be the only one that explicitly requires the parameter season_length

I think that’s fine. The only thing common to all models really is that they have a fixed prediction length; in general, I think it’s fine that models each have their own set of options, related to how they work internally.

This model has a very simple logic, entirely determined by `season_length`: specifying this directly has no ambiguity. Adding indirection to translate `freq` only hides a little bit what the model will end up doing, with no added flexibility.

Then there are the choices for the default seasonality for [hourly](https://github.com/awslabs/gluonts/blob/3ccb6d377a5bf9b27de74a47cdab295f4d61f7a7/src/gluonts/time_feature/seasonality.py#L26) and [daily](https://github.com/awslabs/gluonts/blob/3ccb6d377a5bf9b27de74a47cdab295f4d61f7a7/src/gluonts/time_feature/seasonality.py#L27) data, which are completely arbitrary and end up tying together the behavior of different models or components for no reason.

In summary, I tend to agree that if you want to use the naive seasonal method, you should bring your own seasonality parameter. We do provide some heuristics to get that from frequency, and one is free to use them, but should do that explicitly.

lostella:
I think naive2 could be simplified in a similar way, see https://github.com/awslabs/gluonts/blob/3ccb6d377a5bf9b27de74a47cdab295f4d61f7a7/src/gluonts/ext/naive_2/_predictor.py#L61-L80

jaheba:
@lostella Do you want to adjust naive2 in a separate PR?

jaheba (approved):

PR #2933, opened 2023-06-26
Add data_path to env.

jaheba:
Add a standard location to store data related to GluonTS. Add `GLUONTS_DATA` as an environment variable to set the location directly, but this defaults to `$XDG_DATA_HOME/gluonts` and if that not exists, `~/.local/share/gluonts`.

This ensures that we store data in a (somewhat) standardized location.

### DIFF
--- a/src/gluonts/env.py
+++ b/src/gluonts/env.py
@@ -13,8 +13,19 @@
 
 
 import os
+from pathlib import Path
 
 from .core.settings import Settings
+
+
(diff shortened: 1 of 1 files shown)


lostella:
This effectively changes the default path? Maybe it deserves a description :-)

jaheba:
Added, I will change the behaviour of the dataset repository in a separate PR.

lostella (approved):

PR #2937, opened 2023-06-27
Enable `time_feat` for zebras Periods.

jaheba:
`.values` is a pandas specific attribute. numpy uses `__array__` (among other things) to make `np.array(x)` work for custom types.

*Description of changes:*

### DIFF
--- a/src/gluonts/time_feature/_base.py
+++ b/src/gluonts/time_feature/_base.py
@@ -23,25 +23,31 @@ from pydantic import BaseModel
 TimeFeature = Callable[[pd.PeriodIndex], np.ndarray]
 
 
+def _normalize(xs, num: float):
+    """Scale values of ``xs`` to [-0.5, 0.5]."""
+
+    return np.asarray(xs) / (num - 1) - 0.5
+
+
 def second_of_minute(index: pd.PeriodIndex) -> np.ndarray:
     """
     Second of minute encoded as value between [-0.5, 0.5]
     """
-    return index.second.values / 59.0 - 0.5
+    return _normalize(index.second, num=60)
 
 
 def second_of_minute_index(index: pd.PeriodIndex) -> np.ndarray:
     """
     Second of minute encoded as zero-based index, between 0 and 59.
     """
-    return index.second.astype(float).values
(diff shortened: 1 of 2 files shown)


src/gluonts/time_feature/_base.py
```diff
@@ -23,113 +23,120 @@
...
-    return index.second.values / 59.0 - 0.5
+    return _normalize(index.second, num=60)
 
 
 def second_of_minute_index(index: pd.PeriodIndex) -> np.ndarray:
     """
     Second of minute encoded as zero-based index, between 0 and 59.
     """
-    return index.second.astype(float).values
+    return np.asarray(index.second, dtype="float")
```
lostella: elsewhere `float` is given instead of `"float"`

test/time_feature/test_features.py
```diff
@@ -80,16 +82,21 @@
...
 )
 def test_feature_normalized_bounds(
-    feature: TimeFeature, index: pd.PeriodIndex
+    feature: TimeFeature, pd_index: pd.PeriodIndex
 ):
-    values = feature(index)
-    assert isinstance(values, np.ndarray)
-    for v in values:
-        assert -0.5 <= v <= 0.5
+    zb_index = zb.periods(pd_index[0], pd_index.freqstr, len(pd_index))
```
lostella, line 87: So these tests now only check that the time features work with the zebras index, should we keep testing on the original pandas index?

jaheba: No? The below loop checks both pandas and zebras.

src/gluonts/time_feature/_base.py
jaheba: Something I was wondering, why is this using float at all?

test/time_feature/test_features.py
lostella: woops, missed it

src/gluonts/time_feature/_base.py
lostella: Right, this could probably be `int`

lostella (approved):

PR #2940, opened 2023-07-07
Add missing value imputation to COPDeepAR.

melopeo:
*Description of changes:*
Add the option to provide missing value imputation method to COPDeepAR.


src/gluonts/nursery/temporal_hierarchical_forecasting/model/cop_deepar/_estimator.py
```diff
@@ -208,6 +211,17 @@ def __init__(
 
         print(f"Distribution output: {base_estimator_hps['distr_output']}")
 
+        if "impute_missing_values" not in base_estimator_hps:
+            base_estimator_hps["impute_missing_values"] = impute_missing_values
+
+        if "imputation_method" not in base_estimator_hps:
+            base_estimator_hps["imputation_method"] = imputation_method
```
shubhamkapoor: One problem here is that the given imputation method will be applied to base estimators at all temporal levels. For eg. if you use `RollingMeanValueImputation`, each temporal level might need a different imputation window size that is suitable for its granularity.

melopeo: Thank you for this observation. I think we can handle this in basically the same way as we were handling the lags upper bound via `temporal_hierarchy.agg_multiples`.

shubhamkapoor (commented):
Nice! Can you also look into the failing tests. Otherwise LGTM.

melopeo:
Thank you!
The current style tests fail due to reasons independent of the current PR. Potentially due to mypy updates. @lostella Can you verify this? Otherwise we are ready to merge this one. Thanks!

shubhamkapoor (dismissed):

src/gluonts/nursery/temporal_hierarchical_forecasting/model/cop_deepar/_estimator.py
```diff
@@ -208,6 +211,17 @@ def __init__(
...
+
+        if "num_imputation_samples" not in base_estimator_hps:
+            base_estimator_hps[
```
jaheba: You can use `base_estimator_hps.setdefault("num_imputation_samples", num_imputation_samples)` instead.

src/gluonts/nursery/temporal_hierarchical_forecasting/model/cop_deepar/_estimator.py
```diff
@@ -172,6 +172,9 @@ def __init__(
         return_forecasts_at_all_levels: bool = False,
         naive_reconciliation: bool = False,
         dtype: Type = np.float32,
+        impute_missing_values: bool = False,
+        imputation_method: Optional[MissingValueImputation] = None,
```
jaheba, line 176: Why do we need two parameters for that, shouldn't `imputaion_method=None` mean no imputation?

src/gluonts/nursery/temporal_hierarchical_forecasting/model/cop_deepar/_estimator.py
```diff
@@ -224,6 +238,15 @@ def __init__(
...
             lags_seq = [lag for lag in lags_seq if lag >= num_nodes]
 
+            # adapt window_length if RollingMeanValueImputation is used
+            if isinstance(imputation_method, RollingMeanValueImputation):
+                base_estimator_hps_agg[
+                    "imputation_method"
+                ] = RollingMeanValueImputation(
+                    window_size=imputation_method.window_size // agg_multiple
+                )
+                print(base_estimator_hps_agg["imputation_method"])
```
jaheba: Let's remove the print.

src/gluonts/nursery/temporal_hierarchical_forecasting/model/cop_deepar/_estimator.py
melopeo: Here I am following how `DeepAR` does it: 

- https://github.com/awslabs/gluonts/blob/dev/src/gluonts/mx/model/deepar/_estimator.py#L181C13-L181C13

src/gluonts/nursery/temporal_hierarchical_forecasting/model/cop_deepar/_estimator.py
melopeo: Agreed!

src/gluonts/nursery/temporal_hierarchical_forecasting/model/cop_deepar/_estimator.py
melopeo: Agreed and updated accordingly.

jaheba (approved):

PR #2944, opened 2023-07-17
[Nursery] Cop-DeepAR: Fix evaluation at all levels

rshyamsundar:
*Issue #, if available:*

Allows the evaluation of time series at all levels of the temporal hierarchy. 

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Please tag this pr with at least one of these labels to make our release process faster:** BREAKING, new feature, bug fix, other change, dev setup

PR #2945, opened 2023-07-17
[Nursery] CopDeepAR: Remove freq argument

rshyamsundar:
*Issue #, if available:*

Remove the `freq` argument in the overriding predictor.
*Description of changes:*


src/gluonts/nursery/temporal_hierarchical_forecasting/train_and_eval_cop_deepar.py
```diff
@@ -22,6 +22,9 @@
 )
 
 
+EVALUATE_ALL_LEVELS = True
```
jaheba: Is this change unrelated?

melopeo: In the previous version this parameter was not defined, and it was leading to an error.

melopeo: @rshyamsundar If we keep `EVALUATE_ALL_LEVELS = False` then I think we can merge this quickly and independently from the PR https://github.com/awslabs/gluonts/pull/2944.

melopeo (approved):
LGTM

PR #2948, opened 2023-07-24
Ignore F403 errors in preludes.

jaheba:
`ruff` started to complain about our prelude imports using `*`. Since we don't actually want to import anything but just execute the prelude which executes handlers for `singledispatch` we just ignore `F403`.

See: https://www.flake8rules.com/rules/F403.html

---

### DIFF
--- a/src/gluonts/mx/prelude.py
+++ b/src/gluonts/mx/prelude.py
@@ -11,7 +11,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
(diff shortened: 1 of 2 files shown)


melopeo (approved):

lostella:
@jaheba let’s always have description in PRs, however short. Some context is always useful. Also labels: does this need to be backported?

jaheba:
Added description. I think we can backport this, but there is no rush since this doesn't just anything functional but makes the CI pass on the ruff check.

Issue #2947, opened 2023-07-22
BUG: RecursionError: maximum recursion depth exceeded while calling a Python object


Cameronwood611:
## Description
Getting recursion depth exceeded with what I believe is coming from gluon calling get_batch (see stack trace "[Previous line repeated 964 more times]").

## To Reproduce
This is an issue with the way the forcast_generator unpacks the batch. Only happens after calling model.predict several times and iterating through the returned generator.

This example isn't totally complete because you have to provide your own model, sorry.

```python
import random
import numpy as np
import pandas as pd
from gluonts.torch import PyTorchPredictor
from gluonts.dataset.common import ListDataset, DataEntry
from gluonts.torch import TemporalFusionTransformerEstimator


estimator = TemporalFusionTransformerEstimator(
freq='S',
context_length=10,
prediction_length=5,
trainer_kwargs={'max_epochs': 5}
)
train = ListDataset([{'target': np.array([random.uniform(-1, 1) for _ in range(50)]), 'start': pd.Period('01-01-2023', freq='S')} for _ in range(20)], freq='S')

pred = ListDataset([{'target': np.array([random.uniform(-1, 1) for _ in range(50)]), 'start': pd.Period('01-01-2023', freq='S')} for _ in range(20)], freq='S')
model = estimator.train(train, pred)

x, y, z = np.array([x for x in range(10)]), np.array([y for y in range (10)]), np.array([z for z in range(10)]) #for example
start = pd.Period('01-01-2023', freq='S')
_input = ListDataset([{'target': x, 'start': start},
                          {'target': y, 'start': start},
                          {'target': z, 'start': start}], freq='S')
for i in range(5000):
    pred = model.predict(_input)
    output = [sig.mean for sig in pred]  #this calls on a generator which causes the recursion error 

```

## Error message or code output
```
Exception in thread Controller:
Traceback (most recent call last):
  File "threading.py", line 980, in _bootstrap_inner
  File "project/event_loop.py", line 89, in run
  File "_project/_controller.py", line 54, in recv_msg
  File "project/controller.py", line 180, in recv_msg
  File "project/controller.py", line 320, in __predict_future_path
  File "project/predictor/periodic/predictor.py", line 38, in predict
    x, y, z = self.postprocess()
  File "project/predictor/periodic/predictor.py", line 87, in postprocess
    self.pred = [sig.mean_ts.to_numpy() for sig in self.pred] #type: ignore
[...]

PR #2951, opened 2023-07-28
Fix: prevent accumulation of `SelectFields` in `PyTorchPredictor`

Cameronwood611:
*Issue #, if available:* #2947

*Description of changes:* Have simple check to not accumulate the same field names, resulting in a RecursionError.

@abdulfatir @lostella Please review as this relates to changes you (co-)authored


lostella:
@Cameronwood611 thanks for this; it seems like this fix is breaking the behaviour of models, so I'm not sure it's the right way to go. Is there any way you could share the model you're using for #2947, for debugging purposes? Or even a simpler, smaller dummy model that just runs into the issue. What kind of model are we talking about?

Cameronwood611:
> @Cameronwood611 thanks for this; it seems like this fix is breaking the behaviour of models, so I'm not sure it's the right way to go. Is there any way you could share the model you're using for #2947, for debugging purposes? Or even a simpler, smaller dummy model that just runs into the issue. What kind of model are we talking about?

Yes! I updated the issue linked but I'll also post the example here --sorry about that. I noticed I had to raise the for loop above 1000 iterations (around the default recursion depth) for it to break; not sure if that's related to different model params, not loading in the model, etc. I'm not sure of the side effects this change has but maybe we could look into a certain filter or adding SelectFields into a set? Would be willing to help with this effort.

```python
import random
import numpy as np
import pandas as pd
from gluonts.torch import PyTorchPredictor
from gluonts.dataset.common import ListDataset, DataEntry
from gluonts.torch import TemporalFusionTransformerEstimator


estimator = TemporalFusionTransformerEstimator(
freq='S',
context_length=10,
prediction_length=5,
trainer_kwargs={'max_epochs': 5}
)
train = ListDataset([{'target': np.array([random.uniform(-1, 1) for _ in range(50)]), 'start': pd.Period('01-01-2023', freq='S')} for _ in range(20)], freq='S')

pred = ListDataset([{'target': np.array([random.uniform(-1, 1) for _ in range(50)]), 'start': pd.Period('01-01-2023', freq='S')} for _ in range(20)], freq='S')
model = estimator.train(train, pred)

x, y, z = np.array([x for x in range(10)]), np.array([y for y in range (10)]), np.array([z for z in range(10)]) #for example
start = pd.Period('01-01-2023', freq='S')
_input = ListDataset([{'target': x, 'start': start},
                          {'target': y, 'start': start},
                          {'target': z, 'start': start}], freq='S')
for i in range(5000): #used to be 1000 for it to break..
    pred = model.predict(_input)
    output = [sig.mean for sig in pred]  #this calls on a generator which causes the recursion error 

```

src/gluonts/torch/model/predictor.py
```diff
@@ -73,9 +74,11 @@ def network(self) -> nn.Module:
...
-            self.input_names + self.required_fields, allow_missing=True
-        )
+        if self.provided_fields:
+            self.input_transform += SelectFields(
+                self.input_names + self.required_fields, allow_missing=True
+            )
+            self.provided_fields = True
         inference_data_loader = InferenceDataLoader(
             dataset,
             transform=self.input_transform,
```
lostella: @Cameronwood611 the problem is with the `+=` which updates the predictor state every time `.predict` is invoked. Solution should be to either do here

```python
input_transform = self.input_transform + SelectFields(...)
inference_data_loader = InferenceDataLoader(
    ...,
    transform=input_transform,
    ....,
)
```

or directly in the constructor 

```
self.input_transform = input_transform + SelectFields(...)
```

and remove lines 76-78 from here.

lostella:
@Cameronwood611 thanks for the bug hunting and the fix, if you update it as suggested then it should work fine, and we'll backport it and release it soon

abdulfatir (approved):

Cameronwood611:
> @Cameronwood611 thanks for the bug hunting and the fix, if you update it as suggested then it should work fine, and we'll backport it and release it soon

Thank you!

PR #2952, opened 2023-07-28
[torch] Return a model even if callback has no best model path

kashif:
makes the `train` more robust especially if one passes `max_epoch=0` or `num_batches_per_epoch=0` to the estimators


lostella:
@kashif are `max_epochs==0` and `num_batches_per_epoch==0` the only occasions in which the path would be empty?

lostella (approved):

kashif:
thanks!

kashif:
yes those were the only 2 and perhaps if `batch_size=0`...

PR #2960, opened 2023-08-11
[Hierarchical] Allow computing projection based on a non-Euclidean norm

rshyamsundar:
*Issue #, if available:*

*Description of changes:*
Currently the forecasts are reconciled by projecting them on to the feasible set based on the Euclidean norm. Sometimes it is desirable to weigh the distances (errors) between the unreconciled and reconciled forecasts differently for each level (because there are more number of terms for the bottom level compared to the top level). 

This PR adds an option to provide such weights via a diagonal matrix. 

### DIFF
--- a/src/gluonts/mx/model/deepvar_hierarchical/_estimator.py
+++ b/src/gluonts/mx/model/deepvar_hierarchical/_estimator.py
@@ -73,7 +73,10 @@ def constraint_mat(S: np.ndarray) -> np.ndarray:
     return A
 
 
-def null_space_projection_mat(A: np.ndarray) -> np.ndarray:
+def null_space_projection_mat(
+    A: np.ndarray,
+    D: Optional[np.ndarray],
+) -> np.ndarray:
     """
     Computes the projection matrix for projecting onto the null space of A.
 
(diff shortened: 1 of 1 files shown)


melopeo (changes requested):
Thanks for this PR!

A question I have is: is there any way the user can provide the proposed matrix `D` through the estimator? Currently the matrix `S` is provided, which is used to compute the projection matrix `A`. I guess we need to provide the option to give the matrix `D` as a parameter in the estimator.

src/gluonts/mx/model/deepvar_hierarchical/_estimator.py
```diff
@@ -73,7 +73,10 @@ def constraint_mat(S: np.ndarray) -> np.ndarray:
     return A
 
 
-def null_space_projection_mat(A: np.ndarray) -> np.ndarray:
+def null_space_projection_mat(
+    A: np.ndarray,
+    D: Optional[np.ndarray],
```
melopeo, line 78: Why do we give this matrix as `optional`?
What are the advantages against `D: np.array = None`? With the current approach tests are failing.

rshyamsundar: ya, `=None` is missing actually. Let me fix.

rshyamsundar:
@melopeo Address them in a [new PR](https://github.com/awslabs/gluonts/pull/2961). Had to open a new one as I pushed this to a wrong branch.

PR #2961, opened 2023-08-11
[Hierarchical] Allow computing projection based on a non-Euclidean norm

rshyamsundar:
*Issue #, if available:*

*Description of changes:*
Currently the forecasts are reconciled by projecting them on to the feasible set based on the Euclidean norm. Sometimes it is desirable to weigh the distances (errors) between the unreconciled and reconciled forecasts differently for each level (because there are more number of terms for the bottom level compared to the top level).

This PR adds an option to provide such weights via a diagonal matrix.

### DIFF
--- a/src/gluonts/mx/model/deepvar_hierarchical/_estimator.py
+++ b/src/gluonts/mx/model/deepvar_hierarchical/_estimator.py
@@ -86,7 +86,8 @@ def null_space_projection_mat(
         The constraint matrix A in the equation: Ay = 0 (y being the
         values/forecasts of all time series in the hierarchy).
     D
-        Positive definite matrix (typically a diagonal matrix). Optional.
+        Symmetric positive definite matrix (typically a diagonal matrix).
+        Optional.
         If provided then the distance between the reconciled and unreconciled
         forecasts is calculated based on the norm induced by D. Useful for
         weighing the distances differently for each level of the hierarchy.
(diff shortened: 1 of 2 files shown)


melopeo:
Can we add tests to be sure that the new introduced projection matrix is indeed projecting into the subspace of reconciled samples?

Places where these tests could be added are:

- [test_coherency_error](https://github.com/awslabs/gluonts/blob/dev/test/mx/model/deepvar_hierarchical/test_coherency_error.py)
- [test_reconcile_samples](https://github.com/awslabs/gluonts/blob/dev/test/mx/model/deepvar_hierarchical/test_reconcile_samples.py)

src/gluonts/mx/model/deepvar_hierarchical/_estimator.py
```diff
@@ -82,14 +85,29 @@ def null_space_projection_mat(A: np.ndarray) -> np.ndarray:
...
     """
     num_ts = A.shape[1]
-    return np.eye(num_ts) - A.T @ np.linalg.pinv(A @ A.T) @ A
+    if D is None:
+        return np.eye(num_ts) - A.T @ np.linalg.pinv(A @ A.T) @ A
+    else:
+        assert np.all(
+            np.linalg.eigvals(D) > 0
+        ), "`D` must be positive definite."
+        D_inv = np.linalg.inv(D)
```
melopeo, line 110: This is a good assert check. Can we add an assert to verify that the matrix is symmetric?
If I understand correctly, the proposed oblique projection matrix is derived from the induced norm $\|\|x\|\|^2_A=x^TAx$. 
Using a check on symmetry before the check on eigenvalues would be good as the former is cheaper than the latter.

src/gluonts/mx/model/deepvar_hierarchical/_estimator.py
```diff
@@ -109,6 +127,12 @@ class DeepVARHierarchicalEstimator(DeepVAREstimator):
         Length of the prediction horizon
     S
         Summation or aggregation matrix.
+    D
+        Positive definite matrix (typically a diagonal matrix). Optional.
```
melopeo, line 134: Should we specify "symmetric" positive definite matrix?

melopeo (approved):

PR #2965, opened 2023-08-15
Simplify PyTorchPredictor serde

lostella:
*Description of changes:* Simplify the serialization format for `PyTorchPredictor`. Piggy-backing on the recursive definition of `encode` and `decode` [here](https://github.com/awslabs/gluonts/blob/3ae0ac464c844cddaa85f13adf35b347b66c6a3e/src/gluonts/core/serde/_base.py), this PR removes the need for dedicated file encoding the network and the input transform: these can be stored directly together with the rest of the constructor parameters.

### DIFF
--- a/src/gluonts/model/predictor.py
+++ b/src/gluonts/model/predictor.py
@@ -162,7 +162,6 @@ class RepresentablePredictor(Predictor):
         return equals(self, that)
 
     def serialize(self, path: Path) -> None:
-        # call Predictor.serialize() in order to serialize the class name
         super().serialize(path)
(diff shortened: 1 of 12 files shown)


lostella:
Alternatively, one could base `PyTorchPredictor` on [`RepresentablePredictor`](https://github.com/awslabs/gluonts/blob/3ae0ac464c844cddaa85f13adf35b347b66c6a3e/src/gluonts/model/predictor.py#L136-L178) and get serde (and `__eq__`) for free, at the cost of having to make the constructor `@validated`.

jaheba (approved):

PR #2970, opened 2023-08-17
Add truncated normal distribution for torch distributions

melopeo:
*Description of changes:*
Add truncated normal distribution for torch distributions


src/gluonts/torch/distributions/utils/truncated_normal.py
```diff
@@ -0,0 +1,193 @@
...
+        phi = 0.5 * (1 + (x * CONST_INV_SQRT_2).erf())
+        return phi.clamp(self.eps, 1 - self.eps)
+
+    @staticmethod
+    def _inv_big_phi(x):
+        return CONST_SQRT_2 * (2 * x - 1).erfinv()
+
+    def cdf(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
```
lostella: I think `_validate_args` is to signal that `a` and `b` in the constructor need to be validated, shouldn't have to do with validating that `value` here is admissible

src/gluonts/torch/distributions/utils/truncated_normal.py
```diff
@@ -0,0 +1,193 @@
...
+        if sample_shape is None:
+            sample_shape = torch.Size([])
+        shape = self._extended_shape(sample_shape)
+        p = torch.empty(shape, device=self.a.device).uniform_(
+            self._dtype_min_gt_0, self._dtype_max_lt_1
+        )
+        return self.icdf(p)
+
+
+class TruncatedNormal(TruncatedStandardNormal):
```
lostella: Shouldn't it be the other way around? `TruncatedNormal` has whatever `loc` and `scale`, while `TruncatedStandardNormal`  fixes `loc=0, scale=1`.

Why is `TruncatedStandardNormal` needed at all?

lostella (commented):
@melopeo why does `TruncatedNormal` appear twice, both in `distributions` and `distributions/utils`?

~~Also what is the logic? Is this simply truncating samples between the bounds (when sampling) and raising errors for values outside the bounds when computing log density? Then maybe one could have a generic wrapper to apply to any distribution.~~ (nevermind: I see there's more to it)

src/gluonts/torch/distributions/utils/truncated_normal.py
melopeo: Agreed and updated.

src/gluonts/torch/distributions/utils/truncated_normal.py
melopeo: Agreed. I simplified and merged to a single class.

src/gluonts/torch/distributions/truncated_normal.py
```diff
@@ -0,0 +1,295 @@
...
+CONST_INV_SQRT_2PI = 1 / math.sqrt(2 * math.pi)
+CONST_INV_SQRT_2 = 1 / math.sqrt(2)
+CONST_LOG_INV_SQRT_2PI = math.log(CONST_INV_SQRT_2PI)
+CONST_LOG_SQRT_2PI_E = 0.5 * math.log(2 * math.pi * math.e)
+
+
+class TruncatedNormal(Distribution):
+    """Truncated Normal distribution."""
+
+    """Implements a Truncated Normal distribution with location scaling.
```
lostella: Better have just one docstring. Maybe the first one can be removed?

src/gluonts/torch/distributions/truncated_normal.py
```diff
@@ -0,0 +1,295 @@
...
+        .. math::
+            loc = tanh(loc / upscale) * upscale.
+
+    This behaviour can be disabled by switching off the tanh_loc parameter (see below).
+
+
+    Args:
+        loc (torch.Tensor): normal distribution location parameter
+        scale (torch.Tensor): normal distribution sigma parameter (squared root of variance)
+        upscale (torch.Tensor or number, optional): 'a' scaling factor in the formula:
```
lostella: Could you adapt this to the numpy docstring style? See https://numpydoc.readthedocs.io/en/latest/format.html#parameters

That is

```
Parameters
----------
<name>
    <description>
...
```

lostella (changes requested):

src/gluonts/torch/distributions/truncated_normal.py
melopeo: Agreed and updated.

src/gluonts/torch/distributions/truncated_normal.py
melopeo: Agreed and updated.

lostella: I would also remove the types: just

```
<name>
    <description>
```

The reason is: types are already "documented" as type annotations, sometimes we change them slightly (like making them `Optional` or `Union`) and then the docstring and the type annotation go out of sync unless the docstring is updated accordingly

melopeo: Agreed and updated.

src/gluonts/torch/distributions/truncated_normal.py
```diff
@@ -0,0 +1,291 @@
...
+        normal distribution sigma parameter (squared root of variance)
+    min:
+        minimum value of the distribution. Default = -1.0
+    max:
+        maximum value of the distribution. Default = 1.0
+    upscale:
+        scaling factor. Default = 5.0
+    tanh_loc: if ``True``, the above formula is used for
+        the location scaling, otherwise the raw value is kept.
+        Default is ``False``
```
lostella:
```suggestion
    tanh_loc:
        if ``True``, the above formula is used for
        the location scaling, otherwise the raw value is kept.
        Default is ``False``
```

lostella (approved):
Looks great! Thank you!

PR #2973, opened 2023-08-21
Fix: use `isinstance` instead of type comparison

melopeo:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/core/serde/_base.py
+++ b/src/gluonts/core/serde/_base.py
@@ -315,7 +315,7 @@ def decode(r: Any) -> Any:
 
         assert cls is not None, f"Can not locate {r['class']}."
 
-        if kind == Kind.Type:
+        if isinstance(kind, Kind.Type):
             return cls
 
         args = decode(r.get("args", []))
@@ -331,10 +331,10 @@ def decode(r: Any) -> Any:
 
         raise ValueError(f"Unknown kind {kind}.")
 
-    if type(r) == dict:
+    if isinstance(r, dict):
         return valmap(decode, r)
 
-    if type(r) == list:
+    if isinstance(r, list):
         return list(map(decode, r))
 
     return r
--- a/src/gluonts/core/serde/_base.py
+++ b/src/gluonts/core/serde/_base.py
@@ -309,7 +309,7 @@ def decode(r: Any) -> Any:
     """
 
     # structural recursion over the possible shapes of r
-    if type(r) == dict and "__kind__" in r:
+    if isinstance(r, dict) and "__kind__" in r:
         kind = r["__kind__"]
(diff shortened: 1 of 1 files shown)


lostella (dismissed):

src/gluonts/core/serde/_base.py
```diff
@@ -309,13 +309,13 @@ def decode(r: Any) -> Any:
...
     # structural recursion over the possible shapes of r
-    if type(r) == dict and "__kind__" in r:
+    if isinstance(r, dict) and "__kind__" in r:
         kind = r["__kind__"]
         cls = cast(Any, locate(r["class"]))
 
         assert cls is not None, f"Can not locate {r['class']}."
 
-        if kind == Kind.Type:
+        if kind.value == Kind.Type.value:
```
abdulfatir: What happens if you reset only this line to what it was before?

src/gluonts/core/serde/_base.py
```diff
@@ -309,13 +309,13 @@ def decode(r: Any) -> Any:
...
     # structural recursion over the possible shapes of r
-    if type(r) == dict and "__kind__" in r:
+    if isinstance(r, dict) and "__kind__" in r:
         kind = r["__kind__"]
         cls = cast(Any, locate(r["class"]))
 
         assert cls is not None, f"Can not locate {r['class']}."
 
-        if kind == Kind.Type:
+        if kind == Kind.Type.value:
```
lostella: I think this is not needed

lostella (changes requested):
See comment

src/gluonts/core/serde/_base.py
melopeo: If I understand correctly, this line as it was originally, i.e. `if kind == Kind.Type:` is not passing the style checks. Any suggestions on how to fix it?

lostella: I think it's passing them, is just a comparison between strings as far as I understand

src/gluonts/core/serde/_base.py
melopeo: In the original state this line was not passing the style checks.

src/gluonts/core/serde/_base.py
melopeo: ok. I will reset it to the original case. Maybe I just got it wrong.

src/gluonts/core/serde/_base.py
abdulfatir: I think this line had no issues with the checks. It was only the `type() == ` lines.

abdulfatir (approved):

lostella (approved):

PR #2976, opened 2023-08-22
Fix ArrowDecoder.decode to return instead of yield

lostella:
*Description of changes:* `ArrowDecoder.decode` used to `yield from` something, effectively returning a generator. This meant for example that `ArrowFile.__getitem__` also returns a generator when given an integer index, [see here](https://github.com/awslabs/gluonts/blob/dev/src/gluonts/dataset/arrow/file.py#L162). This PR fixes it to make them return a record from the file instead.

### DIFF
--- a/src/gluonts/dataset/arrow/dec.py
+++ b/src/gluonts/dataset/arrow/dec.py
@@ -32,7 +32,7 @@ class ArrowDecoder:
         )
 
     def decode(self, batch, row_number: int):
-        yield from self.decode_batch(batch.slice(row_number, row_number + 1))
+        return next(self.decode_batch(batch.slice(row_number, row_number + 1)))
 
     def decode_batch(self, batch):
         for row in batch.to_pandas().to_dict("records"):
--- a/test/dataset/test_arrow.py
+++ b/test/dataset/test_arrow.py
@@ -90,7 +90,7 @@ def test_arrow(writer, flatten_arrays):
         assert len(dataset[10:]) == len(data[10:])
         assert_equal(dataset[10:], data[10:])
 
-        assert len(dataset[3:7]) == len(data[3: 7])
+        assert len(dataset[3:7]) == len(data[3:7])
         assert_equal(dataset[3:7], data[3:7])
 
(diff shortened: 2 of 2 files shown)


src/gluonts/dataset/arrow/dec.py
```diff
@@ -32,7 +32,7 @@ def from_schema(cls, schema):
         )
 
     def decode(self, batch, row_number: int):
-        yield from self.decode_batch(batch.slice(row_number, row_number + 1))
+        return next(self.decode_batch(batch.slice(row_number, row_number + 1)))
```
abdulfatir, line 35: Would this correctly return all elements in the batch?

lostella: No, just one

lostella:
@abdulfatir in fact, this was the previous behaviour of the `decode` method, which was changed in #2217

src/gluonts/dataset/arrow/dec.py
abdulfatir: Okay, and is that the correct behavior?

lostella: I think this is the intended behaviour, since the batch is being sliced to contain only one element; all elements in a batch (possibly containing a single element) can be generated via `decode_batch` directly, while this seems more of a conveniency method to directly access a particular element. A docstring would help, but this also seem to be the behaviour prior to #2217.

lostella: `ArrowFile.__getitem__` is unchanged since #2217, so I think it got broken there and went unnoticed

abdulfatir: Ah okay. This is only slicing one element.

abdulfatir: Should we add a test for `__iter__` and `__getitem__` behaviors?

lostella: Indeed: added assertion to existing test, failed before, passes now

abdulfatir (approved):

PR #2986, opened 2023-08-30
[shell] make ForecastConfig freq attribute explicit

melopeo:
*Description of changes:*
Currently `ForecastConfig` requires the attribute `freq`, but this is not stated explicitly. In this PR we make this attribute requirement explicit.


lostella:
@melopeo if this is not a bug fix (hence not urgent) I'd like to wait for @jaheba to chime so we can clarify, if that's ok

src/gluonts/shell/serve/app.py
```diff
@@ -37,6 +37,7 @@
 
 
 class ForecastConfig(BaseModel):
+    freq: str
```
jaheba, line 40: I think this has the downside that now `freq` is required for every requests, even if the given predictor has a fixed `freq`.

I guess one could set it to `None` by default.

melopeo: I am fine with that. I just think that setting this up explicitly will help the user to track quicker what is happening in case something is breaking.

jaheba: Fine with what?

jaheba (approved):

PR #2995, opened 2023-09-04
[torch-DeepAR] allow user to set `max_epochs`

melopeo:
*Description of changes:*
In torch-DeepAR it is currently hardcoded the value of `max_epochs`. In this PR we allow the user to set it up as desired.


abdulfatir:
I think it's not exposed by design. The idea is that the user needs to pass the `trainer_kwargs` argument to overwrite these defaults.

PR #2996, opened 2023-09-05
Add `auto` device for `PyTorchPredictor`

lostella:
*Description of changes:* Add an `auto` device that can be used in `PyTorchPredictor` through the `resolve_device` util. Use `auto` by default in estimators, so that we avoid the issue of serializing a model with `cuda` and then being unable to load it on a `cpu`-only machine.

### DIFF
--- a/src/gluonts/torch/util.py
+++ b/src/gluonts/torch/util.py
@@ -11,10 +11,24 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-from typing import List, Optional, Type
+from typing import List, Optional, Type, Union
 import inspect
 
 import torch
+
+
+def resolve_device(
+    device: Union[str, torch.device]
+) -> Union[str, torch.device]:
+    """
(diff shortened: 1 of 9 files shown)


src/gluonts/torch/util.py
```diff
@@ -11,12 +11,26 @@
...
+) -> Union[str, torch.device]:
+    """
+    Resolves a torch device to the most appropriate one.
+
+    The ``"auto"`` device is resolved to ``"cuda"`` if CUDA is available,
+    and to ``"cpu"`` otherwise. Otherwise the device is unchanged.
+    """
+    if isinstance(device, str) and device == "auto":
+        device = "cuda" if torch.cuda.is_available() else "cpu"
+    return device
```
jaheba:
```suggestion
    if device == "auto":
        if torch.cuda.is_available():
            return "cuda"
        return "cpu"

    return device
```

jaheba (approved):

PR #2997, opened 2023-09-06
Set number of torch threads to 1 in `gluonts.shell`

lostella:
*Description of changes:* The `gluonts.shell` multiprocessing setup interferes in a strange way with tensor operations from `torch`: when trying to serve an example model of mine, requests end up freezing the worker processes, even though the model works fine when invoked manually from the Python REPL.

I think this may be related to [this issue](https://github.com/pytorch/pytorch/issues/82843), and in fact the solution suggested there works.

### DIFF
--- a/src/gluonts/shell/__main__.py
+++ b/src/gluonts/shell/__main__.py
@@ -28,6 +28,13 @@ from .util import Forecaster, forecaster_type_by_name
 
 logger = logging.getLogger(__name__)
 
+try:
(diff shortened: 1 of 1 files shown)


lostella:
No need for this, better solution is to set `OMP_NUM_THREADS=1` when running the shell (thanks @jaheba!)

melopeo:
Where can how can we do that?

PR #3005, opened 2023-09-15
Make WaveNet `forward` configurable

lostella:
*Description of changes:* Expose additional arguments in `forward` to allow customizing prediction length, number of samples, and temperature at run time. Also rename some attributes from `n_xyz` to `num_xyz` for consistency with the rest of the library.

### DIFF
--- a/src/gluonts/torch/model/wavenet/estimator.py
+++ b/src/gluonts/torch/model/wavenet/estimator.py
@@ -73,11 +73,11 @@ class WaveNetEstimator(PyTorchLightningEstimator):
         self,
         freq: str,
         prediction_length: int,
-        n_bins: int = 1024,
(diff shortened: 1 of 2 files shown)


abdulfatir:
Just to confirm: these are only changes to Wavenet but there's actually no way to use them at the moment, correct? As far as I understand, passing these extra options would currently break the predictor.

lostella:
> Just to confirm: these are only changes to Wavenet but there's actually no way to use them at the moment, correct? As far as I understand, passing these extra options would currently break the predictor.

Correct: let's take this step by step

abdulfatir:
Thought: Now I know the answer. I need to approve this PR.

abdulfatir (approved):

PR #3017, opened 2023-10-16
add iTransformer multivariate forecaster

kashif:
Implemented the probabilistic iTransformer model from the paper: https://arxiv.org/pdf/2310.06625.pdf

Screenshot 2023-10-17 at 09 54 07


src/gluonts/torch/model/i_transformer/module.py
```diff
@@ -0,0 +1,158 @@
...
+    Parameters
+    ----------
+    prediction_length
+        Number of time points to predict.
+    context_length
+        Number of time steps prior to prediction time that the model.
+    distr_output
+        Distribution to use to evaluate observations and sample predictions.
+        Default: ``StudentTOutput()``.
+    """
```
lostella, line 59: Seems like a few arguments are missing

src/gluonts/torch/model/i_transformer/module.py
```diff
@@ -0,0 +1,158 @@
...
+
+from gluonts.core.component import validated
+from gluonts.model import Input, InputSpec
+from gluonts.torch.distributions import StudentTOutput
+from gluonts.torch.scaler import StdScaler, MeanScaler, NOPScaler
+
+
+class ITransformerModel(nn.Module):
+    """
+    Module implementing the iTransformer model for multivariate forecasting as described in
```
lostella, line 27: I’m wondering, does the model need to be for multivariate data necessarily? (Haven't read the paper yet)

kashif: yes has to be else we end up with a single vector... let me paste a picture in the issue

kashif:
@lostella I realize that the `input_size` is not really needed and we should not have it... this will allow us to train the same model on different-sized multivariate dim potentially...  I use `input_size` to reshape things but perhaps we can find another way.

lostella:
@kashif yes, if I read the code correctly you should be able to set `input_size = past_target.shape[2]`

test/torch/model/test_estimators.py
```diff
@@ -123,17 +131,35 @@ def test_estimator_constant_dataset(
     estimator_constructor, use_validation_data: bool
 ):
     constant = get_dataset("constant")
-
     estimator = estimator_constructor(constant)
 
+    if isinstance(estimator, ITransformerEstimator):
```
lostella: This will not scale, @kashif how about copying the test (it’s not so long anyway) for the multi variate setting? `test_estimator_multivariate_constant_dataset`

test/torch/model/test_estimators.py
```diff
@@ -263,19 +294,29 @@ def test_estimator_constant_dataset(
 def test_estimator_with_features(estimator_constructor):
     freq = "1h"
     prediction_length = 12
+    estimator = estimator_constructor(freq, prediction_length)
+
+    if isinstance(estimator, ITransformerEstimator):
```
lostella: Same as the above, I guess

examples/iTransformer.ipynb
lostella, line 1: I think this examples folder we should remove eventually. How about keeping his notebook close to the model? In its subpackage I mean. Pip install should anyway ignore it

lostella:
@kashif maybe for tests, we introduce a `test_multivariate_estimators.py` module?

examples/iTransformer.ipynb
kashif: this example folder is nice as its the only one with a multivariate model (showing the grouping and evaluation) as well as plotting some of the dims of the multivariate...

kashif:
ready!

lostella (approved):

lostella:
Thanks @kashif! 🚀

sirdawar:
Trying to use this sample with my own dataset, now when trying to use estimator I get following error:
GluonTSDataError: Input for field "target" does not have the requireddimension (field: target, ndim observed: 1, expected ndim: 2).

Where does it infer to have target dim of 2 and how to change that?

kashif:
you cannot change it as iTransformers is an inherently multivariate model... so the dataset has to be of shape: [batch_size, context_length, multivariate_Dim > 1]

kashif:
recall that the model takes a rep. of variate, makes a set of vectors out of them, and passes them to an unmasked transformer. The transformer as you know is a set-equivariant layer; since there is no ordering of the variates in a multivariate dataset, this layer makes perfect sense. However, iTransformer doesn't make sense in the univariate setting as there would be just a single vector as input to the transformer, which is then just an MLP...

sirdawar:
Thanks! One more thing, how to define the training set length, I am wondering if dataloader is sampling it somehow, since no matter how long time-series I put training time is same.

kashif:
right that is because while training a shared (global) deep learning models where the sizes of the time series can vary, the scheme is to randomly sample some fixed context-length window and subsequent prediction-length window within the training dataset... this way over time the model will learn via all the data. Use the `context_length` argument to the estimator to set this

sirdawar:
Okay, and how to make it learn more over time?

kashif:
do more steps of gradient descent and hope it doesn't overfit...

sirdawar:
Thanks so far! Training works, well but I get error when I want to evaluate it in the step 
```
forecasts = list(forecast_it)
targets = list(ts_it)
```
_IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed_

I am thinking that the way I create dataset might be wrong. For example my dataset goes like 

`<html>
<body>

Date | ObjectName | Consumption_Elec | Temperature | Consumption_Heat
-- | -- | -- | -- | --

2018-01-01 01:00:00 | Object1 | 17.16 | 0.0 | 0.1155
2018-01-01 01:00:00 |  Object1 | 17.16 | 0.0 | 0.2100
...

</body>
</html>`

And I am trying to create a dataset by 
`PandasDataset.from_long_dataframe(train_df, target=["Consumption_Heat","Consumption_Elec"], item_id="ObjectName",feat_dynamic_real='Temperature',timestamp="Date",freq=freq,unchecked=True)`

PR #3022, opened 2023-10-20
Fix WaveNet inputs

### LINKED ISSUE #3021: WaveNet model crashes if `num_feat_static_real > 0`
## Description
If I create `WaveNetEstimator` with `num_feat_static_real > 0`, the model will crash at the forward pass. This happens because `num_feat_static_real` is [used when defining the expected input shape](https://github.com/awslabs/gluonts/blob/dev/src/gluonts/torch/model/wavenet/module.py#L144), but actually isn't received by the model.

## To Reproduce

```python
import numpy as np
import pandas as pd
from gluonts.torch.model.wavenet import WaveNetEstimator

N = 50
freq = "D"
data = [
    {
        "target": np.random.rand(N),
        "feat_static_real": [5.0] * 5,
        "start": pd.Period("2020-01-01", freq=freq),
    }
]

model = WaveNetEstimator(freq=freq, prediction_length=5, num_feat_static_real=5, trainer_kwargs={"accelerator": "cpu"})
model.train(data)
```

```
RuntimeError: Given groups=1, weight of size [24, 39, 1], expected input[32, 35, 20] to have 39 channels, but got 35 channels instead
```

## Environment
- Operating system:
- Python version:
- GluonTS version:
- MXNet version:

shchur:
*Issue #, if available:* #3021

*Description of changes:*
- Pass `feat_static_real` to the WaveNet model
- Fix the `num_features` computation in WaveNet model to account for the observed value indicator

### DIFF
--- a/src/gluonts/torch/model/wavenet/estimator.py
+++ b/src/gluonts/torch/model/wavenet/estimator.py
@@ -54,6 +54,7 @@ from .lightning_module import WaveNetLightningModule
 
 PREDICTION_INPUT_NAMES = [
     "feat_static_cat",
+    "feat_static_real",
     "past_target",
     "past_observed_values",
     "past_time_feat",
--- a/src/gluonts/torch/model/wavenet/lightning_module.py
+++ b/src/gluonts/torch/model/wavenet/lightning_module.py
@@ -53,6 +53,7 @@ class WaveNetLightningModule(pl.LightningModule):
         Execute training step.
         """
         feat_static_cat = batch["feat_static_cat"]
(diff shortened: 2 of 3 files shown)


lostella:
@shchur I'm wondering whether [this test case](https://github.com/awslabs/gluonts/blob/dev/test/torch/model/test_estimators.py#L250-L260) should have caught the issue?

shchur:
@lostella Actually, I think that the model only breaks for `num_feat_static_real` > 1, so this test doesn't catch it.

lostella (approved):

PR #3026, opened 2023-10-24
Add support for Pydantic v1 and v2.

ddelange:
*Issue #, if available:*

*Description of changes:*

No breaking or functional changes, continues to use the v1 codebase that is shipped with pydantic v2.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -23,7 +23,7 @@ apidoc:
 		--module-first \
 		-o api/gluonts \
 		../src/gluonts \
-		../src/gluonts/nursery/*
+		../src/gluonts/nursery/* \
 		../src/gluonts/pydantic.py
 
 	@rm -Rf api/gluonts/modules.rst
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -24,7 +24,7 @@ apidoc:
 		-o api/gluonts \
 		../src/gluonts \
 		../src/gluonts/nursery/*
-                ../src/gluonts/pydantic.py
+		../src/gluonts/pydantic.py
 
 	@rm -Rf api/gluonts/modules.rst
 	@sed -i"" -e "s/gluonts package/API Docs/" api/gluonts/gluonts.rst
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -24,6 +24,7 @@ apidoc:
 		-o api/gluonts \
 		../src/gluonts \
 		../src/gluonts/nursery/*
+                ../src/gluonts/pydantic.py
 
 	@rm -Rf api/gluonts/modules.rst
 	@sed -i"" -e "s/gluonts package/API Docs/" api/gluonts/gluonts.rst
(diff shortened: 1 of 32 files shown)


lostella:
@ddelange thanks for this! I fixed a couple of details that were missing. Kindly asking @jaheba to chime in to review this

src/gluonts/meta/_version.py
jaheba, line 1: Can we have this in a separate PR?

jaheba:
Thanks for the PR!

Maybe it is easier to define a new `pydantic` module from which we import everywhere else. A bit like there is `gluonts.json` as a facade for different json implementations.

That way we might also be able to align on the type and thus make mypy happy.

src/gluonts/meta/_version.py
ddelange: https://github.com/awslabs/gluonts/pull/3028

ddelange:
Hi @jaheba 👋

I've moved the VCS commit out and added a commit with your suggestion. Please have a look:)

ddelange:
There's one mypy error left on my machine that seems unrelated to this PR (but may be due to the versions installed on my system):
```
src/gluonts/zebras/_time_frame.py:449: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
src/gluonts/zebras/_time_frame.py:453: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
src/gluonts/mx/model/gp_forecaster/gaussian_process.py:395: error: Argument 1 to "axis" has incompatible type "list[Any]"; expected "tuple[float, float, float, float] | bool | str | None"  [arg-type]
Found 1 error in 1 file (checked 249 source files)
```

edit: green on CI 🤔

jaheba:
Looks good to me!

The only thing I would add is an `__all__` in `pydantic.py`, which contains all exposed names. Then we also won't need the `# noqa` anymore.

jaheba:
@lostella Not sure we should backport this, since this could potentially be breaking?

lostella:
@jaheba 0.14 is still not out (only rc) so we can still allow some breaking maybe? how breaking is this?

ddelange:
@lostella no breaking changes here, `pydantic.v1` is a 1-1 replacement for v1 `pydantic` :)

jaheba (dismissed):

lostella:
merge pending additional changes from @jaheba to have the docs build ignore problematic modules/packages:
- `gluonts.pydantic`
- `gluonts.nursery`

jaheba:
Let's merge #3030 first, then add `pydantic` in the ignore section of the build as well.

lostella:
#3030 and #3031 are merged, @ddelange could you add similar changes to exclude the pydantic module from the docs build? That should get rid of the failures, and then we can merge 🚀

ddelange:
Like that?

lostella:
Amazing, thanks a lot for the PR @ddelange!

lostella (approved):

PR #3027, opened 2023-10-24
Make `mypy` checks opt-out instead of opt-in, fix several type issues

lostella:
*Description of changes:*
- Fix several type issues through the package
- Mark many other failing checks as "ignore", either because there's too much code to go through otherwise (rotbaum), or because it would be impractical to ensure checks passing
- Update setup.py to use an opt-out mechanism for type checking, instead of opt-in
- Remove all `.typesafe` files, add a `.typeunsafe` file to `gluonts.nursery` to opt it out
- Marking as BREAKING because some code moved from `gluonts.torch.distributions` to `gluonts.torch.model.mqf2`

### DIFF
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -29,7 +29,7 @@ logger = logging.getLogger(__name__)
 def _linear_interpolation(
     xs: np.ndarray, ys: np.ndarray, x: float
 ) -> np.ndarray:
-    assert sorted(xs) == xs
+    assert np.all(sorted(xs) == xs)
     assert len(xs) == len(ys)
     assert len(xs) >= 2
 
--- a/src/gluonts/model/forecast.py
+++ b/src/gluonts/model/forecast.py
@@ -14,7 +14,7 @@
 import re
 import logging
 from dataclasses import field
-from typing import Callable, Dict, List, Optional, Union, Tuple
+from typing import Callable, Dict, List, Optional, Union, Tuple, Sequence
 
 import numpy as np
 import pandas as pd
@@ -436,7 +436,7 @@ class SampleForecast(Forecast):
         self.samples = samples
         self._sorted_samples_value = None
         self._mean = None
-        self._dim = None
+        self._dim: Optional[int] = None
         self.item_id = item_id
         self.info = info
 
@@ -472,6 +472,7 @@ class SampleForecast(Forecast):
         """
         if self._mean is None:
             self._mean = np.mean(self.samples, axis=0)
+        assert self._mean is not None
         return self._mean
 
     @property
@@ -543,7 +544,7 @@ class SampleForecast(Forecast):
         return QuantileForecast(
             forecast_arrays=np.array(
                 [
-                    self.quantile(q) if q != "mean" else self.mean()
+                    self.quantile(q) if q != "mean" else self.mean
                     for q in quantiles
                 ]
             ),
@@ -596,7 +597,7 @@ class QuantileForecast(Forecast):
         ]
         self.item_id = item_id
         self.info = info
-        self._dim = None
+        self._dim: Optional[int] = None
 
         shape = self.forecast_array.shape
         assert shape[0] == len(self.forecast_keys), (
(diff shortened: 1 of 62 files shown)


lostella:
Todos:
- [x] get rid of a circular import I introduced (may need to break some import path there, but for good reasons)
- [x] solve mypy complaints about "duplicate modules" (see following, also [here](https://github.com/awslabs/gluonts/actions/runs/6627134500/job/18001440630?pr=3027#step:6:68))
```
src/gluonts/torch/model/deep_npts/__init__.py: error: Duplicate module named "gluonts.torch.model.deep_npts" (also at "/home/runner/work/gluonts/gluonts/src/gluonts/torch/model/deep_npts/__init__.py")
src/gluonts/torch/model/deep_npts/__init__.py: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules for more info
src/gluonts/torch/model/deep_npts/__init__.py: note: Common resolutions include: a) using `--exclude` to avoid checking one of them, b) adding `__init__.py` somewhere, c) using `--explicit-package-bases` or adjusting MYPYPATH
```
- [x] make the type-checking mechanism opt-out, and opt-out problematic packages in `gluonts.nursery`

src/gluonts/model/forecast.py
```diff
@@ -247,9 +247,12 @@ class Forecast:
     item_id: Optional[str]
     info: Optional[Dict]
     prediction_length: int
-    mean: np.ndarray
     _index = None
 
+    @property
+    def mean(self) -> np.ndarray:
+        raise NotImplementedError()
```
shchur, line 254: Nit:
```suggestion
        raise NotImplementedError
```

src/gluonts/torch/model/lightning_util.py
```diff
@@ -18,5 +18,6 @@
 
 def has_validation_loop(trainer: pl.Trainer):
     if version.parse(pl.__version__) < version.parse("2.0.0"):
+        assert hasattr(trainer._data_connector, "_val_dataloader_source")
```
shchur: We have changed the dependency range for lightning to `>=2.0`, can we just drop this check completely?

src/gluonts/torch/model/wavenet/estimator.py
```diff
@@ -96,7 +96,7 @@ def __init__(
         num_batches_per_epoch: int = 50,
         num_parallel_samples: int = 100,
         negative_data: bool = False,
-        trainer_kwargs: Dict[str, Any] = None,
+        trainer_kwargs: Dict[str, Any] = dict(),
```
shchur: I think we shouldn't use mutable types as default arguments. Maybe change this to 
```python
        trainer_kwargs: Optional[Dict[str, Any]] = None,
```
and also undo the change on L181 below?

src/gluonts/torch/model/wavenet/module.py
```diff
@@ -329,24 +326,20 @@ def base_net(
...
-                queues is not None
-            ), "Queues cannot be empty in prediction mode!"
-
         skip_outs = []
         queues_next = []
         out = inputs
         for i, layer in enumerate(self.residuals):
             skip, out = layer(out)
-            if prediction_mode:
+            if queues is not None:
```
shchur, line 342: That's a nice simplification!

src/gluonts/torch/model/wavenet/estimator.py
lostella: done

src/gluonts/torch/model/wavenet/module.py
lostella: and the fact is: type checking surfaced this, we should have it opt-out instead of opt-in

src/gluonts/torch/model/lightning_util.py
lostella: done

src/gluonts/model/forecast.py
lostella: Keeping it this way for consistency with the code below: if linters/style checkers are fine with it, I'm fine with it!

jaheba (dismissed):

jaheba (approved):

PR #3033, opened 2023-10-30
Allow function as `season_length` argument in `SeasonalNaivePredictor`

lostella:
*Description of changes:* The idea is to allow constructing the predictor with a "season length map" instead of a fixed integer, so that the model can be applied to datasets with different frequencies and expected seasonal patterns. For example, the predictor can be constructed by passing `season_length=gluonts.time_feature.seasonality.get_seasonality`.

### DIFF
--- a/src/gluonts/model/seasonal_naive/_predictor.py
+++ b/src/gluonts/model/seasonal_naive/_predictor.py
@@ -11,6 +11,8 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
+from typing import Callable, Union
+
 import numpy as np
 
 from gluonts.core.component import validated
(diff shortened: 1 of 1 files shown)


src/gluonts/model/seasonal_naive/_predictor.py
```diff
@@ -54,18 +59,25 @@ class SeasonalNaivePredictor(RepresentablePredictor):
...
 
         self.prediction_length = prediction_length
         self.season_length = season_length
         self.imputation_method = imputation_method
 
     def predict_item(self, item: DataEntry) -> Forecast:
+        if isinstance(self.season_length, int):
+            season_length = self.season_length
+        else:
+            season_length = self.season_length(item["start"].freq)
```
jaheba, line 79: This requires that there is a `freq` for the time series, which might not be the case in the future?

lostella: I guess this mechanism is really only useful if you have frequency information attached to the series. If that's not the case, there's no point in using this option, and the model should be configured with an `int` value.

Also: I believe the assumption that `item["start"]` is a `pd.Period` (hence has frequency information) is already baked into this method since it uses [`forecast_start`](https://github.com/awslabs/gluonts/blob/a25ca9dc61cb22f426dbc6004e19394b8130766e/src/gluonts/dataset/util.py#L21)?

jaheba: At the moment we require that all time series have frequency information attached -- but that's an arbitrary requirement.

jaheba (approved):

PR #3051, opened 2023-11-10
ev: Batch data for faster evaluation.

lostella:
*Description of changes:* Add batching to `gluonts.model.evaluation`, reaching speedups of ~6x in some cases (speedup depends also on how many metrics are being evaluated, more metrics => more speedup). The following example is using `m4_daily` data:

```
4227it [00:46, 91.83it/s]
evaluation (batch_size=1): 46.04444639701978
4227it [00:27, 152.03it/s]
evaluation (batch_size=2): 27.80498409701977
4227it [00:15, 277.03it/s]
evaluation (batch_size=5): 15.259674712986453
4227it [00:11, 371.23it/s]
evaluation (batch_size=10): 11.387762713013217
4227it [00:09, 451.05it/s]
evaluation (batch_size=20): 9.372666671988554
4227it [00:08, 512.34it/s]
evaluation (batch_size=50): 8.252159972995287
4227it [00:07, 546.00it/s]
evaluation (batch_size=100): 7.743695029988885
4227it [00:07, 561.83it/s]
evaluation (batch_size=200): 7.524872001988115
4227it [00:07, 576.23it/s]
evaluation (batch_size=500): 7.336895904998528
```

Code:

```python
import timeit
import numpy as np
import pandas as pd

from gluonts.ev.metrics import (
    SMAPE,
    MASE,
    NRMSE,
    ND,
    MeanWeightedSumQuantileLoss,
    AverageMeanScaledQuantileLoss,
    MAECoverage,
)
from gluonts.dataset.repository import get_dataset
from gluonts.model.evaluation import evaluate_forecasts
from gluonts.model import SampleForecast


quantile_levels = [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9]

METRICS = [
    SMAPE(),
    MASE(),
    NRMSE(),
    ND(),
    MeanWeightedSumQuantileLoss(quantile_levels=quantile_levels),
    AverageMeanScaledQuantileLoss(quantile_levels=quantile_levels),
    MAECoverage(quantile_levels=quantile_levels),
]


if __name__ == "__main__":
    data = list(get_dataset("m4_daily").test)
    test_data = split(data, offset=-14)[1].generate_instances(
        prediction_length=14
    )
    inputs = list(test_data.input)
    labels = list(test_data.label)

    forecasts = [
        SampleForecast(
            start_date=entry["start"],
            samples=np.random.normal(size=(100, 14)),
        )
        for entry in labels
    ]

    batch_sizes = [1, 2, 5, 10, 20, 50, 100, 200, 500]

    ref = None

    for batch_size in batch_sizes:
        t0 = timeit.default_timer()
        res = evaluate_forecasts(
            forecasts,
            test_data=test_data,
            metrics=METRICS,
            batch_size=batch_size,
            seasonality=7,
        )
        t1 = timeit.default_timer()
        if ref is None:
            ref = res
        else:
            pd.testing.assert_frame_equal(ref, res)
        print(f"evaluation ({batch_size=}): {t1 - t0}")
```

### DIFF
--- a/src/gluonts/model/evaluation.py
+++ b/src/gluonts/model/evaluation.py
@@ -23,9 +23,10 @@ from tqdm import tqdm
 
 from gluonts.dataset import DataEntry
 from gluonts.dataset.split import TestData
-from gluonts.model import Forecast, Predictor
 from gluonts.ev.ts_stats import seasonal_error
 from gluonts.itertools import batcher, prod
+from gluonts.model import Forecast, Predictor
+from gluonts.time_feature.seasonality import get_seasonality
 
 logger = logging.getLogger(__name__)
 
@@ -60,7 +61,7 @@ def _get_data_batch(
     input_batch: List[DataEntry],
     label_batch: List[DataEntry],
     forecast_batch: List[Forecast],
-    seasonality: int = 1,
+    seasonality: Optional[int] = None,
     mask_invalid_label: bool = True,
     allow_nan_forecast: bool = False,
 ) -> ChainMap:
@@ -74,11 +75,18 @@ def _get_data_batch(
 
     seasonal_error_values = []
     for input_ in input_batch:
+        seasonality_entry = seasonality
+        if seasonality_entry is None:
+            seasonality_entry = get_seasonality(input_["start"].freqstr)
         input_target = input_["target"]
         if mask_invalid_label:
             input_target = np.ma.masked_invalid(input_target)
         seasonal_error_values.append(
-            seasonal_error(input_target, seasonality=seasonality, time_axis=-1)
+            seasonal_error(
+                input_target,
+                seasonality=seasonality_entry,
+                time_axis=-1,
+            )
         )
     other_data["seasonal_error"] = np.array(seasonal_error_values)
 
@@ -96,7 +104,7 @@ def evaluate_forecasts_raw(
     batch_size: int = 100,
     mask_invalid_label: bool = True,
     allow_nan_forecast: bool = False,
-    seasonality: int = 1
+    seasonality: Optional[int] = None,
 ) -> dict:
     """
     Evaluate ``forecasts`` by comparing them with ``test_data``, according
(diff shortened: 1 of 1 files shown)


src/gluonts/model/evaluation.py
```diff
@@ -94,9 +93,10 @@ def evaluate_forecasts_raw(
     test_data: TestData,
     metrics,
     axis: Optional[Union[int, tuple]] = None,
+    batch_size: int = 100,
     mask_invalid_label: bool = True,
     allow_nan_forecast: bool = False,
-    seasonality: Optional[int] = None
+    seasonality: int = 1
```
abdulfatir: Why is the default being changed to 1?

src/gluonts/model/evaluation.py
```diff
@@ -130,16 +130,26 @@ def evaluate_forecasts_raw(
 
     index_data = []
 
-    for input_, label, forecast in tqdm(
-        zip(test_data.input, test_data.label, forecasts)
+    input_batches = batcher(test_data.input, batch_size=batch_size)
+    label_batches = batcher(test_data.label, batch_size=batch_size)
+    forecast_batches = batcher(forecasts, batch_size=batch_size)
+
+    pbar = tqdm()
```
abdulfatir, line 145: Can we also set the total length here?

src/gluonts/model/evaluation.py
lostella: Good point: I had initially done this because I was also computing the seasonal error in batches, that didn’t play well with potentially different seasonalities in a batch; turns out batching there is anyway problematic because of different length, so I should indeed be able to maintain the current behavior.

src/gluonts/model/evaluation.py
lostella: We could, but depending on what dataset type one is using, this may end up iterating the dataset to get its length. I don’t think the current code does it, so maybe let’s do it in a separate change.

src/gluonts/model/evaluation.py
lostella: Done

src/gluonts/model/evaluation.py
jaheba: I guess you could try to call `len` on it as long as it doesn't consume it?

jaheba (approved):

src/gluonts/model/evaluation.py
lostella: Yes, that's the idea. However this may still take time (say the dataset is huge and `len` needs to iterate it), which I would like to avoid at least in this PR

PR #3060, opened 2023-11-24
Fix `iterable.Cached`.

jaheba:
Previously, we only cached elements that were consumed on the first iteration.

### DIFF
--- a/src/gluonts/itertools.py
+++ b/src/gluonts/itertools.py
@@ -317,15 +317,19 @@ class Cached:
     """
 
     iterable: SizedIterable
+    _iterable: Iterable = field(init=False)
     cache: list = field(default_factory=list, init=False)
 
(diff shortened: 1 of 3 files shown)


src/gluonts/itertools.py
```diff
@@ -317,15 +317,19 @@ class Cached:
     """
 
     iterable: SizedIterable
+    _iterable: Iterable = field(init=False)
     cache: list = field(default_factory=list, init=False)
 
+    def __post_init__(self):
+        # ensure we only iterate once over the iterable
+        self._iterable = iter(self.iterable)
```
lostella: Then `_iterable` could be called `iterator` I guess

lostella (commented):
Beautiful! Thanks

lostella (approved):

PR #3061, opened 2023-11-24
Torch: Remove double caching of dataset.

jaheba:
*Issue #, if available:*

*Description of changes:*


lostella (changes requested):
Imports to be removed, otherwise looks good 👍 

Could this be backported as well?

jaheba:
I guess we should also backport this. No need to keep the dataset in memory twice

lostella (approved):

PR #3065, opened 2023-11-28
Fix Rotbaum to handle short series 

leica2023:
*Issue #, if available:*

*Description of changes:*
for items with less than forecast_horizon historical steps, jobs failed due to error 'Number of features of the model must match the input. Model n_features_ is 88 and input n_features is 60'. This PR fixes this error. 


src/gluonts/ext/rotbaum/_preprocess.py
```diff
@@ -454,7 +454,7 @@ def make_features(self, time_series: Dict, starting_index: int) -> List:
             prefix = [None] * abs(starting_index)
         else:
             prefix = []
-        time_series_window = time_series["target"][starting_index:end_index]
+        time_series_window = time_series["target"] if prefix else time_series["target"][starting_index:end_index]
```
lostella: I would include this in the previous `if else` above

lostella:
Thanks @leica2023! I think it would make sense to add a small test that reproduces the issue, and that would fail before this fix.

More at a high level: in this PR, the `make_features` method is being fixed to add padding whenever `target` is too short. This method appears to only be invoked a prediction time: is this going to work as intended? Will the model be trained to understand how to treat those padded series?

lostella:
Also fix the styling issue by applying `black`

leica2023:
> Thanks @leica2023! I think it would make sense to add a small test that reproduces the issue, and that would fail before this fix.
> 
> More at a high level: in this PR, the `make_features` method is being fixed to add padding whenever `target` is too short. This method appears to only be invoked a prediction time: is this going to work as intended? Will the model be trained to understand how to treat those padded series?

This

leica2023:
> Thanks @leica2023! I think it would make sense to add a small test that reproduces the issue, and that would fail before this fix.
> 
> More at a high level: in this PR, the `make_features` method is being fixed to add padding whenever `target` is too short. This method appears to only be invoked a prediction time: is this going to work as intended? Will the model be trained to understand how to treat those padded series?

Correct. The issue happens at prediction time for instances that have history length shorter than context_length. This method is also used in training[ data preparation](https://github.com/awslabs/gluonts/blob/7dbe5e6dc83a69ab6bafa9ff59bb062d3018df5b/src/gluonts/ext/rotbaum/_preprocess.py#L181),  but [this part](https://github.com/awslabs/gluonts/blob/7dbe5e6dc83a69ab6bafa9ff59bb062d3018df5b/src/gluonts/ext/rotbaum/_preprocess.py#L144) ensures ts is long enough to contribute to training samples.

src/gluonts/ext/rotbaum/_preprocess.py
leica2023: Sure, makes sense.

leica2023: Added.

lostella:
@leica2023 need to also format the test scripts:

```
black src test
```

src/gluonts/nursery/spliced_binned_pareto/run_model_example.ipynb
```diff
@@ -45,7 +45,7 @@
     "import os\n",
     "import pprint\n",
     "import random\n",
-    "from scipy import stats \n",
```
lostella: @leica2023 could you revert the changes to notebooks? No need to run `black` on them

test/ext/rotbaum/test_rotbaum_smoke.py
```diff
@@ -59,3 +61,69 @@ def test_rotbaum_smoke(datasets):
...
+                "feat_static_cat": np.array([0.0, 1.0], dtype=float),
+                "past_feat_dynamic_real": np.array(
+                    [[0 for i in range(5)], [750.0 for i in range(5)]]
+                ),
+            },
+        ],
+        freq=freq,
+    )
+
+    predictor = TreePredictor(
```
lostella, line 115: TreePredictor also needs importing?

test/ext/rotbaum/test_rotbaum_smoke.py
```diff
@@ -59,3 +61,69 @@ def test_rotbaum_smoke(datasets):
     predictor = estimator.train(dataset_train)
     forecasts = list(predictor.predict(dataset_test))
     assert len(forecasts) == len(dataset_test)
+
+
+def test_short_history_item_pred():
```
lostella, line 66: @leica2023 would this test fail before the fix contained in the PR? (That is, would it fail if run on the current `dev` branch)

leica2023: correct.

test/ext/rotbaum/test_rotbaum_smoke.py
leica2023: Added this and reverted ipynb styling.

PR #3067, opened 2023-11-30
add validation option to TrainDataset

melopeo:
*Description of changes:*

TrainDatasets currently have only fields for train and test datasets. In this commit we add the optional case for validation dataset.

### DIFF
--- a/src/gluonts/dataset/common.py
+++ b/src/gluonts/dataset/common.py
@@ -73,11 +73,12 @@ class SourceContext(NamedTuple):
 class TrainDatasets(NamedTuple):
     """
     A dataset containing two subsets, one to be used for training purposes, and
-    the other for testing purposes, as well as metadata.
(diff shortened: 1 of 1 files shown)


src/gluonts/dataset/common.py
```diff
@@ -427,6 +433,7 @@ def __call__(self, data: DataEntry) -> DataEntry:
 def load_datasets(
     metadata: Path,
     train: Path,
+    validation: Optional[Path],
```
lostella, line 436: This is breaking the tests I think

lostella:
Hey @melopeo, could you maybe update the OP to explain the motivation behind this? For example, why would one want to fix the validation data, when that could be considered a degree of freedom (some ways of setting it up may be better than others in some use cases; how to do that best may even depend on whether you use it for say model tuning vs early stopping)

PR #3068, opened 2023-12-01
Fix Rotbaum serialization and deserialization

pantanurag555:
Description of changes:
Prior to these changes, serializing a TreePredictor would only save the constructor arguments for the class. A trained TreePredictor creates a model_list within the predictor object that includes _n_ trained tree models for _n_ prediction length. This change pickles that model_list during serialization and reads it during deserialization. This change will enable Rotbaum to perform predictions after serialization and deserialization.
This request also includes a guardrail to prevent prediction failures after serialization and deserialization if dataset doesn't contain feat_static_cat and cardinality was initially set to `auto`. In scenarios of `auto` cardinality, PreprocessOnlyLagFeatures class infers the cardinality during training and sets it within the preprocess object in TreePredictor class. Pickling this object during serialization tends to be very memory intensive (100MB+). To avoid pickling this object, this guardrail also checks whether the dataset has feat_static_cat instead of just relying on the value of cardinality.

### DIFF
--- a/src/gluonts/ext/rotbaum/_predictor.py
+++ b/src/gluonts/ext/rotbaum/_predictor.py
@@ -340,6 +340,30 @@ class TreePredictor(RepresentablePredictor):
                 item_id=ts.get("item_id"),
             )
 
+    def serialize(self, path: Path) -> None:
+        """
+        This function calls parent class serialize() in order to serialize
+        the class name, version information and constuctor arguments. It
+        persists the tree predictor by pickling the model list that is
+        generated when pickling the TreePredictor.
+        """
+        super().serialize(path)
+        with (path / "predictor.pkl").open("wb") as f:
+            pickle.dump(self.model_list, f)
+
+    @classmethod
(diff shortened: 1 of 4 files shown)


src/gluonts/ext/rotbaum/_predictor.py
```diff
@@ -340,6 +343,31 @@ def predict(  # type: ignore
...
+    @classmethod
+    def deserialize(cls, path: Path, **kwargs: Any) -> "TreePredictor":
+        """
+        This function loads and returns the serialized model. It loads
+        the predictor class with the serialized arguments. It then loads
+        the trained model list by reading the pickle file.
+        """
+
+        with (path / "predictor.json").open("r") as fp:
+            predictor = load_json(fp.read())
```
lostella: This should work

```suggestion
        predictor = super().deserialize(path)
        assert isinstance(predictor, cls)
```

pantanurag555: Added in next commit.

lostella (approved):

PR #3093, opened 2023-12-27
Unify QuantileOutput and DistributionOutput

### LINKED ISSUE #3083: Unify QuantileOutput and DistributionOutput
## Description
`QuantileOutput` and `DistributionOutput` objects fulfill similar purposes and should be interchangeable in many cases. 

For example, it should be possible to train all models that produce a `QuantileForecast` (e.g., `TemporalFusionTransformer`) with a `DistributionOutput` head to produce a `DistributionForecast`. Similarly, it should be possible to train models that produce a `DistributionForecast` (`PatchTST`, `SimpleFeedForward`, `DLinear`) with a `QuantileOutput` head.

This feature doesn't affect models that produce a `SampleForecast` (e.g., DeepAR).

shchur:
*Issue #, if available:* closes #3083

*Description of changes:*
- Move loss computation into `Output` object and remove all losses defined in `gluonts.torch.module.loss`.
- Make following torch models compatible with both `DistributionOutput` and `QuantileOutput`:
  - `SimpleFeedForward`
  - `TemporalFusionTransformer`
  - `DLinear`
  - `PatchTST`
  - `LagTST`
- Change return type of the `forward` method for the following models to `Tuple[Tuple[Tensor, ...], Tensor, Tensor]`:
  - `MQCNN` (MXNet)
  - `MQRNN` (MXNet)
  - `TemporalFusionTransformer` (MXNet)
  - `TemporalFusionTransformer` (PyTorch)
- Update the logic inside `QuantileForecastGenerator` to support the new unified signature of `forward` method
- Replace `predict_to_numpy` method with `to_numpy` 

cc @kashif 


src/gluonts/torch/distributions/quantile.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def quantile_loss(self, value: torch.Tensor) -> torch.Tensor:
+        value = value.unsqueeze(-1)
+        return 2 * (
+            (value - self.outputs)
+            * ((value <= self.outputs).float() - self.quantiles)
+        ).abs().sum(dim=-1)
+
+
+class QuantileOutput(Output):
```
shchur: I am a bit unsure about the design here. The `QuantileOutput` object can, in principle, inherit from either
1. `DistributionOutput` or
2. `Output` 

## Justification for Option 1 (inherit from `DistributionOutput`)

Loss functions like [`torch.module.loss.QuantileLoss`](https://github.com/awslabs/gluonts/blob/dev/src/gluonts/torch/modules/loss.py#L85) expect a `torch.distributions.Distribution` object when computing loss. Therefore, we need to create a class `QuantileDistribution` (defined above) that holds the quantile predictions in it. 

This could imply that `QuantileOutput` should inherit from `DistributionOutput` since it produces a `QuantileDistribution` object.

## Justification for Option 2 (inherit from `Output`)

We can differentiate between `Output` and `DistributionOutput` objects as follows:

- `DistributionOutput` objects allow sampling. Therefore, they can be used in sample-based models like DeepAR as well as Seq2Seq models (e.g., TFT, PatchTST)
- `Output` may not necessarily allow sampling. Therefore, they are only suitable for Seq2Seq models (e.g., TFT, PatchTST)

I am now leaning towards Option 2

src/gluonts/model/forecast_generator.py
```diff
@@ -49,7 +49,7 @@ def log_once(msg):
 # numpy conversion differently, use a dispatching function to prevent needing
 # a ForecastGenerators for each framework
```
lostella: Nitpick: the docstring here could use an update

src/gluonts/torch/distributions/distribution_output.py
```diff
@@ -98,6 +102,27 @@ def get_args_proj(self, in_features: int) -> nn.Module:
...
+        scale: Optional[torch.Tensor] = None,
+    ) -> Distribution:
+        raise NotImplementedError()
+
+    @property
+    def value_in_support(self) -> float:
+        r"""
+        A float that will have a valid numeric value when computing the
+        log-loss of the corresponding distribution. By default 0.0.
+        This value will be used when padding data series.
```
lostella: I guess this could also be rephrased to be more precise:
- “loss” instead of “log loss” (not sure how log ended up there)
- No mention of how this value will be used (it’s the user’s concern)

src/gluonts/torch/distributions/quantile.py
lostella: I’m also leaning towards option 2, but this makes me think whether the `QuantileLoss` and `NegativeLogLikelihood` kind of classes are really useful: you can’t really mix and match them with different outputs, they are kind of tied

lostella (commented):
Thanks @shchur, left some comments but I’ll take a deeper look

src/gluonts/torch/distributions/quantile.py
shchur: I agree, it seems that most classes that inherit from `torch.distributions.Distribution` usually only implement one of `log_prob`, `quantile_loss`, `crps`, `energy_score`, so there is tight coupling between loss/output. Also, only `log_prob` is part of the `torch.distributions.Distribution` API, the rest of the methods are only available in GluonTS.

In theory, one could implement multiple methods for the same distribution object - but that would require a lot of work.

A better alternative could be defining `loss` method as part of an `Output` object, and setting it to NLL for `DistrbutionOutput`. This would, however, also require quite a few breaking changes to the API.

lostella: > A better alternative could be defining loss method as part of an Output object, and setting it to NLL for DistrbutionOutput.

Yes, that seems like a better approach. Then subclasses of `Output` define everything about the output projection and how it (and everything else) is learned.

As far as I can see, this looks much cleaner than how things are now (and worth the breaking change imo)

shchur: Okay, then I will update this PR to combine the Output class with loss calculation.

src/gluonts/model/forecast_generator.py
```diff
@@ -116,17 +115,21 @@ def __call__(
...
         for batch in inference_data_loader:
             inputs = select(input_names, batch, ignore_missing=True)
-            outputs = predict_to_numpy(prediction_net, inputs)
-            if output_transform is not None:
-                outputs = output_transform(batch, outputs)
+            (outputs,), loc, scale = prediction_net(*inputs.values())
+            if scale is not None:
+                outputs = outputs * scale[..., None]
+            if loc is not None:
+                outputs = outputs + loc[..., None]
```
lostella: I'm wondering: would it be better to turn `to_numpy` here? Otherwise the type of the output of `prediction_net` depend on the type of `prediction_net`, and we're using indexing/multiplication/addition without really knowing if it will work. Of course, it's just indexing/multiplication/addition, but I'm wondering if it would be somehow clearer if these objects were `np.ndarray` already here

src/gluonts/model/forecast_generator.py
```diff
@@ -116,17 +115,21 @@ def __call__(
...
+                log_once(OUTPUT_TRANSFORM_NOT_SUPPORTED_MSG)
             if num_samples:
                 log_once(NOT_SAMPLE_BASED_MSG)
 
             i = -1
-            for i, output in enumerate(outputs):
+            for i, output in enumerate(to_numpy(outputs)):
                 yield QuantileForecast(
-                    output,
+                    output.T,
```
lostella, line 133: why move the transposition here, from inside the model?

test/torch/model/test_modules.py
```diff
@@ -77,7 +78,7 @@ def assert_shapes_and_dtypes(tensors, shapes, dtypes):
             TemporalFusionTransformerModel(
                 context_length=24,
                 prediction_length=12,
-                quantiles=[0.2, 0.25, 0.5, 0.9, 0.95],
+                distr_output=QuantileOutput([0.2, 0.25, 0.5, 0.9, 0.95]),
```
lostella, line 81: I guess this class now also supports other parametric distribution families, right? Should some test case be added for that?

src/gluonts/torch/model/mqf2/distribution.py
```diff
@@ -292,7 +292,7 @@ def __init__(
         es_num_samples: int = 50,
         beta: float = 1.0,
     ) -> None:
-        super().__init__(self)
+        super().__init__(self, beta=beta)
```
lostella, line 295: Then I guess there's no need to set `self.beta` further below (line 306)

src/gluonts/model/forecast_generator.py
shchur: I agree, your suggestion feels cleaner. Implemented it.

src/gluonts/model/forecast_generator.py
shchur: QuantileForecast expects array of shape `[num_quantiles, prediction_length]`, but all models usually produce the output of shape `[batch_size, prediction_length, *additional_dims]` - both in case of `DistributionOutput` and `QuantileOutput`. I think it's better to keep model output shape consistent and do the transpose here.

src/gluonts/torch/model/mqf2/distribution.py
shchur: Good catch!

test/torch/model/test_modules.py
shchur: We have a similar test in https://github.com/awslabs/gluonts/pull/3093/files#diff-636e8492b38a2a549ef86320a8d00ca2b531aff6e31b4d61260ce6ed5ea7d132R104, do you think that's sufficient?

lostella: Yes, sorry, somehow I missed that

lostella (approved):

PR #3096, opened 2024-01-03
add TiDE model

maxc01:
*Description of changes:*

Implemented TiDE model for long-horizon forecasting: https://arxiv.org/abs/2304.08424

### DIFF
--- a/src/gluonts/torch/model/tide/estimator.py
+++ b/src/gluonts/torch/model/tide/estimator.py
@@ -121,7 +121,7 @@ class TiDEEstimator(PyTorchLightningEstimator):
         This must be set if ``num_feat_static_cat > 0`` (default: None).
     embedding_dimension
         Dimension of the embeddings for categorical features
-        (default: ``[min(16, (cat+1)//2) for cat in cardinality]``).
+        (default: ``[16 for cat in cardinality]``).
     layer_norm
         Enable layer normalization or not (default: False).
     lr
--- a/src/gluonts/torch/model/tide/estimator.py
+++ b/src/gluonts/torch/model/tide/estimator.py
@@ -22,10 +22,7 @@ from gluonts.dataset.field_names import FieldName
 from gluonts.dataset.loader import as_stacked_batches
 from gluonts.itertools import Cyclic
 from gluonts.model.forecast_generator import DistributionForecastGenerator
-from gluonts.torch.modules.loss import DistributionLoss, NegativeLogLikelihood
 from gluonts.time_feature import (
-    TimeFeature,
-    time_features_from_frequency_str,
     minute_of_hour,
     hour_of_day,
     day_of_month,
@@ -42,17 +39,12 @@ from gluonts.transform import (
     AsNumpyArray,
     AddObservedValuesIndicator,
     AddTimeFeatures,
-    AddAgeFeature,
     VstackFeatures,
     InstanceSplitter,
     ValidationSplitSampler,
     TestSplitSampler,
     ExpectedNumInstanceSampler,
-    MissingValueImputation,
-    DummyValueImputation,
     InstanceSampler,
-    InstanceSplitter,
-    SelectFields,
 )
 
 from gluonts.torch.model.estimator import PyTorchLightningEstimator
@@ -60,7 +52,6 @@ from gluonts.torch.model.predictor import PyTorchPredictor
 from gluonts.torch.distributions import (
     DistributionOutput,
     StudentTOutput,
-    NormalOutput,
 )
 
 from .lightning_module import TiDELightningModule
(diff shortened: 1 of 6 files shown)


src/gluonts/torch/model/tide/estimator.py
```diff
@@ -0,0 +1,419 @@
...
+    feat_proj_hidden_dim
+        Size of the feature projection layer (default: 40).
+    encoder_hidden_dim
+        Size of the dense encoder layer (default: 40).
+    decoder_hidden_dim
+        Size of the dense decoder layer (default: 40).
+    temporal_hidden_dim
+        Size of the temporal decoder layer (default: 40).
+    distr_hidden_dim
+        Size of the distribution projection layer (default: 40).
```
kashif: i think you can safely make this dim smaller by default.. since for point forecast it works when this is `1`, so it should also work fine with it say `10` or even `4`

src/gluonts/torch/model/tide/estimator.py
```diff
@@ -0,0 +1,419 @@
...
+        num_feat_static_real: int = 0,
+        num_feat_static_cat: int = 0,
+        cardinality: Optional[List[int]] = None,
+        embedding_dimension: Optional[List[int]] = None,
+        layer_norm: bool = False,
+        lr: float = 1e-3,
+        weight_decay: float = 1e-8,
+        scaling: Optional[str] = "mean",
+        distr_output: DistributionOutput = StudentTOutput(),
+        loss: DistributionLoss = NegativeLogLikelihood(),
```
lostella: Just a heads-up: in light of #3093, this should be removed (I'm reviewing #3093 first, we can then follow up on the required changes)

lostella:
@maxc01 PR #3093 was merged, could you update this model to reflect the changes there?

src/gluonts/torch/model/tide/estimator.py
lostella: @maxc01 what is the number of parameters in the default configuration?

kashif: the orignal model is not probablilistic

kashif: i did the same for patchTST too...

lostella: I mean total number of the model's parameters, in its layers

src/gluonts/torch/model/tide/module.py
```diff
@@ -0,0 +1,429 @@
...
+        decoder_hidden_dim: int,
+        temporal_hidden_dim: int,
+        distr_hidden_dim: int,
+        decoder_output_dim: int,
+        dropout_rate: float,
+        num_layers_encoder: int,
+        num_layers_decoder: int,
+        layer_norm: bool,
+        distr_output: DistributionOutput,
+        scaling: str,
```
lostella: I think it would be nice to have default values also at the module level, so that this class could be easily used directly. You can set the defaults that are implied by the default estimator config, maybe?

src/gluonts/torch/model/tide/estimator.py
maxc01: @kashif seems the default hidden dim is too hight... lowered it down to 4.

maxc01: @lostella the number of trainable parameters depends on the context size and the prediction length, there is no such a fixed number, for example, if prediction length=168 and context length is also 168, the trainable parameters is 133K.

src/gluonts/torch/model/tide/estimator.py
maxc01: @lostella adapted the changes.

src/gluonts/torch/model/tide/module.py
maxc01: Will this lead to some duplications? My understanding is that the estimator, which is also the interface to users,  is responsible for setting these parameters, while the model code in module.py takes whatever it receives from the interface (estimator.py). If a user is also expected to use this model in module.py directly, I totally agree that these default values should also be there. Could you give such a use case?

lostella:
@maxc01 a few style and type checks failing:

```
src/gluonts/torch/__init__.py:31:25: F401 [*] `.model.tide.TiDEEstimator` imported but unused
```

```
src/gluonts/torch/model/tide/module.py:292: error: Argument 1 to "sum" has incompatible type "Optional[List[int]]"; expected "Iterable[bool]"  [arg-type]
src/gluonts/torch/model/tide/module.py:301: error: Argument "embedding_dims" to "FeatureEmbedder" has incompatible type "Optional[List[int]]"; expected "List[int]"  [arg-type]
```

src/gluonts/torch/model/tide/module.py
lostella: I think it may make sense for a certain segmente of users, that just want the pytorch module and provide data loaders themselves for training/inference.

But for the PR it's fine to avoid the duplication, and maybe in the mean time we find a nice way to have shared default settings between estimator/module, across the library.

lostella (approved):
Thanks @maxc01 ! 🚀

PR #3112, opened 2024-01-29
CI: chain requirements to ensure compatibility

Borda:
*Issue #, if available:* atm, all install requirements are installed as stand-alone, which means if there is a version collision, `pip` does not care about it, and so for eventual version compatibility, a collision may not be discovered

*Description of changes:*
just install all needed requirements for the particular job at once so pip sees all required packages and finds an intersection among all :flags: 

cc: @jaheba @kashif @lostella kind ling as it seems there are no revivers by default :chipmunk: 


lostella:
@Borda thanks! I guess let's wait for all checks to be green before coming back to this

lostella (dismissed):
Thanks @Borda! Just needs rebasing

Borda:
> Just needs rebasing

@lostella done; just curious why rebasing and not just merging the `dev` branch?

lostella:
> > Just needs rebasing
> 
> @lostella done; just curious why rebasing and not just merging the `dev` branch?

whatever works :-) what I meant is to update the branch so that CI workflows run

requirements/requirements-rotbaum-extra-methods.txt
```diff
@@ -1,2 +1,2 @@
 lightgbm>=2.3.1
-scikit-garden>=0.1.2
+scikit-garden @ https://github.com/scikit-garden/scikit-garden/archive/refs/heads/master.zip
```
Borda: the last published version is from 2017 but the repo has some updates from 2021

Borda:
If possible, I would suggest dropping `scikit-garden` as it has been a stale project since 2017, and there seems to be no such intention for updating at least to installable without problems, see https://github.com/scikit-garden/scikit-garden/issues/105#issuecomment-1923509842

kashif:
we can also move the pytorch version to 2.x no?

Borda:
> we can also move the pytorch version to 2.x no?

probably but lets make it as separate PR :chicken:

.github/workflows/pypi.yml
```diff
@@ -16,10 +16,6 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: '3.8'
-    - name: Install dependencies
-      run: |
-        python -m pip install -U pip
-        python -m pip install setuptools wheel
```
lostella: I’m not sure about the changes here, can we revert these?

lostella:
> If possible, I would suggest dropping `scikit-garden` as it has been a stale project since 2017, and there seems to be no such intention for updating at least to installable without problems, see [scikit-garden/scikit-garden#105 (comment)](https://github.com/scikit-garden/scikit-garden/issues/105#issuecomment-1923509842)

Yeah we should get rid of that

.github/workflows/pypi.yml
```diff
@@ -16,10 +16,6 @@ jobs:
...
     - name: Create `dist` folder
```
Borda:
```suggestion
    - name: Install dependencies
      run: |
        python -m pip install -U pip
        python -m pip install setuptools wheel
    - name: Create `dist` folder
```

.github/workflows/pypi.yml
Borda: done

lostella (approved):
🚀

PR #3115, opened 2024-01-31
extend exception for invalid `offset_name`

Borda:
*Issue #, if available:* it is quite hard to understand why the exception was raised when you do not see the values

*Description of changes:* extend the exception, which will help a user understand what went wrong...

cc: @jaheba, @lostella, @kashif, @rshyamsundar

### DIFF
--- a/src/gluonts/time_feature/lag.py
+++ b/src/gluonts/time_feature/lag.py
@@ -146,7 +146,7 @@ def get_lags_for_frequency(
             + _make_lags_for_hour(offset.n / (60 * 60))
         )
     else:
-        raise Exception("invalid frequency")
+        raise ValueError(f"invalid frequency | `freq_str={freq_str}` -> `offset_name={offset_name}`")
 
     # flatten lags list and filter
     lags = [


kashif:
lets fix the issue, which is i believe that pandas has deprecated some of the freq strings. 
Screenshot 2024-01-29 at 11 44 14
Screenshot 2024-01-29 at 11 46 01

Borda:
> lets fix the issue, which is i believe that pandas has deprecated some of the freq strings.

That is a good point; if you do not mind, I will split it into two PRs as these springs are used in multiple places, and this PR keeps focused on the exception enrichment... So, I would take the conversion of the string just after :)

Borda:
@kashif, based on your reaction, if you are fine with keeping this PR simple, would you mind approving Ci and reviewing it? :)

kashif:
I am not a maintainer so have no such power... but I believe they will review it

src/gluonts/time_feature/lag.py
```diff
@@ -146,7 +146,7 @@ def _make_lags_for_month(multiple, num_cycles=3):
             + _make_lags_for_hour(offset.n / (60 * 60))
         )
     else:
-        raise Exception("invalid frequency")
+        raise ValueError(f"invalid frequency | `freq_str={freq_str}` -> `offset_name={offset_name}`")
```
lostella, line 149:
```suggestion
        raise ValueError(f"invalid frequency: {freq_str=}, {offset_name=}")
```

lostella:
@Borda thanks for this! Just left a comment on improving the message, let me know what you think.

There’s a lot of red in the workflows, I’ll take a look and try to unblock PRs

lostella (approved):

Borda:
> thanks for this! Just left a comment on improving the message, let me know what you think.

yes, that looks better, but seems you merged it before I could update it

> There’s a lot of red in the workflows, I’ll take a look and try to unblock PRs

PR #3111, opened 2024-01-29
migrate style check to precommit

Borda:
*Issue #, if available:* #2637 partially

*Description of changes:* moving some style checks to common pre-commit, which can be efficiently run by the developer locally and also for PR runs only on the diff, not the full project, so the iteration shall be much faster... In addition, a [bot](https://github.com/marketplace/pre-commit-ci) can be installed to this repository that may send some automated fixes related to the open PR so a developer will not think about styles too much :flamingo:  

BTW, it would be good to migrate also the `just license` so the whole workflow can be dropped

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

cc: @jaheba @kashif @lostella kind ling as it seems there are no revivers by default :chipmunk: 

**Please tag this pr with at least one of these labels to make our release process faster:** BREAKING, new feature, bug fix, other change, dev setup

PR #3131, opened 2024-02-20
lint: freeze & run Black version `24.02`

Borda:
*Issue #, if available:*
set a single black version to ensure reproducibility
UPDATE: seem that the latest Black would need to be applied

*Description of changes:*
freeze Black version, but the better way is in #3111

cc: @jaheba @kashif @lostella


Borda:
@jaheba @kashif @lostella all it is generated by Blck, but maybe you will feel more confident running/applying it on your own... :flamingo:

Borda:
This test is failing but only on py3.8 so it is strange and maybe flaky?
`FAILED test/mx/model/deep_factor/test_model.py::test_accuracy[constant-True] - assert 0.3266171349419488 <= 0.3`

lostella (approved):

lostella:
Flaky test

Borda:
> Flaky test

playing with flaky over here: #3132

.github/workflows/style_type_checks.yml
```diff
@@ -9,22 +9,16 @@ jobs:
...
         pip install .
-        pip install click black mypy
-        pip install types-python-dateutil
-        pip install types-waitress
-        pip install types-PyYAML
-    - name: Style and type checks
-      run: |
-        just black
-        just mypy
+        # todo: install also `black[jupyter]`
```
lostella, line 16: Why this todo?

.github/workflows/style_type_checks.yml
```diff
@@ -9,22 +9,16 @@ jobs:
...
+        pip install click "black==24.01" "mypy==1.8.0" \
```
lostella: I think it’s better to keep these to the latest version

lostella (changes requested):

.github/workflows/style_type_checks.yml
Borda: that is Back's suggesting, and I think it would be good to enable it
but as it was not enabled yet so, I did not want to introduce it without a discussion

.github/workflows/style_type_checks.yml
Borda:
```suggestion
        pip install click "black==24.02" "mypy==1.8.0" \
```

.github/workflows/style_type_checks.yml
Borda: so once it will be installed it will also lint ipython notebooks

Borda:
@lostella mind checking it again, pls :chipmunk: 
all seems to be :green_circle:

.github/workflows/style_type_checks.yml
lostella: @Borda I meant to just use `black` without version specifier

Borda: No, it is with specific pls see the actual PR's diff

Borda:
@lostella, if you would have a chance to check it, it would be great to finish it :flamingo:

Borda:
@lostella, is there still anything I need to change? :chipmunk:

lostella:
@Borda maybe it's better not to freeze the black version. For example, vulnerabilities are found in black versions prior to 24.3.0, see https://github.com/awslabs/gluonts/security/dependabot/17

Borda:
@lostella I see your point but from developer's perspective it is easier to have it frozen so you won't si sudden changes or broken lint, which would be unrelated to a contributor work, but instead move the black version to `requorements/lint.txt` and dependable which would periodically check and bump via PR so you will have it still up-to-date but fully under control... :) if you want I can do that in another PR :)

lostella:
> but instead move the black version to `requorements/lint.txt` and dependable which would periodically check and bump via PR so you will have it still up-to-date but fully under control

yes, I think that would be much better!

Borda:
> yes, I think that would be much better!

OK so, could we please land this with just update and I will prepare another with the automated process so it is separately and easier to review...

lostella (approved):

Borda:
@lostella in fact this pre-commit if the bot is installed can do this version check and update, see: #3111

PR #3151, opened 2024-03-23
Fix loader for the M5 dataset

shchur:
*Issue #, if available:*
- Fix loader for the M5 dataset

*Description of changes:*


src/gluonts/dataset/repository/_m5.py
```diff
@@ -61,6 +61,7 @@ def generate_m5_dataset(
             "d",
         ],
         axis=1,
+        errors="ignore",
```
lostella, line 64: is this needed?

lostella (approved):
Why the change to ignore errors?

PR #3155, opened 2024-04-02
Fix loaders for M5 & ETT datasets

shchur:
*Description of changes:*
- Fix how `item_id` is obtained for M5 and ETT datasets
- Fix `lxml` dependency range

### DIFF
--- a/requirements/requirements-docs.txt
+++ b/requirements/requirements-docs.txt
@@ -3,6 +3,7 @@ ipykernel~=6.5
 nbconvert~=6.5.1
 nbsphinx~=0.8.8
 notedown
+lxml~=5.1.0
 pytest-runner~=2.11
 recommonmark
 sphinx~=4.0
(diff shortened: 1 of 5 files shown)


src/gluonts/json.py
```diff
@@ -26,7 +26,8 @@
 character if set to `True`.
 """
 
-__all__ = [  # noqa
+# ruff: noqa: F822
```
shchur, line 29: New ruff version complains about this, requires a fix

requirements/requirements-docs.txt
```diff
@@ -3,6 +3,7 @@ ipykernel~=6.5
 nbconvert~=6.5.1
 nbsphinx~=0.8.8
 notedown
+lxml~=5.1.0
```
shchur, line 6: Newly released `lxml-5.2.0` breaks the doc build process, need to cap the dependency

canerturkmen (approved):

PR #3156, opened 2024-04-02
Fix item_id for M5 dataset

shchur:
*Description of changes:*
- Fix the `item_id` calculation for the M5 dataset in case `id` column is missing in the original dataset

### DIFF
--- a/src/gluonts/dataset/repository/_m5.py
+++ b/src/gluonts/dataset/repository/_m5.py
@@ -112,16 +112,17 @@ def generate_m5_dataset(
         len(state_ids_un),
     ]
 
-    # Build target series
-    train_ids = (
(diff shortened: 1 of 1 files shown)


src/gluonts/dataset/repository/_m5.py
```diff
@@ -112,16 +112,17 @@ def generate_m5_dataset(
         len(state_ids_un),
     ]
 
-    # Build target series
-    train_ids = (
-        sales_train_validation["item_id"].str
+    # Compute unique ID in case `id` column is missing
+    sales_train_validation["id"] = (
```
abdulfatir: Should we add a check for the existence of `id` col first?

abdulfatir (approved):

PR #3162, opened 2024-04-23
Add Seasonal Aggregate Predictor

rshyamsundar:
*Issue #, if available:*

*Description of changes:*
Another simple model whose prediction is an aggregate of past values according to the seasonality. The number of seasons as well as the aggregate function can be provided. The special case `num_seasons = 1` corresponds to the `SeasonalNaive` predictor.

Added several test cases for different aggregate functions including the following edge cases:
* not enough data is available,
* missing values are not imputed.


src/gluonts/model/seasonal_agg/_predictor.py
```diff
@@ -0,0 +1,130 @@
...
+        self,
+        prediction_length: int,
+        season_length: int,
+        num_seasons: int,
+        agg_fun: Callable = np.nanmean,
+        imputation_method: MissingValueImputation = LastValueImputation(),
+    ) -> None:
+        super().__init__(prediction_length=prediction_length)
+
+        assert (
```
shubhamkapoor, line 75: I didn't understand `not isinstance(season_length, int)` part of the assert statement. Shouldn't it be `isinstance(season_length, int) and season_length>0`?

Same for `num_seasons` as well.

src/gluonts/model/seasonal_agg/_predictor.py
```diff
@@ -0,0 +1,130 @@
...
+        self.season_length = season_length
+        self.num_seasons = num_seasons
+        self.agg_fun = agg_fun
+        self.imputation_method = imputation_method
+
+    def predict_item(self, item: DataEntry) -> Forecast:
+        if isinstance(self.season_length, int):
+            season_length = self.season_length
+        else:
+            season_length = self.season_length(item["start"].freq)
```
shubhamkapoor, line 93: Can `self.season_length` be a callable? Its type in the signature is `int`.

src/gluonts/model/seasonal_agg/_predictor.py
rshyamsundar: Here `season_length` can also be `Callable`, similar to the seasonal naive predictor. Updated the signature.

Addressed the issue for `num_seasons`, however. Thanks.

src/gluonts/model/seasonal_agg/_predictor.py
rshyamsundar: yes, updated the signature.

src/gluonts/model/seasonal_agg/_predictor.py
```diff
@@ -0,0 +1,133 @@
...
+        value with the last value that was not missing.
+    """
+
+    @validated()
+    def __init__(
+        self,
+        prediction_length: int,
+        season_length: Union[int, Callable],
+        num_seasons: int,
+        agg_fun: Callable = np.nanmean,
```
shubhamkapoor, line 70: Just an idea: I was thinking that do we need `agg_fun` here? Unlike `SeasonalNaive` where we have one sample values per prediction step, here we have a list of values coming from different `num_seasons` for each prediction step. This can be thought of as prediction samples. Since we are using `SampleForecast` object as the return type, can we simply apply these aggregations like `mean`, `median (P50)`, `min (P1)`, and `max (P100)` directly on the `SampleForecast` object.

rshyamsundar: I thought about it, but then the actual predictions of the model depends on some post-processing of `SampleForecast`, which might require some tinkering if we are evaluating it along with others. For example if we want to include the seasonal average method in some evaluation pipeline that accepts `Forecast` object and returns evaluation metrics, we need to tweak the output of `predict` method before evaluation otherwise the evaluation does not make sense.

I think it is better to have a simple contract for the models here: `predict` returns a `Forecast` object which in this case is a point forecast. We are using `SampleForecast` to represent point forecast is a different issue that we may need to address.

shubhamkapoor (approved):

src/gluonts/model/seasonal_agg/_predictor.py
```diff
@@ -0,0 +1,133 @@
...
+        Defaults to :py:class:`LastValueImputation` which replaces each missing
+        value with the last value that was not missing.
+    """
+
+    @validated()
+    def __init__(
+        self,
+        prediction_length: int,
+        season_length: Union[int, Callable],
+        num_seasons: int,
```
lostella, line 69: I think the naming here is confusing (same as in the naive seasonal predictor): `season_length` should really be `num_seasons` (how many seasons compose a period), and `num_seasons` should be something like `num_aggregations` or `aggregation_length` or something.

However I see the point in keeping this consistent with naive seasonal. Maybe though we rename `num_seasons` to something that relates it to "aggregation" more? I like `aggregation_length`

PR #3167, opened 2024-05-05
PatchTST: Add support for time features

rshyamsundar:
*Description of changes:*
Adding support for including time features in the `PachTST` model. The original `PatchTST` paper does not talk about time-based features and there are several ways one could include them.

Here we go with this simple approach (and maintain the simplicity of the `PatchTST` model):
* Similar to target time series, time series features, if provided, are converted to patches and appended to the target patches.
* Since `PatchTST` is encoder only model, we combine `past_time_feat` and `future_time_feat` (this dichotomy is an artefact of `gluonts`) and (left) shift them by `prediction_length` so that the features are aligned with the `target` input.

Added tests for the `PatchTSTEstimator` with features enabled.


kashif:
@rshyamsundar the "lag-tst" model can also be tried as it allows one to take those features naturally

melopeo (approved):
LGTM

PR #3186, opened 2024-05-31
Fix incorrect input routing for models

### LINKED ISSUE #3185: Error calling make_evaluation_predictions with TFT using past_feat_dynamic_real after update to 0.15.0
## Description
The issue is caused by changes to `QuantileForecastGenerator`. Some of the inputs to TFT forward method are optional. New `QuantileForecastGenerator` changed the way the input is passed to the model, breaking cases where some additional features are passed, but not all. In my case I had problems when dataset contains `past_feat_dynamic_real` without `feat_dynamic_cat`

New implementation of QuantileForecastGenerator calls:
```python
    inputs = select(input_names, batch, ignore_missing=True)
    (outputs,), loc, scale = prediction_net(*inputs.values())
```

This causes problems if some of the `input_names` are not present in dataset. The inputs are assigned to wrong function parameters. The problem can be fixed by changing the last line to:

```python
(outputs,), loc, scale = prediction_net(**inputs)
```
In the such case the names are passed along with the values, so that the inputs and parameters don't get mixed.

## To Reproduce

```python
# Using dataset like this will cause the described problem
static_feature_cols = ['sf1', 'sf2']
dynamic_feature_cols = ['df1', 'df2']

test = PandasDataset.from_long_dataframe(df, 
                                       target='target',
                                       item_id = "series_idx",
                                       timestamp= "date",
                                       static_feature_columns=static_feature_cols,
                                       past_feat_dynamic_real=dynamic_feature_cols,
                                       freq = "1D",
                                       )




```

```
TypeError: linear(): argument 'input' (position 1) must be Tensor, not NoneType
```

## Environment
- Operating system: Linux
- Python version: 3.8
- GluonTS version: 0.15.0
- MXNet version: ---

shchur:
Fixes #3185

*Description of changes:*

There is currently a bug where the model inputs may be routed incorrect by the forecast generator. This effectively results in `past_feat_dynamic_real` and `past_feat_dynamic_cat` being ignored by the TFT model.

MWE:
```python
from unittest import mock
import numpy as np
import pandas as pd
from gluonts.torch.model.tft import TemporalFusionTransformerEstimator

freq = "D"
N = 50
data = [
    {"target": np.arange(N), "past_feat_dynamic_real": np.random.rand(1, N).astype("float32"), "start": pd.Period("2020-01-01", freq=freq)}
]

predictor = TemporalFusionTransformerEstimator(prediction_length=1, freq=freq, past_dynamic_dims=[1], trainer_kwargs={"max_epochs": 1}).train(data)

with mock.patch("gluonts.torch.model.tft.module.TemporalFusionTransformerModel._preprocess") as mock_fwd:
    try:
        fcst = list(predictor.predict(data))
    except:
        pass   
    call_kwargs = mock_fwd.call_args[1]

call_kwargs["feat_dynamic_cat"]  
# tensor([[[0.8073]]])
call_kwargs["past_feat_dynamic_real"]  
# None
```

The bug occurs because model inputs are passed as positional arguments instead of keyword arguments.

### DIFF
--- a/src/gluonts/model/forecast_generator.py
+++ b/src/gluonts/model/forecast_generator.py
@@ -84,7 +84,8 @@ def make_distribution_forecast(distr, *args, **kwargs) -> Forecast:
 
 def make_predictions(prediction_net, inputs: dict):
     # MXNet predictors only support positional arguments
-    if prediction_net.__class__.__module__.startswith("gluonts.mx"):
+    class_name = prediction_net.__class__.__module__
+    if class_name.startswith("gluonts.mx") or class_name.startswith("mxnet"):
         return prediction_net(*inputs.values())
     else:
         return prediction_net(**inputs)
--- a/src/gluonts/model/forecast_generator.py
+++ b/src/gluonts/model/forecast_generator.py
@@ -115,7 +115,7 @@ class QuantileForecastGenerator(ForecastGenerator):
     ) -> Iterator[Forecast]:
         for batch in inference_data_loader:
             inputs = select(input_names, batch, ignore_missing=True)
-            (outputs,), loc, scale = prediction_net(*inputs.values())
+            (outputs,), loc, scale = prediction_net(**inputs)
             outputs = to_numpy(outputs)
             if scale is not None:
                 outputs = outputs * to_numpy(scale[..., None])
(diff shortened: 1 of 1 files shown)


src/gluonts/model/forecast_generator.py
```diff
@@ -82,6 +82,14 @@ def make_distribution_forecast(distr, *args, **kwargs) -> Forecast:
     raise NotImplementedError
 
 
+def make_predictions(prediction_net, inputs: dict):
+    # MXNet predictors only support positional arguments
+    if prediction_net.__class__.__module__.startswith("gluonts.mx"):
```
shchur: I couldn't find a more elegant way to use different logic for MXNet and PyTorch models :/

I tried `@singledispatch`, but that doesn't work for subclasses (i.e., we'd need to define it for all subclasses of `pl.LightningModule` in GluonTS, and same for MXNet).

lostella (approved):

PR #3218, opened 2024-09-13
implement GPD from wikipedia definition with 3 params

kashif:
Implemeneted the Generalized Pareto Distribution class from the Wikipedia definition and moved the original implementation as Exponentiated GP (ExGeneralizedPareto)


kashif:
@elenaehrlich would you have some time to review this PR?

PR #3228, opened 2024-10-17
Drop Python 3.8 from workflows, run on 3.11

lostella:
*Issue #, if available:* Python 3.8 reached EOL, and some packages are not distributed for it in their latest versions.

*Description of changes:*
- update github workflows to run on Python (up to) 3.11
- fix tests and testutils that would break on 3.11
- delete some unused workflows
- update sphinx version used in the docs build workflow


test/torch/distribution/test_negative_binomial.py
```diff
@@ -50,7 +50,7 @@ def test_custom_neg_bin_cdf(total_count, probs, value):
 @pytest.mark.parametrize("probs", [0.1, 0.5, 0.8])
 @pytest.mark.parametrize("total_count", [3, 7, 100])
 @pytest.mark.parametrize("value", [0.1, 0.5, 0.9])
-def test_custom_neg_bin_icdf(total_count, probs, value):
+def test_custom_studentt_icdf(total_count, probs, value):
```
canerturkmen: why studentt? (the function body refers to NegBin still.

.github/workflows/tests-prophet.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
       max-parallel: 4
       fail-fast: false
       matrix:
-        python-version: ['3.8']
+        python-version: ['3.12']
```
canerturkmen: prophet itself only appears to support until 3.11

https://github.com/facebook/prophet/blob/main/python/pyproject.toml

.github/workflows/tests-r.yml
```diff
@@ -8,7 +8,7 @@ jobs:
       max-parallel: 4
       fail-fast: false
       matrix:
-        python-version: ['3.9']
+        python-version: ['3.12']
```
canerturkmen: why 3.12 in some places and 3.11 in others? should 3.11 be our standard latest especially in extras?

test/torch/distribution/test_negative_binomial.py
lostella: ha! rebasing issue I suppose

.github/workflows/tests-r.yml
lostella: I think that makes sense, updating

test/core/test_serde_dataclass.py
```diff
@@ -20,14 +20,13 @@
 @serde.dataclass
 class Estimator:
     prediction_length: int
-    context_length: int = serde.OrElse(
-        lambda prediction_length: prediction_length * 2
-    )
```
lostella, line 25: Python 3.11 would complain here, that the default value in a dataclass cannot be mutable

test/conftest.py
```diff
@@ -30,7 +30,7 @@
 
 try:
     import mxnet as mx
-except ImportError:
```
lostella, line 33: The import above can raise `OSError` as well, so I'm catching that too to turn off `mx`

test/core/test_serde_dataclass.py
```diff
@@ -20,14 +20,13 @@
...
+    context_length: int = serde.EVENTUAL
 
     use_feat_static_cat: bool = True
     cardinality: List[int] = serde.EVENTUAL
 
-    def __eventually__(self, cardinality):
+    def __eventually__(self, context_length, cardinality):
```
canerturkmen, line 28: 🪄

canerturkmen (approved):
LGTM! Thanks!

PR #3233, opened 2024-11-04
Update gluonts.shell requirements

lostella:
*Issue #, if available:* 
- https://github.com/awslabs/gluonts/security/dependabot/31
- https://github.com/awslabs/gluonts/security/dependabot/32

*Description of changes:*
- Update waitress to the minimum recommended version as per the above issues
- Update flask to the latest release
- Enable gluonts.shell tests in the PyTorch tests workflow

### DIFF
--- a/.github/workflows/tests-torch.yml
+++ b/.github/workflows/tests-torch.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Install dependencies
       run: |
         pip install ".[arrow,torch,shell,test]" \
-          -r requirements/requirements-extras-m-competitions.txt \
+          -r requirements/requirements-extras-m-competitions.txt
         pip install --no-deps -r requirements/requirements-extras-cpflows.txt
     - name: Test with pytest
       run: |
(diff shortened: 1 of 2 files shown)


lostella:
This needs dropping Python 3.8 from workflows, so waiting on https://github.com/awslabs/gluonts/pull/3228

.github/workflows/tests-torch.yml
```diff
@@ -21,10 +21,9 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        pip install ".[arrow]" \
+        pip install ".[arrow,torch,shell]" \
```
abdulfatir, line 24: Why shell here?

lostella: To enable shell tests (they run only if these packages are there https://github.com/awslabs/gluonts/blob/dev/test/shell/require-packages.txt)

abdulfatir (approved):

PR #3234, opened 2024-11-05
Fix testutils for gluonts.shell on recent python versions

lostella:
*Issue #, if available:* Running tests for gluonts.shell raises
```
E   ValueError: mutable default <class 'gluonts.shell.serve.Settings'> for field settings is not allowed: use default_factory
```
on Python 3.11.

*Description of changes:* Fix the above issues. Also, one conditional mxnet import fails with an `OSError` on my machine which is not caught, so I'm adding it to the try/catch.

### DIFF
--- a/src/gluonts/testutil/shell.py
+++ b/src/gluonts/testutil/shell.py
@@ -21,7 +21,7 @@ import time
 import typing
 import waitress
 from contextlib import closing, contextmanager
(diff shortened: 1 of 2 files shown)


abdulfatir (approved):

lostella:
Superseded by fixes in #3228

PR #3259, opened 2025-06-24
Make observed_value_field optional in TFTInstanceSplitter

abdulfatir:
*Issue #, if available:* N/A

*Description of changes:* Previously, `TFTInstanceSplitter` assumed that the `observed_value_field` is always present in the entry. This PR fixes that. 

### DIFF
--- a/requirements/requirements-extras-statsforecast.txt
+++ b/requirements/requirements-extras-statsforecast.txt
@@ -1,4 +1,4 @@
 # scipy cap can be removed once this is resolved: https://github.com/statsmodels/statsmodels/issues/9584
-scipy<=1.15.3; python_version > "3.7.0"
+scipy<1.16.0; python_version > "3.7.0"
 scipy~=1.7.3; python_version <= "3.7.0"
 statsforecast~=1.0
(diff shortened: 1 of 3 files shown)


shchur (dismissed):
LGTM!

abdulfatir:
@shchur @lostella there's an issue with `statsmodels` accessing private API of `scipy` which is breaking our CI. Do the version caps look reasonable to you? See: https://github.com/statsmodels/statsmodels/issues/9584

requirements/requirements-extras-statsforecast.txt
```diff
@@ -1 +1,4 @@
+# scipy cap can be removed once this is resolved: https://github.com/statsmodels/statsmodels/issues/9584
+scipy<1.16.0; python_version > "3.7.0"
```
shchur, line 2: The version cap looks reasonable to me

requirements/requirements-extras-statsforecast.txt
```diff
@@ -1 +1,4 @@
...
+scipy~=1.7.3; python_version <= "3.7.0"
```
shchur, line 3: I would just skip this completely: Python 3.8 is already EOL for a while so we don't need to worry about it

shchur (approved):

requirements/requirements-extras-statsforecast.txt
lostella: I guess it's just a bit redundant, but since there may be EOL versions of Python running and doing stuff 🤖 doesn't hurt to leave it like this

lostella (approved):

PR #3265, opened 2025-08-13
Recover from a checkpoint if training failed with an exception

shchur:
*Issue #, if available:*

*Description of changes:*
- When training models that rely on the `StundentTOutput` (e.g., PatchTST, DeepAR), we occasionally run into numerical issues that turn all models weights to NaNs in the middle of the training:
    ```
    ValueError: Expected parameter df (Tensor of shape (64, 179)) of distribution Chi2() to satisfy the constraint GreaterThan(lower_bound=0.0), but found invalid values:
    tensor([[nan, nan, nan,  ..., nan, nan, nan],
            [nan, nan, nan,  ..., nan, nan, nan],
            [nan, nan, nan,  ..., nan, nan, nan],
            ...,
            [nan, nan, nan,  ..., nan, nan, nan],
            [nan, nan, nan,  ..., nan, nan, nan],
            [nan, nan, nan,  ..., nan, nan, nan]])
    ```
    Currently, this completely breaks the training process. 
    
    With this PR, we attempt to recover the model weights from a checkpoint if an exception is encountered during training.
 
    A potential alternative to this PR would be to fix the numerical issues plaguing the StudentT distribution by, e.g., clipping parameters like the scale. However, this might result in regressions for some existing users. Also, it's not very clear when and why this problem occurs, so even if we introduce clipping now, there might still be other datasets where it won't be sufficient.
- Bump `lightning` upper bound to `<2.6`.

### DIFF
--- a/requirements/requirements-pytorch.txt
+++ b/requirements/requirements-pytorch.txt
@@ -1,7 +1,7 @@
 torch>=1.9,<3
-lightning>=2.2.2,<2.5
+lightning>=2.2.2,<2.6
 # Capping `lightning` does not cap `pytorch_lightning`, so we cap manually
-pytorch_lightning>=2.2.2,<2.5
+pytorch_lightning>=2.2.2,<2.6
 # scipy cap can be removed once this is resolved: https://github.com/statsmodels/statsmodels/issues/9584
 scipy<1.16.0; python_version > "3.7.0"
 scipy~=1.7.3; python_version <= "3.7.0"
--- a/test/torch/model/test_estimators.py
+++ b/test/torch/model/test_estimators.py
@@ -20,6 +20,7 @@ import pytest
 import pandas as pd
 import numpy as np
 from lightning import seed_everything
+from lightning.pytorch.callbacks import Callback
 
 from gluonts.dataset.repository import get_dataset
 from gluonts.model.predictor import Predictor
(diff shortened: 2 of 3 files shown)


requirements/requirements-pytorch.txt
```diff
@@ -1,7 +1,7 @@
 torch>=1.9,<3
-lightning>=2.2.2,<2.5
+lightning>=2.2.2,<2.6
 # Capping `lightning` does not cap `pytorch_lightning`, so we cap manually
-pytorch_lightning>=2.2.2,<2.5
+pytorch_lightning>=2.2.2,<2.6
```
shchur, line 4: PyTorch tests for Python 3.11 fail with `torch-2.8`. Bumping lightning seems to fix the issue.

test/torch/model/test_estimators.py
```diff
@@ -400,3 +401,29 @@ def test_estimator_with_features(estimator_constructor):
...
+
+def test_estimator_recovers_if_exception_encountered_during_training():
+    class RaiseOnEpoch(Callback):
+        def __init__(self):
+            self.raised = False
+
+        def on_train_epoch_start(self, trainer, pl_module):
+            if trainer.current_epoch == 3:
+                self.raised = True
+                raise ValueError("Surprise!")
```
lostella, line 414: 🎉

lostella (approved):
Looks good to me, thanks @shchur!

PR #3279, opened 2026-02-20
docs(time_feature): clarify get_seasonality returns calendar defaults

### LINKED ISSUE #3278: Misleading function name: get_seasonality(freq) does not estimate or detect seasonality
I would like to raise a concern regarding the function:
gluonts.time_feature.seasonality.get_seasonality(freq)

The current name suggests that the function detects, estimates, or otherwise infers seasonality from data.

However, after reviewing the implementation, the function does not analyze any time series. It simply returns a hard-coded calendar heuristic based on the pandas frequency string.

For example:
"H" → returns 24
"15min" → returns 96
"M" → returns 12
"D" → returns 1

There is:
No ACF
No periodogram
No decomposition
No statistical inference of any kind

The function maps frequency aliases to predefined calendar constants and divides by the interval multiple. If the division is not exact, it falls back to 1.

This is not “getting seasonality” in any statistical sense. It is imposing a deterministic convention.

Why this matters

The returned value propagates into:
Seasonal naïve baselines
MASE scaling
Evaluation pipelines
Benchmark comparisons

Because of the function name, users may reasonably assume that some form of data-driven inference is taking place, which is not the case.

The concern is not the heuristic itself — deterministic defaults are often necessary in large-scale systems. The issue is that the naming suggests inference rather than a static rule.

Suggestion
Consider renaming the function to something more explicit, for example:
calendar_default_seasonal_period(freq)
default_seasonality_from_frequency(freq)
heuristic_seasonal_period(freq)

Alternatively, add explicit documentation clarifying that:
The function does not inspect data
The returned value is a calendar-based convention

It should not be interpreted as detected seasonality

Clear naming is especially important in forecasting libraries, where terminology carries strong statistical meaning.
Thank you for considering this clarification.

abhu85:
## Summary

This PR addresses the concern raised in #3278 that the `get_seasonality` function name is misleading.

The function name suggests it detects or estimates seasonality from data, but it actually returns predetermined calendar-based heuristics (e.g., hourly -> 24, monthly -> 12). Users may reasonably assume some form of data-driven inference is taking place, which is not the case.

## Changes

Added comprehensive documentation to both `get_seasonality` and `DEFAULT_SEASONALITIES` that explicitly clarifies:

- The function does **NOT** analyze time series data
- It returns static calendar conventions (not detected seasonality)
- Values are used for MASE scaling, seasonal naive baselines, etc.
- The returned value should not be interpreted as detected seasonality

## Why Documentation Instead of Renaming?

After reviewing the codebase, I found `get_seasonality` is imported and used in 20+ locations across evaluation metrics, model implementations, and tests. Renaming would be a breaking change affecting many users.

The documentation approach:
- Zero breaking changes
- Clearly communicates the function's behavior
- Follows NumPy-style docstring conventions
- Includes examples showing expected behavior

## Test Plan

- [x] All 26 existing seasonality tests pass
- [x] All 4 doctests in the updated module pass
- [x] No functional changes - only documentation added

```bash
pytest test/time_feature/test_seasonality.py -v  # 26 passed
python -m doctest src/gluonts/time_feature/seasonality.py -v  # 4 tests passed
```

Fixes #3278

### DIFF
--- a/src/gluonts/time_feature/seasonality.py
+++ b/src/gluonts/time_feature/seasonality.py
@@ -20,6 +20,20 @@ from gluonts.time_feature import norm_freq_str
 logger = logging.getLogger(__name__)
 
 
+#: Default calendar-based seasonality values for common pandas frequencies.
+#:
+#: These are heuristic defaults based on common calendar conventions, not
+#: values inferred from data. They are used when a seasonality parameter
+#: is required but no data-driven estimation is performed.
+#:
+#: - Seconds ("S", "s"): 3600 (one hour of seconds)
+#: - Minutes ("T", "min"): 1440 (one day of minutes)
+#: - Hours ("H", "h"): 24 (one day of hours)
+#: - Days ("D"): 1 (no sub-daily seasonality assumed)
+#: - Weeks ("W"): 1 (no sub-weekly seasonality assumed)
+#: - Months ("M", "ME"): 12 (one year of months)
+#: - Business days ("B"): 5 (one week of business days)
+#: - Quarters ("Q", "QE"): 4 (one year of quarters)
 DEFAULT_SEASONALITIES = {
     "S": 3600,  # 1 hour
     "s": 3600,  # 1 hour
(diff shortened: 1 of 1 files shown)


abhu85:
Friendly ping for review. This is a documentation-only change addressing the concern in #3278 that `get_seasonality` name is misleading - clarifies it returns calendar defaults, not detected seasonality.

PR #3280, opened 2026-03-01
Add QuantileOutput support to DeepAR

timoschowski:
## Summary

- Add support for `QuantileOutput` in DeepAR, enabling direct quantile regression (pinball loss) as an alternative to distribution-based outputs trained with NLL loss
- Refactor `forward()` into `_forward_distribution()` (existing path, unchanged) and `_forward_quantile()` (new path using median for autoregressive feedback)
- Wire `QuantileForecastGenerator` in the estimator's `create_predictor()` so end-to-end training and inference works out of the box
- Add unit tests and two comparison example scripts (synthetic data + electricity dataset)

## Motivation

DeepAR currently only supports distribution-based outputs (e.g. `StudentTOutput`, `NormalOutput`). This PR gives users a simpler, non-parametric alternative — `QuantileOutput` — that directly predicts quantile values via pinball loss, similar to what MQ-CNN already supports. This is useful when users want specific quantile forecasts without assuming a parametric distribution family.

## Usage

```python
from gluonts.torch.distributions.quantile_output import QuantileOutput

estimator = DeepAREstimator(
    freq="H",
    prediction_length=24,
    distr_output=QuantileOutput(quantiles=[0.1, 0.5, 0.9]),
)
predictor = estimator.train(training_data)
```

## Changes

| File | Change |
|------|--------|
| `src/gluonts/torch/model/deepar/module.py` | Widen `distr_output` type to `Output`, dispatch `forward()` to distribution/quantile paths, add `_forward_quantile()`, guard `output_distribution()` and `log_prob()` |
| `src/gluonts/torch/model/deepar/estimator.py` | Widen `distr_output` type, select `QuantileForecastGenerator` vs `SampleForecastGenerator` in `create_predictor()` |
| `test/torch/model/test_deepar_modules.py` | Add `test_deepar_quantile_output()` — shapes, loss, log_prob error |
| `examples/deepar_quantile_comparison.py` | Comparison on synthetic sine data: NormalOutput vs QuantileOutput |
| `examples/deepar_electricity_studentt_vs_quantile.py` | Comparison on electricity dataset: StudentTOutput vs QuantileOutput with GluonTS Evaluator |

## Design decisions

1. **Median for autoregressive feedback**: The P50 quantile (or closest to 0.5) is fed back into the RNN at each prediction step
2. **Return normalized predictions**: `_forward_quantile()` returns quantile values in scale-normalized space; `QuantileForecastGenerator` handles scale multiplication — consistent with how `QuantileOutput.loss()` already works
3. **`num_parallel_samples` ignored for QuantileOutput**: Quantile predictions are deterministic, no sampling needed
4. **Branching via `isinstance`**: Simple, localized, fully backward-compatible — no changes to the `Output` class hierarchy

## Test plan

- [x] `python -m pytest test/torch/model/test_deepar_modules.py -v` — all 11 tests pass (4 existing distribution tests + 6 RNN input tests + 1 new quantile test)
- [x] `examples/deepar_quantile_comparison.py` — trains both models, produces forecasts, prints metrics, saves plot
- [x] `examples/deepar_electricity_studentt_vs_quantile.py` — end-to-end on electricity dataset with GluonTS Evaluator metrics

🤖 Generated with [Claude Code](https://claude.com/claude-code)


timoschowski:
deepar_electricity_studentt_vs_quantile
deepar_quantile_comparison

carlocav:
Hello. Is it possible to use quantile loss for computing the validation loss and keep likelihood for the training loss? Deepar usually is unstable, and minimizing a validation quantile loss while still training the model to predict the full distribution rather than quantiles could be useful. This is implemented in nixtla neuralforecast which however lacks essential functions such as generating dependent sample paths

PR #3283, opened 2026-03-16
Update package to use pyproject.toml, support Python 3.10 to 3.14

lostella:
*Issue #, if available:*

*Description of changes:* The goal of this PR is to move completely to `pyproject.toml` for the project metadata, allowing to fully manage the project using `uv`.

- Remove `setup.py`, `setup.cfg`, `MANIFEST.in`, and `requirements/` in favor of `pyproject.toml`
- Update supported Python range to 3.10–3.14 across the board
- Remove the `nursery` optional-dependency extra; keep the nursery test workflow using the `torch` extra instead
- Fix `cpflows` dependency (upstream) that was pinning `six==1.15.0` (broken on Python 3.12)
- Bump minimum `pandas` to `>=1.4` (first version supporting Python 3.10)
- Update CI workflows to use `uv` instead of `pip`/`setup.py`
- Update actions versions: `actions/checkout@v7`, `astral-sh/setup-uv@v8.2.0`, `github/codeql-action@v3`, `extractions/setup-just@v2`,
  `aws-actions/configure-aws-credentials@v4`
- Remove obsolete workflows (mxnet nightly, win32 nightly, flake8)
- Add new `lints.yml` workflow (ruff + docformatter)
- Update code to fix formatting and type-checking issues
- Disable mxnet model tests
- Set package version on the dev branch to `0.17.0.dev0`, per [PEP 440 dev release
  conventions](https://packaging.python.org/en/latest/specifications/version-specifiers/)
- Update development docs (`devsetup.md`, `CONTRIBUTING.md`, `README.md`) to reflect `uv`-based workflow
- Update `dev_setup.sh` minimum Python check to 3.10

This will likely take a while to merge, given the size and since all GH workflows need to be updated.


pyproject.toml
```diff
@@ -1,5 +1,275 @@
+[project]
+name = "gluonts"
+version = "0.16.0"
```
lostella, line 3: This is wrong

pyproject.toml
```diff
@@ -1,5 +1,275 @@
...
+    "time series",
+    "forecasting",
+    "machine learning",
+    "deep learning",
+    "probabilistic forecasting",
+]
+
+dependencies = [
+    "numpy>=1.24,<2.2; python_version >= '3.12'",
+    "numpy>=1.16,<2.2; python_version < '3.12'",
```
lostella, line 38: These need updating

.devtools/githooks/pre-commit
```diff
@@ -36,7 +36,7 @@ def type_check() -> int:
                 returned a non-zero exit code. Fix the type errors listed above
                 and then run
 
-                    python setup.py type_check
+                    just mypy
```
shchur: Minor: Looks like we switched from using `just` in `.github/workflows/style_type_checks.yml`, but added `just` here. Should we stick to one option?

pyproject.toml
```diff
@@ -1,5 +1,263 @@
...
+    "holidays~=0.9",
+    "matplotlib~=3.6",
+]
+
+torch = [
+    "torch>=1.13,<3",
+    "lightning>=2.2.2,<2.6",
+    "pytorch_lightning>=2.2.2,<2.6",
+    "scipy<1.16.0",
+    "cpflows @ git+https://github.com/lostella/CP-Flow.git@gluonts-dependency",
```
shchur, line 106: Note for later: this looks dangerous 😬 - maybe completely drop this dependency (with an informative error msg inside the corresponding model) or move it to an extra?

shchur (dismissed):
LGTM!

.devtools/githooks/pre-commit
lostella: Good catch, my intention was to remove references to `just` from pretty much everywhere, except where it's really convenient (building docs). Will do a second pass for this

pyproject.toml
lostella: Yes, 100%. I think it's better to address this immediately: try-catch in the specific model on the `cpflows` import, put this as an extra dependency, somehow put a guard the tests as well

shchur (dismissed):
I didn't spot any issues, looks good

Issue #3282, opened 2026-03-16
[Security] tarfile.extractall without member validation


sudabg:
**Severity**: HIGH (Bandit B202 - tarfile_unsafe_members)
**File**: `src/gluonts/util.py`

## Vulnerability

`tarfile.extractall()` without member validation allows path traversal (zip slip attack).

## Fix

```python
import os

def _is_within_directory(directory, target):
    abs_directory = os.path.realpath(directory)
    abs_target = os.path.realpath(target)
    return abs_target.startswith(abs_directory + os.sep) or abs_target == abs_directory

def safe_extract(tar, path='.', members=None, *, numeric_owner=False):
    for member in tar.getmembers():
        member_path = os.path.join(path, member.name)
        if not _is_within_directory(path, member_path):
            raise Exception(f'Path traversal attempt: {member.name}')
    tar.extractall(path, members, numeric_owner=numeric_owner)
```

Replace `tar.extractall(...)` with `safe_extract(tar, ...)`.

## References
- CWE-22: Path Traversal
- Bandit B202

lostella:
The `gluonts.util` module is actually ok (safe_extract is already there). Other parts are not using it though. Addressing in #3285

PR #3285, opened 2026-03-17
Fix path traversal

lostella:
*Issue #, if available:* #3282

*Description of changes:*


abdulfatir (dismissed):
Looks good to me, but I am not sure if zipfile extractall actually is vulnerable. See: https://github.com/python/cpython/issues/84940

lostella:
Good find. Reverted the zip-related change (not broken, don't fix it)

PR #3288, opened 2026-04-03
Fix QuantileForecast for multivariate forecasting

kashif:
- Fix QuantileForecastGenerator to use `np.moveaxis` instead of `.T` for correct handling of 3D arrays `(prediction_length, target_dim, num_quantiles)`
- Add `copy_aggregate` method to `QuantileForecast` for `MultivariateEvaluator` support
- Add `test_copy_aggregate` for both `SampleForecast` and `QuantileForecast`

*Issue #, if available:*

*Description of changes:*


src/gluonts/model/forecast.py
```diff
@@ -644,6 +644,20 @@ def quantile(self, inference_quantile: Union[float, str]) -> np.ndarray:
                 inference_quantile,
             )
 
+    def copy_aggregate(self, agg_fun: Callable) -> "QuantileForecast":
+        if len(self.forecast_array.shape) == 2:
+            forecast_array = self.forecast_array
+        else:
+            # Aggregate over target dimension axis
+            forecast_array = agg_fun(self.forecast_array, axis=2)
```
lostella: I'm wondering whether the `else` branch here should only apply to the case where `len(self.forecast_array.shape) == 3`, else raise `ValueError`. Thoughts?

src/gluonts/model/forecast_generator.py
```diff
@@ -141,16 +145,24 @@ def __call__(
...
             i = -1
             for i, output in enumerate(outputs):
+                forecast_array = np.moveaxis(output, -1, 0)
+                forecast_keys = list(self.quantiles)
+                if self.median_idx is not None:
+                    forecast_array = np.concatenate(
+                        [forecast_array, forecast_array[self.median_idx:self.median_idx+1]],
+                        axis=0,
+                    )
+                    forecast_keys = forecast_keys + ["mean"]
```
lostella: I'm not a fan of this solution: I think `QuantileForecast` is meant to contain quantiles as the name suggests. If the median is to be used where the mean is actually needed, maybe it is up to the caller to decide that?

test/model/test_forecast.py
```diff
@@ -136,6 +136,28 @@ def test_copy_dim(name):
...
+def test_copy_aggregate(name):
+    forecast = MULTIVARIATE_FORECASTS[name]
+    agg_forecast = forecast.copy_aggregate(np.sum)
+
+    assert agg_forecast.dim() == 1
+    assert agg_forecast.start_date == forecast.start_date
+    assert agg_forecast.item_id == forecast.item_id
+    assert agg_forecast.info == forecast.info
+
+    if name == "SampleForecast":
```
lostella: Better (typechecker-friendly)

```suggestion
    if isinstance(forecast, SampleForecast) and isinstance(agg_forecast, SampleForecast):
```

test/model/test_forecast.py
```diff
@@ -136,6 +136,28 @@ def test_copy_dim(name):
...
+        assert np.array_equal(
+            agg_forecast.samples,
+            forecast.samples.sum(axis=2),
+        )
+    else:
```
lostella:
```suggestion
    elif isinstance(forecast, QuantileForecast) and isinstance(agg_forecast, QuantileForecast):
```

test/model/test_forecast.py
```diff
@@ -136,6 +136,28 @@ def test_copy_dim(name):
...
+        assert np.array_equal(
+            agg_forecast.forecast_array,
+            forecast.forecast_array.sum(axis=2),
+        )
+
```
lostella, line 171:
```suggestion
    else:
        raise ValueError("Wrong forecast types")
```

src/gluonts/model/forecast_generator.py
```diff
@@ -141,16 +145,24 @@ def __call__(
 
             i = -1
             for i, output in enumerate(outputs):
+                forecast_array = np.moveaxis(output, -1, 0)
```
lostella, line 150: Maybe worth adding a comment line here, since shape assumptions are somewhat implicit

```suggestion
                # Univariate case:
                #     (prediction_length, num_quantiles) -> (num_quantiles, prediction_length)
                # Multivariate case:
                #     (prediction_length, target_dim, num_quantiles) -> (num_quantiles, prediction_length, target_dim)
                forecast_array = np.moveaxis(output, -1, 0)
```

(Please validate that these are the intended shapes!)

src/gluonts/model/forecast.py
kashif: yes good idea instead of silently failing

src/gluonts/model/forecast_generator.py
kashif: ok makes sense, so now QuantileForecastGenerator only emits the quantiles it was asked for

src/gluonts/model/forecast_generator.py
```diff
@@ -115,6 +115,10 @@ class QuantileForecastGenerator(ForecastGenerator):
     @validated()
     def __init__(self, quantiles: List[str]) -> None:
         self.quantiles = quantiles
+        self.median_idx = next(
+            (i for i, q in enumerate(quantiles) if float(q) == 0.5),
+            None,
+        )
```
lostella: Not needed anymore, I think?

kashif:
btw i get tons of warning now when i run the multivariate evaluator:
```
Running evaluation: 7it [00:00, 109.58it/s]
Running evaluation: 0it [00:00, ?it/s]WARNING:gluonts.model.forecast:The mean prediction is not stored in the forecast data; the median is being returned instead. This behaviour may change in the future.
WARNING:gluonts.model.forecast:The mean prediction is not stored in the forecast data; the median is being returned instead. This behaviour may change in the future.
WARNING:gluonts.model.forecast:The mean prediction is not stored in the forecast data; the median is being returned instead. This behaviour may change in the future.
WARNING:gluonts.model.forecast:The mean prediction is not stored in the forecast data; the median is being returned instead. This behaviour may change in the future.
WARNING:gluonts.model.forecast:The mean prediction is not stored in the forecast data; the median is being returned instead. This behaviour may change in the future.
WARNING:gluonts.model.forecast:The mean prediction is not stored in the forecast data; the median is being returned instead. This behaviour may change in the future.
WARNING:gluonts.model.forecast:The mean prediction is not stored in the forecast data; the median is being returned instead. This behaviour may change in the future.
WARNING:gluonts.model.forecast:The mean prediction is not stored in the forecast data; the median is being returned instead. This behaviour may change in the future.
WARNING:gluonts.model.forecast:The mean prediction is not stored in the forecast data; the median is being returned instead. This behaviour may change in the future.
WARNING:gluonts.model.forecast:The mean prediction is not stored in the forecast data; the median is being returned instead. This behaviour may change in the future.
WARNING:gluonts.model.forecast:The mean prediction is not stored in the forecast data; the median is being returned instead. This behaviour may change in the future.
WARNING:gluonts.model.forecast:The mean prediction is not stored in the forecast data; the median is being returned instead. This behaviour may change in the future.
WARNING:gluonts.model.forecast:The mean prediction is not stored in the forecast data; the median is being returned instead. This behaviour may change in the future.
WARNING:gluonts.model.forecast:The mean prediction is not stored in the forecast data; the median is being returned instead. This behaviour may change in the future.
Running evaluation: 7it [00:00, 107.79it/s]
...
```

lostella:
I guess this is what I meant: the caller does the imputation median -> mean if they want. The issue is that too many warnings are produced. There should be a way for the warning to appear only once, I suppose

kashif:
yes let me make it a warn once

kashif:
@lostella  ok now we warn once

kashif:
@lostella  gentle reminder thanks!

lostella (approved):

kashif:
thank you!

Issue #3263, opened 2025-07-19
bug: unexpected in-place modification of dataframe index inside class method


AzulGarza:
hey! just a heads up, these lines:

https://github.com/awslabs/gluonts/blob/7bb6800a149a6caac7ec1725132c73bf06a57e2b/src/gluonts/dataset/pandas.py#L284-L286

are modifying the dataframe in-place by setting its index. since this happens inside a class method, it might be unexpected for whoever passed the dataframe, especially if they rely on its original structure later. in my case, i iterate over the same dataframe later on, but after calling the method (second print), the index has been altered, which breaks assumptions about its structure. 

<img width="479" height="244" alt="Image" src="https://github.com/user-attachments/assets/c32bf11d-4197-4447-ba61-3a3750e41f14" />

to avoid side effects, we probably want to work on a copy instead (or at least document that this changes the input). otherwise, it can silently break downstream code.

PR #3298, opened 2026-05-09
PandasDataset.from_long_dataframe: don't mutate caller's DataFrame index

jbbqqf:
*Issue #, if available:* #3263

*Description of changes:*

## Summary

`PandasDataset.from_long_dataframe` mutates the caller's `DataFrame.index` as a side effect. This PR takes a shallow copy at the top of the constructor so the caller's frame is left intact, and pins the contract with two regression tests.

## Context

Reported in #3263. Both lines flagged in the issue are visible side effects on the caller-owned object:

```python
# src/gluonts/dataset/pandas.py (before)
if timestamp is not None:
    dataframe.index = pd.to_datetime(dataframe[timestamp])  # mutates caller

if not isinstance(dataframe.index, DatetimeIndexOpsMixin):
    dataframe.index = pd.to_datetime(dataframe.index)        # mutates caller
```

The reporter's case is iterating the same `df` after constructing the dataset and seeing a different index than the one they passed in.

## Changes

- `src/gluonts/dataset/pandas.py`: take a shallow copy (`dataframe = dataframe.copy(deep=False)`) at the top of `from_long_dataframe` before any index reassignment. Shallow is enough — the underlying column data is not modified, only the index wrapper is rebound, and `df.index = X` on a shallow copy does not propagate back to the original (verified empirically).
- `test/dataset/test_pandas.py`: two regression tests covering both branches (timestamp-column path and auto-`to_datetime` path). Each asserts the caller's `index`, columns, and values are byte-identical after the call.

A short inline comment in the constructor cites #3263 so a future reviewer reading the diff cold sees the rationale.

## Reproduce BEFORE/AFTER yourself (copy-paste)

```bash
git clone https://github.com/awslabs/gluonts.git /tmp/repro && cd /tmp/repro
python3.11 -m venv .venv && source .venv/bin/activate
pip install -e . pytest pandas

# --- BEFORE (origin/dev) ---
git checkout origin/dev
git fetch https://github.com/jbbqqf/gluonts.git feat/3263-pandas-no-mutate
git checkout FETCH_HEAD -- test/dataset/test_pandas.py   # borrow the new tests only
pytest test/dataset/test_pandas.py -x -q -k does_not_mutate
# Expected: 2 failures — the caller's df.index has been mutated to DatetimeIndex

# --- AFTER (this PR) ---
git checkout FETCH_HEAD
pytest test/dataset/test_pandas.py -x -q
# Expected: all 31 tests pass; the new 2 mutation-guard tests included
```

## What I ran locally

- `pytest test/dataset/test_pandas.py -q` → **31 passed in 0.5s** on the branch.
- Same suite on `origin/dev` with the new tests cherry-picked → **2 failed, 29 passed** (the regression tests fail loudly).

## Edge cases tested

| # | Scenario | Input | Expected | Verified by |
|---|----------|-------|----------|-------------|
| 1 | `timestamp` column path | long df with explicit `time` column | caller's `df.index` (RangeIndex), columns, values unchanged after construction | `test_from_long_dataframe_does_not_mutate_caller_index` |
| 2 | Auto-`to_datetime` path | long df with string index, no `timestamp` arg | caller's string `df.index` unchanged after construction | `test_from_long_dataframe_does_not_mutate_caller_when_index_not_datetime` |
| 3 | Existing happy path | every case in `test_pandas.py` | dataset entries identical to before this PR | the existing 29 tests continue to pass |

## Risk / blast radius

Tiny. The change is local to `from_long_dataframe`; the shallow copy adds an O(1) reference bump, no data copy. Downstream code in the same function continues to operate on the (now copy-owned) index/columns, so behaviour from the dataset's perspective is unchanged. Direct construction via `PandasDataset(...)` (the `__post_init__` / `_pair_to_dataentry` path) has a similar mutation pattern but is out of scope here — the linked issue specifically targets `from_long_dataframe` (the cited line range is `pandas.py:284-286`); a follow-up could harden the other path.

## Release note

```release-note
PandasDataset.from_long_dataframe no longer mutates the caller's DataFrame index.
```

---

*PR drafted with assistance from Claude Code. The change was reviewed manually against gluonts dev and the reproducer block above was used during development; it is the same one a reviewer can paste verbatim.*


lostella (approved):
Thanks @jbbqqf!

PR #3304, opened 2026-06-15
[Security] Resolve security issue for torch.load

jalencato:
*Issue #, if available:*

*Description of changes:*

Add weight-only option for torch.load

### DIFF
--- a/requirements/requirements-pytorch.txt
+++ b/requirements/requirements-pytorch.txt
@@ -1,4 +1,4 @@
-torch>=1.9,<3
+torch>=1.13,<3
 lightning>=2.2.2,<2.6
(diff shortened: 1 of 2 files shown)


lostella (changes requested):
Thanks @jalencato, looks good to me overall. It seems like this option was added in torch 1.13, could you bump the version floor in https://github.com/awslabs/gluonts/blob/dev/requirements/requirements-pytorch.txt#L1 ?

jalencato:
> Thanks @jalencato, looks good to me overall. It seems like this option was added in torch 1.13, could you bump the version floor in https://github.com/awslabs/gluonts/blob/dev/requirements/requirements-pytorch.txt#L1 ?

Update the requirements.txt. Should resolve the PR.

lostella (approved):

Issue #3070, opened 2023-12-02
Numerical overflow in StdScaler


shchur:
## Description
StdScaler is prone to numerical overflow because of the squaring operation [here](https://github.com/awslabs/gluonts/blob/dev/src/gluonts/torch/scaler.py#L172).

## To Reproduce

```python
import torch
from gluonts.torch.scaler import StdScaler

x = torch.empty(5, 10).normal_(mean=0, std=1e20)
weights = torch.ones_like(x)
scaler = StdScaler(keepdim=True)
x, loc, scale = scaler(x, weights)  # scale contains inf values
```

This leads to models that use StdScaler encountering NaN values, e.g.,
```python
import pandas as pd
from gluonts.torch.model.tft import TemporalFusionTransformerEstimator

data = [{"target": [1e20 * i for i in range(50)], "start": pd.Period("2020-01-01", "D")}]
estimator = TemporalFusionTransformerEstimator(
    prediction_length=30, freq="D", trainer_kwargs={"max_epochs": 1, "accelerator": "cpu"}
)
predictor = estimator.train(data)  # all weights are NaN, forecasts are also all NaN
```
### What doesn't work
- Using `torch.masked.masked_tensor` to hide values with 0 weight

## Environment
- Operating system: Linux
- Python version: 3.10
- GluonTS version: 0.14.2

(Add as much information about your environment as possible, e.g. dependencies versions.)

lostella:
@shchur do you think you can take a stab at this?

PR #3312, opened 2026-07-04
Fix StdScaler overflow for large-valued inputs

K2alyan:
*Issue #, if available:* #3070

## What was happening

`StdScaler` computes weighted variance by squaring the centered values directly in float32. For inputs with values around 1e20 (which isn't unusual in financial or physical datasets), `(x - mean)^2` lands around 1e40 — past float32's ceiling of ~3.4e38. The result is `inf` scale, which cascades into `NaN` model weights and predictions.

The minimal reproducer from the issue:

```python
import torch
from gluonts.torch.scaler import StdScaler

x = torch.empty(5, 10).normal_(mean=0, std=1e20)
_, _, scale = StdScaler(keepdim=True)(x, torch.ones_like(x))
print(scale)  # tensor([[inf], [inf], [inf], [inf], [inf]])
```

After this fix:

```python
print(scale)  # tensor([[9.8e+19], [1.0e+20], ...])  — finite, correct
```

## The fix

Cast to float64 before squaring, take the sqrt there, then cast back. The standard deviation itself (~1e20) is representable in float32; it's the intermediate variance (~1e40) that isn't. The cast is inside `torch.no_grad()` so there's no gradient overhead.

```python
# before
variance = (((data - loc) * weights) ** 2).sum(...) / denominator

# after
variance = (
    (data.double() - loc.double()).pow(2) * weights
).sum(...) / denominator
scale = torch.sqrt(variance + self.minimum_scale).to(data.dtype)
```

The rewrite also corrects the formula: the original code squared `weights` alongside the deviations (`((data - loc) * weights) ** 2`). For the binary 0/1 indicators GluonTS uses this produces the same numbers, but it doesn't match the weighted-variance definition. Changed to `(data - loc).pow(2) * weights`.

## Tests

Added three `StdScaler` tests that were missing from the file:

- constant series (variance=0, scale falls back to sqrt(minimum_scale), output is zero)
- partial observation (loc and scale computed from observed window only)
- the exact overflow reproducer from #3070 — asserts `scale` and `scaled` are finite


lostella (approved):
Thanks @K2alyan!

PR #3324, opened 2026-07-20
Serde: restrict decode to known-serializable types

jalencato:
`serde.decode` previously instantiated any class it could resolve from a `class` name, guarded only by a short list of disallowed builtins. This replaces that list with an allowlist: `decode` now only instantiates types that `encode` is known to produce.

Allowed targets:
- classes that opt into serde (`@validated`, `serde.dataclass`, and subclasses of `Stateless`/`Stateful`/`PurePath`/`BaseModel`)
- NamedTuples and types that define `__getnewargs_ex__`
- a fixed set of builtins (`tuple`, `set`, `getattr`, `partial`, ...)
- the constructors emitted by the registered numpy/pandas/mxnet/zebras encoders, listed by name

`Kind.Type` is unchanged: it returns the located object without calling it.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/gluonts/core/serde/_base.py
+++ b/src/gluonts/core/serde/_base.py
@@ -11,6 +11,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
+import dataclasses
 import textwrap
 from enum import Enum
 from functools import singledispatch, partial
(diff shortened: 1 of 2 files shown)


lostella (approved):

src/gluonts/core/serde/_base.py
```diff
@@ -295,6 +296,85 @@ def encode_partial(v: partial) -> Any:
...
+    ):
+        return True
+
+    # classes that expose their init kwargs, or are reconstructable via the
+    # pickle `__getnewargs_ex__` protocol -- both are encoding paths `encode`
+    # relies on for such types
+    if hasattr(cls, "__init_passed_kwargs__") or hasattr(
+        cls, "__getnewargs_ex__"
+    ):
+        return True
```
lostella, line 374: I'm not sure how much of a concern this is, but any class could expose these attributes and pass the check

PR #3330, opened 2026-07-30
fix(util): reject tar link members escaping extraction target

melopeo:
*Description of changes:*

safe_extractall validated only member names, so a symlink (or hardlink) member whose target pointed outside the destination passed validation. extractall then created the link and wrote a following member through it, allowing arbitrary file write (path traversal).

Inspect link members in will_extractall_into and reject any whose resolved target escapes the destination. Also pass filter="data" on Python 3.12+ as defense in depth. Adds regression tests for symlink/hardlink escape and a benign internal symlink.

### DIFF
--- a/src/gluonts/util.py
+++ b/src/gluonts/util.py
@@ -101,9 +101,7 @@ def will_extractall_into(tar: tarfile.TarFile, path: Path) -> None:
             try:
                 link_target.relative_to(path)
             except ValueError:
-                raise PermissionError(
-                    f"'{member.name}' links out of target."
-                )
+                raise PermissionError(f"'{member.name}' links out of target.")
 
 
 def safe_extractall(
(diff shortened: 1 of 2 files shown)


src/gluonts/util.py
```diff
@@ -88,6 +89,20 @@ def will_extractall_into(tar: tarfile.TarFile, path: Path) -> None:
         except ValueError:
             raise PermissionError(f"'{member.name}' extracts out of target.")
 
+        # Reject links whose target escapes the destination.
+        if member.issym() or member.islnk():
```
shchur, line 95: Why not just reject any link? I don't think there is a valid reason to include symlinks in the types of archives that GluonTS might want to deal with (e.g. benchmark datasets)
```suggestion
        if member.issym() or member.islnk():
            raise ValueError("...")
```

melopeo: Good call — switched to rejecting any link member outright, which is simpler and also handles chained-link cases. Kept PermissionError for consistency with the existing validation; let me know if you'd prefer ValueError.

shchur (approved):
LGTM!

PR #3346, opened 2026-09-04
fix(torch): narrow training-recovery catch and preserve the traceback (#3295)

### LINKED ISSUE #3295: Torch estimator: broad except Exception in checkpoint recovery masks unrelated failures and loses optimizer/RNG state
## Context

[#3265](https://github.com/awslabs/gluonts/pull/3265) added a try/except around `trainer.fit` in `src/gluonts/torch/model/estimator.py` so that training can survive the `StudentTOutput` NaN `ValueError` described in the PR description. The current implementation (dev @ a0af774, lines 210-240) is:

```python
try:
    trainer.fit(
        model=training_network,
        train_dataloaders=training_data_loader,
        val_dataloaders=validation_data_loader,
        ckpt_path=ckpt_path,
    )
except Exception as e:
    if checkpoint.best_model_path == "":
        logger.error("Training failed with no checkpoint available")
        raise
    else:
        logger.warning(
            "Recovering from a checkpoint after training failed "
            f"with the following exception:\n {e}"
        )

if checkpoint.best_model_path != "":
    logger.info(f"Loading best model from {checkpoint.best_model_path}")
    best_model = training_network.__class__.load_from_checkpoint(
        checkpoint.best_model_path
    )
else:
    best_model = training_network
```

## Observations

1. **The catch is too wide.** `except Exception` matches the targeted `ValueError`, but also CUDA OOM (`torch.cuda.OutOfMemoryError`), disk-full errors from dataloader checkpoints, shape mismatches from mis-configured user features, pickling failures, and Lightning's own `MisconfigurationException`. After any of these the training log shows a single WARNING and the function returns a stale checkpoint from a much earlier epoch, which users then evaluate and possibly promote to production. The failure signal that should have been a hard stop becomes a silent accuracy regression.

2. **The original traceback is discarded.** `e` is formatted with `f\"{e}\"`, so only the exception's `args` are logged. Users lose the file/line/frames they need to decide whether the exception was NaN-loss (fine to recover from) or a real bug. `logger.exception(...)` or `logger.warning(..., exc_info=True)` both preserve the stack.

3. **Recovery restores only model weights.** `training_network.__class__.load_from_checkpoint(...)` rebuilds the module state dict; it does not restore the optimizer, LR scheduler, RNG, or dataloader iterator. For a short run that is fine — but the warning message does not say so, and users reading \"recovering from a checkpoint\" reasonably expect the full training state. A longer-running pipeline that wraps `train(...)` and then continues with additional gradient steps will silently diverge from a run that never hit the exception.

4. **No retry / attempt counter.** Nothing prevents a user from wrapping `train(...)` in their own retry loop and silently burning through N failed runs, each of which recovers from the very first checkpoint and reports success.

## Suggested direction

- Narrow the catch to the exceptions this code was designed to recover from. `(ValueError, RuntimeError)` covers both the NaN-loss `ValueError` described in #3265 and the \"Expected parameter ... to satisfy the constraint\" `RuntimeError` variant that PyTorch sometimes raises instead. Let `torch.cuda.OutOfMemoryError`, `KeyboardInterrupt`, `SystemExit`, and Lightning's `MisconfigurationException` propagate.
- Log with `logger.exception(...)` / `exc_info=True` so the traceback lives in the log.
- If full trainer state is worth restoring, Lightning already does it: instead of `load_from_checkpoint` after the catch, pass `ckpt_path=checkpoint.best_model_path` on a re-`trainer.fit(...)` call. If that is out of scope, at minimum document (in the warning message and the estimator docstring) that only weights are restored.
- Consider an optional `max_recoveries` knob — default 1 — so a flapping training is surfaced to the user by the second failure instead of silently truncating every attempt.

I am happy to send a PR if the direction above is agreeable; I wanted to open the issue first since the trade-off around \"narrow the catch vs. keep the safety net wide\" is a design call for the maintainers.

warren830:
## Issue

Fixes #3295.

## Problem

`PyTorchLightningEstimator.train_model` wraps `trainer.fit` in a `try/except` (added in #3265) to survive the `StudentTOutput` NaN `ValueError`. But the catch was `except Exception`, so it also swallowed **unrelated** failures — CUDA OOM (`torch.cuda.OutOfMemoryError`), dataloader/disk errors, shape mismatches from mis-configured features, Lightning's `MisconfigurationException`, etc. After any of those, `train_model` logged a single WARNING and returned a **stale best checkpoint from a much earlier epoch**, turning a hard failure into a silent accuracy regression. The original traceback was also discarded (`f"{e}"` logs only the exception args).

## Fix

- Narrow the catch to `(ValueError, RuntimeError)` — this covers the targeted NaN `ValueError` from #3265 and the "Expected parameter … to satisfy the constraint" `RuntimeError` variant PyTorch sometimes raises instead. Everything else (OOM, `KeyboardInterrupt`/`SystemExit`, misconfiguration) now propagates as the hard stop it is.
- Log the recovered-from failure with `exc_info=True` so the full traceback stays in the log.
- The warning message now states explicitly that only model **weights** are restored on recovery, not optimizer/scheduler/RNG state.

## Tests

Added `test/torch/model/test_estimator_recovery.py` (mocks `pl.Trainer`, no real training):

- a `ValueError` with a checkpoint available **is recovered from**;
- an unrelated failure (`MemoryError`, `KeyboardInterrupt`) **propagates** even when a checkpoint exists;
- a failure with **no** checkpoint re-raises.

Strict red→green verified: reverting the catch to `except Exception` makes the unrelated-failure test fail (the exception is masked); the fix makes it pass. `black --line-length 79` clean.

This PR intentionally does not change the weights-only recovery into a full-trainer-state restore or add a `max_recoveries` knob (both raised in the issue as design calls) — happy to follow up if maintainers want them.

### DIFF
--- a/src/gluonts/torch/model/estimator.py
+++ b/src/gluonts/torch/model/estimator.py
@@ -214,14 +214,30 @@ class PyTorchLightningEstimator(Estimator):
                 val_dataloaders=validation_data_loader,
                 ckpt_path=ckpt_path,
             )
-        except Exception as e:
+        except (ValueError, RuntimeError) as e:
+            # This recovery path exists for the numerical-instability
+            # failures training can hit late in a run -- e.g. the
+            # ``StudentTOutput`` NaN ``ValueError`` from #3265, or the
+            # "Expected parameter ... to satisfy the constraint"
+            # ``RuntimeError`` PyTorch sometimes raises instead. Only those
+            # are caught here: unrelated failures (CUDA OOM, dataloader
+            # errors, misconfiguration, ``KeyboardInterrupt``/``SystemExit``)
+            # are not ``ValueError``/``RuntimeError`` and propagate as the
+            # hard stops they are, instead of being masked by a stale
+            # checkpoint from a much earlier epoch.
             if checkpoint.best_model_path == "":
                 logger.error("Training failed with no checkpoint available")
                 raise
             else:
+                # ``exc_info=True`` keeps the full traceback in the log so the
+                # recovered-from failure can still be diagnosed. Note only the
+                # model weights are restored below, not the optimizer, LR
+                # scheduler, RNG, or dataloader state.
                 logger.warning(
-                    "Recovering from a checkpoint after training failed "
-                    f"with the following exception:\n {e}"
+                    "Recovering from the best checkpoint after training "
+                    "failed with the following exception (only model weights "
+                    "are restored, not optimizer/scheduler/RNG state):",
+                    exc_info=True,
                 )
 
         if checkpoint.best_model_path != "":
--- a/test/torch/model/test_estimator_recovery.py
+++ b/test/torch/model/test_estimator_recovery.py
@@ -1,0 +1,119 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+"""
+Tests for the training-failure recovery path in
+``PyTorchLightningEstimator.train_model``.
+
+The recovery ``try/except`` around ``trainer.fit`` exists to survive the
+numerical-instability failures training can hit late in a run (e.g. the
+``StudentTOutput`` NaN ``ValueError`` from #3265). It must recover from those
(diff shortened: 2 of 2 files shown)


warren830:
Hi maintainers — flagging that this PR has not had CI run yet: as a first-time contributor the fork workflows are still awaiting approval, so there are currently zero checks on it.

Status as it stands today:

- Branch is **up to date with `dev`** (0 commits behind), so approving now would run CI against current base.
- Scope is 2 files, +138/-3: narrowing the training-recovery `except` in `src/gluonts/torch/model/estimator.py` and adding `test/torch/model/test_estimator_recovery.py`.
- The change addresses #3295: the broad catch currently swallows unrelated failures during recovery and discards the original traceback, so a real bug surfaces as a confusing recovery error. The fix narrows the caught type and re-raises with the traceback preserved.
- The added test fails on the unpatched code and passes with the fix, so it is a real red-to-green guard.

A single "Approve and run workflows" would give this real CI signal. Happy to adjust the caught exception set if you would prefer a different boundary — that is the one design point where maintainer preference matters more than mine.

